PR #427, opened 2017-01-06
Better handling of mouse events in the table body.

### LINKED ISSUE #291: table (react) eats all click events
### Bug report

- __Package version(s)__:  1.1.0
- __Browser and OS versions__: MacOS 10.11.6 & Ubuntu 16.10 both with Chrome 54.0.2840.98. Mac 10.11.6 with Safari 10.0.1

#### Steps to reproduce

1. Create a table
1. Attach an onClick or onMouseDown/Up handler to an element rendered in that table
1. Click on that element and the handler will never fire
1. You can best demonstrate this by going to the demo (http://blueprintjs.com/docs/#components.table-js.example-editable) and trying to use your mouse to move the cursor to the beginning of the field after you activate it for editing.

cannot-click

(All that mouse hovering is me trying to click to move the cursor around)

#### Actual behavior

Mouse click events are not propagated to children cells (even if you set the `SelectionModes` to `NONE`.)

#### Expected behavior

Children elements of the table to be able to receive and respond to click events.

themadcreator:
Fixes #291

Previously, mouse events in the table body were used for selecting regions. With
the addition of interactive cells such as `EditableCell` components, we must now
provide full mouse interactivity to parts of the table body.

Previously, two instances of `Draggable` were contending for mouse events. The
first was waiting for double clicks to focus the `EditableCell`. The second was
listening for clicks and drags for region selection. Both of these were invoking
`preventDefault`, which was preventing text selection from working within the
`EditableCell` even when it was in edit mode.

This change adds new props to `Draggable` to control the use of `preventDefault`
and `stopPropagation` on the mouse events. We also added an `interactive` prop to
the `Cell` component, which applies a z-index that brings the cell above the
region layer.

Now, the table body will listen for clicks and drags for selection.
`EditableCell`s will listen for double clicks to start editing. When editing is
begun, the cell become "interactive" and it is moved to a z-index above the
selection regions layer. Also during edit mode, the `preventDefault` prop is
disabled and `stopPropagation` is enabled to prevent the table body from messing
with selected regions. This prevents an issue where the user starts dragging a
text selection in an `EditableCell` and ends up selecting regions.

In addition, we performed some cleanup on the table body:

Consolidated ghost and non-ghost render methods, which were nearly identical.

Moved selection interaction component to top-level of table body instead of
attaching listeners to individual cells. This will reduce the total number of
event listeners attached to DOM elements. However, it did necessitate the
creation of a small wrapper to use the `ContextMenuTarget` decorator.


themadcreator:
Of particular interest will be the "Editable Column Names and Cells" preview example

packages/table/src/cell/cell.tsx
```diff
@@ -17,9 +17,21 @@ export interface ICellProps extends IIntentProps, IProps {
...
+     */
+    interactive?: boolean;
+
+    /**
      * An optional native tooltip that is displayed on hover
      */
     tooltip?: string;
+
+    /**
+     * If true (the default), the cell contents will be wrapped in a div with
```
giladgray: could use the `@default true` syntax supported by the docs?

packages/table/src/cell/cell.tsx
```diff
@@ -28,10 +40,18 @@ export const emptyCellRenderer = (_rowIndex: number, _columnIndex: number) => <C
...
+    };
+
     public render() {
-        const { style, tooltip, className } = this.props;
-        const content = (<div className="bp-table-truncated-text">{this.props.children}</div>);
-        const classes = classNames("bp-table-cell", className, Classes.intentClass(this.props.intent));
-        return (<div className={classes} style={style} title={tooltip}>{content}</div>);
+        const { className, intent, interactive, style, tooltip, truncated } = this.props;
+        const content = truncated ?
+            <div className="bp-table-truncated-text">{this.props.children}</div> : this.props.children;
```
giladgray: there's also `.pt-text-overflow-ellipsis` in core styles

packages/table/src/cell/cell.tsx
```diff
@@ -28,10 +40,18 @@ export const emptyCellRenderer = (_rowIndex: number, _columnIndex: number) => <C
...
+        const classes = classNames(
+            "bp-table-cell", className, Classes.intentClass(intent), {
```
giladgray: `this.props.className` should come last when using `classNames()`

packages/table/src/cell/editableCell.tsx
```diff
@@ -40,45 +41,85 @@ export interface IEditableCellProps extends ICellProps {
...
+        this.setState({ isEditing: false });
+        if (this.props.onCancel) {
+            this.props.onCancel(value);
+        }
+    }
+
+    private handleConfirm = (value: string) => {
+        this.setState({ isEditing: false });
+        if (this.props.onConfirm) {
+            this.props.onConfirm(value);
```
giladgray: core utils has `safeInvoke(this.props.onConfirm, value)` for this pattern https://github.com/palantir/blueprint/blob/master/packages/core/src/common/utils.ts#L18

packages/table/src/common/_variables.scss
```diff
@@ -49,3 +49,9 @@ $resize-horizontal-cursor: ns-resize !default;
 
 // Actions
 $action-cursor: pointer !default;
+
+/*
+Z-index layers
+*/
+$interactive-cell-z-index: 10;
+$region-layer-z-index: 9;
```
giladgray: 1. add `!default` so these can be customized
2. make them based on `$pt-z-index-content [- 1]` ?

packages/table/src/common/contextMenuTargetWrapper.tsx
```diff
@@ -0,0 +1,31 @@
...
+import * as React from "react";
+
+export interface IContextMenuTargetWrapper extends IProps {
+    renderContextMenu: (e: React.MouseEvent<HTMLElement>) => JSX.Element;
+    style: React.CSSProperties;
+}
+
+/**
+ * Since the ContextMenuTarget uses the `onContextMenu` prop instead of event
+ * attachment API, the prop can be lost. This wrapper helps us maintain context
```
giladgray: what do you mean "event attachment API?" like `document.addEventListener`?

packages/table/src/interactions/draggable.tsx
```diff
@@ -86,11 +86,28 @@ export interface IDragHandler {
...
     onDoubleClick?: (event: MouseEvent) => void;
+
+    /**
+     * Default true, this prevents mouse events from performing their default
+     * operation such as text selection.
+     */
+    preventDefault?: boolean;
+
+    /**
+     * Default false, this prevents the event from propagating up to parent
```
giladgray: use `@default false` syntax

packages/table/src/common/_variables.scss
themadcreator: What's `!default` ?

packages/table/src/common/contextMenuTargetWrapper.tsx
themadcreator: Yeah, `element.addEventListener("contextmenu", ...)` versus `onContextMenu`

giladgray: but React... like of course it doesn't do that.

packages/table/src/common/_variables.scss
giladgray: https://robots.thoughtbot.com/sass-default

giladgray: > You can assign to variables if they aren’t already assigned by adding the !default flag to the end of the value. This means that if the variable has already been assigned to, it won’t be re-assigned, but if it doesn’t have a value yet, it will be given one.

themadcreator: Neato

packages/table/src/cell/cell.tsx
```diff
@@ -19,6 +19,15 @@ export interface ICellProps extends IIntentProps, IProps {
     style?: React.CSSProperties;
 
     /**
+     * If true, the cell will be rendered above overlay layers to enable mouse
+     * interactions within the cell.
+     *
```
giladgray: don't need the blank line here (see `tooltip` below) but if you like it then whatevs

packages/table/src/cell/editableCell.tsx
```diff
@@ -40,45 +41,81 @@ export interface IEditableCellProps extends ICellProps {
     onConfirm?: (value: string) => void;
 }
 
-export class EditableCell extends React.Component<IEditableCellProps, {}> {
-    private cellElement: HTMLElement;
+export interface IEditableCellState {
+    /**
+     * Stores the editing state of the cell
+     */
```
giladgray: there's no strong need to document state fields cuz they're only used in the component, and this comment doesn't supply new information given the prop name.

packages/table/src/cell/editableCell.tsx
```diff
@@ -40,45 +41,81 @@ export interface IEditableCellProps extends ICellProps {
...
+    isEditing: boolean;
+}
+
+@PureRender
+export class EditableCell extends React.Component<IEditableCellProps, IEditableCellState> {
+    public constructor(props: IEditableCellProps, context?: any) {
+        super(props, context);
+        this.state = {
+            isEditing: false,
+        };
```
giladgray: preferable to do the simple `public state = { ... }` instead of defining the whole constructor (unless of course you need the constructor but pretty sure you don't here)

packages/table/src/cell/editableCell.tsx
```diff
@@ -40,45 +41,81 @@ export interface IEditableCellProps extends ICellProps {
...
+
         return (
-            <div
-                className={classNames(className, Classes.intentClass(intent), "bp-table-cell")}
-                style={style}
-                title={tooltip}
-                ref={this.handleCellRef}
-            >
-                <Draggable onDoubleClick={this.handleCellDoubleClick}>
+            <Cell {...this.props} truncated={false} interactive={interactive}>
```
giladgray: 🎉  _neato_

packages/table/src/cell/editableCell.tsx
```diff
@@ -40,45 +41,81 @@ export interface IEditableCellProps extends ICellProps {
...
+    }
+
+    private handleConfirm = (value: string) => {
+        this.setState({ isEditing: false });
+        Utils.safeInvoke(this.props.onConfirm, value);
+    }
+
+    private handleCellActivate = (_event: MouseEvent) => {
+        // Cancel edit of active cell when clicking away
+        if (!this.state.isEditing && document.activeElement instanceof HTMLElement && document.activeElement.blur) {
```
giladgray: boolean coercion in the last clause. 
could argue that the `instanceof` check is enough because [MDN](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/blur)

packages/table/src/cell/editableCell.tsx
```diff
@@ -40,45 +41,81 @@ export interface IEditableCellProps extends ICellProps {
...
+            document.activeElement.blur();
+        }
+        return true;
     }
 
     private handleCellDoubleClick = (_event: MouseEvent) => {
-        if (this.cellElement == null) {
+        const cellElement = ReactDOM.findDOMNode(this) as HTMLElement;
```
giladgray: why did you switch from `ref` to `findDOMNode`? i'm undecided on which is preferable.

packages/table/src/common/contextMenuTargetWrapper.tsx
giladgray: how about saying "uses prop instead of adding event listeners" (paraphrased) for clarity?

packages/table/src/interactions/draggable.tsx
```diff
@@ -10,6 +10,7 @@ import * as PureRender from "pure-render-decorator";
 import * as React from "react";
 import * as ReactDOM from "react-dom";
 
+import { Utils } from "../common/utils";
```
giladgray: future issue: this is _super_ awkward usage. why not just export individual functions like we do in other utils files?

packages/table/src/interactions/draggable.tsx
```diff
@@ -86,11 +86,31 @@ export interface IDragHandler {
...
+     * This prevents mouse events from performing their default operation such
+     * as text selection.
+     *
+     * @default true
+     */
+    preventDefault?: boolean;
+
+    /**
+     * This prevents the event from propagating up to parent elements.
+     *
```
giladgray: calling out unnecessary whitespace again, but it's your call.

packages/table/src/tableBody.tsx
```diff
@@ -125,53 +146,22 @@ export class TableBody extends React.Component<ITableBodyProps, {}> {
...
-                onSelection={onSelection}
-                selectedRegions={selectedRegions}
-                selectedRegionTransform={selectedRegionTransform}
-            >
-                {React.cloneElement(cell, { style } as ICellProps)}
-            </DragSelectable>
-        );
+        const rect = isGhost ? grid.getGhostCellRect(rowIndex, columnIndex) : grid.getCellRect(rowIndex, columnIndex);
+        const style = Object.assign({}, cell.props.style, Rect.style(rect));
+        return React.cloneElement(cell, { style, key } as ICellProps);
```
giladgray: two clones, one object: `Utils.assignClasses` and right here. for every cell.

is it possible to merge those into one final clone, given that the first one just changes `className`?

packages/table/src/cell/editableCell.tsx
themadcreator: but isn't there an issue re-using the same object for every instance? I know state objects aren't supposed to be mutated directly, but setting this in the constructor always creates a new object for the instance just to be safe. Mostly I was just following the pattern i saw elsewhere, is this not the way?

packages/table/src/cell/editableCell.tsx
themadcreator: Hmm, i wanted to further verify that the blur method was supported even if the element is an HTMLElement. We did the same thing for the focus method below

packages/table/src/interactions/draggable.tsx
themadcreator: That's fair, I guess I've always structured utils as an object. But I see your point. For another PR, though

packages/table/src/cell/editableCell.tsx
giladgray: the compiler writes a constructor for you and puts all those instance properties inside it. same thing with bound `private func = () => {}` properties.

from `/docs/docs/docs.js` (can't link to lines):
image

packages/table/src/cell/editableCell.tsx
themadcreator: I switched from rendering a regular `<div>` to a custom component `<Cell>`. The ref in that case would be a `Cell`, but i need the `HTMLElement`. So this just seems more direct.

packages/table/src/cell/editableCell.tsx
themadcreator: well okay. It's less code anyway.

packages/table/src/cell/editableCell.tsx
giladgray: oh snap yeah. makes perfect sense.

packages/table/src/cell/editableCell.tsx
giladgray: understandable! i'll bet @adidahiya wants you to check `!= null`

PR #435, opened 2017-01-09
Add support for cell loading state

michael-yx-wu:
#### Checklist

- [x] Include tests
- [x] Update documentation

- Are docs sufficient / good?
- The example is located in the `Table > Loading States > Cell` section


packages/docs/src/styles/_sections.scss
```diff
@@ -244,7 +244,9 @@
 }
 
 #{section-id("components.table-js")} {
-  .kss-example {
```
michael-yx-wu: instead of modifying this class, we can just modify the bp-table-container class directly with the same effect

adidahiya:
can you add an example of this feature to the table preview page as well?

also the latest docs preview doesn't work; webpack complains about a missing module in https://992-71939872-gh.circle-artifacts.com/0/home/ubuntu/blueprint/packages/docs/dist/index.html

packages/table/src/_docs.scss
```diff
@@ -124,6 +124,28 @@ Styleguide components.table-js.example-editable
 */
 
 /*
+Loading States
```
llorca: Loading **s**tates, we're all sentence casing now 🔨

packages/table/src/cell/_common.scss
```diff
@@ -18,7 +18,12 @@ $dark-cell-background-color: $dark-gray3 !default;
 $dark-cell-border-color: $pt-dark-divider-black !default;
 $dark-cell-text-color: $pt-dark-text-color !default;
 
+@mixin fade-in() {
```
llorca: Is this worth a mixin? It's so short

packages/table/src/_docs.scss
michael-yx-wu: Oops!

packages/table/src/cell/_common.scss
michael-yx-wu: I think it is. I use it in a couple places in the original PR.

packages/table/src/cell/cell.tsx
```diff
@@ -25,13 +27,28 @@ export interface ICellProps extends IIntentProps, IProps {
 export type ICellRenderer = (rowIndex: number, columnIndex: number) => React.ReactElement<ICellProps>;
 
 export const emptyCellRenderer = (_rowIndex: number, _columnIndex: number) => <Cell />;
+export const loadingCellRenderer = () => <Cell loading={true} />;
+
+export const CELL_CLASSNAME = "bp-table-cell";
```
llorca: That should go into some kind of Classes.ts file

michael-yx-wu:
@adidahiya This is the first part of the table loading feature. Future PR(s) will include supporting showing loading states on individual `EditableCell`s, `ColumnHeaderCell`s, and `RowHeader`s. There will also be another PR that allows you to set "bulk" options to show loading states on all column headers, row headers, and table body cells. 

Should all of these use cases get their own examples as well?

packages/table/src/cell/cell.tsx
michael-yx-wu: Sure. I'm going to make a new Classes.ts file in the Table package then

adidahiya:
Feel free to split things up into small PRs; that is preferred. I don't think you need to go overboard with granular examples for every possible configuration.

michael-yx-wu:
Okay. I'll add an example to the preview when I make the PR for the "bulk" loading options on the `Table` component itself.

michael-yx-wu:
screen shot 2017-01-09 at 2 31 41 pm

@llorca noticed that on Chrome the border-radius doesn't always seems render properly. This seems to be related to odd height and percentage width. Has anyone seen this before? @giladgray @adidahiya

packages/core/src/common/_progress.scss
```diff
@@ -0,0 +1,23 @@
+// Copyright 2017 Palantir Technologies, Inc. All rights reserved.
```
giladgray: this file should be `progress/_common.scss`

packages/core/src/common/classes.ts
```diff
@@ -63,6 +63,8 @@ export const INPUT_GROUP = "pt-input-group";
 
 export const LABEL = "pt-label";
 
+export const LOADING = "pt-loading";
```
giladgray: this belongs in the "modifiers" block at the top, with ACTIVE

packages/core/src/components/skeleton/_skeleton.scss
```diff
@@ -43,24 +43,7 @@ Markup:
...
-    border-color: $skeleton-color-start;
-    background-color: $skeleton-color-start;
-  }
-
-  50% {
-    border-color: $skeleton-color-end;
-    background-color: $skeleton-color-end;
-  }
-}
+@import "../../common/progress";
```
giladgray: update path

packages/table/src/cell/_common.scss
```diff
@@ -18,7 +18,12 @@ $dark-cell-background-color: $dark-gray3 !default;
 $dark-cell-border-color: $pt-dark-divider-black !default;
 $dark-cell-text-color: $pt-dark-text-color !default;
 
+@mixin text-fade-in() {
+  transition: color $pt-transition-duration * 3;
```
giladgray: this probably works better as a variable? `$text-fade-in-transition`

packages/table/src/cell/cell.tsx
```diff
@@ -25,13 +27,30 @@ export interface ICellProps extends IIntentProps, IProps {
...
-        const classes = classNames("bp-table-cell", className, Classes.intentClass(this.props.intent));
-        return (<div className={classes} style={style} title={tooltip}>{content}</div>);
+        const { style, loading, tooltip, className } = this.props;
+
+        const classes = classNames(
+            CELL_CLASSNAME,
+            Classes.intentClass(this.props.intent),
+            {
+               [Classes.LOADING]: loading,
+            },
```
giladgray: style nit: i typically inline one-prop objects. `{ [key]: val }`

packages/table/src/cell/cell.tsx
giladgray: tracked in #438

packages/table/src/common/loadableContent.tsx
```diff
@@ -0,0 +1,45 @@
...
+ * Licensed under the BSD-3 License as modified (the “License”); you may obtain a copy
+ * of the license at https://github.com/palantir/blueprint/blob/master/LICENSE
+ * and https://github.com/palantir/blueprint/blob/master/PATENTS
+ */
+
+import * as React from "react";
+
+import { Classes } from "@blueprintjs/core";
+
+export interface ILoadable {
```
giladgray: `ILoadingProps` for consistency with `IIntentProps`, etc?

packages/table/src/common/loadableContent.tsx
```diff
@@ -0,0 +1,45 @@
...
+    /**
+     * Render the loading state.
```
giladgray: not great docs...

packages/table/src/common/loadableContent.tsx
```diff
@@ -0,0 +1,45 @@
...
+     * @default false
+     */
+    loading?: boolean;
+}
+
+export interface ILoadableContentProps extends ILoadable {
+    /**
+     * Show a variable length skeleton when rendering the loading state.
```
giladgray: explain what "variable length" means here (using word in its own definition). 
something about choosing a random width

packages/table/src/common/loadableContent.tsx
```diff
@@ -0,0 +1,45 @@
...
+}
+
+export class LoadableContent extends React.Component<ILoadableContentProps, {}> {
+    private skeletonLength: number;
+    private style: React.CSSProperties;
+
+    public constructor(props: ILoadableContentProps) {
+        super(props);
+        this.skeletonLength = props.variableLength ? 75 - Math.floor(Math.random() * 11) * 5 : 100;
+        this.style = { width: `${this.skeletonLength}%` };
```
giladgray: store either `skeletonLength` or `style` cuz one is easily derivable from the other. your choice.

packages/table/src/common/loadableContent.tsx
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    public render() {
+        if (this.props.loading) {
+            return <div className={`${Classes.SKELETON}`} style={this.style} />;
+        }
+
+        return React.Children.only(this.props.children);
```
giladgray: this is probably worth documenting in the usage of this component: it requires exactly one child.

packages/table/test/cellTests.tsx
```diff
@@ -30,6 +30,12 @@ describe("Cell", () => {
         expect(cell.find(".inner").text()).to.equal("Purple");
     });
 
+    it("renders loading state", () => {
+        const cellHarness = harness.mount( <Cell loading={true} />);
```
giladgray: remove leading whitespace: `mount(<Cell`

packages/table/src/common/loadableContent.tsx
giladgray: also that it doesn't support string content.

though we could make it support string content, that might be valuable (see popover line 268).

packages/table/src/common/loadableContent.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+    public constructor(props: ILoadableContentProps) {
+        super(props);
+        this.skeletonLength = props.variableLength ? 75 - Math.floor(Math.random() * 11) * 5 : 100;
+        this.style = { width: `${this.skeletonLength}%` };
+    }
+
+    public render() {
+        if (this.props.loading) {
+            return <div className={`${Classes.SKELETON}`} style={this.style} />;
```
giladgray: no need to interpolate the one string. `className={Classes.SKELETON}`

packages/table/test/loadableContentTests.tsx
```diff
@@ -0,0 +1,50 @@
...
+        const loadableContentHarness = harness.mount(
+           <LoadableContent loading={false}>
+               <span>{someText}</span>
+           </LoadableContent>,
+        );
+
+        expect(loadableContentHarness.element.textContent).to.equal(someText);
+    });
+
+    it("does not render child when loading", () => {
```
giladgray: "renders skeleton instead of child when loading"

packages/table/src/common/_loading.scss
```diff
@@ -0,0 +1,42 @@
+// Copyright 2017 Palantir Technologies, Inc. All rights reserved.
+
+@import "~@blueprintjs/core/src/common/progress";
+@import "~@blueprintjs/core/src/common/variables";
+
+@mixin cell-loading() {
```
giladgray: use this mixin directly in block below beginning line 31. could just inline it.

packages/table/src/common/_loading.scss
```diff
@@ -0,0 +1,42 @@
+// Copyright 2017 Palantir Technologies, Inc. All rights reserved.
```
giladgray: @themadcreator what do you think of this file living in `common/` dir? 
it's not a true common file because it defines CSS rules (not just sass stuff). but it affects both cells and headers.

packages/core/src/components/progress/_common.scss
```diff
@@ -9,3 +9,20 @@ $progress-track-color: rgba($gray1, 0.2) !default;
 $progress-head-color: rgba($gray1, 0.8) !default;
 $dark-progress-track-color: rgba($black, 0.5) !default;
 $dark-progress-head-color: $gray3 !default;
+
+$skeleton-animation-duration: $pt-transition-duration * 20 !default;
+$skeleton-color-start: rgba($gray4, 0.2) !default;
+$skeleton-color-end: rgba($gray1, 0.2) !default;
+
+@keyframes glow {
```
giladgray: this is not kosher for a `_common` file. will be rendered to CSS twice. i would define this directly in `_skeleton.scss`

packages/core/src/components/progress/_common.scss
```diff
@@ -9,3 +9,20 @@ $progress-track-color: rgba($gray1, 0.2) !default;
 $progress-head-color: rgba($gray1, 0.8) !default;
 $dark-progress-track-color: rgba($black, 0.5) !default;
 $dark-progress-head-color: $gray3 !default;
+
+$skeleton-animation-duration: $pt-transition-duration * 20 !default;
+$skeleton-color-start: rgba($gray4, 0.2) !default;
+$skeleton-color-end: rgba($gray1, 0.2) !default;
```
giladgray: these vars don't belong here because they start with `skeleton-`. perhaps `skeleton/_common.scss`

packages/core/src/components/progress/_common.scss
```diff
@@ -9,3 +9,20 @@ $progress-track-color: rgba($gray1, 0.2) !default;
 $progress-head-color: rgba($gray1, 0.8) !default;
 $dark-progress-track-color: rgba($black, 0.5) !default;
 $dark-progress-head-color: $gray3 !default;
+
+$skeleton-animation-duration: $pt-transition-duration * 20 !default;
```
giladgray: i'd refactor to make this even more easily reusable (and then you won't have to stylelint-disable):

```scss
$skeleton-animation: ($pt-transition-duration * 20) linear infinite glow !default;
```

packages/table/examples/cellLoadingExample.tsx
```diff
@@ -0,0 +1,115 @@
...
+
+import * as React from "react";
+
+import { RadioGroup } from "@blueprintjs/core";
+import BaseExample, { handleStringChange } from "@blueprintjs/core/examples/common/baseExample";
+
+import { Cell, Column, Table } from "../src";
+
+// tslint:disable-next-line:no-var-requires
+const bigSpaceRocks = require("./pha.json") as IBigSpaceRock[];
```
giladgray: 1. awkward to use interface before it's defined.
2. can we give the file a better name? what is "pha?"

packages/table/src/cell/_common.scss
```diff
@@ -18,12 +18,15 @@ $dark-cell-background-color: $dark-gray3 !default;
 $dark-cell-border-color: $pt-dark-divider-black !default;
 $dark-cell-text-color: $pt-dark-text-color !default;
 
+$text-fade-in-transition: color $pt-transition-duration * 3;
```
giladgray: hrmm i take it back. make the variable be just the `-duration` and add ` !default` to the end.

packages/table/src/cell/cell.tsx
```diff
@@ -25,13 +27,28 @@ export interface ICellProps extends IIntentProps, IProps {
 export type ICellRenderer = (rowIndex: number, columnIndex: number) => React.ReactElement<ICellProps>;
 
 export const emptyCellRenderer = (_rowIndex: number, _columnIndex: number) => <Cell />;
+export const loadingCellRenderer = () => <Cell loading={true} />;
```
giladgray: this guy's still here?

packages/table/src/common/_loading.scss
```diff
@@ -0,0 +1,40 @@
...
+
+@import "~@blueprintjs/core/src/common/variables";
+@import "~@blueprintjs/core/src/components/progress/common";
```
giladgray: should import `skeleton/common`

packages/table/test/loadableContentTests.tsx
```diff
@@ -0,0 +1,67 @@
...
+        const loadableContentHarness = harness.mount(
+           <LoadableContent loading={false}>
+               <span>{someText}</span>
+           </LoadableContent>,
+        );
+
+        expect(loadableContentHarness.element.textContent).to.equal(someText);
+    });
+
+    it("does not render string content", () => {
```
giladgray: not simply "does not render," but "throws error on string content"

same with next test

packages/core/src/components/progress/_common.scss
michael-yx-wu: nice! that's a good idea

michael-yx-wu: will need to stylelint disable here since we're moving the keyframes directly into _skeleton.scss

packages/table/examples/cellLoadingExample.tsx
michael-yx-wu: 1. okay, will reorder lines
2. ripped this from slate docs. "potentially hazardous asteroids"

packages/table/src/cell/cell.tsx
michael-yx-wu: oops, removing

packages/table/src/cell/_common.scss
michael-yx-wu: hm in that case i'd like to use this variable inside `_loading.scss` as well, though it feels weird to have that file import from this one

packages/table/src/cell/_common.scss
```diff
@@ -18,12 +18,15 @@ $dark-cell-background-color: $dark-gray3 !default;
 $dark-cell-border-color: $pt-dark-divider-black !default;
 $dark-cell-text-color: $pt-dark-text-color !default;
 
+$fade-in-duration: $pt-transition-duration * 3;
```
llorca: `$cell-transition-duration`, or even abstract it all to `$cell-transition`

packages/table/src/common/loadableContent.tsx
```diff
@@ -0,0 +1,55 @@
...
+import * as React from "react";
+
+import { Classes } from "@blueprintjs/core";
+
+export interface ILoadingProps {
+    /**
+     * If true, show a fixed height loading skeleton instead of the cell content.
+     * @default false
+     */
+    loading?: boolean;
```
michael-yx-wu: Do docs here need to be more elaborate? Note that these docs will be repeated by any component props that extends this interface

packages/table/examples/cellLoadingExample.tsx
giladgray: lol sold! 🚢

packages/table/src/common/loadableContent.tsx
```diff
@@ -31,8 +31,13 @@ export class LoadableContent extends React.Component<ILoadableContentProps, {}>
...
     public constructor(props: ILoadableContentProps) {
         super(props);
-        const skeletonLength = props.variableLength ? 75 - Math.floor(Math.random() * 11) * 5 : 100;
-        this.style = { width: `${skeletonLength}%` };
+        this.calculateStyle(props.loading);
+    }
+
+    public componentWillReceiveProps(nextProps: ILoadableContentProps) {
+        if (!this.props.loading && nextProps.loading || this.props.variableLength !== nextProps.variableLength) {
+            this.calculateStyle(nextProps.variableLength);
```
themadcreator: Since you use Math.random() every time in calculateStyle, won't this make the skeleton rectangle change size if the loading state switches back and forth?

packages/table/src/common/_loading.scss
themadcreator: I suppose the loading stuff living here is okay. Maybe a separate subdirectory should be considered as well.

As far as the SASS, I think i would prefer to make the rules below a mixin and then have bp-table-cell and bp-table-header mix them in in their respective .scss files.

packages/table/src/common/loadableContent.tsx
themadcreator: This description seems fine to me

packages/table/src/common/loadableContent.tsx
michael-yx-wu: Yes, that's the intended behavior. It seem weird to have rectangles be the same the random length when switching.

packages/table/src/common/_loading.scss
```diff
@@ -0,0 +1,38 @@
...
+    // loading state
+    // stylelint-disable-next-line declaration-no-important
+    color: transparent !important;
+    pointer-events: none;
+  }
+
+  .pt-skeleton {
+    opacity: 0;
+    height: $pt-grid-size / 2;
+    animation: skeleton-animation($cell-transition-duration * 2),
```
giladgray: set `animation-delay: 0 ($cell-transition-duration * 2);` separately. also swap order so un-delayed transition is first (time order).

packages/core/src/components/skeleton/_common.scss
```diff
@@ -0,0 +1,11 @@
+// Copyright 2017 Palantir Technologies, Inc. All rights reserved.
+
+@import "../../common/variables";
+
+$skeleton-animation: ($pt-transition-duration * 20) linear infinite glow !default;
+$skeleton-color-start: rgba($gray4, 0.2) !default;
+$skeleton-color-end: rgba($gray1, 0.2) !default;
+
+@function skeleton-animation($delay: 0) {
+  @return ($pt-transition-duration * 20) linear infinite $delay glow;
```
giladgray: remove this function and just set `animation-delay` separately when necessary.

packages/table/examples/cellLoadingExample.tsx
```diff
@@ -0,0 +1,122 @@
...
+import { Cell, Column, Table } from "../src";
+
+interface IBigSpaceRock {
+    id: number;
+    name: string;
+    year: number;
+    absoluteMagnitude: number;
+    discoverer: string;
+    observatory: string;
+    [key: string]: string | number;
```
giladgray: this doesn't seem necessary. there are no arbitrary keys.

packages/table/src/common/loadableContent.tsx
giladgray: if this interface will be shared across several components then the docs should be generalized so they can be reused in different contexts (for instance, don't talk about fixed-height or cells).

or, just define this prop separately in each interface with more complete and relevant documentation.

packages/table/src/_docs.scss
```diff
@@ -124,6 +124,30 @@ Styleguide components.table-js.example-editable
 */
 
 /*
+Loading states
+
+When fetching or updating data, it may be desirable to show a loading state.
```
giladgray: add:

The Table components provide fine-grain loading control of loading row headers, column headers, or individual cells.

packages/table/src/_docs.scss
```diff
@@ -124,6 +124,30 @@ Styleguide components.table-js.example-editable
...
+Weight: 4
+
+Styleguide components.table-js.loading
+*/
+
+/*
+Cell
+
+Although the most common use cases will probably involve showing the loading state on full columns,
+full rows, or the entire body, `Cell`s expose a `loading` prop for granular control of which cells
```
giladgray: remove everything before "`Cell`s expose..."

packages/table/src/_docs.scss
```diff
@@ -124,6 +124,30 @@ Styleguide components.table-js.example-editable
...
+should show a loading state. Below is a table of the largest potentially hazardous asteroid
+(based on absolute magnitude) discovered in a given year. Try selecting a different preset
+configuration.
```
giladgray: "preset loading configuration"

packages/table/examples/cellLoadingExample.tsx
michael-yx-wu: removing, but in newer version of tsc you'll see this error: 
```
message: 'Index signature of object type implicitly has an 'any' type.'
```

giladgray: should be safe to remove this line

packages/table/src/common/_loading.scss
```diff
@@ -32,7 +32,7 @@
   .pt-skeleton {
     opacity: 0;
     height: $pt-grid-size / 2;
-    animation: skeleton-animation($cell-transition-duration * 2),
-      $cell-transition-duration linear forwards skeleton-fade-in;
+    animation: $skeleton-animation, $cell-transition-duration linear forwards skeleton-fade-in;
+    animation-delay: $cell-transition-duration, 0;
```
giladgray: i would flip the order so the 0-delay transition is first. then it reads in the order it renders.

packages/table/src/common/loadableContent.tsx
```diff
@@ -0,0 +1,57 @@
...
+import * as React from "react";
+
+import { Classes } from "@blueprintjs/core";
+
+export interface ILoadingProps {
+    /**
+     * If true, show the component's loading state. When set on a component that supports
+     * `ICellProps`, content will be hidden behind a fixed-height skeleton. When set on all other
+     * components, children components that support `ICellProps` will display their loading state.
+     * @default false
```
michael-yx-wu: This interface will only be used on components that fall into these two categories. But having this indiscriminately in the docs feels weird. Better to just remove this then?

michael-yx-wu: *and put directly in the props?

PR #467, opened 2017-01-12
[WIP] Date range input

### LINKED ISSUE #249: Feature request: <DateRangeInput/>
Same like `<DateInput/>` for DateRange , for now i used popover. but i think it is good to have it as a component

## TODOs

@cmslewis: I'll use this space to track TODOs for this component (currently living on `feature/date-range-input`).

### Part 9: Obey and control the focused field
- [x] The currently focused field dictates which end of the date range the next click will modify.
- [x] Move focus between fields intelligently after a date is selected.
- [x] Add unit tests to verify the behavior.

### Part 10: Basic documentation
- [x] Add documentation for `DateRangeInput`
    - [x] Basic description of the component (hovered range, selected range, typing, validation)
    - [x] Uncontrolled mode
    - [x] Controlled mode (add disclaimer that it isn't fully implemented yet)
    - [x] JavaScript API (props, etc. - can get this for free)
    - [x] Basic example with controls (at this point, mostly just need a `format` selector)

### Part 11: Merge `feature/date-range-input` into `master`
- [ ] Handle merge conflicts with @leebyp's changes from #575

### Eventually
- [ ] Add `allowSingleDayRange` prop
- [ ] Add `selectAllOnFocus` prop
- [ ] Show min/max dates as placeholder on focus in an empty field
- [ ] (e.g.) When selected range is `[<startDate>, <endDate>]` and end field is focused, clicking on `<startDate>` in calendar should _clear_ the start date and _leave_ the end date in place. (Currently, it clears the selected range and then sets the clicked date as a new end date.) Should work symmetrically for start field too.  
- [ ] More robust controlled mode (how should we handle hover events, if at all?)
- [ ] (Refactor) Pull date-range state-change logic into its own `DateRangeSelectionStrategy` class  

### Further out
- [ ] When focused in an input field, allow keyboard incrementing of each of the date, month, or year in isolation (similar to `TimeInput`).

cmslewis:
# [!] This is just a proof of concept to play around with. Don't merge.

#### Fixes #249

@llorca @giladgray

Figured it would be easier to open a temporary PR and share a preview link (EDIT: ugh, still have to write tests to enable that) than to ask you guys to pull and run the branch yourselves. What do you think of the UX here? Also curious about the props API around separators - it'd be nice if the component could parse different separators (e.g. "-" and "to") but render only one kind of separator when displaying the formatted date range (in this case, "to"). These are configurable here via `separators` and `displaySeparator`, which seems...meh.

Overall, not sure I'm loving this single-field date range input. Feels like it isn't on rails enough, plus the behavior when you nullify the start or end date is particularly weird. There's also a lot of code that's not-quiiiite-identical to code in DateInput, but close enough. I'm sure we could pull some common functionality out into a dateInputCore.tsx file; I just didn't put any effort into that yet.

There's still some incorrect functionality around invalid dates. But I'm mostly curious about UX for the moment.

Here's a screen capture (ignore the weirdness at the end when refocusing after inputting an invalid date):

date-range-input

LMK what you think. Also LMK if you'd prefer a different forum for this discussion. 

#### Checklist

- [ ] Include tests
- [ ] Update documentation

- 
- 
- 

- 
- 
-

### DIFF
--- a/packages/datetime/examples/dateRangeInputExample.tsx
+++ b/packages/datetime/examples/dateRangeInputExample.tsx
@@ -1,0 +1,19 @@
+/*
+ * Copyright 2016 Palantir Technologies, Inc. All rights reserved.
+ * Licensed under the BSD-3 License as modified (the “License”); you may obtain a copy
+ * of the license at https://github.com/palantir/blueprint/blob/master/LICENSE
+ * and https://github.com/palantir/blueprint/blob/master/PATENTS
+ */
+
+import BaseExample from "@blueprintjs/core/examples/common/baseExample";
+import * as React from "react";
+
+import { DateRangeInput } from "../src";
+
+export class DateRangeInputExample extends BaseExample<{}> {
+    protected renderExample() {
+        return (
+            <DateRangeInput />
+        );
+    }
+}
--- a/packages/datetime/examples/dateRangeInputExample.tsx
+++ b/packages/datetime/examples/dateRangeInputExample.tsx
@@ -13,7 +13,7 @@ import { DateRangeInput } from "../src";
 export class DateRangeInputExample extends BaseExample<{}> {
     protected renderExample() {
         return (
-            <DateRangeInput />
+            <DateRangeInput format={"MM/DD/YYYY"} />
         );
     }
 }
--- a/packages/datetime/examples/index.ts
+++ b/packages/datetime/examples/index.ts
@@ -7,6 +7,7 @@
 
 export * from "./dateInputExample";
 export * from "./datePickerExample";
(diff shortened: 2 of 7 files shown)


leebyp:
is the idea behind having one input that we want to allow autoparsing for someone typing "past week" etc.?

otherwise tabbing between two inputs is more intuitive to me than "knowing" to type "to" / "-", though this is pretty much a design decision i guess

cmslewis:
> is the idea behind having one input that we want to allow autoparsing for someone typing "past week" etc.?

@leebyp - that was one of the original motivations, yes. spoke with @llorca; going to try a two-field approach now.

PR #471, opened 2017-01-12
Update table tsconfig to use new compiler flags

michael-yx-wu:
Let's use consistent compiler options across packages. Somewhat related: it also seems like `tslint:latest` has disabled the rules for these new flags to avoid being redundant.


adidahiya:
Instead of adding the same options in multiple places, can you use tsconfig.json inheritance with the `"extends"` keyword? https://www.typescriptlang.org/docs/handbook/tsconfig-json.html

We ought to have a `tsconfig.base.json` somewhere which defines a consistent set of compiler checks that we use across the packages. See https://github.com/palantir/blueprint/issues/376

michael-yx-wu:
Looks like that's only available for Typescript `2.1.x`. Should we track this and just duplicate the flags for now so that this doesn't block on #376?

adidahiya:
@michael-yx-wu yeah, that's fine; but I imagine bumping to 2.1 in this PR could be pretty simple too

michael-yx-wu:
Okay, I'll give that a shot!

giladgray:
per https://github.com/ivogabe/gulp-typescript/issues/459, this is not the correct approach.

Issue #298, opened 2016-12-02
Missing micro-interactions in table


llorca:
- [ ] Keyboard shortcuts (using `<Hotkey>`)
  - [x] Copy `cmd+c`
  - [ ] ~~Paste `cmd+v`~~ (can't do because table is data agnostic, we don't know what data you're putting in there)
  - [x] Move selection `up`/`right`/`bottom`/`left`
  - [ ] Select the cell immediately to the right `tab`
  - [ ] Select the cell immediately below `return`
  - [ ] Select the cell immediately to the left `shift+tab`
  - [ ] Select the cell immediately above `shift+return`
  - [ ] Type any character to edit

<br>

- [ ] Right-click menu
  - [x] When the right-click menu is open, we should block scrolling
  - [x] Right-click menu should be dismissed when clicking outside of it
  - [ ] Right-clicking on a selection shouldn't change the selection
    - [ ] Right-clicking on a cell/column/row should select said cell/column/row

<br>

- [ ] Editable (header) cells
  - [ ] Clicking outside of an cell being edited should disable editing
  - [ ] Ability to click on the edited text to place cursor and make a text selection with the mouse
  - [ ] Make editable cells and editable header cells consistent

<br>

- [ ] Selection
  - [ ] Clicking on "..." in cell shouldn't select the cell, similarly to other "..." (header for instance)
  - [ ] Column headers and row headers should reflect the current selection
  - [ ] Clicking on a selected cell shouldn't deselect it. Cmd+clicking on a cell should deselect it. (#1073)
  - [ ] Range select rows/columns `shift+click`
  - [ ] Add rows/columns/cells to selection `cmd+click`
  - [ ] Add to selection rectangle `cmd/shift+up`/`cmd/shift+right`/`cmd/shift+bottom`/`cmd/shift+left`
  - [ ] Clicking on a cell and dragging out of the table should widen selecting the active selection region

samdenicola:
@llorca These look great -- can't wait! 

Any thoughts on adding onCellDoubleClick/onRowDoubleClick callbacks? I'm trying to build a "library"-esque table component where rows are a single item in one's inventory with different columns being attributes of the item. Users should be able to double click to "open" up an item. My solution so far feels a bit hacky -- and is ugly without one of the above proposed fixes: "Clicking on a selected cell shouldn't deselect it. Cmd+clicking on a cell should deselect it."
```
<Cell>
  <div
    className={'clickable-cell'}
    onDoubleClick={() => {
[...]

PR #542, opened 2017-01-25
Add support for copying table cells with cmd+c hotkey

gscshoyru:
Part of #298 

#### Checklist

- [x] Include tests
- [x] Update documentation

You can specify a getCellData in the props and then
cmd+c will copy selected cells.

What I'm doing here is sane, the shortcut only works when it's supposed to, etc.

### DIFF
--- a/packages/core/src/compatibility/browser.ts
+++ b/packages/core/src/compatibility/browser.ts
@@ -10,7 +10,6 @@ const browser = {
     isEdge: !!userAgent.match(/Edge/),
     isInternetExplorer: (!!userAgent.match(/Trident/) || !!userAgent.match(/rv:11/)),
     isWebkit: !!userAgent.match(/AppleWebKit/),
-    // isWebkit: Object.prototype.toString.call((window as any).HTMLElement).indexOf("Constructor") > 0,
  };
 
 export const Browser = {
--- a/packages/core/src/compatibility/browser.ts
+++ b/packages/core/src/compatibility/browser.ts
@@ -9,9 +9,11 @@ const userAgent = typeof navigator !== "undefined" ? navigator.userAgent : "";
 const browser = {
     isEdge: !!userAgent.match(/Edge/),
     isInternetExplorer: (!!userAgent.match(/Trident/) || !!userAgent.match(/rv:11/)),
+    isWebkit: Object.prototype.toString.call((window as any).HTMLElement).indexOf("Constructor") > 0,
  };
 
 export const Browser = {
     isEdge: () => browser.isEdge,
     isInternetExplorer: () => browser.isInternetExplorer,
+    isWebkit: () => browser.isWebkit,
 };
--- a/packages/core/src/compatibility/browser.ts
+++ b/packages/core/src/compatibility/browser.ts
@@ -9,7 +9,8 @@ const userAgent = typeof navigator !== "undefined" ? navigator.userAgent : "";
 const browser = {
     isEdge: !!userAgent.match(/Edge/),
     isInternetExplorer: (!!userAgent.match(/Trident/) || !!userAgent.match(/rv:11/)),
-    isWebkit: Object.prototype.toString.call((window as any).HTMLElement).indexOf("Constructor") > 0,
+    isWebkit: !!userAgent.match(/AppleWebKit/),
+    // isWebkit: Object.prototype.toString.call((window as any).HTMLElement).indexOf("Constructor") > 0,
  };
 
 export const Browser = {
--- a/packages/table/preview/index.tsx
+++ b/packages/table/preview/index.tsx
@@ -44,7 +44,11 @@ ReactDOM.render(<Nav selected="index" />, document.getElementById("nav"));
 
 function getTableComponent(numCols: number, numRows: number, columnProps?: any, tableProps?: any) {
     // combine table overrides
-    const tablePropsWithDefaults = { numRows, ...tableProps };
+    const getCellData = (row: number, col: number) => {
+        return Utils.toBase26Alpha(col) + (row + 1);
+    };
+
+    const tablePropsWithDefaults = Object.assign({numRows, getCellData}, ...tableProps);
 
     // combine column overrides
     const columnPropsWithDefaults = {
@@ -271,6 +275,7 @@ class RowSelectableTable extends React.Component<{}, {}> {
     public render() {
         return (<div>
             <Table
+                renderBodyContextMenu={renderBodyContextMenu}
                 numRows={7}
                 isRowHeaderShown={false}
                 onSelection={this.handleSelection}
--- a/packages/table/preview/index.tsx
+++ b/packages/table/preview/index.tsx
@@ -48,11 +48,7 @@ function getTableComponent(numCols: number, numRows: number, columnProps?: any, 
         return Utils.toBase26Alpha(col) + (row + 1);
     };
 
-    const onCopy = () => {
-        console.log("did copy");
-    }
-
-    const tablePropsWithDefaults = {numRows, getCellClipboardData, onCopy, ...tableProps};
+    const tablePropsWithDefaults = {numRows, getCellClipboardData, ...tableProps};
 
     // combine column overrides
     const columnPropsWithDefaults = {
--- a/packages/table/preview/index.tsx
+++ b/packages/table/preview/index.tsx
@@ -44,7 +44,7 @@ ReactDOM.render(<Nav selected="index" />, document.getElementById("nav"));
 
 function getTableComponent(numCols: number, numRows: number, columnProps?: any, tableProps?: any) {
     // combine table overrides
-    const getCellData = (row: number, col: number) => {
+    const getCellClipboardData = (row: number, col: number) => {
         return Utils.toBase26Alpha(col) + (row + 1);
     };
 
@@ -52,7 +52,7 @@ function getTableComponent(numCols: number, numRows: number, columnProps?: any, 
         console.log("did copy");
     }
 
-    const tablePropsWithDefaults = {numRows, getCellData, onCopy, ...tableProps};
+    const tablePropsWithDefaults = {numRows, getCellClipboardData, onCopy, ...tableProps};
 
     // combine column overrides
     const columnPropsWithDefaults = {
--- a/packages/table/preview/index.tsx
+++ b/packages/table/preview/index.tsx
@@ -48,7 +48,11 @@ function getTableComponent(numCols: number, numRows: number, columnProps?: any, 
         return Utils.toBase26Alpha(col) + (row + 1);
     };
 
-    const tablePropsWithDefaults = Object.assign({numRows, getCellData}, ...tableProps);
+    const onCopy = () => {
+        console.log("did copy");
+    }
+
+    const tablePropsWithDefaults = {numRows, getCellData, onCopy, ...tableProps};
 
     // combine column overrides
     const columnPropsWithDefaults = {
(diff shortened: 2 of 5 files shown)


packages/table/src/table.tsx
```diff
@@ -51,6 +53,14 @@ export interface ITableProps extends IProps, IRowHeights, IColumnWidths {
     fillBodyWithGhostCells?: boolean;
 
     /**
+     * If exists, a callback that returns the data for a specific cell. This need not
+     * match the value displayed in the `<Cell>` component. The value will be
+     * invisibly added as `textContent` into the DOM before copying. If not exists,
+     * copy via hotkeys (mod+c) will not work
```
giladgray: trailing period please

packages/table/src/table.tsx
```diff
@@ -349,6 +360,30 @@ export class Table extends AbstractComponent<ITableProps, ITableState> {
...
+            e.preventDefault();
+            e.stopPropagation();
+
+            const cells = Regions.enumerateUniqueCells(selectedRegions, grid.numRows, grid.numCols);
+            const sparse = Regions.sparseMapCells(cells, getCellData);
+            Clipboard.copyCells(sparse);
+        };
+
+        return <Hotkeys>
+            <Hotkey label="copy selected table cells" group="table" combo="mod+c" onKeyDown={doCopy} />
```
giladgray: "Copy selected..." sentence casing

packages/table/src/table.tsx
```diff
@@ -349,6 +360,30 @@ export class Table extends AbstractComponent<ITableProps, ITableState> {
         );
     }
 
+    public renderHotkeys() {
+        const doCopy = (e: KeyboardEvent) => {
```
giladgray: why is this defined here instead of `private handleCopyHotkey`?

packages/table/preview/index.tsx
```diff
@@ -45,7 +45,11 @@ ReactDOM.render(<Nav selected="index" />, document.getElementById("nav"));
 
 function getTableComponent(numCols: number, numRows: number, columnProps?: any, tableProps?: any) {
     // combine table overrides
-    const tablePropsWithDefaults = { numRows, ...tableProps };
+    const getCellData = (row: number, col: number) => {
+        return Utils.toBase26Alpha(col) + (row + 1);
+    };
+
+    const tablePropsWithDefaults = Object.assign({numRows, getCellData}, ...tableProps);
```
giladgray: use the object spread notation that you just reverted. https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-1.html#object-spread-and-rest

`{ numRows, getCellData, ...tableProps }`

giladgray:
looks like it's time to increase table test coverage... can't keep lowering the threshold.

gscshoyru:
That's fine I should add tests anyways; just wanted eyes on it first :)

packages/table/preview/index.tsx
adidahiya: @giladgray can you enable the `ban` rule to ban Object.assign?

leebyp:
you can put the browser checks into the compatibility file introduced here https://github.com/palantir/blueprint/pull/590

packages/table/src/table.tsx
```diff
@@ -660,6 +711,22 @@ export class Table extends AbstractComponent<ITableProps, ITableState> {
...
+            return (
+                <Hotkey
+                    label="Copy selected table cells"
+                    group="table"
+                    combo="mod+c"
+                    onKeyDown={this.handleCopy}
+                />
+            );
+        } else {
+            return null;
```
giladgray: `return undefined` please

packages/table/test/harness.ts
```diff
@@ -13,6 +13,80 @@ import * as ReactDOM from "react-dom";
...
+    CHROME,
+    EDGE,
+    FIREFOX,
+    IE,
+    UNKNOWN,
+    WEBKIT,
+}
+
+/**
+ * Use feature detection to determine current browser.
```
giladgray: lol this is kinda backwards... you're supposed to use feature detection _instead of_ browser detection: ensure that the feature you want exists, rather than asserting the browser "make." it's version-independent and much less prone to false positives.

is there a more reliable way of doing these checks that we can shim in here quickly?

at the very least move this to its own file.

gscshoyru: I stole all of this from hotkeys test framework, that's how it does it. I couldn't figure out how to share test stuff between them. Is there a good way to do this? I'll totally do that. 

And yeah agree that it sucks but...

packages/table/preview/index.tsx
```diff
@@ -45,7 +45,15 @@ ReactDOM.render(<Nav selected="index" />, document.getElementById("nav"));
 
 function getTableComponent(numCols: number, numRows: number, columnProps?: any, tableProps?: any) {
     // combine table overrides
-    const tablePropsWithDefaults = { numRows, ...tableProps };
+    const getCellData = (row: number, col: number) => {
+        return Utils.toBase26Alpha(col) + (row + 1);
+    };
+
+    const onCopy = () => {
+        console.log("did copy");
```
gscshoyru: Oooh better idea. 👍

packages/table/src/table.tsx
```diff
@@ -51,6 +53,14 @@ export interface ITableProps extends IProps, IRowHeights, IColumnWidths {
     fillBodyWithGhostCells?: boolean;
 
     /**
+     * If exists, a callback that returns the data for a specific cell. This need not
+     * match the value displayed in the `<Cell>` component. The value will be
+     * invisibly added as `textContent` into the DOM before copying. If not exists,
+     * copy via hotkeys (mod+c) will not work.
```
gscshoyru: This works the same as it does for the context menu copy -- you need to give that a callback too. The reason why is what you want rendered (with divs and styles and pretty things) may be radically different from what you want copied (the data). I've talked to @themadcreator at length about this, and this seems like the best way :-/

packages/table/preview/index.tsx
gscshoyru: Oh wait I can't do this because it's not passing that info in -- it just gets success. I'm not sure what I'd pass to it actually. Lemme think about it.

packages/table/src/table.tsx
```diff
@@ -408,6 +436,29 @@ export class Table extends AbstractComponent<ITableProps, ITableState> {
...
+        // prevent "real" copy from being called
+        e.preventDefault();
+        e.stopPropagation();
+
+        const cells = Regions.enumerateUniqueCells(selectedRegions, grid.numRows, grid.numCols);
+        const sparse = Regions.sparseMapCells(cells, getCellData);
+        if (sparse != null) {
+            const success = Clipboard.copyCells(sparse);
+            if (onCopy != null) {
+                onCopy(success);
```
giladgray: use `Utils.safeInvoke` from the core package for this pattern please

packages/table/src/table.tsx
```diff
@@ -660,6 +711,22 @@ export class Table extends AbstractComponent<ITableProps, ITableState> {
         });
     }
 
+    private maybeRenderCopyHotkey() {
+        const {getCellData} = this.props;
```
giladgray: spaces in braces

packages/table/src/table.tsx
giladgray: ping

packages/table/src/table.tsx
gscshoyru: What is this for? I can see that it exists; I just don't understand why.

giladgray: safely invokes a maybe-defined callback. it's precisely this pattern (if function exists then call it) expressed as one type-safe statement.

packages/table/src/table.tsx
themadcreator: For example, @adidahiya, if they select a whole column, we'd want to copy all cells in the columns, but due to virtualized rendering, there's no guarantee that all the cells in that column will have ever been rendered. Furthermore, some cells have truncated content (like JSON content), so copying the actual cell text would be incomplete.

themadcreator: But we could consider renaming this `getCellClipboardData` since without the naming, users might think this is used to fill cells when rendering.

packages/table/src/table.tsx
themadcreator: It's to make the classic "check null, then invoke" code go from 3 lines to one.

packages/table/src/table.tsx
```diff
@@ -660,6 +711,22 @@ export class Table extends AbstractComponent<ITableProps, ITableState> {
...
+        if (getCellData != null) {
+            return (
+                <Hotkey
+                    label="Copy selected table cells"
+                    group="table"
```
themadcreator: Should group be capitalized? You can see the result my pressing `shift + ?` when the table is focused:
image

packages/table/test/harness.ts
```diff
@@ -13,6 +13,80 @@ import * as ReactDOM from "react-dom";
...
+        if ((typeof navigator !== "undefined") && /Mac|iPod|iPhone|iPad/.test(navigator.platform)) {
+            metaKey = true;
+        } else {
+            ctrlKey = true;
+        }
+    }
+
+    (event as any).initKeyboardEvent(eventType, true, true, window, key, 0, ctrlKey, false, false, metaKey);
+
+    // Hack around these readonly properties in WebKit and Chrome
```
themadcreator: I'm very sorry =(

packages/table/test/selectionTests.tsx
```diff
@@ -35,6 +37,21 @@ describe("Selection", () => {
...
+            return Utils.toBase26Alpha(col) + (row + 1);
+        };
+        const copyCellsStub = sinon.stub(Clipboard, "copyCells").returns(true);
+        const table = harness.mount(createTableOfSize(3, 7, {}, {getCellData, onCopy}));
+
+        table.find(TH_SELECTOR).mouse("mousedown").mouse("mouseup");
+        table.find(TH_SELECTOR).focus();
+        table.find(TH_SELECTOR).keyboard("keydown", "C", true);
+        expect(copyCellsStub.lastCall.args).to.deep.equal([[["A1"], ["A2"], ["A3"], ["A4"], ["A5"], ["A6"], ["A7"]]]);
+        expect(onCopy.lastCall.args).to.deep.equal([true]);
```
themadcreator: 👍

packages/table/preview/index.tsx
themadcreator: Maybe the on copy method itself should have an argument that contains the sparse copied cells?

themadcreator:
Okay, my requests have been addressed, though build is red

packages/core/src/compatibility/browser.ts
```diff
@@ -9,9 +9,11 @@ const userAgent = typeof navigator !== "undefined" ? navigator.userAgent : "";
 const browser = {
     isEdge: !!userAgent.match(/Edge/),
     isInternetExplorer: (!!userAgent.match(/Trident/) || !!userAgent.match(/rv:11/)),
+    isWebkit: Object.prototype.toString.call((window as any).HTMLElement).indexOf("Constructor") > 0,
```
giladgray: probably `>= 0`?

also make sure `window` is defined first

packages/core/src/compatibility/browser.ts
```diff
@@ -9,9 +9,12 @@ const userAgent = typeof navigator !== "undefined" ? navigator.userAgent : "";
 const browser = {
     isEdge: !!userAgent.match(/Edge/),
     isInternetExplorer: (!!userAgent.match(/Trident/) || !!userAgent.match(/rv:11/)),
+    isWebkit: !!userAgent.match(/AppleWebKit/),
+    // isWebkit: Object.prototype.toString.call((window as any).HTMLElement).indexOf("Constructor") > 0,
```
giladgray: please remove commented code

PR #189, opened 2016-11-17
Numeric Input Component

cmslewis:
#### PR checklist

- [x] New feature
  - [x] Includes tests
- [x] Documentation update

#### What changes did you make?

Implemented a numeric stepper component that can accept a numeric input value and modify it via helpful key bindings and increment buttons.

numeric-stepper-v2

Supported hotkeys are:

* ` ↑ `  - increment the value by 1 
* ` ↓ ` - decrement the value by 1
* ` Shift + ↑ `  - increment the value by 10
* ` Shift + ↓ ` - decrement the value by 10
* ` Alt/Option + ↑ ` - increment the value by 0.1
* ` Alt/Option + ↓ ` - decrement the value by 0.1

The keys also work when clicking the increment buttons:

* ` Click '+' ` - increment the value by 1
* ` Click '-' ` - decrement the value by 1
* ` Shift + Click '+' ` - increment the value by 10
* ` Shift + Click '-' ` - decrement the value by 10 
* ` Alt/Option + Click '+' ` - increment the value by 0.1
* ` Alt/Option + Click '-' ` - decrement the value by 0.1

#### Is there anything you'd like reviewers to focus on?

(Updated 2/2/17) At this point, this thing has been code-reviewed pretty thoroughly. Barring any glaring issues with code or documentation, I'd like to get this merged ASAP and tweak it in future PRs if we have to. I've verified through myriad unit tests and a high-powered example that this component is functional, extensible, and bug-free AFAIK. Should be good to go.

PR #559, opened 2017-01-27
Fix z-index stack for .pt-button-group and .pt-control-group elements

cmslewis:
(Sweeping up a couple final issues from #537, specifically from [these changes](https://github.com/palantir/blueprint/pull/537/files#diff-9f4ffc04872ae8416007b7851351b176) to `_control-group.scss`.)

### Before

(Note: I discovered these issues while working on the `NumericInput` component in #189.)

- If a control group has multiple consecutive buttons with an intent styling, the focus outlines will be improperly layered behind neighboring buttons. See here:

screen shot 2017-01-26 at 6 52 35 pm

- For buttons in a `.pt-control-group > .pt-button-group` with _no_ intent, a hovered neighbor would appear above a focused neighbor's outline:

image

Here's a GIF of the issues in action:

_Case 1: The buttons are **not** nested in a `.pt-button-group`:_

_Case 2: The buttons **are** nested in a `.pt-button-group`:_

 
## After:
After some z-index massaging, everything behaves as expected:

_Case 1: The buttons are **not** nested in a `.pt-button-group`:_

_Case 2: The buttons **are** nested in a `.pt-button-group`:_

### Dark Theme

Looks good as well.

_Case 1: The buttons are **not** nested in a `.pt-button-group`:_

_Case 2: The buttons **are** nested in a `.pt-button-group`:_

- Does this handle all the cases?
- Should we write unit tests for CSS edge cases like this?
- Should we start using variables for z-indexes within `_control-group.scss`? (I think so)

### DIFF
--- a/packages/core/src/components/forms/_control-group.scss
+++ b/packages/core/src/components/forms/_control-group.scss
@@ -109,6 +109,10 @@ Styleguide components.forms.control-group
 
     &[class*="pt-intent-"] {
       z-index: 3;
+
+      &:focus {
+        z-index: 6;
+      }
     }
   }
 
--- a/packages/core/src/components/forms/_control-group.scss
+++ b/packages/core/src/components/forms/_control-group.scss
@@ -110,6 +110,12 @@ Styleguide components.forms.control-group
     &[class*="pt-intent-"] {
       z-index: 3;
 
+      // ensure hovering over an element won't lift it above a neighboring
+      // element's focus outline
+      &:hover {
+        z-index: 5;
+      }
+
       &:focus {
         z-index: 6;
       }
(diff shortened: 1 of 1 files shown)


cmslewis:
Ah, found one last little bug that arises when hovering on neighbors of focused buttons:

image

This also manifested itself with the numeric input. Will push a fix momentarily.

llorca:
Very cool 🎉  That last bug is something expected. Mouse interactions take precedence on keyboard interactions.

Can you show what your examples look like in dark theme? I suspect we'll have to make some adjustments there

cmslewis:
@llorca whoops, just saw your comment and had already pushed fixes to keep the focus outline on top. How do the new GIFS look? The new behavior of keeping focus outlines always on top actually feels better to me, but happy to defer.

llorca:
@cmslewis now there are issues in both light and dark theme with focus non-intent buttons being on top of intent buttons

cmslewis:
@llorca - yes, that's what I'd expect. Feels weird to have a focus state underneath another element no matter the intent of either, no? Let's sync up in the office tomorrow.

llorca:
This is actually still a WIP, definitely not merging today

cmslewis:
@llorca - Whoops, thought this PR contained our new work, and I renamed it incorrectly. I'll close this PR and open a new one with the improved z-index corrections.

PR #575, opened 2017-01-30
Independent Daterangepicker months

### LINKED ISSUE #432: UX: Alter daterange picker UI to show startmonth/endmonth instead of two consecutive months
#### Actual behavior

Currently the daterange picker looks like this:
image

In the field, users often want to select a date range from today to a date far in the future (like the next full calendar year). Like so:
image

##### Example 1
1. select start date
1. selected end date
1. change month
1. select end date (FAIL) This doesn't work because is starts a new range.

##### Example 2
A user wants to search say... 3 years in the past. They navigate the left cal to 2014 (of the current month) and select today's date. Then they go to the right calendar to click today's date (of this year, because that's what it looked like a second ago).

They get confused that the right calendar has updated (and sometimes they don't notice because they were so sure that it wasn't going to change).

#### Expected behavior

Furthermore, many users first impression was that the two months represented separate startdate/enddate calendars.

Preferably, there should be an optional mode where the two calendar months are disconnected, allowing the user to easily select start date in one and end date in the other.

Further, the calendars could have some minor coordination to make sure the left calendar is always chronologically before the right one.

@haosharon Created a copy of daterangepicker that does this on GHE. That might be a good place to start.

leebyp:
#### Fixes #432 

- [x] Fix existing tests
- [x] Add new tests
- [x] Update documentation

- DateRangePicker months set to be contiguous through prop
- Extracted date comparison logic into separate class for readability


cmslewis:
@llorca I guess it depends how valuable the interaction of shuffling both months together might be; that's the capability that would go away if we hide the buttons when the months are contiguous. I don't have a good sense for that at this point.

llorca:
@thisisalessandro would love to get your design input

thisisalessandro:
A few notes:
- Disabling the arrows might be equally confusing: the mental gap between noticing that the arrows are disabled and understanding that it's because the months are contingent is big. Having the months "move" together is at least practical
- Maybe the first click should always be interpreted as a start date. Right now if you pick a date and then pick another one that precedes the former, the first one counted as an end date and the second one as a start date. I think the most common case for pick a date prior to the one you just selected is because you picked a wrong start date, not because you are thinking in reverse
- Related to☝️ : when we print the dates we represent the first click as a start date
- Not super convinced by the fact that disabled days don't have gray background. Especially in cases like this:
screen shot 2017-01-31 at 4 16 52 pm

leebyp:
re- @thisisalessandro 's points

1. Yeah that was my worry too. I think this is intrinsically linked with the decision of whether you can move the left date ahead of the right one **directly with the dropdown**.
2. This is related but should definitely not be in this PR - related to things @cmslewis is doing.
3. -
4. With the increased flexibility in the component now, I'm actually down to just set `showOutsideDays` to `false` - then it should be obvious which are disabled.

llorca:
I'm not opposed to having `showOutsideDays` to `false` by default if you feel strongly 👍

cmslewis:
@llorca @leebyp @thisisalessandro - just touched on outside days in my first PR on the road to merging my  `DateRangeInput` work. I kind of think we should just get rid of them. Reasoning here:
https://github.com/palantir/blueprint/pull/586/files#r98813760

thisisalessandro:
I'd be down with removing them. They are only useful if you can see one month at a time.

cmslewis:
That can actually happen in the DateRangePicker when minDate and maxDate are within the same calendar month. Not sure how high-priority that case is though.

On Jan 31, 2017, at 8:11 PM, Alessandro Mingione <notifications@github.com<mailto:notifications@github.com>> wrote:


I'd be down with removing them. They are only useful if you can see one month at a time.

llorca:
This is cool, we want it. Can we have it as a simple prop/option on the date range picker?

leebyp:
@llorca the optional prop is implemented in #678

packages/datetime/src/_daterangepicker.scss
```diff
@@ -55,11 +55,26 @@ Styleguide components.datetime.daterangepicker.js
...
+      min-width: $pt-grid-size * 43;
+    }
+
+    .DayPicker-Month:not(:last-child) {
+      margin-right: $pt-grid-size;
+    }
+
+    ~ .DayPicker {
+      border-left: 1px solid $pt-divider-black;
+      min-width: $pt-grid-size * 44;
```
cmslewis: Why are the min-widths different?

packages/datetime/src/common/classes.ts
```diff
@@ -18,6 +18,7 @@ export const DATEPICKER_MONTH_SELECT = "pt-datepicker-month-select";
 export const DATEPICKER_YEAR_SELECT = "pt-datepicker-year-select";
 
 export const DATERANGEPICKER = "pt-daterangepicker";
+export const DATERANGEPICKER_SEQUENTIAL = "pt-daterangepicker-sequential";
```
cmslewis: This class still abides by the old `isSequential` naming scheme. Can we change to `pt-daterangepicker-contiguous` or something?

packages/datetime/src/common/dateUtils.ts
```diff
@@ -107,3 +109,21 @@ export function getDateTime(date: Date, time: Date) {
             time.getHours(), time.getMinutes(), time.getSeconds(), time.getMilliseconds());
     }
 }
+
+export function getDatePreviousMonth(date: Date): Date {
+    const monthIsJanuary = date.getMonth() === Months.JANUARY;
+    if (monthIsJanuary) {
```
cmslewis: nit: IMO the equality expression above is clear enough without this extra variable.

packages/datetime/src/common/monthAndYear.ts
```diff
@@ -0,0 +1,74 @@
...
+        return compareMonthAndYear(this, displayMonth) > 0;
+    }
+
+    public isSame(displayMonth: MonthAndYear): boolean {
+        return compareMonthAndYear(this, displayMonth) === 0;
+    }
+}
+
+// returns -ve if left < right
+// returns +ve if left > right
```
cmslewis: nit: the `...ve` suffix took a sec for me to interpret. [I guess I'm not mathy enough](https://en.wiktionary.org/wiki/%2Bve). Mild preference for `negative` and `positive`.

packages/datetime/test/dateRangePickerTests.tsx
```diff
@@ -127,8 +129,147 @@ describe("<DateRangePicker>", () => {
...
+    describe("left/right calendar when not sequential", () => {
+        it("only shows one calendar when minDate and maxDate are in the same month", () => {
+            const contiguousCalendarMonths = false;
+            const minDate = new Date(2015, Months.DECEMBER, 1);
+            const maxDate = new Date(2015, Months.DECEMBER, 15);
+
+            renderDateRangePicker({ contiguousCalendarMonths, maxDate, minDate });
+            assert.lengthOf(document.getElementsByClassName("DayPicker"), 1);
+            assert.lengthOf(document.getElementsByClassName(".DayPicker-NavButton--prev"), 0);
+            assert.lengthOf(document.getElementsByClassName(".DayPicker-NavButton--next"), 0);
```
cmslewis: We don't want the dots in the class name when using `document.getElementsByClassName`. Won't return anything.

packages/datetime/test/dateRangePickerTests.tsx
```diff
@@ -57,43 +57,45 @@ describe("<DateRangePicker>", () => {
...
+            assert.equal(dateRangePicker.state.leftView.getMonth(), today.getMonth());
         });
 
         it("is a day between minDate and maxDate if only maxDate/minDate set and today is not in range", () => {
             const maxDate = new Date(2005, Months.JANUARY);
             const minDate = new Date(2000, Months.JANUARY);
             renderDateRangePicker({ maxDate, minDate });
-            const { displayMonth, displayYear } = dateRangePicker.state;
-            assert.isTrue(DateUtils.isDayInRange(new Date(displayYear, displayMonth), [minDate, maxDate]));
+            const leftDisplayMonth = dateRangePicker.state.leftView;
```
cmslewis: nit: Pondering to myself it `const leftView = ...` would be clearer to me than `const leftDisplayMonth = ...`, since the display month includes a year. Thoughts @leebyp?

packages/datetime/src/_daterangepicker.scss
```diff
@@ -55,11 +55,26 @@ Styleguide components.datetime.daterangepicker.js
 .pt-daterangepicker {
   white-space: nowrap;
 
+  &.pt-daterangepicker-sequential {
+    .DayPicker {
+      min-width: $pt-grid-size * 43;
```
leebyp: Extracted out the width for DayPicker since it's used to calculate the other widths. And made the css a bit more verbose to show what's going on. I haven't moved the shortcuts out, since it is arbitrary.

packages/datetime/test/dateRangePickerTests.tsx
leebyp: Ah yes thanks, this was leftover from refactors - updated a couple of other places i missed too.

packages/datetime/test/dateRangePickerTests.tsx
leebyp: haha oops...

leebyp:
@cmslewis should be easier to read now

cmslewis:
@leebyp love the "Show shortcuts" toggle and the CSS variables! The `monthAndYear` variables are still a little tricky to get right, but I'm okay with them as is. My only remaining nit is to give the toggle labels a little more room if possible:


Otherwise, looks good IMO.

llorca:
Shouldn't it be "Constrain" instead of "Contrain"?

PR #586, opened 2017-02-01
[Date Range Input] - Part 1: Show a selection preview on hover

cmslewis:
#### Checklist

- [x] Include tests
- [x] Update documentation (should update automatically to show `onHoverChange` in the props list)

##### What
On hover in the `DateRangePicker`, show a visual preview of the range you're about to select. Additionally, hide "outside" dates, because they introduce too much weirdness in hover interactions.

##### Why?
The upcoming `DateRangeInput` implementation will need to display the hovered range in its input fields so that the user remains fully aware what the effect of the next mouse click will be (e.g. the component will need to show the hovered dates in the input fields and move focus between the fields based on the hovered date vs. the selected date/range). This PR just gets the visual hover preview working.

- Are the CSS styles sane and minimal?
- Are there any weird bugs with hovered ranges vis-a-vis selected dates/ranges?
- Are there any bugs with `allowSingleDayRange`?
- Is it reasonable to call `handleDayMouseEnter` from `handleDayClick`? Another option may be to update `hoverValue` within `getStateChange`, but my `hoverValue` was emerging as `undefined` in my `modifiers` when I tried to update it through that code path. I'm also not sure how 1:1 the update logic will be between hover and selection in more complex strategies.

### DIFF
--- a/packages/datetime/src/_daterangepicker.scss
+++ b/packages/datetime/src/_daterangepicker.scss
@@ -94,34 +94,16 @@ Styleguide components.datetime.daterangepicker.js
   // very long selectors ahead!
   // stylelint-disable max-line-length
 
-  .DayPicker-Day--hover-range:not(.DayPicker-Day--hover-range-end) {
-    border-top-right-radius: 0;
-    border-bottom-right-radius: 0;
-  }
-
-  .DayPicker-Day--hover-range:not(.DayPicker-Day--hover-range-start) {
-    border-top-left-radius: 0;
-    border-top-right-radius: 0;
-  }
-
   .DayPicker-Day--selected-range-start:not(.DayPicker-Day--outside):not(.DayPicker-Day--selected-range-end):not(.DayPicker-Day--hovered-range-end),
-  .DayPicker-Day--hovered-range-start:not(.DayPicker-Day--outside):not(.DayPicker-Day--selected-range-end):not(.DayPicker-Day--hovered-range-end) {
+  .DayPicker-Day--hovered-range-start:not(.DayPicker-Day--outside):not(.DayPicker-Day--hovered-range-end) {
     border-top-right-radius: 0;
     border-bottom-right-radius: 0;
-
-    &.DayPicker-Day--hovered-range {
-      border-radius: 0;
-    }
   }
 
-  .DayPicker-Day--selected-range-end:not(.DayPicker-Day--outside):not(.DayPicker-Day--selected-range-start),
+  .DayPicker-Day--selected-range-end:not(.DayPicker-Day--outside):not(.DayPicker-Day--selected-range-start):not(.DayPicker-Day--hovered-range-start),
   .DayPicker-Day--hovered-range-end:not(.DayPicker-Day--outside):not(.DayPicker-Day--hovered-range-start) {
     border-top-left-radius: 0;
     border-bottom-left-radius: 0;
-
-    &.DayPicker-Day--hovered-range {
-      border-radius: 0;
-    }
   }
 
   // stylelint-enable max-line-length
--- a/packages/datetime/src/_daterangepicker.scss
+++ b/packages/datetime/src/_daterangepicker.scss
@@ -63,22 +63,65 @@ Styleguide components.datetime.daterangepicker.js
     margin-right: $pt-grid-size;
   }
 
+  .DayPicker-Day--hovered-range:not(.DayPicker-Day--outside) {
+    border-radius: 0;
+
+    // need to disable hover-styles for all variants of selected dates
+    // stylelint-disable max-line-length
+    &:not(.DayPicker-Day--selected):not(.DayPicker-Day--selected-range):not(.DayPicker-Day--selected-range-start):not(.DayPicker-Day--selected-range-end) {
+      background-color: $light-gray4;
+
+      &:hover {
+        background-color: $light-gray1;
+      }
+    }
+    // stylelint-enable max-line-length
+  }
+
   .DayPicker-Day--selected-range:not(.DayPicker-Day--outside) {
     border-radius: 0;
     background-color: $light-gray4;
+
+    &:hover {
+      background-color: $light-gray1;
+    }
+  }
+
+  .DayPicker-Day--outside {
+    visibility: hidden;
   }
 
   // very long selectors ahead!
   // stylelint-disable max-line-length
 
-  .DayPicker-Day--selected-range-start:not(.DayPicker-Day--outside):not(.DayPicker-Day--selected-range-end) {
+  .DayPicker-Day--hover-range:not(.DayPicker-Day--hover-range-end) {
     border-top-right-radius: 0;
     border-bottom-right-radius: 0;
   }
 
-  .DayPicker-Day--selected-range-end:not(.DayPicker-Day--outside):not(.DayPicker-Day--selected-range-start) {
+  .DayPicker-Day--hover-range:not(.DayPicker-Day--hover-range-start) {
+    border-top-left-radius: 0;
+    border-top-right-radius: 0;
+  }
+
+  .DayPicker-Day--selected-range-start:not(.DayPicker-Day--outside):not(.DayPicker-Day--selected-range-end):not(.DayPicker-Day--hovered-range-end),
+  .DayPicker-Day--hovered-range-start:not(.DayPicker-Day--outside):not(.DayPicker-Day--hovered-range-end) {
+    border-top-right-radius: 0;
+    border-bottom-right-radius: 0;
+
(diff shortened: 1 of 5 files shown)


packages/datetime/src/_daterangepicker.scss
```diff
@@ -63,20 +63,45 @@ Styleguide components.datetime.daterangepicker.js
...
   .DayPicker-Day--selected-range:not(.DayPicker-Day--outside) {
     border-radius: 0;
     background-color: $light-gray4;
+
+    &:hover {
+      background-color: $light-gray1;
+    }
+  }
+
+  .DayPicker-Day--outside {
```
cmslewis: I've hidden "outside" dates (dates in Month N's calendar that are actually part of Month N±1) for a few reasons:

1. When you click on an outside date, it selects the corresponding date in the neighboring month. To show an accurate preview on hover, you'd presumably want to highlight those actual dates that would be included in the selection post-click. But this feels super weird. Look:



Granted, there _is_ a missing style revealed by this GIF: a hovered-range endpoint that isn't selected or below the mouse cursor. Even with that, I think this would feel weird.

2. It's not clear whether you should highlight an outside date when it falls within a hovered range. I got feedback from other designers in 101F that they were confused that the outside dates weren't also highlighted when a hovered range spanned them:

image

3. (Not really relevant to this PR's scope) We may eventually consider adding a "disabled" style to dates that precede a selected start date to prevent the selection of invalid, earlier end dates. I'd want those disabled dates to look exactly the way outside dates look now.

thisisalessandro: I'd be down with removing them. They are only useful if you can see one month at a time.

llorca: Bye outside dates 👋

leebyp:
@cmslewis set the `minDate` / `maxDate` to disable some dates, and the dates in between selected and the cursor are still highlighted

screen shot 2017-02-01 at 13 19 14 2

packages/datetime/test/dateRangePickerTests.tsx
```diff
@@ -511,11 +723,31 @@ describe("<DateRangePicker>", () => {
...
+    /**
+     * Returns the hovered range excluding endpoints.
+     */
+    function getHoveredRangeDayElements() {
+        const selectedRange = DateClasses.DATERANGEPICKER_DAY_HOVERED_RANGE;
+        return document.queryAll(`.${selectedRange}:not(.${DateClasses.DATEPICKER_DAY_OUTSIDE})`);
+    }
+
+    function getHoveredRangeStartDayElement() {
+        // use .query to return just the first element
```
leebyp: probably dont need a comment since we do this everywhere, non?

cmslewis: Deleted!

cmslewis:
@leebyp - Fixed hovering over disabled dates. Good catch!

cmslewis:
@giladgray @llorca @leebyp - Would love a CR on this when you get a moment. I already have Part 2 ready when this merges.

packages/datetime/test/dateRangePickerTests.tsx
```diff
@@ -311,6 +492,37 @@ describe("<DateRangePicker>", () => {
...
         });
 
+        it("onHoverChange fired with correct values when a day is clicked", () => {
+            const dateRange = [new Date(2015, Months.JANUARY, 1), new Date(2015, Months.JANUARY, 5)] as DateRange;
+            renderDateRangePicker({ initialMonth: new Date(2015, Months.JANUARY, 1) });
+            assert.isTrue(onDateRangePickerHoverChangeSpy.notCalled);
+            clickDay(1);
+            assert.isTrue(onDateRangePickerHoverChangeSpy.calledOnce);
+            assert.isTrue(DateUtils.areSameDay(dateRange[0], onDateRangePickerHoverChangeSpy.args[0][0][0]));
+            assert.isNull(onDateRangePickerHoverChangeSpy.args[0][0][1]);
```
leebyp: i know we do this elsewhere too, but is there a reason we dont use `calledWith`?

packages/datetime/src/dateRangePicker.tsx
```diff
@@ -56,6 +57,13 @@ export interface IDateRangePickerProps extends IDatePickerBaseProps, IProps {
     onChange?: (selectedDates: DateRange) => void;
 
     /**
+     * Called when the user changes the hovered date range, either from mouseenter or mouseleave.
+     * When triggered from mouseenter, it will pass the date range that would result from next click.
+     * When triggered from mouseleave, it will pass `[null, null]`.
```
leebyp: does it make sense that when im moving hovering between days, i keep getting `[null, null]` fired in between? genuine questions, not rhetoric

leebyp:
couple new comments, rest LGTM

packages/datetime/test/dateRangePickerTests.tsx
cmslewis: Good point. I copied this syntax from an existing test in this file, but `calledWith` seems semantically clearer to me in general.

packages/datetime/src/dateRangePicker.tsx
cmslewis: @leebyp After building the whole `DateRangeInput` prototype on another branch, I can say this approach works empirically despite its clunkiness. That said, I'd love to figure out a way around this. The problem is that `mouseleave` on the old hovered date triggers before `mouseenter` on the new hovered date—maybe we can cancel the invocation of `this.props.onHoverChange` in `handleDayMouseLeave` if a `mouseenter` happens within the next `X` milliseconds? That feels kinda gross though.

packages/datetime/src/_daterangepicker.scss
```diff
@@ -63,20 +63,45 @@ Styleguide components.datetime.daterangepicker.js
...
-  .DayPicker-Day--selected-range-start:not(.DayPicker-Day--outside):not(.DayPicker-Day--selected-range-end) {
+  .DayPicker-Day--selected-range-start:not(.DayPicker-Day--outside):not(.DayPicker-Day--selected-range-end):not(.DayPicker-Day--hovered-range-end),
+  .DayPicker-Day--hovered-range-start:not(.DayPicker-Day--outside):not(.DayPicker-Day--hovered-range-end) {
     border-top-right-radius: 0;
     border-bottom-right-radius: 0;
   }
 
-  .DayPicker-Day--selected-range-end:not(.DayPicker-Day--outside):not(.DayPicker-Day--selected-range-start) {
+  .DayPicker-Day--selected-range-end:not(.DayPicker-Day--outside):not(.DayPicker-Day--selected-range-start):not(.DayPicker-Day--hovered-range-start),
+  .DayPicker-Day--hovered-range-end:not(.DayPicker-Day--outside):not(.DayPicker-Day--hovered-range-start) {
```
giladgray: do we still need to care about `--outside` since we're hiding them?

packages/datetime/src/_daterangepicker.scss
```diff
@@ -63,20 +63,45 @@ Styleguide components.datetime.daterangepicker.js
     margin-right: $pt-grid-size;
   }
 
+  .DayPicker-Day--hovered-range:not(.DayPicker-Day--outside) {
+    border-radius: 0;
+
+    // need to disable hover-styles for all variants of selected dates
+    // stylelint-disable max-line-length
+    &:not(.DayPicker-Day--selected):not(.DayPicker-Day--selected-range):not(.DayPicker-Day--selected-range-start):not(.DayPicker-Day--selected-range-end) {
```
giladgray: yuck! why do you have to do this? 
i wonder if we can find another way (though perhaps in a separate PR).
can we just define this one first so the others will override by virtue of coming later?

packages/datetime/src/dateRangePicker.tsx
giladgray: how about passing `undefined` instead of `[null, null]`?

cmslewis: @giladgray - I'm all about it. Changing now.

packages/datetime/src/_daterangepicker.scss
cmslewis: Selected styles are defined in another file with less-specific selectors, so that makes things tricky. Can definitely return to this in a later PR to see if there's a better approach.

packages/datetime/src/_daterangepicker.scss
cmslewis: I included these before we made a final decision. It's low lift enough to include the styles for outside days just in case, but I'd also be happy to remove all outside-day styles in one fell swoop in a separate PR.

Issue #237, opened 2016-11-24
Input group right element button dancing


Reaverart:
### Bug Report

- __Package version(s)__: (1.1.0)
- __Browser and OS versions__: (chrome 54, osX 10.11.6)

#### Describe the issue:

If input group lays inside of control group a rightElement button moving on click. 
It happens because of this rule https://github.com/palantir/blueprint/blob/master/packages/core/src/components/forms/_control-group.scss#L130
So 'pt-input-action' becomes 0 width and button moves right, because it aligns to left if no side specified
```
<div class="pt-control-group">
  <div class="pt-input-group">
    <span class="pt-icon pt-icon-search"></span>
    <input type="text" placeholder="Search..." class="pt-input" style="padding-right: 30px;">
    <span class="pt-input-action">
      <button class="pt-button pt-icon-small-cross pt-intent-warning pt-minimal" type="button"></button>
    </span>
  </div>
</div>
```
#### Actual behavior:

button moves right
![repro](https://cloud.githubusercontent.com/assets/443450/20614990/ffe0c54e-b289-11e6-88c6-e62b71d4e9e7.gif)
_EDIT: GIF added by @cmslewis_

#### Expected behavior:

button should stay on same position

cmslewis:
Thanks for filing this bug, @Reaverart. Really appreciate the code sample too - definitely helps us move faster to repro the issue. Seems quick enough that we should hopefully get a fix in shortly.

cmslewis:
Fixed by #239

PR #592, opened 2017-02-01
Fix z-index stack for .pt-button-group and .pt-control-group elements

cmslewis:
#### Checklist

- [ ] Include tests (it would be awesome to add unit tests to verify the z-index stacking order, but we'd need [`chai-enzyme`](https://github.com/producthunt/chai-enzyme) to access computed styles for mounted React components - should we add it as a dependency?).
- [ ] Update documentation (@llorca - can you edit examples to your liking?)

- Clearly specify the z-index stack of all possible `.pt-button-group` elements to ensure sharp edges in all cases and states:

```
  (lowest z-index)

  default buttons
    :disabled, .pt-disabled
    normal
    :focus
    :hover
    :active, .pt-active

  intent buttons
    :disabled, .pt-disabled
    normal
    :focus
    :hover
    :active, .pt-active

  (highest z-index)
```

- Clearly specify the z-index stack of all possible `.pt-control-group` elements to ensure sharp edges in all cases and states:

```
  (lowest z-index)

  input fields
     :disabled, .pt-disabled
     normal

  regular buttons / pt-select > select
    :disabled, .pt-disabled
    normal
    :focus
    :hover
    :active, .pt-active

  pt-input intent

  intent buttons (selects can't be intented yet, but if they could be, they'd be here too)
    :disabled, .pt-disabled
    normal
    :focus
    :hover
    :active, .pt-active

  input fields (NOT readonly, :disabled, .pt-disabled)
     :focus

  intent input fields (NOT readonly, :disabled, .pt-disabled)
     :focus

  input-group children

  .pt-select::after (i.e. the caret)

  (highest z-index)
```

_Definitely look at the diff in "Split" mode, not "Unified" mode._

- How does the new z-index stack feel?
- Is it implemented correctly?
- Are we good with the z-index variable names? Thought about combining them, but the two group types had different enough stacking semantics that it was clearer to spell everything out separately. 
- Are there any weird edge cases with certain combinations of grouped elements?
- Are there any regressions?

### DIFF
--- a/packages/core/src/components/button/_button-group.scss
+++ b/packages/core/src/components/button/_button-group.scss
@@ -126,18 +126,8 @@ Styleguide components.button-group.css
 //
 //   (highest z-index)
 
-$-button-group-z-index-stack:
-  "button-disabled",
-  "button-default",
-  "button-focus",
-  "button-hover",
-  "button-active",
-
-  "intent-button-disabled",
-  "intent-button-default",
-  "intent-button-focus",
-  "intent-button-hover",
-  "intent-button-active";
+// boost all states of intent buttons above all states of default buttons
+$button-group-intent-boost: length($button-state-stack);
 
 .pt-button-group {
   display: inline-flex;
@@ -146,7 +136,7 @@ $-button-group-z-index-stack:
     // ensure button can never be smaller than its default size
     flex: 0 0 auto;
     position: relative;
-    z-index: index($button-group-z-index-stack, "button-default");
+    z-index: index($button-state-stack, "default");
 
     // the ordering of these z-index CSS rules is particular because of CSS
     // selector specificity. specifically, disabled styles should have
@@ -154,42 +144,42 @@ $-button-group-z-index-stack:
     // disabled buttons from reordering elements in the stack.
 
     &:focus {
-      z-index: index($button-group-z-index-stack, "button-focus");
+      z-index: index($button-state-stack, "focus");
     }
 
     &:hover {
-      z-index: index($button-group-z-index-stack, "button-hover");
+      z-index: index($button-state-stack, "hover");
     }
 
     &:active,
     &.pt-active {
-      z-index: index($button-group-z-index-stack, "button-active");
+      z-index: index($button-state-stack, "active");
     }
 
     &:disabled,
     &.pt-disabled {
-      z-index: index($button-group-z-index-stack, "button-disabled");
+      z-index: index($button-state-stack, "disabled");
     }
 
     &[class*="pt-intent-"] {
-      z-index: index($button-group-z-index-stack, "intent-button-default");
+      z-index: index($button-state-stack, "default") + $button-group-intent-boost;
 
       &:focus {
-        z-index: index($button-group-z-index-stack, "intent-button-focus");
+        z-index: index($button-state-stack, "focus") + $button-group-intent-boost;
       }
 
       &:hover {
-        z-index: index($button-group-z-index-stack, "intent-button-hover");
+        z-index: index($button-state-stack, "hover") + $button-group-intent-boost;
       }
 
       &:active,
       &.pt-active {
-        z-index: index($button-group-z-index-stack, "intent-button-active");
+        z-index: index($button-state-stack, "active") + $button-group-intent-boost;
       }
 
       &:disabled,
       &.pt-disabled {
-        z-index: index($button-group-z-index-stack, "intent-button-disabled");
+        z-index: index($button-state-stack, "disabled") + $button-group-intent-boost;
       }
     }
   }
--- a/packages/core/src/components/button/_button-group.scss
+++ b/packages/core/src/components/button/_button-group.scss
@@ -126,17 +126,18 @@ Styleguide components.button-group.css
 //
 //   (highest z-index)
 
-$button-group-z-index-button-disabled: 1;
-$button-group-z-index-button-default: 2;
-$button-group-z-index-button-focus: 3;
-$button-group-z-index-button-hover: 4;
-$button-group-z-index-button-active: 5;
+$-button-group-z-index-stack:
+  "button-disabled",
+  "button-default",
+  "button-focus",
+  "button-hover",
+  "button-active",
 
-$button-group-z-index-intent-button-disabled: 6;
-$button-group-z-index-intent-button-default: 7;
-$button-group-z-index-intent-button-focus: 8;
-$button-group-z-index-intent-button-hover: 9;
-$button-group-z-index-intent-button-active: 10;
+  "intent-button-disabled",
+  "intent-button-default",
+  "intent-button-focus",
+  "intent-button-hover",
+  "intent-button-active";
 
 .pt-button-group {
   display: inline-flex;
@@ -145,7 +146,7 @@ $button-group-z-index-intent-button-active: 10;
     // ensure button can never be smaller than its default size
     flex: 0 0 auto;
     position: relative;
-    z-index: $button-group-z-index-button-default;
+    z-index: index($button-group-z-index-stack, "button-default");
 
     // the ordering of these z-index CSS rules is particular because of CSS
     // selector specificity. specifically, disabled styles should have
@@ -153,42 +154,42 @@ $button-group-z-index-intent-button-active: 10;
     // disabled buttons from reordering elements in the stack.
 
     &:focus {
-      z-index: $button-group-z-index-button-focus;
+      z-index: index($button-group-z-index-stack, "button-focus");
     }
 
     &:hover {
-      z-index: $button-group-z-index-button-hover;
+      z-index: index($button-group-z-index-stack, "button-hover");
     }
 
     &:active,
     &.pt-active {
-      z-index: $button-group-z-index-button-active;
+      z-index: index($button-group-z-index-stack, "button-active");
     }
 
     &:disabled,
     &.pt-disabled {
-      z-index: $button-group-z-index-button-disabled;
+      z-index: index($button-group-z-index-stack, "button-disabled");
     }
 
     &[class*="pt-intent-"] {
-      z-index: $button-group-z-index-intent-button-default;
+      z-index: index($button-group-z-index-stack, "intent-button-default");
 
       &:focus {
-        z-index: $button-group-z-index-intent-button-focus;
+        z-index: index($button-group-z-index-stack, "intent-button-focus");
       }
 
       &:hover {
-        z-index: $button-group-z-index-intent-button-hover;
+        z-index: index($button-group-z-index-stack, "intent-button-hover");
       }
 
       &:active,
       &.pt-active {
-        z-index: $button-group-z-index-intent-button-active;
+        z-index: index($button-group-z-index-stack, "intent-button-active");
       }
 
       &:disabled,
       &.pt-disabled {
-        z-index: $button-group-z-index-intent-button-disabled;
+        z-index: index($button-group-z-index-stack, "intent-button-disabled");
       }
     }
   }
--- a/packages/core/src/components/button/_button-group.scss
+++ b/packages/core/src/components/button/_button-group.scss
@@ -56,6 +56,40 @@ Markup:
 Styleguide components.button-group.css
 */
 
+//
+// for best visual results, button group elements should be stacked in the
+// following order to ensure sharp edges in all cases and states:
+//
+//   (lowest z-index)
+//
+//   default buttons
+//     :disabled, .pt-disabled
+//     normal
+//     :focus
+//     :hover
+//     :active, .pt-active
+//
+//   intent buttons
+//     :disabled, .pt-disabled
+//     normal
+//     :focus
+//     :hover
+//     :active, .pt-active
+//
+//   (highest z-index)
+
+$button-group-z-index-button-disabled: 1;
+$button-group-z-index-button-default: 2;
+$button-group-z-index-button-focus: 3;
+$button-group-z-index-button-hover: 4;
+$button-group-z-index-button-active: 5;
+
+$button-group-z-index-intent-button-disabled: 6;
+$button-group-z-index-intent-button-default: 7;
+$button-group-z-index-intent-button-focus: 8;
+$button-group-z-index-intent-button-hover: 9;
+$button-group-z-index-intent-button-active: 10;
+
 .pt-button-group {
   display: inline-flex;
 
@@ -63,45 +97,51 @@ Styleguide components.button-group.css
     // ensure button can never be smaller than its default size
     flex: 0 0 auto;
     position: relative;
-    z-index: 0;
+    z-index: $button-group-z-index-button-default;
 
-    // the ordering of these z-index CSS rules is particular
-    // because of CSS selector specificity
+    // the ordering of these z-index CSS rules is particular because of CSS
+    // selector specificity. specifically, disabled styles should have
+    // precedence over hover and active styles to prevent mouse interactions on
+    // disabled buttons from reordering elements in the stack.
+
     &:focus {
-      z-index: 1;
+      z-index: $button-group-z-index-button-focus;
     }
 
     &:hover {
-      z-index: 2;
-    }
-
-    // intent buttons are always above any other button state
-    &[class*="pt-intent-"] {
-      z-index: 4;
-
-      &:hover {
-        z-index: 5;
-      }
-
-      &:active,
-      &.pt-active {
-        z-index: 6;
-      }
-
-      &:disabled,
-      &.pt-disabled {
-        z-index: 3;
-      }
+      z-index: $button-group-z-index-button-hover;
     }
 
     &:active,
     &.pt-active {
-      z-index: 3;
+      z-index: $button-group-z-index-button-active;
     }
 
     &:disabled,
     &.pt-disabled {
-      z-index: 0;
+      z-index: $button-group-z-index-button-disabled;
+    }
+
+    &[class*="pt-intent-"] {
+      z-index: $button-group-z-index-intent-button-default;
+
+      &:focus {
+        z-index: $button-group-z-index-intent-button-focus;
+      }
+
+      &:hover {
+        z-index: $button-group-z-index-intent-button-hover;
+      }
+
+      &:active,
+      &.pt-active {
+        z-index: $button-group-z-index-intent-button-active;
+      }
+
+      &:disabled,
+      &.pt-disabled {
+        z-index: $button-group-z-index-intent-button-disabled;
+      }
     }
   }
 
--- a/packages/core/src/components/button/_button-group.scss
+++ b/packages/core/src/components/button/_button-group.scss
@@ -54,54 +54,6 @@ Markup:
 .pt-minimal - Use minimal buttons. Note that these minimal buttons will not automatically
               truncate text in an ellipsis because of the divider line added in CSS.
 
-<div class="docs-button-group-example-set">
-  <h4>Regular</h4>
-  <div class="pt-button-group">
-    <button tabindex="0" class="pt-button" disabled>:disabled</button>
-    <a tabindex="0" class="pt-button" role="button">Normal</a>
-  </div>
-  <div class="pt-button-group">
-    <a tabindex="0" class="pt-button pt-disabled" role="button">.pt-disabled</a>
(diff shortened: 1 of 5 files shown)


packages/core/src/components/forms/_control-group.scss
```diff
@@ -76,69 +174,126 @@ Styleguide components.forms.control-group
...
+  // similarly to button groups, elements in control groups are stacked in a
+  // very particular order for best visual results. in each level of selector
+  // specificity, we define disabled styles last so that they override all other
+  // equally-specific styles (e.g. we don't want mouse interactions or focus
+  // changes to change the z-index of a disabled element).
+
   .pt-button,
   .pt-input,
-  .pt-select select {
+  .pt-select {
```
cmslewis: Moved from old line 107.

packages/core/src/components/forms/_control-group.scss
```diff
@@ -76,69 +174,126 @@ Styleguide components.forms.control-group
...
     border-radius: inherit;
 
     &:focus {
-      z-index: 1;
+      z-index: $control-group-z-index-input-focus;
+      border-radius: $pt-border-radius;
     }
-  }
 
-  &:not(.pt-vertical) > * {
```
cmslewis: Moved this block to the bottom of the file to keep z-index-related stuff in one clump.

packages/core/src/components/forms/_control-group.scss
```diff
@@ -76,69 +174,126 @@ Styleguide components.forms.control-group
...
-    margin-right: -$button-border-width;
-  }
+    // define disabled styles last so that they override all other
+    // equally-specific styles.
```
cmslewis: Will delete one of these explanations (it's also explained in the top-level comment above).

packages/core/src/components/forms/_control-group.scss
```diff
@@ -185,11 +340,4 @@ Styleguide components.forms.control-group
       border-radius: 0 0 $pt-border-radius $pt-border-radius;
     }
   }
-
-  .pt-dark & {
```
cmslewis: Moved to line 212 to keep z-index-related rules in one clump.

packages/core/src/components/forms/_control-group.scss
```diff
@@ -76,69 +174,126 @@ Styleguide components.forms.control-group
...
+    // define disabled styles last so that they override all other
+    // equally-specific styles.
+    [readonly],
+    :disabled,
+    .pt-disabled {
+      z-index: $control-group-z-index-input-disabled;
+      border-radius: 0;
+    }
 
-  > :first-child {
```
cmslewis: Moved to line 289.

packages/core/src/components/forms/_control-group.scss
```diff
@@ -76,69 +174,126 @@ Styleguide components.forms.control-group
...
-    border-radius: $pt-border-radius 0 0 $pt-border-radius;
-  }
+    &[class*="pt-intent"] {
+      z-index: $control-group-z-index-intent-input-default;
 
-  > :last-child {
```
cmslewis: Moved to line 294.

packages/core/src/components/forms/_control-group.scss
```diff
@@ -76,69 +174,126 @@ Styleguide components.forms.control-group
...
 
+  // similarly to button groups, elements in control groups are stacked in a
+  // very particular order for best visual results. in each level of selector
+  // specificity, we define disabled styles last so that they override all other
+  // equally-specific styles (e.g. we don't want mouse interactions or focus
+  // changes to change the z-index of a disabled element).
+
   .pt-button,
   .pt-input,
-  .pt-select select {
```
cmslewis: Moved to line 226.

packages/core/src/components/forms/_control-group.scss
```diff
@@ -76,69 +174,126 @@ Styleguide components.forms.control-group
...
-      z-index: 3;
+    &:focus {
+      // establish new stacking context so focus state covers neighbors
+      position: relative;
+      z-index: $control-group-z-index-button-focus;
     }
-  }
 
-  .pt-input {
-    position: relative;
```
cmslewis: Moved to the top of the file.

packages/core/src/components/forms/_control-group.scss
```diff
@@ -76,69 +174,126 @@ Styleguide components.forms.control-group
...
+        z-index: $control-group-z-index-intent-input-focus;
+      }
+    }
   }
 
-  // similarly to button groups, elements in control groups are stacked
-  // in a very particular order for best visual results
   .pt-button,
-  .pt-select {
-    position: relative;
```
cmslewis: Moved to the top of the file.

packages/core/src/components/forms/_control-group.scss
```diff
@@ -76,69 +174,126 @@ Styleguide components.forms.control-group
...
 
-    &[class*="pt-intent-"] {
-      z-index: 3;
+    &:active {
+      z-index: $control-group-z-index-button-active;
     }
 
-    &:not([readonly], :disabled, .pt-disabled):focus {
-      z-index: 6;
-      border-radius: $pt-border-radius;
```
cmslewis: Moved to line 197.

packages/core/src/components/forms/_control-group.scss
```diff
@@ -76,69 +174,126 @@ Styleguide components.forms.control-group
...
+    &[readonly],
+    &:disabled,
+    &.pt-disabled {
+      z-index: $control-group-z-index-button-disabled;
     }
-  }
 
-  // establish new stacking context so focus state covers neighbors
-  .pt-button:focus,
-  select:focus {
```
cmslewis: Now handled with higher z-indexes in the same stacking context.

packages/core/src/components/forms/_control-group.scss
```diff
@@ -76,69 +174,126 @@ Styleguide components.forms.control-group
...
   .pt-input-group > .pt-button,
-  .pt-input-group > .pt-input-action,
+  .pt-input-group > .pt-input-action {
+    z-index: $control-group-z-index-input-group-children;
+  }
+
+  // keep the select-menu carets on top of everything always
   .pt-select::after {
-    z-index: 7;
+    z-index: $control-group-z-index-select-caret;
```
cmslewis: Moved the carets to a z-index all their own that explicitly sits them on top of everything, just to be absolutely safe.

packages/docs/src/styles/_sections.scss
```diff
@@ -54,7 +61,7 @@
   display: flex;
   flex-direction: column;
   justify-content: space-between;
-  height: $pt-grid-size * 13;
+  height: $pt-grid-size * 50;
```
cmslewis: Feels weird that we have to explicitly set a height. Why not just have docs containers expand and contract automatically with the height of the child contents?

adidahiya:
Not opposed to adding chai-enzyme as a devDependency. Sounds like it could be useful.

packages/docs/src/styles/_sections.scss
giladgray: they should do that... casual flex box. why do you have to set a height?

packages/core/src/components/forms/_control-group.scss
```diff
@@ -76,69 +174,126 @@ Styleguide components.forms.control-group
...
+  }
+
+  // have consecutive elements share a border
+  &:not(.pt-vertical) > * {
+    margin-right: -$button-border-width;
```
llorca: That's actually on true for light theme.

The visuals look different in dark theme, depending on whether you use a button group or single buttons for consecutive buttons (top is bad, bottom is good):
image

So we can keep that line for the light theme, but we need to take care of dark theme. Proposal (to be refactor in the appropriate place):

```
.pt-dark &:not(.pt-vertical) {
  > * {
    margin-right: 0;
  }

  > .pt-button + .pt-button {
    margin-left: $button-border-width;
  }
}
```

packages/core/src/components/forms/_control-group.scss
```diff
@@ -66,6 +102,68 @@ Weight: 4
...
+//   intent input fields (NOT readonly, disabled, pt-disabled)
+//      :focus
+//
+//   input-group children
+//
+//   .pt-select::after (the caret)
+//
+//   (highest z-index)
+
+$control-group-z-index-input-disabled: 1;
```
llorca: 2 Qs about maintainability:

- if our visual design changes and we ever need to update the stacking of elements in control groups, we can simply modify the values of these variables, right? No need to mess with the selectors and CSS specificities below?

- these variables wouldn't get exported, and I'm pretty sure they wouldn't be used anywhere else. Maybe we should use a map object instead of single variables for every value?

packages/docs/src/styles/_sections.scss
```diff
@@ -38,6 +38,13 @@
 
 // Section-Specific Styles
 
+#{section-id("components.button-group")} {
+  .docs-button-group-example-set .pt-button-group {
+    margin-top: 20px;
```
llorca: `$pt-grid-size * 2`

packages/docs/src/styles/_sections.scss
llorca: It's already the case. Setting a height here will evenly space out the children so there's spacing in between them. Without this line:
image

So what we can do instead is remove this line and add an additional rule in the same block:
```
.pt-control-group-example {
  > .pt-control.group:not(:last-child) {
    margin-bottom: $pt-grid-size * 2;
  }
}
```

packages/core/src/components/forms/_control-group.scss
cmslewis: Fixed and incorporated as follows:

```

// have consecutive elements share a border
&:not(.pt-vertical) {
  > * {
    margin-right: -$button-border-width;
  }

  .pt-dark & {
    > * {
      margin-right: 0;
    }

    // conseuctive buttons within a button group look okay out of the box, but
    // we need need to make a small correction for non-grouped buttons
    > .pt-button + .pt-button {
      margin-left: $button-border-width;
    }
  }
}
```

packages/docs/src/styles/_sections.scss
cmslewis: Fixed

packages/core/src/components/forms/_control-group.scss
llorca: Great! For clarity, maybe also mention that we're basically replicating what's happening in dark button groups?

cmslewis: Done 👍

packages/docs/src/styles/_sections.scss
cmslewis: Derp, don't know why I didn't think to try that. Thanks, @llorca.

packages/core/src/components/forms/_control-group.scss
cmslewis: - Yep, just change the z-indexes! 

- Moving everything into a list so we can just use the `index()` helper.

packages/core/src/components/button/_button-group.scss
```diff
@@ -53,56 +53,145 @@ Markup:
...
+  "button-default",
+  "button-focus",
+  "button-hover",
+  "button-active",
+
+  "intent-button-disabled",
+  "intent-button-default",
+  "intent-button-focus",
+  "intent-button-hover",
+  "intent-button-active";
```
giladgray: i see two duplicated lists here. what happened to a single `intent-boost` variable?

then you go `index(list, "disabled") + $intent-boost` (pseudocode).

giladgray: then you could create a mixin that wraps up lines 151-172 and reuse it in the intent block by passing a single fixed "booster" value. `@include z-index-state-layers($boost)`

cmslewis: This was meant to be just an incremental change.

packages/core/src/components/button/_button-group.scss
```diff
@@ -53,56 +53,145 @@ Markup:
...
+//   intent buttons
+//     :disabled, .pt-disabled
+//     normal
+//     :focus
+//     :hover
+//     :active, .pt-active
+//
+//   (highest z-index)
+
+$-button-group-z-index-stack:
```
llorca: Remove extra `-` before `button`. Also, can we just go for `$button-group-stack` or something simpler? Is `z-index` really necessary in the name?

packages/core/src/components/button/_button-group.scss
cmslewis: @giladgray chatted and worked through a potential refactor with Antoine. The `intent-boost` approach works well for button groups, but it's more tricky than helpful in control groups, where differently ordered `.pt-input` layers get in the way.

One approach we're considering now is to use a single list of named layers that is shared between control groups and button groups, since the relative ordering of elements doesn't actually change between the two contexts.

llorca: 👍 TLDR: 1 list with everything (`$control-group-stack` or something), and button groups use a subset of that list via `index()`

packages/core/src/components/button/_button-group.scss
cmslewis: @llorca Fixed. I ended up creating a single `$control-group-stack` variable, putting it in `forms/_common.scss`, and importing it into `_control-group.scss` and `_button-group.scss`. LMK how that looks.

packages/core/src/components/forms/_control-group.scss
```diff
@@ -76,69 +113,135 @@ Styleguide components.forms.control-group
...
+    > * {
+      margin-right: -$button-border-width;
+    }
+
+    .pt-dark & {
+      > * {
+        margin-right: 0;
+      }
+
+      // conseuctive buttons within a button group look okay out of the box, but
```
llorca: - consecutive
- dark theme

llorca:
I'm a fan of this approach! Much cleaner, less code, 1 reusable list. Big comment explanation no longer needed, code speaks for itself 💯

packages/core/src/components/forms/_control-group.scss
```diff
@@ -30,34 +31,70 @@ child must be marked individually as `.pt-large` for uniform large appearance.
...
-        <option value="2">Requests</option>
-        <option value="3">Projects</option>
-      </select>
-    </div>
     <div class="pt-input-group">
-      <span class="pt-icon pt-icon-search"></span>
-      <input type="text" class="pt-input" value="from:ggray to:allorca" />
+      <span class="pt-icon pt-icon-people"></span>
+      <input type="text" class="pt-input pt-intent-success" placeholder="I am a text input..." style="padding-right:94px" />
+      <div class="pt-input-action">
```
giladgray: `.pt-input-action` shouldn't be needed when wrapping a `.pt-button`. `InputGroup` always uses it because it can't know about the contents but should be safe to remove in these static HTML examples.

packages/core/src/components/forms/_control-group.scss
```diff
@@ -30,34 +31,70 @@ child must be marked individually as `.pt-large` for uniform large appearance.
...
     </div>
+    <button class="pt-button pt-intent-warning">Button</button>
+    <button class="pt-button pt-intent-success">Button</button>
+    <button class="pt-button">Button</button>
   </div>
   <div class="pt-control-group">
     <div class="pt-input-group">
       <span class="pt-icon pt-icon-people"></span>
-      <input type="text" class="pt-input" placeholder="Find collaborators..." style="padding-right:94px" />
+      <input type="text" class="pt-input pt-intent-success" placeholder="I am a text input..." style="padding-right:94px" />
```
giladgray: why the placeholder change? kinda breaks the example metaphor

packages/core/src/components/forms/_control-group.scss
```diff
@@ -30,34 +31,70 @@ child must be marked individually as `.pt-large` for uniform large appearance.
...
           can view<span class="pt-icon-standard pt-icon-caret-down pt-align-right"></span>
         </button>
       </div>
     </div>
-    <button class="pt-button pt-intent-primary">Add</button>
+    <div class="pt-button-group">
+      <button class="pt-button pt-intent-warning">Button</button>
+      <button class="pt-button pt-intent-success">Button</button>
+      <button class="pt-button">Button</button>
+    </div>
```
giladgray: @cmslewis are we planning to ship all these example changes? on one hand, they're useful for testing, but on the other they're somewhat unsightly and are usages that we may not want to promote so visibly.

packages/core/src/components/forms/_common.scss
```diff
@@ -26,6 +26,33 @@ $dark-input-shadow-color-focus: $pt-intent-primary !default;
...
+  "intent-button-focus",
+  "intent-button-hover",
+  "intent-button-active",
+
+  "intent-input-default",
+  "input-focus",
+  "intent-input-focus",
+  "input-group-children",
+
+  "select-caret";
```
giladgray: @cmslewis planning to remove this in a future increment?

giladgray:
⭐️ the single stack variable approach works nicely!

packages/core/src/components/forms/_control-group.scss
llorca: On that note, the 3 examples we currently serve in the docs are more interesting indeed. Once we're confident everything looks good, let's bring back those examples, and maybe we can think of additional 1 or 2 to showcase as well

packages/core/src/components/forms/_control-group.scss
cmslewis: Ha, fixed.

packages/core/src/components/forms/_control-group.scss
cmslewis: Fixed. 👍

packages/core/src/components/forms/_control-group.scss
cmslewis: In the testing phase, it was tricky to tell at a glance what was a button and what was an input when either was disabled:

image

Explicit placeholder text helped with that. Totally not wedded to these examples; I'd prefer CSS unit tests in this instance.

packages/core/src/components/forms/_control-group.scss
cmslewis: Yep, totally down to revert to our prior examples before merging.

packages/core/src/components/forms/_control-group.scss
cmslewis: Just kiddingggg, I need to revert. Forgot that we need these `.pt-input-action` wrappers in all cases to address issues like #237. If you delete the `.pt-input-action` wrapper, you get a weird flicker on focus.

packages/core/src/components/forms/_common.scss
cmslewis: Removed!

cmslewis:
@giladgray and @llorca. LMK if this looks good to you now; if so, I'll revert to the previous examples and merge.

packages/core/src/components/forms/_common.scss
```diff
@@ -26,6 +26,31 @@ $dark-input-shadow-color-focus: $pt-intent-primary !default;
 // second box-shadow of $pt-input-box-shadow
 $input-box-shadow-focus: inset 0 1px 1px rgba($black, $pt-drop-shadow-opacity) !default;
 
+// for best visual results, button group and control group elements should be
+// stacked in the following order to ensure sharp edges in all cases and states:
+
+$control-group-stack:
```
cmslewis: Sweet. Doing so now.

cmslewis:
@giladgray Oh no! When I reverted the examples to the originals, I found a caret bug on focus:

image

Having `position: relative` on `.pt-select` doesn't help; looks like we need that `select-caret` layer in `$control-group-stack` after all. Happy to revisit in a future PR though.

packages/core/src/components/forms/_control-group.scss
```diff
@@ -76,69 +77,136 @@ Styleguide components.forms.control-group
     flex: 0 0 auto;
   }
 
+  // similarly to button groups, elements in control groups are stacked in a
+  // very particular order for best visual results. in each level of selector
+  // specificity, we define disabled styles last so that they override all other
+  // equally-specific styles (e.g. we don't want mouse interactions or focus
+  // changes to change the z-index of a disabled element).
```
giladgray: does it make sense to use `!important` for disabled styles like @leebyp is doing in #565?

PR #609, opened 2017-02-02
[Date Range Input] - Part 2: Basic Input HTML/CSS

cmslewis:

Implemented the basic HTML and CSS for the `DateRangeInput` (without the popover).

- I defined an empty `IDateRangeInputProps` interface for this PR. Will add prop definitions over time as I implement them.
- ~I've set Part 1 as the base branch until that PR merges into `feature/date-range-input`.~
- The unit test here is just a placeholder; it's present to ensure this PR meets the minimum coverage threshold.

### DIFF
--- a/packages/datetime/src/common/classes.ts
+++ b/packages/datetime/src/common/classes.ts
@@ -17,6 +17,10 @@ export const DATEPICKER_FOOTER = "pt-datepicker-footer";
 export const DATEPICKER_MONTH_SELECT = "pt-datepicker-month-select";
 export const DATEPICKER_YEAR_SELECT = "pt-datepicker-year-select";
 
+export const DATERANGEINPUT = "pt-daterangeinput";
+export const DATERANGEINPUT_FIELD = "pt-daterangeinput-field";
+export const DATERANGEINPUT_TRIGGER = "pt-daterangeinput-trigger";
+
 export const DATERANGEPICKER = "pt-daterangepicker";
 export const DATERANGEPICKER_DAY_SELECTED_RANGE = "DayPicker-Day--selected-range";
 export const DATERANGEPICKER_DAY_HOVERED_RANGE = "DayPicker-Day--hovered-range";
--- a/packages/datetime/src/common/classes.ts
+++ b/packages/datetime/src/common/classes.ts
@@ -19,6 +19,9 @@ export const DATEPICKER_YEAR_SELECT = "pt-datepicker-year-select";
 
 export const DATERANGEPICKER = "pt-daterangepicker";
 export const DATERANGEPICKER_DAY_SELECTED_RANGE = "DayPicker-Day--selected-range";
+export const DATERANGEPICKER_DAY_HOVERED_RANGE = "DayPicker-Day--hovered-range";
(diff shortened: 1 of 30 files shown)


packages/docs/src/styles/_examples.scss
```diff
@@ -65,7 +65,7 @@ $options-margin: $pt-grid-size * 3;
     margin-bottom: $options-margin;
   }
 
-  .pt-input-group + .pt-input-group {
+  :not(.pt-control-group) .pt-input-group + .pt-input-group {
```
cmslewis: The previous global examples style broke the `DateRangeInput`, because the component includes two sibling `.pt-input-group`s:

image

Needed to tweak the global style to fix this.

packages/datetime/src/common/classes.ts
```diff
@@ -17,6 +17,10 @@ export const DATEPICKER_FOOTER = "pt-datepicker-footer";
 export const DATEPICKER_MONTH_SELECT = "pt-datepicker-month-select";
 export const DATEPICKER_YEAR_SELECT = "pt-datepicker-year-select";
 
+export const DATERANGEINPUT = "pt-daterangeinput";
+export const DATERANGEINPUT_FIELD = "pt-daterangeinput-field";
```
cmslewis: `pt-daterangeinput-input` is more correct, but it sounds a little silly. Happy to change to that if we think it's better though.

llorca:
What do people think about removing the icon on the right?

packages/datetime/src/common/classes.ts
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015 Palantir Technologies, Inc. All rights reserved.
+ * Copyright 2017 Palantir Technologies, Inc. All rights reserved.
```
giladgray: we're not in the practice of updating copyright years. @adidahiya comments?

packages/datetime/src/dateRangeInput.tsx
```diff
@@ -0,0 +1,69 @@
...
+                    className={DateClasses.DATERANGEINPUT_FIELD}
+                    inputRef={this.setStartDateInputRef}
+                    placeholder="Start date"
+                    type="text"
+                />
+                <InputGroup
+                    className={DateClasses.DATERANGEINPUT_FIELD}
+                    inputRef={this.setEndDateInputRef}
+                    placeholder="End date"
+                    type="text"
```
giladgray: how about `leftInputProps` and `rightInputProps` for passing arbitrary props to these two fields? that'll cover like 1000 bases ⚾️.

giladgray: maybe `start/end` instead of `left/right`

cmslewis:
This PR got in a bad state. Closing this and opening a new one.

packages/datetime/src/dateRangeInput.tsx
cmslewis: Sounds good. Doing this in a new PR.

cmslewis:
> What do people think about removing the icon on the right?

Let's chat about this in the new PR.

cmslewis:
See #614 for a second attempt at this PR.

PR #613, opened 2017-02-03
Wordsmith docs overview page

adidahiya:


cmslewis:
Let's edit the repo description in Github too?

image

adidahiya:
@cmslewis updated the landing page. I'll update the repo description when this merges.

packages/docs/src/styleguide.md
```diff
@@ -36,8 +38,8 @@ relevant sections in the sidebar for details about the styling & component APIs
 ### TypeScript
 
 Blueprint is written in [TypeScript](https://www.typescriptlang.org/), a statically typed superset
-of JavaScript that compiles to plain JavaScript. Many of the code samples throughout this site and
-all interactive examples are also written in TypeScript. TypeScript code looks exactly like ES2015
+of JavaScript that compiles to plain JavaScript. All the code samples throughout this site and
+all interactive examples are also written in TypeScript. TypeScript code looks exactly like ES2015+
```
giladgray: whats the + for?

packages/docs/src/styleguide.md
```diff
@@ -1,16 +1,18 @@
...
-Development occurs in [palantir/blueprint](https://github.com/palantir/blueprint).
-Please file issues on the [GitHub issue tracker](https://github.com/palantir/blueprint/issues).
+Development and issue tracking occurs in [github.com/palantir/blueprint](https://github.com/palantir/blueprint).
 
-Releases are tagged and documented [here](https://github.com/palantir/blueprint/releases)
+Use the [__blueprintjs__ tag on Stack Overflow](http://stackoverflow.com/questions/tagged/blueprintjs)
+for support requests.
+
+Releases are tagged and documented [here on Github](https://github.com/palantir/blueprint/releases)
 (more info about releases [on the wiki](https://github.com/palantir/blueprint/wiki/Releases-&-Artifacts)).
```
giladgray: pretty sure this wiki page does not exist

packages/docs/src/styleguide.md
adidahiya: because there's ES2016, ES2017...

packages/docs/src/styleguide.md
adidahiya: Not part of this PR

packages/docs/src/styleguide.md
```diff
@@ -1,16 +1,18 @@
-Blueprint is a React UI toolkit for the web.
+Blueprint is a React-based UI toolkit for the web.
```
llorca: I think it's way cleaner without the verboseness of the "-based" (not really necessary). If you feel strongly, I'm cool with here, but I prefer keeping the tagline everywhere as is for now. Note that changing it in this file only isn't enough, there are other places where it appears on GitHub, the landing, different services... Let's wait for 2.0 please, and we can work out and even better one.

packages/docs/src/styleguide.md
llorca: It exists internally, I looked at it, seems pretty unnecessary and out-of-date. Let's remove this now?

packages/docs/src/styleguide.md
```diff
@@ -23,9 +25,9 @@ Don't forget to include the main CSS stylesheet too!
 
 ### Beyond core styles
 
-The Blueprint team maintains multiple packages of styles and JavaScript components on NPM (in the scope
-`@blueprintjs`). These are listed in the header under _Releases_. They can be installed with
-similar commands as above. Once you've included any additional package(s) in your application, explore the
+Blueprint is a collection of packages of styles and JavaScript components&mdash;the full list appears in this
```
llorca: First read I thought the "full list" referred to the "components" and not the "packages". Suggestion: `the full list appears` -> `those packages appear`?

packages/docs/src/styleguide.md
giladgray: yeah that's what i thought, just looked like a typo at first

packages/docs/src/styleguide.md
llorca: Looks like I'm the only one thinking like this, so nevermind.

Can you make sure you caught all instances of that tagline and that it's changed everywhere?

PR #614, opened 2017-02-03
[Date Range Input] - Part 2: Basic HTML/CSS, example, and unit test

cmslewis:
> Second attempt at PR #609, which got in a bad state

Implemented the basic HTML and CSS for the `DateRangeInput` (without the popover), plus an example and a unit test.

- The unit test here is just a placeholder; it's present to ensure this PR meets the minimum coverage threshold.
- @llorca and I decided to nix (read: not implement) the calendar icon for now in preparation for nixing the `openOnFocus` prop eventually. That prop doesn't really make sense on this component, since this input wouldn't be particularly usable sans popover.

#### Screenshot

image

image

### DIFF
--- a/packages/datetime/examples/dateRangeInputExample.tsx
+++ b/packages/datetime/examples/dateRangeInputExample.tsx
@@ -12,8 +12,6 @@ import { DateRangeInput } from "../src";
 
 export class DateRangeInputExample extends BaseExample<{}> {
     protected renderExample() {
-        return (
-            <DateRangeInput />
-        );
+        return <DateRangeInput />;
     }
 }
--- a/packages/datetime/examples/dateRangeInputExample.tsx
+++ b/packages/datetime/examples/dateRangeInputExample.tsx
@@ -1,0 +1,19 @@
+/*
+ * Copyright 2017 Palantir Technologies, Inc. All rights reserved.
+ * Licensed under the BSD-3 License as modified (the “License”); you may obtain a copy
+ * of the license at https://github.com/palantir/blueprint/blob/master/LICENSE
+ * and https://github.com/palantir/blueprint/blob/master/PATENTS
+ */
+
+import BaseExample from "@blueprintjs/core/examples/common/baseExample";
+import * as React from "react";
+
+import { DateRangeInput } from "../src";
+
+export class DateRangeInputExample extends BaseExample<{}> {
+    protected renderExample() {
+        return (
+            <DateRangeInput />
+        );
+    }
+}
--- a/packages/datetime/src/_daterangeinput.scss
+++ b/packages/datetime/src/_daterangeinput.scss
@@ -1,0 +1,46 @@
+// Copyright 2017 Palantir Technologies, Inc. All rights reserved.
+// Licensed under the BSD-3 License as modified (the “License”); you may obtain a copy
+// of the license at https://github.com/palantir/blueprint/blob/master/LICENSE
+// and https://github.com/palantir/blueprint/blob/master/PATENTS
+
+@import "../../core/src/common/variables";
+
+/*
+Date range input
+
+@react-example DateRangeInputExample
+
+Styleguide components.datetime.daterangeinput
+*/
+
+.pt-daterangeinput {
+  position: relative;
+}
+
+.pt-daterangeinput-icon-wrapper {
+
+  // .pt-input-group already styles immediate-children .pt-buttons with
+  // these position styles. however, the .pt-button is a grandchild of the
+  // .pt-input-group in this case, so we need to restate the styles here.
+  .pt-button {
+    position: absolute;
+    top: 0;
+
+    // .pt-control-group styles .pt-button with position: relative when the
+    // button gains focus. that causes the button to move down on click in this
+    // case, so we disable that behavior.
+    &:focus {
+      position: absolute;
+    }
+  }
+
+  // we're applying the .pt-input class to a div here, not a proper <input>. the
+  // .pt-button above is the only child and is absolutely positioned, so we need
+  // to explicitly specify the size of the .pt-input to be a square. we also
+  // need to clear its padding, as that's not needed here.
+  .pt-input {
+    width: $pt-input-height;
+    height: $pt-input-height;
+    padding: 0;
+  }
+}
--- a/packages/datetime/src/_daterangeinput.scss
+++ b/packages/datetime/src/_daterangeinput.scss
@@ -12,7 +12,3 @@ Date range input
 
 Styleguide components.datetime.daterangeinput
 */
-
-.pt-daterangeinput {
-  position: relative;
-}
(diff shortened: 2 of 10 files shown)


packages/datetime/src/common/classes.ts
```diff
@@ -17,6 +17,10 @@ export const DATEPICKER_FOOTER = "pt-datepicker-footer";
 export const DATEPICKER_MONTH_SELECT = "pt-datepicker-month-select";
 export const DATEPICKER_YEAR_SELECT = "pt-datepicker-year-select";
 
+export const DATERANGEINPUT = "pt-daterangeinput";
+export const DATERANGEINPUT_FIELD = "pt-daterangeinput-field";
```
cmslewis: // Copied from the other PR (sorry for the split contexts)

`pt-daterangeinput-input` is more correct, but it sounds a little silly. Happy to change to that if we think it's better though.

packages/docs/src/styles/_examples.scss
```diff
@@ -65,7 +65,7 @@ $options-margin: $pt-grid-size * 3;
     margin-bottom: $options-margin;
   }
 
-  .pt-input-group + .pt-input-group {
+  :not(.pt-control-group) .pt-input-group + .pt-input-group {
```
cmslewis: // Copied from the other PR (sorry for the split contexts)

The previous global examples style broke the `DateRangeInput`, because the component includes two sibling `.pt-input-group`s:

image

Needed to tweak the global style to fix this.

packages/datetime/src/dateRangeInput.tsx
```diff
@@ -0,0 +1,75 @@
...
+} from "@blueprintjs/core";
+
+import * as DateClasses from "./common/classes";
+import {
+    IDatePickerBaseProps,
+} from "./datePickerCore";
+
+export interface IDateRangeInputProps extends IDatePickerBaseProps, IProps {
+    startInputProps?: IInputGroupProps;
+    endInputProps?: IInputGroupProps;
```
cmslewis: @giladgray - here you go! ^

cmslewis:
@llorca I want to keep the `DateRangeInput` API consistent with the `DateInput` API for the V1. Since the latter supports an `openOnFocus` prop, I want to add that here too. When that prop is false, we'll need a button of some kind that can open the popover. So I'd argue to keep the icon for now, then maybe we can nix the icon from both `DateRangeInput` and `DateInput` in a single PR later on.

packages/datetime/src/dateRangeInput.tsx
```diff
@@ -0,0 +1,75 @@
...
+                    className={DateClasses.DATERANGEINPUT_FIELD}
+                    inputRef={this.setEndDateInputRef}
+                    placeholder="End date"
+                    type="text"
+                    {...this.props.endInputProps}
+                />
+                <div className={classNames(DateClasses.DATERANGEINPUT_ICON_WRAPPER, Classes.INPUT_GROUP)}>
+                    <div className={Classes.INPUT}>
+                        <Button
+                            className={classNames(Classes.MINIMAL, "pt-icon-calendar")}
```
michael-yx-wu: use the iconClass function: https://github.com/palantir/blueprint/blob/master/packages/core/src/common/classes.ts#L183

packages/datetime/src/dateRangeInput.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+    public render() {
+        return (
+            <div className={Classes.CONTROL_GROUP}>
+                <InputGroup
+                    className={DateClasses.DATERANGEINPUT_FIELD}
+                    inputRef={this.setStartDateInputRef}
+                    placeholder="Start date"
+                    type="text"
+                    {...this.props.startInputProps}
```
cmslewis: Hmm, on second glance, I guess we'd want to combine any custom `className` contents with the `DateClasses.DATERANGEINPUT_FIELD` class above. Probably (maybe?) want to strip out the `inputRef` prop too. But does it make sense to allow developers to freely override `placeholder`, particularly with a function instead of a string? That seems like it could be nice.

packages/datetime/src/dateRangeInput.tsx
cmslewis: Ah, forgot about that. 👍

cmslewis:
Ping @giladgray @llorca for a quick CR and (ideally) approval, so I can move on to Part 3 and 4 for this week's release.

llorca:
Can we have the section showing very last, after date input?
image

packages/datetime/examples/dateRangeInputExample.tsx
```diff
@@ -0,0 +1,19 @@
...
+ */
+
+import BaseExample from "@blueprintjs/core/examples/common/baseExample";
+import * as React from "react";
+
+import { DateRangeInput } from "../src";
+
+export class DateRangeInputExample extends BaseExample<{}> {
+    protected renderExample() {
+        return (
```
llorca: 1-line that statement?

packages/datetime/src/_daterangeinput.scss
```diff
@@ -0,0 +1,46 @@
...
+@react-example DateRangeInputExample
+
+Styleguide components.datetime.daterangeinput
+*/
+
+.pt-daterangeinput {
+  position: relative;
+}
+
+.pt-daterangeinput-icon-wrapper {
```
llorca: Talked offline, let's get rid of it. It's an awkward toggle that we've had in `DateInput` through the `openOnFocus` prop, but it's not useful since the date input doesn't do anything more than a regular input

packages/datetime/examples/dateRangeInputExample.tsx
cmslewis: Done

cmslewis:
@llorca moved the example to the bottom of the "Date & time" section in the docs. :+1

packages/datetime/src/dateRangeInput.tsx
```diff
@@ -0,0 +1,61 @@
...
+            </div>
+        );
+    }
+
+    private setStartDateInputRef = (el: HTMLInputElement) => {
+        this.startDateInputRef = el;
+    }
+
+    private setEndDateInputRef = (el: HTMLInputElement) => {
+        this.endDateInputRef = el;
```
giladgray: you'll want to `safeInvoke(this.props.startInputProps.inputRef)` here and above. maybe add `defaultProps` for those two of empty objects?

packages/datetime/test/dateRangeInputTests.tsx
```diff
@@ -0,0 +1,22 @@
...
+
+import { InputGroup } from "@blueprintjs/core";
+import { DateRangeInput } from "../src/index";
+
+describe("<DateRangeInput>", () => {
+    it("renders with two InputGroup children", () => {
+        const component = mount(<DateRangeInput />);
+        expect(component.childAt(0).type()).to.equal(InputGroup);
+        expect(component.childAt(1).type()).to.equal(InputGroup);
+        expect(component.children().length).to.equal(2);
```
giladgray: how about `expect(component.find(InputGroup).length).to.equal(2)`? 
no need to validate the exact children, that's likely to bite you later.

cmslewis:
@giladgray addressed your changes. 👍

Issue #646, opened 2017-02-08
Disabled should take precedence over readonly in inputs


tgreenwatts:
Current:
<img width="419" alt="screen shot 2017-02-08 at 12 13 55 pm" src="https://cloud.githubusercontent.com/assets/24398123/22755489/383aa97e-edf8-11e6-87f5-c98506e42ee9.png">

Expected:
<img width="421" alt="screen shot 2017-02-08 at 12 12 33 pm" src="https://cloud.githubusercontent.com/assets/24398123/22755492/39b9f4a8-edf8-11e6-9362-b4698bace943.png">

The input is both readonly and disabled 

@llorca

PR #652, opened 2017-02-09
Fix readonly & disabled styles for inputs

llorca:
Before:
screen shot 2017-02-08 at 4 33 51 pm
screen shot 2017-02-08 at 4 34 30 pm

After (regular then primary):
screen shot 2017-02-08 at 4 19 50 pm
screen shot 2017-02-08 at 4 19 55 pm
screen shot 2017-02-08 at 4 20 16 pm
screen shot 2017-02-08 at 4 20 22 pm


packages/core/src/components/forms/_control-group.scss
```diff
@@ -100,20 +100,19 @@ Styleguide components.forms.control-group
       border-radius: $pt-border-radius;
     }
 
-    [readonly],
-    :disabled,
-    .pt-disabled {
-      z-index: index($control-group-stack, "input-disabled");
-      border-radius: 0;
```
llorca: This line was unnecessary so I removed it. I moved the block below `pt-intent` so that it takes precedence

llorca:
**Fixes #646** 

(GitHub won't let me update the PR description at the moment...)

packages/core/src/components/forms/_control-group.scss
leebyp: the whole block wasn't actually doing anything without the `&` before the attribs...

llorca: Correct, I forgot to mention that! Still had to move the block

PR #655, opened 2017-02-09
Fix tree contextmenu test

leebyp:
Hm... this worked when i tried uncommenting it locally?


JKillian:
@leebyp, hmm, maybe just something funky on my machine then? Anyway, merged your PR and as long as mine passes on CircleCI we should be good to go!

PR #665, opened 2017-02-10
Added New UI Icons

pkwi:
20px and 16px icons added: follower, following


packages/core/resources/icons/icons.json
```diff
@@ -2693,5 +2693,19 @@
...
+    "className": "pt-icon-follower",
+    "tags": "person, human, male, female, character, customer, individual, social",
+    "group": "action",
+    "content": "\\e760"
+  },
+  {
+    "name": "Following",
+    "className": "pt-icon-following",
+    "tags": "person, human, male, female, character, customer, individual, social",
+    "group": "action",
```
giladgray: i think these should be in `"interface"` with the other people:

image

PR #678, opened 2017-02-13
Set independent months in DateRangePicker through props, and add tests

leebyp:
- Enable independent months feature though props
- Add independent DateRangePicker months tests
- Update docs

`isSequential` naming?

### DIFF
--- a/packages/datetime/src/dateRangePicker.tsx
+++ b/packages/datetime/src/dateRangePicker.tsx
@@ -168,9 +168,12 @@ export class DateRangePicker
         const { disabledDays, selectedDays } = this.states;
 
         if (isSequential || isShowingOneMonth) {
-            // use the left DayPicker when we only show one
+            // use the left DayPicker when we only need one
+            const classes = classNames(DateClasses.DATEPICKER, DateClasses.DATERANGEPICKER, className, {
+                [DateClasses.DATERANGEPICKER_SEQUENTIAL]: !isShowingOneMonth,
+            });
             return (
-                <div className={classNames(DateClasses.DATEPICKER, DateClasses.DATERANGEPICKER, className)}>
+                <div className={classes}>
                     {this.maybeRenderShortcuts()}
                     <DayPicker
                         captionElement={this.renderSingleDayPickerCaption()}
--- a/packages/datetime/src/dateRangePicker.tsx
+++ b/packages/datetime/src/dateRangePicker.tsx
@@ -48,10 +48,11 @@ export interface IDateRangePickerProps extends IDatePickerBaseProps, IProps {
     defaultValue?: DateRange;
 
     /**
-     * Constrains months to be sequential
+     * Whether displayed months in the calendar are contiguous.
+     * If false, each side of the calendar can move independently to non-contiguous months.
      * @default true
      */
-    isSequential?: boolean;
+    contigiousCalendarMonths?: boolean;
 
     /**
      * Called when the user selects a day.
@@ -89,7 +90,7 @@ export class DateRangePicker
 
     public static defaultProps: IDateRangePickerProps = {
         allowSingleDayRange: false,
-        isSequential: true,
+        contigiousCalendarMonths: true,
         maxDate: getDefaultMaxDate(),
         minDate: getDefaultMinDate(),
         shortcuts: true,
@@ -162,12 +163,12 @@ export class DateRangePicker
 
     public render() {
         const modifiers = combineModifiers(this.modifiers, this.props.modifiers);
-        const { className, isSequential, locale, localeUtils, maxDate, minDate } = this.props;
+        const { className, contigiousCalendarMonths, locale, localeUtils, maxDate, minDate } = this.props;
         const isShowingOneMonth = DateUtils.areSameMonth(this.props.minDate, this.props.maxDate);
         const { leftView, rightView } = this.state;
         const { disabledDays, selectedDays } = this.states;
 
-        if (isSequential || isShowingOneMonth) {
+        if (contigiousCalendarMonths || isShowingOneMonth) {
             // use the left DayPicker when we only need one
             const classes = classNames(DateClasses.DATEPICKER, DateClasses.DATERANGEPICKER, className, {
                 [DateClasses.DATERANGEPICKER_SEQUENTIAL]: !isShowingOneMonth,
@@ -176,7 +177,7 @@ export class DateRangePicker
                 <div className={classes}>
                     {this.maybeRenderShortcuts()}
                     <DayPicker
-                        captionElement={this.renderSingleDayPickerCaption()}
+                        captionElement={this.renderSingleCaption()}
                         disabledDays={disabledDays}
                         fromMonth={minDate}
                         initialMonth={leftView.getFullDate()}
@@ -287,7 +288,7 @@ export class DateRangePicker
         );
     }
 
-    private renderSingleDayPickerCaption() {
+    private renderSingleCaption() {
         const { maxDate, minDate } = this.props;
         return (
             <DatePickerCaption
@@ -401,7 +402,7 @@ export class DateRangePicker
 
     private updateLeftView(leftView: MonthAndYear) {
         let rightView = this.state.rightView.clone();
-        if (rightView.isBefore(leftView) || rightView.isSame(leftView)) {
+        if (!leftView.isBefore(rightView)) {
             rightView = leftView.getNextMonth();
         }
         this.setViews(leftView, rightView);
@@ -409,7 +410,7 @@ export class DateRangePicker
 
     private updateRightView(rightView: MonthAndYear) {
         let leftView = this.state.leftView.clone();
-        if (leftView.isAfter(rightView) || leftView.isSame(rightView)) {
+        if (!rightView.isBefore(rightView)) {
             leftView = rightView.getPreviousMonth();
         }
         this.setViews(leftView, rightView);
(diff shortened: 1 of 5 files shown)


packages/datetime/src/dateRangePicker.tsx
```diff
@@ -48,6 +48,12 @@ export interface IDateRangePickerProps extends IDatePickerBaseProps, IProps {
     defaultValue?: DateRange;
 
     /**
+     * Constrains months to be sequential
+     * @default true
+     */
+    isSequential?: boolean;
```
cmslewis: This effect of this prop isn't immediately clear from its name, and we could also afford to improve the description. Maybe something like this:

```
/**
 * Whether to keep displayed months contiguous in the calendar.
 * If false, each side of the calendar can move independently to non-contiguous months.
 * @default true
 */
keepDisplayMonthsContiguous?: boolean;
```

?

packages/datetime/src/dateRangePicker.tsx
```diff
@@ -155,25 +162,30 @@ export class DateRangePicker
...
         const modifiers = combineModifiers(this.modifiers, this.props.modifiers);
-        const { className, locale, localeUtils, maxDate, minDate } = this.props;
+        const { className, isSequential, locale, localeUtils, maxDate, minDate } = this.props;
         const isShowingOneMonth = DateUtils.areSameMonth(this.props.minDate, this.props.maxDate);
         const { leftView, rightView } = this.state;
         const { disabledDays, selectedDays } = this.states;
 
-        if (isShowingOneMonth) {
-            // use the left DayPicker when we only show one
+        if (isSequential || isShowingOneMonth) {
```
cmslewis: How extensible is this approach to the use case where consumers may want to show only one month at a time in the `DateRangePicker` (by passing in their own `numberOfMonths` prop, e.g.)? We've gotten this request before from a project where users were getting confused by every variant that involved two display months.

packages/datetime/src/dateRangePicker.tsx
```diff
@@ -275,7 +287,7 @@ export class DateRangePicker
         );
     }
 
-    private renderOneMonthCaption() {
+    private renderSingleDayPickerCaption() {
```
cmslewis: `...DayPicker...` or `...DatePicker...`?

packages/datetime/src/dateRangePicker.tsx
```diff
@@ -389,16 +401,16 @@ export class DateRangePicker
 
     private updateLeftView(leftView: MonthAndYear) {
         let rightView = this.state.rightView.clone();
-        if (rightView.isBefore(leftView)) {
-            rightView = rightView.getNextMonth();
+        if (rightView.isBefore(leftView) || rightView.isSame(leftView)) {
```
cmslewis: Simpler: `if (!leftView.isBefore(rightView)) {`?

packages/datetime/src/dateRangePicker.tsx
```diff
@@ -389,16 +401,16 @@ export class DateRangePicker
...
+            rightView = leftView.getNextMonth();
         }
         this.setViews(leftView, rightView);
     }
 
     private updateRightView(rightView: MonthAndYear) {
         let leftView = this.state.leftView.clone();
-        if (leftView.isAfter(rightView)) {
-            leftView = leftView.getPreviousMonth();
+        if (leftView.isAfter(rightView) || leftView.isSame(rightView)) {
```
cmslewis: Simpler: `if (!rightView.isBefore(leftView)) {`?

cmslewis: I'd also support adding an `isSameOrBefore` method, like Moment.js's [`isSameOrBefore`](https://momentjs.com/docs/#/query/is-same-or-before/).

packages/datetime/test/dateRangePickerTests.tsx
```diff
@@ -134,6 +134,145 @@ describe("<DateRangePicker>", () => {
         });
     });
 
+    describe("left/right calendar when not sequential", () => {
+        it("only shows one calendar when minDate and maxDate are the same month", () => {
```
cmslewis: nit: ...are **in** the same month

packages/datetime/test/dateRangePickerTests.tsx
```diff
@@ -134,6 +134,145 @@ describe("<DateRangePicker>", () => {
...
+            const minDate = new Date(2015, Months.DECEMBER, 1);
+            const maxDate = new Date(2015, Months.DECEMBER, 15);
+
+            renderDateRangePicker({ isSequential, maxDate, minDate });
+            assert.lengthOf(document.getElementsByClassName("DayPicker"), 1);
+            assert.lengthOf(document.getElementsByClassName(".DayPicker-NavButton--prev"), 0);
+            assert.lengthOf(document.getElementsByClassName(".DayPicker-NavButton--next"), 0);
+        });
+
+        it("left calendar bound between minDate and (maxDate - 1 month)", () => {
```
cmslewis: Would prefer to have a verb somewhere in this title; the goal of the test isn't crystal clear at a glance.

packages/datetime/test/dateRangePickerTests.tsx
```diff
@@ -134,6 +134,145 @@ describe("<DateRangePicker>", () => {
...
+            const minDate = new Date(2015, Months.JANUARY, 1);
+            const maxDate = new Date(2015, Months.DECEMBER, 15);
+
+            renderDateRangePicker({ isSequential, maxDate, minDate });
+            const monthSelects = getMonthSelect().children;
+            assert.equal(monthSelects[0].getAttribute("value"), Months.JANUARY);
+            assert.equal(monthSelects[monthSelects.length - 1].getAttribute("value"), Months.NOVEMBER);
+        });
+
+        it("right calendar bound between (minDate + 1 month) and maxDate", () => {
```
cmslewis: Same

packages/datetime/test/dateRangePickerTests.tsx
```diff
@@ -134,6 +134,145 @@ describe("<DateRangePicker>", () => {
...
+            const minDate = new Date(2015, Months.JANUARY, 1);
+            const maxDate = new Date(2015, Months.DECEMBER, 15);
+
+            renderDateRangePicker({ isSequential, maxDate, minDate });
+            const monthSelects = getMonthSelect(false).children;
+            assert.equal(monthSelects[0].getAttribute("value"), Months.FEBRUARY);
+            assert.equal(monthSelects[monthSelects.length - 1].getAttribute("value"), Months.DECEMBER);
+        });
+
+        it("left calendar can altered independent of right calendar", () => {
```
cmslewis: nit: fix grammar:

left calendar can **be** altered independent**ly** of right calendar

packages/datetime/test/dateRangePickerTests.tsx
cmslewis: Oh wait, is `bound` a verb here? If so, maybe add an `is` before it to clarify.

packages/datetime/test/dateRangePickerTests.tsx
```diff
@@ -134,6 +134,145 @@ describe("<DateRangePicker>", () => {
...
+            const prevBtn = document.queryAll(".DayPicker-NavButton--prev");
+            const nextBtn = document.queryAll(".DayPicker-NavButton--next");
+
+            TestUtils.Simulate.click(prevBtn[0]);
+            assert.equal(dateRangePicker.state.leftView.getMonth(), Months.APRIL);
+            TestUtils.Simulate.click(nextBtn[0]);
+            assert.equal(dateRangePicker.state.leftView.getMonth(), Months.MAY);
+        });
+
+        it("right calendar can altered independent of left calendar", () => {
```
cmslewis: Same

packages/datetime/test/dateRangePickerTests.tsx
```diff
@@ -134,6 +134,145 @@ describe("<DateRangePicker>", () => {
...
+            const prevBtn = document.queryAll(".DayPicker-NavButton--prev");
+            const nextBtn = document.queryAll(".DayPicker-NavButton--next");
+
+            TestUtils.Simulate.click(nextBtn[1]);
+            assert.equal(dateRangePicker.state.rightView.getMonth(), Months.JULY);
+            TestUtils.Simulate.click(prevBtn[1]);
+            assert.equal(dateRangePicker.state.rightView.getMonth(), Months.JUNE);
+        });
+
+        it("right calendar is shifted if left calendar is equal or past right calendar with month dropdown", () => {
```
cmslewis: Would prefer a slight rewording here to make the cause => effect relationship a little clearer. I'm actually starting to make this adjustment in my own test descriptions. Here's one option:

`Changing left calendar via month dropdown to equal or exceed right calendar, shifts the right calendar`.

Same feedback for the following tests.

leebyp: i can see how that's more readable - updated

packages/datetime/src/dateRangePicker.tsx
leebyp: changed to `singleCaption` to mimic the other `leftCaption` and `rightCaption` methods

packages/datetime/src/dateRangePicker.tsx
leebyp: we can add another prop with `numberOfMonths` yup (though we can discuss on a new issue thread) and the logic should just work out of the box - the css will need some tweaking to support that

PR #683, opened 2017-02-14
[Date Range Input] - Part 6: Controlled Mode

cmslewis:
#### Checklist

- [x] Include tests

Implemented controlled mode.

- _New feature_: Introduce a `value` prop.
- _New feature_: Introduce a `defaultValue` prop.
- Add prop descriptions for all existing props. 
- Add unit tests (and some reorganize existing ones into `when controlled` and `when uncontrolled` blocks).
- Harden existing unit tests (now that we can control the initial display month via `defaultValue` or `value`).

- [ ] My nit questions over in the "Files changed" tab.
- [ ] Polishing copy in prop descriptions

### DIFF
--- a/packages/datetime/src/dateRangeInput.tsx
+++ b/packages/datetime/src/dateRangeInput.tsx
@@ -51,7 +51,7 @@ export interface IDateRangeInputState {
 export class DateRangeInput extends AbstractComponent<IDateRangeInputProps, IDateRangeInputState> {
     public static defaultProps: IDateRangeInputProps = {
         endInputProps: {},
-        format: "MM/DD/YYYY",
+        format: "YYYY-MM-DD",
         maxDate: getDefaultMaxDate(),
         minDate: getDefaultMinDate(),
         startInputProps: {},
--- a/packages/datetime/src/dateRangeInput.tsx
+++ b/packages/datetime/src/dateRangeInput.tsx
@@ -37,6 +37,7 @@ import {
 } from "./dateRangePicker";
 
 export interface IDateRangeInputProps extends IDatePickerBaseProps, IProps {
+    defaultValue?: DateRange;
     endInputProps?: IInputGroupProps;
     format?: string;
     onChange?: (selectedRange: DateRange) => void;
@@ -77,9 +78,7 @@ export class DateRangeInput extends AbstractComponent<IDateRangeInputProps, IDat
     public constructor(props: IDateRangeInputProps, context?: any) {
         super(props, context);
 
-        const [selectedStart, selectedEnd] = (this.props.value != null)
-            ? fromDateRangeToMomentArray(this.props.value)
-            : [moment(null), moment(null)];
+        const [selectedStart, selectedEnd] = this.getInitialRange();
 
         this.state = {
             isOpen: false,
@@ -153,6 +152,23 @@ export class DateRangeInput extends AbstractComponent<IDateRangeInputProps, IDat
 
     private handlePopoverClose = () => {
         this.setState({ isOpen: false });
+    }
+
+    private getInitialRange = (): moment.Moment[] => {
+        const { defaultValue, value } = this.props;
+
+        let initialStart: moment.Moment;
+        let initialEnd: moment.Moment;
+
+        if (value != null) {
+            [initialStart, initialEnd] = fromDateRangeToMomentArray(value);
+        } else if (defaultValue != null) {
+            [initialStart, initialEnd] = fromDateRangeToMomentArray(defaultValue);
+        } else {
+            [initialStart, initialEnd] = [moment(null), moment(null)];
+        }
+
+        return [initialStart, initialEnd];
     }
 
     private getSelectedRange = () => {
--- a/packages/datetime/src/dateRangeInput.tsx
+++ b/packages/datetime/src/dateRangeInput.tsx
@@ -21,6 +21,7 @@ import {
 
 import {
     DateRange,
+    fromDateRangeToMomentArray,
     fromDateToMoment,
     fromMomentToDate,
     isMomentNull,
@@ -40,6 +41,7 @@ export interface IDateRangeInputProps extends IDatePickerBaseProps, IProps {
     format?: string;
     onChange?: (selectedRange: DateRange) => void;
     startInputProps?: IInputGroupProps;
+    value?: DateRange;
 }
 
 export interface IDateRangeInputState {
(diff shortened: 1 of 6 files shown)


packages/datetime/src/dateRangeInput.tsx
```diff
@@ -148,6 +196,23 @@ export class DateRangeInput extends AbstractComponent<IDateRangeInputProps, IDat
         this.setState({ isOpen: false });
     }
 
+    private getInitialRange = (props = this.props): moment.Moment[] => {
```
cmslewis: nit: Maybe just call it `getInitialMomentRange` instead of declaring a return type?

packages/datetime/src/dateRangeInput.tsx
```diff
@@ -148,6 +196,23 @@ export class DateRangeInput extends AbstractComponent<IDateRangeInputProps, IDat
...
+
+        let initialStart: moment.Moment;
+        let initialEnd: moment.Moment;
+
+        if (value != null) {
+            [initialStart, initialEnd] = fromDateRangeToMomentArray(value);
+        } else if (defaultValue != null) {
+            [initialStart, initialEnd] = fromDateRangeToMomentArray(defaultValue);
+        } else {
+            [initialStart, initialEnd] = [moment(null), moment(null)];
```
cmslewis: nit: Could just assign these values on two separate lines, but as is, the structure nicely parallels that of the declarations above. Alternatively, could do `fromDateRangeToMomentArray([null, null])`

packages/datetime/src/dateRangeInput.tsx
```diff
@@ -36,10 +37,44 @@ import {
...
+
+    /**
+     * Props to pass to the end-date input.
+     */
     endInputProps?: IInputGroupProps;
+
+    /**
+     * The format of each date in the date range. See options
+     * here: http://momentjs.com/docs/#/displaying/format/
+     * @default "MM/DD/YYYY"
```
leebyp: `YYYY/MM/DD`?

packages/datetime/src/dateRangeInput.tsx
```diff
@@ -36,10 +37,44 @@ import {
...
     onChange?: (selectedRange: DateRange) => void;
+
+    /**
+     * Props to pass to the start-date input.
+     */
     startInputProps?: IInputGroupProps;
+
+    /**
+     * The currently selected date range. If this prop is present, the component acts in a controlled manner.
+     * To display no date range in the input fields, pass `null` to the value prop. To display an invalid date error
```
leebyp: API question: `null` or `[null, null]`?

packages/datetime/src/dateRangeInput.tsx
```diff
@@ -128,9 +166,19 @@ export class DateRangeInput extends AbstractComponent<IDateRangeInputProps, IDat
...
+            this.setState({ selectedStart, selectedEnd });
+        }
+        super.componentWillReceiveProps(nextProps);
+    }
+
     private handleDateRangePickerChange = (selectedRange: DateRange) => {
-        const [selectedStart, selectedEnd] = selectedRange.map(fromDateToMoment);
-        this.setState({ selectedStart, selectedEnd });
+        if (this.props.value === undefined) {
+            const [selectedStart, selectedEnd] = selectedRange.map(fromDateToMoment);
```
leebyp: `fromDateRangeToMomentArray`?

packages/datetime/src/dateRangeInput.tsx
leebyp: can't you just return all of these directly?

packages/datetime/test/dateRangeInputTests.tsx
```diff
@@ -34,43 +36,99 @@ describe("<DateRangeInput>", () => {
...
+
+            const { root, getDayElement } = wrap(<DateRangeInput defaultValue={defaultValue} />);
+            root.setState({ isOpen: true });
+
+            getDayElement(24).simulate("click");
 
-        getDayElement(22).simulate("click");
-        getDayElement(24).simulate("click");
+            expect(getStartInputText(root)).to.equal("2017-01-22");
+            expect(getEndInputText(root)).to.equal("2017-01-24");
```
leebyp: More rigorous testing with selecting start, selecting end would be nice. Pretty much exactly as you do in the controlled tests below. And `defaultValue` in itself is good to make sure the value syncs up with the input as expected.

packages/datetime/test/dateRangeInputTests.tsx
```diff
@@ -34,43 +36,99 @@ describe("<DateRangeInput>", () => {
...
+        });
 
-        // TODO: Make these checks more rigorous once controlled mode is supported.
-        expect(getStartInputText(root)).to.be.not.empty;
-        expect(getEndInputText(root)).to.be.not.empty;
+        it("Clearing the date range clears the inputs, and invokes onChange with [null, null]", () => {
```
leebyp: Dunno if it's already on your radar, but this reminded me of the `allowSingleDayRange` from DateRangePicker. Not sure if necessary to support.

packages/datetime/test/dateRangeInputTests.tsx
```diff
@@ -89,6 +147,22 @@ describe("<DateRangeInput>", () => {
...
 
+    function assertDateRangesEqual(actual: DateRange, expected: string[]) {
+        const [expectedStart, expectedEnd] = expected;
+        const [actualStart, actualEnd] = actual.map((date: Date) => {
+            if (date == null) {
+                return null;
+            }
+            return [
+                date.getFullYear(),
+                padWithZeroes((date.getMonth() + 1) + "", 2),
```
leebyp: `.toString()`

packages/datetime/src/dateRangeInput.tsx
cmslewis: I generally prefer having a single `return` at the end of a function, but yeah: returning rather than assigning gets rid of the mutability here, which is nicer. 👍

packages/datetime/src/dateRangeInput.tsx
cmslewis: Fixed

packages/datetime/src/dateRangeInput.tsx
cmslewis: Fixed

packages/datetime/test/dateRangeInputTests.tsx
cmslewis: Fixed. This was copied from `dateInputTests.tsx`. Went ahead and added a new `toHyphenatedDateString` function to a new `dateTestUtils.tsx` file in `test/common`.

packages/datetime/test/dateRangeInputTests.tsx
cmslewis: Planning to add that support later. 👍

packages/datetime/test/dateRangeInputTests.tsx

packages/datetime/src/dateRangeInput.tsx
cmslewis: @giladgray @llorca - opinions? IIRC, there's also the consideration that we're trying to prefer `undefined` to `null` as a return value.

PR #691, opened 2017-02-16
Expose more CSS classes in JS

tgreenwatts:
#### Checklist

- [ ] Include tests
- [ ] Update documentation

Add `pt-icon` and `pt-text-overflow-ellipsis` to `Classes` 

Not sure if `pt-text-overflow-ellipsis` should be considered a "modifier" or in a new category 

@giladgray for review

### DIFF
--- a/packages/core/src/common/classes.ts
+++ b/packages/core/src/common/classes.ts
@@ -24,6 +24,8 @@ export const FIXED = "pt-fixed";
 export const FIXED_TOP = "pt-fixed-top";
 export const VERTICAL = "pt-vertical";
 export const ROUND = "pt-round";
+
+// text utilities
 export const TEXT_OVERFLOW_ELLIPSIS = "pt-text-overflow-ellipsis";
 
 // components
--- a/packages/core/src/common/classes.ts
+++ b/packages/core/src/common/classes.ts
@@ -24,6 +24,7 @@ export const FIXED = "pt-fixed";
 export const FIXED_TOP = "pt-fixed-top";
 export const VERTICAL = "pt-vertical";
 export const ROUND = "pt-round";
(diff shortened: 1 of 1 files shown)


palantirtech:
Thanks for your interest in [palantir/blueprint](https://github.com/palantir/blueprint), @tgreenwatts! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

packages/core/src/common/classes.ts
```diff
@@ -24,6 +24,7 @@ export const FIXED = "pt-fixed";
 export const FIXED_TOP = "pt-fixed-top";
 export const VERTICAL = "pt-vertical";
 export const ROUND = "pt-round";
+export const TEXT_OVERFLOW_ELLIPSIS = "pt-text-overflow-ellipsis";
```
llorca: I don't think this is a modifier, more like a utility class. I'll let @giladgray confirm!

cmslewis: I would create a new section:

```
...
export const ROUND = "pt-round";

// text utilities
export const TEXT_OVERFLOW_ELLIPSIS = "pt-text-overflow-ellipsis";

// components
...
```

cmslewis: @palantir/blueprint Should we just put every CSS API class into this file? I'm thinking we might as well.

leebyp: It would be useful to get autocomplete for the other ones in the typography. If we're doing one, no reason not to do the others too if we're maintaining this file

tgreenwatts:
Thanks @adidahiya for merging while I was ooto!

PR #694, opened 2017-02-16
Hide background color on dark minimal disabled intent button hover

llorca:
Before:
image

After:
image

### DIFF
--- a/packages/core/src/components/button/_common.scss
+++ b/packages/core/src/components/button/_common.scss
@@ -393,6 +393,7 @@ $button-intents: (
 
     &:disabled,
     &.pt-disabled {
+      background: none;
       color: rgba($dark-text-color, 0.5);
     }
   }


packages/core/src/components/button/_common.scss
```diff
@@ -393,6 +393,7 @@ $button-intents: (
 
     &:disabled,
     &.pt-disabled {
+      background: none;
```
cmslewis: In the `datetime` package, we typically handle this by adding `:not(.pt-disabled)` in all of the hover and active styles. I prefer what you have here though.

llorca: Yeah, we used to do that in some places. The advantage is that it's less lines of code, but it's harder to maintain. I prefer keeping it super explicit, even if it adds a few more lines

PR #696, opened 2017-02-16
Clean up some docs casing

llorca:
Add some more sentence casing


packages/core/src/components/navbar/_navbar.scss
```diff
@@ -55,7 +55,7 @@ $dark-navbar-background-color: $dark-gray5 !default;
 
 .pt-navbar {
   position: relative;
-  z-index: $pt-z-index-content;
+  z-index: $pt-z-index-overlay;
```
cmslewis: @llorca Can we revert this change and put it in a different PR? Just to conceptually separate different kinds of changes and prevent the other stuff from blocking on this.

PR #698, opened 2017-02-16
Use same color for colored icon and text in menus, callouts

llorca:
With this PR, colored icons get the same color value than the text next to it, in menus, callouts... Somehow, this was already the case for minimal buttons, so nothing to change there.

Before:
image
image

After:
image
image

cc @pkwi

### DIFF
--- a/packages/core/src/components/forms/_input-group.scss
+++ b/packages/core/src/components/forms/_input-group.scss
@@ -197,7 +197,11 @@ $input-button-height-large: $pt-button-height !default;
       }
 
       .pt-icon {
-        color: $color;
+        color: map-get($pt-intent-text-colors, $intent);
+      }
+
+      .pt-dark & .pt-icon {
+        color: map-get($pt-dark-intent-text-colors, $intent);
       }
     }
   }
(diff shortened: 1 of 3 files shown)


pkwi:
Looking good!

leebyp:
@llorca should input icons have the same colour change?

llorca:
Oh yes missed those. @pkwi?

pkwi:
yeah, same for colour for input icons

llorca:
@pkwi input intents never have colored text, and the current icon color is shade 3:
image

Does it still need to be updated?

pkwi:
I think they should get updated too. It makes sense to keep these icons consistent.

packages/core/src/components/forms/_input-group.scss
```diff
@@ -197,7 +197,11 @@ $input-button-height-large: $pt-button-height !default;
       }
 
       .pt-icon {
-        color: $color;
+        color: map-get($pt-intent-text-colors, $intent);
+      }
+
+      .pt-dark & .pt-icon {
```
giladgray: move this inside the block above. `.pt-icon { .pt-icon & { ... } }`

PR #711, opened 2017-02-21
[Date Range Input] - Part 7: Typing dates into input fields

cmslewis:
#### Checklist

- [x] Include tests

- _New feature:_ Can now type dates into the input fields, and the picker selection will update (if uncontrolled).
- _Bugfix:_ Fix parallel issue I noticed in `DateInput` where editing the input field in controlled mode overwrote a controlled value. Now, if the user changes the text in an input field in controlled mode, (1) the calendar selection will not change, and (2) the controlled value will be repopulated in the field on blur. (And to be clear, I fixed this just for `DateRangeInput`, not for `DateInput` too).

_Changes NOT included:_
- **I didn't implement input validation yet**. It's perfectly possible to input backward or otherwise invalid date ranges. The next PR will deal with input validation.
- **I didn't implement focus-handling yet.** The focused field has no bearing on which end of the date range will be specified by the next click. A near-future PR will deal with focus handling. 

- What's the best way to support generic `onBlur`, `onFocus`, and `onChange` handlers for the two input fields, given that each field needs to process the same logic with different pieces of `state`?

    In my prototype from a couple weeks ago, I passed boundary-specific `state` keys (e.g. `"isStartInputFocused"`, `"selectedEnd"`) as separate string parameters from `handleStartInputBlur` (e.g.) into `handleGenericInputBlur` (e.g.).

    The `IStateKeysAndValuesObject` here looks a little smelly at a glance, but it feels nicer than my previous approach.

    Or maybe it's better to pull the logic out into a couple child components (called `DateRangeInputField`?) Not sure how much mileage that'll get us, but happy to take a look if we're not okay with the current implementation.

- How does the controlled-mode _Bugfix_ sound?

#### Screenshot

_Uncontrolled Mode_

_Controlled Mode_

### DIFF
--- a/packages/datetime/src/dateRangeInput.tsx
+++ b/packages/datetime/src/dateRangeInput.tsx
@@ -409,7 +409,7 @@ export class DateRangeInput extends AbstractComponent<IDateRangeInputProps, IDat
                 values: {
                     controlledValue: (controlledRange != null) ? controlledRange[0] : undefined,
                     inputString: this.state.startInputString,
-                    isFocused: this.state.isStartInputFocused,
+                    isInputFocused: this.state.isStartInputFocused,
                     selectedValue: this.state.selectedStart,
                 },
             } as IStateKeysAndValuesObject;
@@ -423,7 +423,7 @@ export class DateRangeInput extends AbstractComponent<IDateRangeInputProps, IDat
                 values: {
                     controlledValue: (controlledRange != null) ? controlledRange[1] : undefined,
                     inputString: this.state.endInputString,
-                    isFocused: this.state.isEndInputFocused,
+                    isInputFocused: this.state.isEndInputFocused,
                     selectedValue: this.state.selectedEnd,
                 },
             } as IStateKeysAndValuesObject;
--- a/packages/datetime/src/dateRangeInput.tsx
+++ b/packages/datetime/src/dateRangeInput.tsx
@@ -171,6 +171,7 @@ export class DateRangeInput extends AbstractComponent<IDateRangeInputProps, IDat
                         placeholder="Start date"
                         {...this.props.startInputProps}
                         inputRef={this.refHandlers.startInputRef}
+                        onChange={this.handleInputChange}
                         onClick={this.handleInputClick}
                         onFocus={this.handleInputFocus}
                         value={startInputString}
@@ -179,6 +180,7 @@ export class DateRangeInput extends AbstractComponent<IDateRangeInputProps, IDat
                         placeholder="End date"
                         {...this.props.endInputProps}
                         inputRef={this.refHandlers.endInputRef}
+                        onChange={this.handleInputChange}
                         onClick={this.handleInputClick}
                         onFocus={this.handleInputFocus}
                         value={endInputString}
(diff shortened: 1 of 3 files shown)


cmslewis:
Update: I'm playing with pulling input-field logic into a separate component. Will report back with results.

llorca:
This may not be for this PR, but note that the input loses focus after selecting a start date

cmslewis:
@llorca - yep, will handle focus-state management in a future PR. 👍

packages/datetime/src/dateRangeInput.tsx
```diff
@@ -190,8 +221,114 @@ export class DateRangeInput extends AbstractComponent<IDateRangeInputProps, IDat
...
+
+    private handleEndInputBlur = (e: React.FormEvent<HTMLInputElement>) => {
+        this.handleInputBlur(e, DateRangeBoundary.END);
+    }
+
+    private handleInputBlur = (_e: React.FormEvent<HTMLInputElement>, boundary: DateRangeBoundary) => {
+        const { keys, values } = this.getStateKeysAndValuesForBoundary(boundary);
+
+        const maybeNextValue = this.dateStringToMoment(values.inputString);
+        const isValueControlled = this.props.value !== undefined;
```
leebyp: fan of starting to use a private method to keep track of this across our components - started doing this here: https://github.com/palantir/blueprint/blob/master/packages/datetime/src/dateRangePicker.tsx#L101

thoughts?

packages/datetime/src/dateRangeInput.tsx
```diff
@@ -190,8 +221,114 @@ export class DateRangeInput extends AbstractComponent<IDateRangeInputProps, IDat
...
+
+        if (values.inputString == null || values.inputString.length === 0) {
+            if (isValueControlled) {
+                this.setState({
+                    [keys.isInputFocused]: false,
+                    [keys.inputString]: this.getFormattedDateString(values.controlledValue),
```
leebyp: yup, i agree with this interaction

packages/datetime/src/dateRangeInput.tsx
```diff
@@ -211,17 +348,74 @@ export class DateRangeInput extends AbstractComponent<IDateRangeInputProps, IDat
...
+                    inputString: "endInputString",
+                    isInputFocused: "isEndInputFocused",
+                    selectedValue: "selectedEnd",
+                },
+                values: {
+                    controlledValue: (controlledValue != null) ? controlledValue[1] : undefined,
+                    inputString: this.state.endInputString,
+                    isFocused: this.state.isEndInputFocused,
+                    selectedValue: this.state.selectedEnd,
+                },
```
leebyp: The `DateRangeInputField` you mention in the description will be another step towards keeping consistencies between `DateInput` and `DateRangeInput` though I am wary we might get into another rabbit hole

packages/datetime/src/dateRangeInput.tsx
```diff
@@ -81,10 +82,36 @@ export interface IDateRangeInputProps extends IDatePickerBaseProps, IProps {
...
+    startInputString?: string;
+    endInputString?: string;
+
     selectedEnd?: moment.Moment;
     selectedStart?: moment.Moment;
 };
 
+interface IStateKeysAndValuesObject {
+    keys: {
+        inputString: string;
```
leebyp: i think you can put the actual strings here to enforce the type

packages/datetime/src/dateRangeInput.tsx
leebyp: even if we split it out, will this parent still need to utilise a lot of the internals from the hypothetical new component? and the code be just as complicated?

packages/datetime/src/dateRangeInput.tsx
cmslewis: Yeah, I thought about that too. We should just be careful to note that `null` may be a meaningful value for `this.props.value`, so not every `isControlled()` implementation will be the same (some will need a strict `!== undefined` check).

packages/datetime/src/dateRangeInput.tsx
cmslewis: Update: I tried splitting each field out into a common, fully-controlled component. The LOC decreased in `dateRangeInput.tsx`, but things actually got hairier and less clear, plus there's still some boilerplate with respect to updating the appropriate state values in the parent component. Wasn't loving it.

packages/datetime/src/dateRangeInput.tsx
cmslewis: What do you mean exactly?

PR #718, opened 2017-02-22
[Table] Add focus cell to blueprint table

gscshoyru:

Adds focus cell to table, allow focus cell to be moved around with arrow keys

This is the framework for the rest of everything we need to do with focus cell -- make sure the behavior and setup makes sense. 

#### Screenshot

screen shot 2017-02-22 at 5 44 24 pm

### DIFF
--- a/packages/table/src/interactions/selectable.tsx
+++ b/packages/table/src/interactions/selectable.tsx
@@ -10,7 +10,7 @@ import * as React from "react";
 import { DragEvents } from "../interactions/dragEvents";
 import { Draggable, ICoordinateData, IDraggableProps } from "../interactions/draggable";
 import { IFocusedCellCoordinates } from "../layers/focusCell";
-import { IRegion, RegionCardinality, Regions  } from "../regions";
+import { IRegion, RegionCardinality, Regions } from "../regions";
 
 export type ISelectedRegionTransform = (region: IRegion, event: MouseEvent, coords?: ICoordinateData) => IRegion;
 
@@ -27,7 +27,7 @@ export interface ISelectableProps {
      * focused cell coordinates. This should be considered the new focused cell
      * state for the entire table.
      */
-    onFocus: (focusedCellCoordinates: IFocusedCellCoordinates) => void;
+    onFocus: (focusedCell: IFocusedCellCoordinates) => void;
 
     /**
      * When the user selects something, this callback is called with a new
@@ -106,17 +106,11 @@ export class DragSelectable extends React.Component<IDragSelectableProps, {}> {
 
         if (regionCardinality === RegionCardinality.FULL_TABLE) {
             focusCellCoordinates = { col: 0, row: 0 };
-        }
-
-        if (regionCardinality === RegionCardinality.FULL_COLUMNS) {
+        } else if (regionCardinality === RegionCardinality.FULL_COLUMNS) {
             focusCellCoordinates = { col: region.cols[0], row: 0 };
-        }
-
-        if (regionCardinality === RegionCardinality.FULL_ROWS) {
+        } else if (regionCardinality === RegionCardinality.FULL_ROWS) {
             focusCellCoordinates = { col: 0, row: region.rows[0] };
-        }
-
-        if (regionCardinality === RegionCardinality.CELLS) {
+        } else if (regionCardinality === RegionCardinality.CELLS) {
             focusCellCoordinates = { col: region.cols[0], row: region.rows[0] };
         }
 
--- a/packages/table/src/interactions/selectable.tsx
+++ b/packages/table/src/interactions/selectable.tsx
@@ -9,7 +9,8 @@ import * as PureRender from "pure-render-decorator";
 import * as React from "react";
 import { DragEvents } from "../interactions/dragEvents";
 import { Draggable, ICoordinateData, IDraggableProps } from "../interactions/draggable";
-import { IRegion, Regions } from "../regions";
+import { IFocusedCellCoordinates } from "../layers/focusCell";
+import { IRegion, RegionCardinality, Regions  } from "../regions";
 
 export type ISelectedRegionTransform = (region: IRegion, event: MouseEvent, coords?: ICoordinateData) => IRegion;
 
@@ -20,6 +21,13 @@ export interface ISelectableProps {
      * and a mouse drag will select the current column/row/cell only.
      */
     allowMultipleSelection: boolean;
+
+    /**
+     * When the user focuses something, this callback is called with a new
+     * focused cell coordinates. This should be considered the new focused cell
+     * state for the entire table.
+     */
+    onFocus: (focusedCellCoordinates: IFocusedCellCoordinates) => void;
 
     /**
      * When the user selects something, this callback is called with a new
@@ -92,6 +100,27 @@ export class DragSelectable extends React.Component<IDragSelectableProps, {}> {
         if (!Regions.isValid(region)) {
             return false;
         }
+
+        const regionCardinality = Regions.getRegionCardinality(region);
+        let focusCellCoordinates: IFocusedCellCoordinates = null;
+
+        if (regionCardinality === RegionCardinality.FULL_TABLE) {
+            focusCellCoordinates = { col: 0, row: 0 };
+        }
+
+        if (regionCardinality === RegionCardinality.FULL_COLUMNS) {
+            focusCellCoordinates = { col: region.cols[0], row: 0 };
+        }
+
+        if (regionCardinality === RegionCardinality.FULL_ROWS) {
+            focusCellCoordinates = { col: 0, row: region.rows[0] };
+        }
+
+        if (regionCardinality === RegionCardinality.CELLS) {
+            focusCellCoordinates = { col: region.cols[0], row: region.rows[0] };
+        }
+
+        this.props.onFocus(focusCellCoordinates);
 
         if (this.props.selectedRegionTransform != null) {
             region = this.props.selectedRegionTransform(region, event);
--- a/packages/table/src/interactions/selectable.tsx
+++ b/packages/table/src/interactions/selectable.tsx
@@ -7,9 +7,9 @@
 
 import * as PureRender from "pure-render-decorator";
 import * as React from "react";
+import { IFocusedCellCoordinates } from "../common/cell";
 import { DragEvents } from "../interactions/dragEvents";
 import { Draggable, ICoordinateData, IDraggableProps } from "../interactions/draggable";
-import { IFocusedCellCoordinates } from "../layers/focusCell";
 import { IRegion, RegionCardinality, Regions } from "../regions";
 
 export type ISelectedRegionTransform = (region: IRegion, event: MouseEvent, coords?: ICoordinateData) => IRegion;
@@ -23,7 +23,7 @@ export interface ISelectableProps {
     allowMultipleSelection: boolean;
 
     /**
-     * When the user focuses something, this callback is called with a new
+     * When the user focuses something, this callback is called with new
      * focused cell coordinates. This should be considered the new focused cell
      * state for the entire table.
      */
@@ -73,6 +73,23 @@ export class DragSelectable extends React.Component<IDragSelectableProps, {}> {
         return event.button === 0;
     }
 
+    private static getFocusCellCoordinatesFromRegion(region: IRegion) {
+        const regionCardinality = Regions.getRegionCardinality(region);
+
+        switch (regionCardinality) {
+            case RegionCardinality.FULL_TABLE:
+                return { col: 0, row: 0 };
+            case RegionCardinality.FULL_COLUMNS:
+                return { col: region.cols[0], row: 0 };
+            case RegionCardinality.FULL_ROWS:
+                return { col: 0, row: region.rows[0] };
+            case RegionCardinality.CELLS:
+                return { col: region.cols[0], row: region.rows[0] };
+            default:
+                return null;
+        }
+    }
+
     public render() {
         const draggableProps = this.getDraggableProps();
         return (
@@ -101,18 +118,7 @@ export class DragSelectable extends React.Component<IDragSelectableProps, {}> {
             return false;
         }
 
-        const regionCardinality = Regions.getRegionCardinality(region);
-        let focusCellCoordinates: IFocusedCellCoordinates = null;
-
-        if (regionCardinality === RegionCardinality.FULL_TABLE) {
-            focusCellCoordinates = { col: 0, row: 0 };
-        } else if (regionCardinality === RegionCardinality.FULL_COLUMNS) {
-            focusCellCoordinates = { col: region.cols[0], row: 0 };
-        } else if (regionCardinality === RegionCardinality.FULL_ROWS) {
-            focusCellCoordinates = { col: 0, row: region.rows[0] };
-        } else if (regionCardinality === RegionCardinality.CELLS) {
-            focusCellCoordinates = { col: region.cols[0], row: region.rows[0] };
-        }
+        const focusCellCoordinates = DragSelectable.getFocusCellCoordinatesFromRegion(region);
 
         this.props.onFocus(focusCellCoordinates);
 
--- a/packages/table/src/layers/_layers.scss
+++ b/packages/table/src/layers/_layers.scss
@@ -40,6 +40,10 @@ $region-selected-color: $pt-intent-primary !default;
   }
 }
 
+.bp-table-focus-region {
+  border: 2px solid $blue3;
+}
+
 .bp-table-column-headers .bp-table-region {
   border-bottom: none;
 }
--- a/packages/table/src/layers/focusCell.ts
+++ b/packages/table/src/layers/focusCell.ts
@@ -1,0 +1,11 @@
+/**
+ * Copyright 2017 Palantir Technologies, Inc. All rights reserved.
+ * Licensed under the BSD-3 License as modified (the “License”); you may obtain a copy
+ * of the license at https://github.com/palantir/blueprint/blob/master/LICENSE
+ * and https://github.com/palantir/blueprint/blob/master/PATENTS
+ */
+
+export interface IFocusedCellCoordinates {
+    col: number;
+    row: number;
+}
--- a/packages/table/src/regions.ts
+++ b/packages/table/src/regions.ts
@@ -7,6 +7,7 @@
 
 import * as Classes from "./common/classes";
 import { Utils } from "./common/utils";
+import { IFocusedCellCoordinates } from "./layers/focusCell";
 
 /**
  * `Region`s contain sets of cells. Additionally, a distinction is drawn, for
@@ -477,7 +478,11 @@ export class Regions {
         return true;
     }
 
-    public static joinStyledRegionGroups(selectedRegions: IRegion[], otherRegions: IStyledRegionGroup[]) {
+    public static joinStyledRegionGroups(
+        selectedRegions: IRegion[],
+        otherRegions: IStyledRegionGroup[],
+        focusedCellCoordinates: IFocusedCellCoordinates,
+    ) {
         let regionGroups: IStyledRegionGroup[] = [];
         if (otherRegions != null) {
             regionGroups = regionGroups.concat(otherRegions);
@@ -486,6 +491,13 @@ export class Regions {
             regionGroups.push({
                 className: Classes.TABLE_SELECTION_REGION,
                 regions: selectedRegions,
+            });
+        }
+
+        if (focusedCellCoordinates != null) {
+            regionGroups.push({
+                className: Classes.TABLE_FOCUS_REGION,
+                regions: [Regions.cell(focusedCellCoordinates.row, focusedCellCoordinates.col)],
             });
         }
         return regionGroups;
--- a/packages/table/src/regions.ts
+++ b/packages/table/src/regions.ts
@@ -5,9 +5,9 @@
  * and https://github.com/palantir/blueprint/blob/master/PATENTS
  */
 
+import { IFocusedCellCoordinates } from "./common/cell";
 import * as Classes from "./common/classes";
 import { Utils } from "./common/utils";
-import { IFocusedCellCoordinates } from "./layers/focusCell";
 
 /**
  * `Region`s contain sets of cells. Additionally, a distinction is drawn, for
--- a/packages/table/src/regions.ts
+++ b/packages/table/src/regions.ts
@@ -481,7 +481,7 @@ export class Regions {
     public static joinStyledRegionGroups(
         selectedRegions: IRegion[],
         otherRegions: IStyledRegionGroup[],
-        focusedCellCoordinates: IFocusedCellCoordinates,
+        focusedCell: IFocusedCellCoordinates,
     ) {
         let regionGroups: IStyledRegionGroup[] = [];
         if (otherRegions != null) {
@@ -494,10 +494,10 @@ export class Regions {
             });
         }
 
-        if (focusedCellCoordinates != null) {
+        if (focusedCell != null) {
             regionGroups.push({
                 className: Classes.TABLE_FOCUS_REGION,
-                regions: [Regions.cell(focusedCellCoordinates.row, focusedCellCoordinates.col)],
+                regions: [Regions.cell(focusedCell.row, focusedCell.col)],
             });
         }
         return regionGroups;
--- a/packages/table/src/table.tsx
+++ b/packages/table/src/table.tsx
@@ -1103,7 +1103,6 @@ export class Table extends AbstractComponent<ITableProps, ITableState> {
     }
 
     private handleFocusMove = (e: KeyboardEvent, direction: "up" | "down" | "left" | "right") => {
-
         e.preventDefault();
         e.stopPropagation();
 
--- a/packages/table/src/table.tsx
+++ b/packages/table/src/table.tsx
@@ -47,7 +47,7 @@ export interface ITableProps extends IProps, IRowHeights, IColumnWidths {
      * spreadsheet.
      * @default false
      */
-    allowFocus?: boolean;
+    enableFocus?: boolean;
 
     /**
      * If `false`, only a single region of a single column/row/cell may be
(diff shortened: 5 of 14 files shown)


packages/table/src/table.tsx
```diff
@@ -316,8 +344,16 @@ export class Table extends AbstractComponent<ITableProps, ITableState> {
 
         const selectedRegions = (props.selectedRegions == null) ? [] as IRegion[] : props.selectedRegions;
 
+        let focusedCellCoordinates: IFocusedCellCoordinates = null;
```
adidahiya: try not to initialize anything to `null` unless required by a 3rd party API, use `undefined` instead

packages/table/src/layers/focusCell.ts
```diff
@@ -0,0 +1,11 @@
+/**
+ * Copyright 2017 Palantir Technologies, Inc. All rights reserved.
+ * Licensed under the BSD-3 License as modified (the “License”); you may obtain a copy
+ * of the license at https://github.com/palantir/blueprint/blob/master/LICENSE
+ * and https://github.com/palantir/blueprint/blob/master/PATENTS
+ */
+
+export interface IFocusedCellCoordinates {
```
adidahiya: does this need to have "focus" semantics in it? why not just `ICellCoordinates`?

themadcreator: Agreed. Just `ICellCoordinate` would work.

This could also be just a tuple of numbers. like `type ICellCoordinate = [number, number]` but it might be easy to forget the order [row, col] vs [col, row].

packages/table/src/regions.ts
```diff
@@ -422,6 +427,13 @@ export class Regions {
                 regions: selectedRegions,
             });
         }
+
+        if (focusedCellCoordinates != null) {
+            regionGroups.push({
+                className: Classes.TABLE_FOCUS_REGION,
+                regions: [Regions.cell(focusedCellCoordinates.row, focusedCellCoordinates.col)],
+            });
+        }
```
themadcreator: Yep, pretty much how i expect this to work.

packages/table/src/table.tsx
```diff
@@ -41,6 +42,14 @@ import { TableBody } from "./tableBody";
 
 export interface ITableProps extends IProps, IRowHeights, IColumnWidths {
     /**
+     * If `true`, there will be a single "focused" cell at all times, 
+     * which can be used to interact with the table as though it is a
+     * spreadsheet.
+     * @default false
```
themadcreator: Since we enable selection by default, should we also enable focusing?

packages/table/src/table.tsx
```diff
@@ -159,11 +173,19 @@ export interface ITableProps extends IProps, IRowHeights, IColumnWidths {
     numRows?: number;
 
     /**
+     * If defined, will set the focused cell state. If defined, this
+     * changes the focused cell to controlled mode, meaning you in charge of
+     * setting the focus in response to events in the `onFocus`
+     * callback.
+     */
+    focusedCellCoordinates?: IFocusedCellCoordinates;
```
themadcreator: would simply `focusedCell` be clear enough?

packages/table/src/table.tsx
```diff
@@ -316,8 +344,16 @@ export class Table extends AbstractComponent<ITableProps, ITableState> {
 
         const selectedRegions = (props.selectedRegions == null) ? [] as IRegion[] : props.selectedRegions;
 
+        let focusedCellCoordinates: IFocusedCellCoordinates = null;
+        if (props.focusedCellCoordinates != null) {
+            focusedCellCoordinates = props.focusedCellCoordinates;
+        } else if (props.allowFocus) {
+            focusedCellCoordinates = { col: 0, row: 0 };
```
themadcreator: you're not guaranteed to actually have any cells. You can set numRows to 0

themadcreator: Also, i can unfocus/blur a cell, so maybe this should be nullable

packages/table/src/table.tsx
gscshoyru: I'm 👎 -- and this is because the focus cell is probably _not_ what most consumers of a table library are expecting. If we turn this on-by-default then upon upgrade all of our consumers will suddenly have a focus cell -- whether they want to or not.

packages/table/src/layers/focusCell.ts
gscshoyru: Hm...

So at some point I need to add a third number (or property?) to this to identify which selected region the focused cell belongs to. I could make a type that extends the ICellCoordinates type -- but I am still going to need that third property, and another name for the thing.

packages/table/src/table.tsx
gscshoyru: I mean, it's not the actual cell? 

You're probably right though.

packages/table/src/table.tsx
gscshoyru: Do we want to unfocus a cell? I can't think of a good usecase -- spreadsheets don't let you do this, they always have one.

packages/table/src/table.tsx
gscshoyru: Makes sense.

packages/table/src/table.tsx
giladgray: agreed, seems safer to ship with it off.

packages/table/src/table.tsx
giladgray: similarly, is it possible to focus more than one cell? like a rectangular group?

packages/table/src/layers/focusCell.ts
adidahiya: cool, then I would recommend extending `ICellCoordinates` with an empty interface and leaving a comment explaining that more things are coming

tgreenwatts:
A few questions from me:
This only applies to using the arrow keys correct? No shift, tab, enter, etc 
This doesn't work on the "Column Type Formats" table, expected? Same goes for Add/Remove and Named examples
The editable table I can actually tab between cells, but it doesn't update the focus cell. I can't use the arrow keys - this is expected because editable cells right?
Does this do anything with scrolling? On the Big Table, I arrowed out of the current view and expected it to scroll the table but it didn't

packages/table/src/interactions/selectable.tsx
```diff
@@ -9,7 +9,8 @@ import * as PureRender from "pure-render-decorator";
 import * as React from "react";
 import { DragEvents } from "../interactions/dragEvents";
 import { Draggable, ICoordinateData, IDraggableProps } from "../interactions/draggable";
-import { IRegion, Regions } from "../regions";
+import { IFocusedCellCoordinates } from "../layers/focusCell";
+import { IRegion, RegionCardinality, Regions  } from "../regions";
```
cmslewis: nit: delete extra space before `}`

packages/table/src/interactions/selectable.tsx
```diff
@@ -22,6 +23,13 @@ export interface ISelectableProps {
     allowMultipleSelection: boolean;
 
     /**
+     * When the user focuses something, this callback is called with a new
+     * focused cell coordinates. This should be considered the new focused cell
+     * state for the entire table.
```
cmslewis: We can polish this language a bit to make it more clear. Maybe:

```
When the user focuses a cell, this callback is called with the new
focused-cell coordinates. This should be considered the new
focused-cell state for the entire table.
```

packages/table/src/interactions/selectable.tsx
```diff
@@ -93,6 +101,27 @@ export class DragSelectable extends React.Component<IDragSelectableProps, {}> {
...
+            focusCellCoordinates = { col: region.cols[0], row: 0 };
+        }
+
+        if (regionCardinality === RegionCardinality.FULL_ROWS) {
+            focusCellCoordinates = { col: 0, row: region.rows[0] };
+        }
+
+        if (regionCardinality === RegionCardinality.CELLS) {
+            focusCellCoordinates = { col: region.cols[0], row: region.rows[0] };
+        }
```
cmslewis: Can we do `if/else-if` for all of these? Doesn't have the nice whitespace between, but makes it clearer that these are all disjoint cases.

packages/table/src/interactions/selectable.tsx
```diff
@@ -93,6 +101,27 @@ export class DragSelectable extends React.Component<IDragSelectableProps, {}> {
...
 
+        const regionCardinality = Regions.getRegionCardinality(region);
+        let focusCellCoordinates: IFocusedCellCoordinates = null;
+
+        if (regionCardinality === RegionCardinality.FULL_TABLE) {
+            focusCellCoordinates = { col: 0, row: 0 };
+        }
+
+        if (regionCardinality === RegionCardinality.FULL_COLUMNS) {
+            focusCellCoordinates = { col: region.cols[0], row: 0 };
```
cmslewis: Just want to make sure I understand: when we've selected a region with `FULL_COLUMNS`, is the top-left cell _in the body_ returned as the focused cell, or is the top-left _column-header cell_ returned? Same question for the `FULL_ROWS` case below.

packages/table/src/layers/focusCell.ts
cmslewis: ping on this @gscshoyru

llorca:
> The editable table I can actually tab between cells, but it doesn't update the focus cell. I can't use the arrow keys - this is expected because editable cells right?

In this case, I'd say tab should go to the next cell, and that next cell should get out of edit mode. You'll be able to get back into edit mode simply by typing characters. For enter, I'd say the same thing, but then it becomes tricky when you need multiline content (note that neither Excel or Spreadsheet seem to support typing in multiline content... or am I missing something?)

So if we do need to support multiline content, maybe we have the first tab/enter that gets you out of edit mode (while remaining on the same cell), and the next tab/enter moves you to the next cell? Then it frees up shift+enter from the editable text component to write multiline content while editing?

packages/table/src/table.tsx
```diff
@@ -159,6 +173,14 @@ export interface ITableProps extends IProps, IRowHeights, IColumnWidths {
     numRows?: number;
 
     /**
+     * If defined, will set the focused cell state. If defined, this
+     * changes the focused cell to controlled mode, meaning you in charge of
```
cmslewis: typo: `meaning you *are* in charge of`

packages/table/src/table.tsx
```diff
@@ -159,6 +173,14 @@ export interface ITableProps extends IProps, IRowHeights, IColumnWidths {
     numRows?: number;
 
     /**
+     * If defined, will set the focused cell state. If defined, this
```
cmslewis: Awkward to say `If defined,` twice. Better wording here?

packages/table/src/table.tsx
cmslewis: I mildly prefer `focusedCell` but would be fine either way.

packages/table/src/table.tsx
cmslewis: ping @gscshoyru

packages/table/src/table.tsx
cmslewis: I wouldn't think it's possible to focus more than one cell. e.g. If you click + drag to select a region of cells, the focused cell would be the cell on which you clicked to initiate the drag.

packages/table/src/table.tsx
```diff
@@ -357,12 +394,15 @@ export class Table extends AbstractComponent<ITableProps, ITableState> {
         newRowHeights = Utils.assignSparseValues(newRowHeights, rowHeights);
 
         const newSelectedRegions = (selectedRegions == null) ? this.state.selectedRegions : selectedRegions;
+        const newFocusedCellCoordinates =
+            (focusedCellCoordinates == null) ? this.state.focusedCellCoordinates : focusedCellCoordinates;
```
cmslewis: nit: when ternaries exceed the max line length, I tend to prefer:

```
const newFocusedCellCoordinates = (focusedCellCoordinates == null)
    ? this.state.focusedCellCoordinates
    : focusedCellCoordinates;
```

so you can read both cases easily.

packages/table/src/table.tsx
```diff
@@ -390,7 +430,8 @@ export class Table extends AbstractComponent<ITableProps, ITableState> {
     }
 
     public renderHotkeys() {
-        const hotkeys = [this.maybeRenderCopyHotkey(), this.maybeRenderSelectAllHotkey()];
+        const hotkeys =
+            [this.maybeRenderCopyHotkey(), this.maybeRenderSelectAllHotkey(), this.maybeRenderFocusHotkeys()];
```
cmslewis: nit: would prefer this formatting at this point:

```
const hotkeys = [
    this.maybeRenderCopyHotkey(),
    this.maybeRenderSelectAllHotkey(),
    this.maybeRenderFocusHotkeys(),
];
```

packages/table/src/table.tsx
```diff
@@ -1004,6 +1092,53 @@ export class Table extends AbstractComponent<ITableProps, ITableState> {
         this.handleSelection([]);
     }
 
+    private handleFocusMove = (e: KeyboardEvent, direction: "up" | "down" | "left" | "right") => {
```
cmslewis: Hmm, should we create a `RelativeDirection` enum or something? May not be necessary, but feels 1% cleaner.

packages/table/src/table.tsx
```diff
@@ -1004,6 +1092,53 @@ export class Table extends AbstractComponent<ITableProps, ITableState> {
...
+        e.preventDefault();
+        e.stopPropagation();
+
+        const {focusedCellCoordinates} = this.state;
```
cmslewis: nit: use 1-space padding when destructuring: `{ focusedCellCoordinates }`. @adidahiya is there a `tslint` rule for this already? if not, worth adding one?

packages/table/src/table.tsx
```diff
@@ -1004,6 +1092,53 @@ export class Table extends AbstractComponent<ITableProps, ITableState> {
...
+        const newFocusedCellCoordinates = { col: focusedCellCoordinates.col, row: focusedCellCoordinates.row };
+        const { grid } = this;
+        if (direction === "up") {
+            newFocusedCellCoordinates.row--;
+        }
+        if (direction === "down") {
```
cmslewis: I'd prefer `if/else-if`s for the direction checks, since they're all disjoint.

packages/table/src/table.tsx
```diff
@@ -1004,6 +1092,53 @@ export class Table extends AbstractComponent<ITableProps, ITableState> {
...
+            newFocusedCellCoordinates.row++;
+        }
+        if (direction === "left") {
+            newFocusedCellCoordinates.col--;
+        }
+        if (direction === "right") {
+            newFocusedCellCoordinates.col++;
+        }
+        if (newFocusedCellCoordinates.row < 0 || newFocusedCellCoordinates.row >= grid.numRows ||
```
cmslewis: And I'd prefer a newline before this to communicate that this `if` statement is a separate thing.

packages/table/src/table.tsx
```diff
@@ -1004,6 +1092,53 @@ export class Table extends AbstractComponent<ITableProps, ITableState> {
...
 
+    private handleFocusMove = (e: KeyboardEvent, direction: "up" | "down" | "left" | "right") => {
+        e.preventDefault();
+        e.stopPropagation();
+
+        const {focusedCellCoordinates} = this.state;
+        const newFocusedCellCoordinates = { col: focusedCellCoordinates.col, row: focusedCellCoordinates.row };
+        const { grid } = this;
+        if (direction === "up") {
+            newFocusedCellCoordinates.row--;
```
cmslewis: nit: I've worked on codebases where `++` and `--` were considered an antipattern unless used in a `for` loop. I tend to prefer an explicit `+= 1` and `-= 1`.

packages/table/src/table.tsx
```diff
@@ -1004,6 +1092,53 @@ export class Table extends AbstractComponent<ITableProps, ITableState> {
...
+        }
+
+        // only set focusedRegion state if not specified in props
+        if (this.props.focusedCellCoordinates == null) {
+            this.setState({ focusedCellCoordinates } as ITableState);
+        }
+
+        const { onFocus } = this.props;
+        if (onFocus != null) {
+            onFocus(focusedCellCoordinates);
```
cmslewis: Use `Utils.safeInvoke` (from `packages/core`). Then this becomes simply:

```
Utils.safeInvoke(this.props.onFocus, focusedCellCoordinates);
```

gscshoyru:
@tgreenwatts just arrow keys for right now. Any other behavior is based on already existing stuff that will probably need to be included/subsumed by this behavior. 

The tables that don't have a focus cell don't have allowFocus turned on on them.

It does not currently scroll as you move around. It probably should at some point in the very near future, but as present this isn't implemented.

tgreenwatts:
Cool filed the 2 follow up pieces then so it's clear they're unfinished: #915 and #914

packages/table/src/interactions/selectable.tsx
gscshoyru: You can't focus the header cells (just like in excel, etc). It'll return the top-left cell, in the body, that corresponds with the thing you just clicked on. Same behavior as excel, etc.

packages/table/src/table.tsx
adidahiya: Just remove the second "If defined". "This changes...".

packages/table/src/table.tsx
adidahiya: this would be clearer as a `switch` statement

packages/table/src/table.tsx
adidahiya: worth adding to the `whitespace` rule, yeah. we check import statements but not destructuring

packages/table/src/interactions/selectable.tsx
themadcreator: or how about a switch?

packages/table/src/table.tsx
```diff
@@ -41,6 +42,14 @@ import { TableBody } from "./tableBody";
 
 export interface ITableProps extends IProps, IRowHeights, IColumnWidths {
     /**
+     * If `true`, there will be a single "focused" cell at all times,
+     * which can be used to interact with the table as though it is a
+     * spreadsheet.
+     * @default false
+     */
+    allowFocus?: boolean;
```
themadcreator: I think i prefer the name `enableFocus` or `focusable`. Is there other precedent for `allow*`?

packages/table/src/table.tsx
```diff
@@ -260,12 +282,18 @@ export interface ITableState {
      * An array of Regions representing the selections of the table.
      */
     selectedRegions?: IRegion[];
+
+    /**
+     * The coordinates of the currently focused table cell
+     */
+    focusedCellCoordinates?: IFocusedCellCoordinates;
```
themadcreator: Same comment as above. I'm not too adamant, though

packages/table/src/table.tsx
themadcreator: Focus is always at most one cell

packages/table/src/table.tsx
themadcreator: meh, it's not like the single-instruction inc/dec is ambiguous or difficult to understand. it's only during assignment like `tmp = other++;` when things go wrong.

packages/table/src/table.tsx
```diff
@@ -1004,6 +1092,53 @@ export class Table extends AbstractComponent<ITableProps, ITableState> {
...
+        }
+        if (direction === "left") {
+            newFocusedCellCoordinates.col--;
+        }
+        if (direction === "right") {
+            newFocusedCellCoordinates.col++;
+        }
+        if (newFocusedCellCoordinates.row < 0 || newFocusedCellCoordinates.row >= grid.numRows ||
+            newFocusedCellCoordinates.col < 0 || newFocusedCellCoordinates.col >= grid.numCols) {
+            return;
```
themadcreator: what are the implications of not invoking the callbacks here? do we want to notify listeners that focus was removed?

packages/table/src/table.tsx
gscshoyru: allowMultipleSelection is also a thing -- but I can change this regardless.

packages/table/src/table.tsx
gscshoyru: It's not removed, it's just not changed. Same thing happens with selection, I think -- if there's no change to these things there's nothing to call the callbacks with.

packages/table/src/interactions/selectable.tsx
```diff
@@ -22,6 +23,13 @@ export interface ISelectableProps {
     allowMultipleSelection: boolean;
 
     /**
+     * When the user focuses something, this callback is called with a new
+     * focused cell coordinates. This should be considered the new focused cell
```
giladgray: does not compute: "a new focused cell coordinates."

packages/table/src/interactions/selectable.tsx
```diff
@@ -93,6 +101,21 @@ export class DragSelectable extends React.Component<IDragSelectableProps, {}> {
...
+        const regionCardinality = Regions.getRegionCardinality(region);
+        let focusCellCoordinates: IFocusedCellCoordinates = null;
+
+        if (regionCardinality === RegionCardinality.FULL_TABLE) {
+            focusCellCoordinates = { col: 0, row: 0 };
+        } else if (regionCardinality === RegionCardinality.FULL_COLUMNS) {
+            focusCellCoordinates = { col: region.cols[0], row: 0 };
+        } else if (regionCardinality === RegionCardinality.FULL_ROWS) {
+            focusCellCoordinates = { col: 0, row: region.rows[0] };
+        } else if (regionCardinality === RegionCardinality.CELLS) {
```
giladgray: i'd pull this out to a helper function and implement with a `switch` that simply `return`s in each case.

packages/table/src/layers/_layers.scss
```diff
@@ -40,6 +40,10 @@ $region-selected-color: $pt-intent-primary !default;
   }
 }
 
+.bp-table-focus-region {
+  border: 2px solid $blue3;
```
giladgray: wonder if we've got variables for this. `$pt-intent-primary` at least? @llorca ?

packages/table/src/table.tsx
```diff
@@ -41,6 +42,14 @@ import { TableBody } from "./tableBody";
 
 export interface ITableProps extends IProps, IRowHeights, IColumnWidths {
     /**
+     * If `true`, there will be a single "focused" cell at all times,
+     * which can be used to interact with the table as though it is a
+     * spreadsheet.
```
giladgray: needs better docs for `false` case. 

how does the focus cell behave if this is `false`? is it even supported?

packages/table/src/table.tsx
```diff
@@ -789,6 +834,49 @@ export class Table extends AbstractComponent<ITableProps, ITableState> {
         }
     }
 
+    private _handleFocusMoveLeft = (e: KeyboardEvent) => this.handleFocusMove(e, "left");
+    private _handleFocusMoveRight = (e: KeyboardEvent) => this.handleFocusMove(e, "right");
+    private _handleFocusMoveUp = (e: KeyboardEvent) => this.handleFocusMove(e, "up");
+    private _handleFocusMoveDown = (e: KeyboardEvent) => this.handleFocusMove(e, "down");
```
giladgray: no need for `_` prefix, we don't use this convention in the codebase because `private` is typically enough.

packages/table/src/table.tsx
```diff
@@ -1004,6 +1092,56 @@ export class Table extends AbstractComponent<ITableProps, ITableState> {
...
+        this.handleFocus(newFocusedCell);
+    }
+
+    private handleFocus = (focusedCell: IFocusedCellCoordinates) => {
+        if (!this.props.enableFocus) {
+            // don't set focus state if focus is not allowed
+            return;
+        }
+
+        // only set focusedRegion state if not specified in props
```
giladgray: "focusedCell state" ?

packages/table/test/selectableTests.tsx
```diff
@@ -31,13 +31,15 @@ describe("DragSelectable", () => {
 
     it("does click selection", () => {
         const onSelection = sinon.spy();
+        const onFocus = sinon.spy();
```
giladgray: do we actually need these spies? considering that not a single one of them seems to be used?

seems to me you could revert this whole file.

packages/table/src/table.tsx
gscshoyru: There is no focus cell if it is false. Focus doesn't exist if it's false.

packages/table/test/selectableTests.tsx
gscshoyru: onFocus is required much the same way that onSelection is required -- there needs to be something specified here regardless. 

Should probably be testing them though.

packages/table/src/table.tsx
giladgray: unless the user sets `focusedCell` and puts it in controlled mode, it seems.

packages/table/test/selectableTests.tsx
giladgray: why is it required??

gscshoyru: Selection can be disabled as well. The determination of whether or not the focus or selection is valid happens in the table parent -- it does the filtering. Not the child. At least that's how the selection logic works, which I borrowed from.

gscshoyru: Also it's a lot more straightforward to always pass down a callback and filter at the top than it is to conditionally pass down a callback and call it depending on existence at the bottom.

packages/table/src/table.tsx
gscshoyru: ...true. 

Suggestions here? Currently the combination of the two does inconsistent things. None of the callbacks work unless it's turned on. Should we just ignore the controlled mode version if it's false? Actually that seems most consistent I'm just going to do that.

packages/table/test/selectableTests.tsx
giladgray: ok cuz this is an internal component?

giladgray:
@gscshoyru resolve merge conflicts please

cmslewis:
See here:

gscshoyru:
@cmslewis essentially, already tracked here: https://github.com/palantir/blueprint/issues/843

PR #728, opened 2017-02-24
Update Sketch Kit

llorca:
- Remove old fonts
- Add callouts

image


cmslewis:
The intented callout titles are a different color from the intented callout icons. Per #698, should they be the same?

PR #812, opened 2017-03-09
Non ideal docs

ryanmcnamara:
It would appear that the docs for this component were
....
8adea640-0ea2-0132-70b0-0add9426c766
....
"non-ideal"


adidahiya:
lol

PR #822, opened 2017-03-10
Add select all styles and interactions to table

gscshoyru:

Add styles for select-all for table

Added select-all interactions -- clicking on the upper left hand corner, or pressing mod+a 

The "display: table" style change is weird because I have no idea why it works -- does anyone know a better way to make it have the height of its child? 

The "table" method in regions.ts is one of the most hillarious functions I've ever written. 

I added "full table" selection to all -- should we add a ROWS_COLUMNS_AND_CELLS selection mode to the thing? Or leave as-is? 

#### Screenshot

mar-10-2017 14-12-46

### DIFF
--- a/packages/table/src/regions.ts
+++ b/packages/table/src/regions.ts
@@ -41,6 +41,7 @@ export enum RegionCardinality {
  */
 export const SelectionModes = {
     ALL: [
+        RegionCardinality.FULL_TABLE,
         RegionCardinality.FULL_COLUMNS,
         RegionCardinality.FULL_ROWS,
         RegionCardinality.CELLS,
@@ -161,6 +162,13 @@ export class Regions {
      */
     public static column(col: number, col2?: number): IRegion  {
         return { cols: this.normalizeInterval(col, col2) };
+    }
+
+    /**
+     * Returns a region containing the entire table.
+     */
+    public static table(): IRegion  {
+        return {};
     }
 
     /**
--- a/packages/table/src/regions.ts
+++ b/packages/table/src/regions.ts
@@ -119,6 +119,17 @@ export class Regions {
      *     }
      *
      * In this case, this method would return RegionCardinality.FULL_COLUMNS.
+     *
+     * If both rows and columns are unbounded, then the region covers the
+     * entire table. Therefore, a region like this:
+     *
+     *     {
+     *         rows: null,
+     *         cols: null
+     *     }
+     *
+     * will return RegionCardinality.FULL_TABLE.
+     *
      * An example of a region containing a single cell in the table would be:
      *
      *     {
--- a/packages/table/src/regions.ts
+++ b/packages/table/src/regions.ts
@@ -118,7 +118,7 @@ export class Regions {
      *         cols: [1, 2]
      *     }
      *
-     * In this case, this method would return RegionCardinality.FULL_COLUMNS.
+     * In this case, this method would return `RegionCardinality.FULL_COLUMNS`.
      *
      * If both rows and columns are unbounded, then the region covers the
      * entire table. Therefore, a region like this:
@@ -128,7 +128,7 @@ export class Regions {
      *         cols: null
      *     }
      *
-     * will return RegionCardinality.FULL_TABLE.
+     * will return `RegionCardinality.FULL_TABLE`.
      *
      * An example of a region containing a single cell in the table would be:
      *
@@ -137,7 +137,7 @@ export class Regions {
      *         cols: [2, 2]
      *     }
      *
-     * In this case, this method would return RegionCardinality.CELLS.
+     * In this case, this method would return `RegionCardinality.CELLS`.
      */
     public static getRegionCardinality(region: IRegion) {
         if (region.cols != null && region.rows != null) {
--- a/packages/table/src/table.scss
+++ b/packages/table/src/table.scss
@@ -88,6 +88,9 @@ $menu-z-index: $row-z-index + 1 !default;
 }
 
 .bp-table-row-headers {
+  // let header's height resize based on height of child element
+  display: table;
+
   flex: 0 0 auto;
   // allow absolute position for region overlay
   position: relative;
--- a/packages/table/src/table.tsx
+++ b/packages/table/src/table.tsx
@@ -386,9 +386,14 @@ export class Table extends AbstractComponent<ITableProps, ITableState> {
     }
 
     public renderHotkeys() {
-        return <Hotkeys>
-            {this.maybeRenderCopyHotkey()}
-        </Hotkeys>;
+        let hotkeys = [this.maybeRenderCopyHotkey(), this.maybeRenderSelectAllHotkey()];
+        // remove null elements
+        hotkeys = hotkeys.filter((element) => element != null);
+        return (
+            <Hotkeys>
+                {hotkeys}
+            </Hotkeys>
+        );
     }
 
     /**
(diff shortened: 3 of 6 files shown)


packages/table/src/table.scss
```diff
@@ -88,6 +88,9 @@ $menu-z-index: $row-z-index + 1 !default;
 }
 
 .bp-table-row-headers {
+  // let header's height resize based on height of child element
+  display: table;
```
adidahiya: does this obviate the `flex` style below?

packages/table/src/table.tsx
```diff
@@ -386,9 +386,14 @@ export class Table extends AbstractComponent<ITableProps, ITableState> {
     }
 
     public renderHotkeys() {
-        return <Hotkeys>
-            {this.maybeRenderCopyHotkey()}
-        </Hotkeys>;
+        let hotkeys = [this.maybeRenderCopyHotkey(), this.maybeRenderSelectAllHotkey()];
+        // remove null elements
+        hotkeys = hotkeys.filter((element) => element != null);
```
adidahiya: just inline in JSX below and use undefined instead of null: `{hotkeys.filter((element) => element !== undefined)}`

packages/table/src/table.scss
gscshoyru: No; at least I don't think so -- display: flex is parent behavior, flex behavior defines behavior of the child. And display: table does behavior of children again. 

That said: if you can figure out how to make the row headers take on the size of its children the way the column headers does, using flexbox rather than this, I would prefer to use that instead.

llorca:
Went through styling with Geoff yesterday, the least we can do is add a `cursor: se-resize;` on hover to the very top left area

llorca:
Also, if possible, let's try to get the top left area included in the blue selection region

packages/table/src/table.tsx
themadcreator: Doesn't our linter reject lambdas in JSX?

packages/table/src/regions.ts
```diff
@@ -164,6 +165,13 @@ export class Regions {
     }
 
     /**
+     * Returns a region containing the entire table.
+     */
+    public static table(): IRegion  {
+        return {};
```
themadcreator: Seems like this might make some headaches for us in the future. Can you please update the docs examples about regions?

themadcreator: This one: https://github.com/palantir/blueprint/blob/master/packages/table/src/regions.ts#L110

tgreenwatts:
probably out of scope of this PR, but should the selected line be within the border or outside? @llorca ? 
image


Also why is there a dot in the upper left corner box?

gscshoyru:
...Because the method used to hide the border at the top of the row headers has a hole in it. Nice catch. I'll fix.

llorca:
@tgreenwatts within, how it is now is the only option. We clip everything beyond that

packages/table/src/regions.ts
```diff
@@ -119,6 +119,17 @@ export class Regions {
...
+     *
+     * If both rows and columns are unbounded, then the region covers the
+     * entire table. Therefore, a region like this:
+     *
+     *     {
+     *         rows: null,
+     *         cols: null
+     *     }
+     *
+     * will return RegionCardinality.FULL_TABLE.
```
giladgray: backticks around `RegionCardinality.FULL_TABLE`

packages/table/src/regions.ts
```diff
@@ -119,6 +119,17 @@ export class Regions {
      *     }
      *
      * In this case, this method would return RegionCardinality.FULL_COLUMNS.
```
giladgray: backticks here too. and maybe elsewhere...

packages/table/src/table.tsx
```diff
@@ -386,12 +386,11 @@ export class Table extends AbstractComponent<ITableProps, ITableState> {
...
 
     public renderHotkeys() {
-        let hotkeys = [this.maybeRenderCopyHotkey(), this.maybeRenderSelectAllHotkey()];
-        // remove null elements
-        hotkeys = hotkeys.filter((element) => element != null);
+        const hotkeys = [this.maybeRenderCopyHotkey(), this.maybeRenderSelectAllHotkey()];
         return (
             <Hotkeys>
-                {hotkeys}
+                {/* filter to remove null elements */}
```
giladgray: unnecessary comment, i think. this is what your `filter` does.

packages/table/src/table.tsx
```diff
@@ -826,6 +830,33 @@ export class Table extends AbstractComponent<ITableProps, ITableState> {
...
+            }
+            const cardinality = Regions.getRegionCardinality(region);
+            const style = grid.getRegionStyle(region);
+
+            switch (cardinality) {
+                case RegionCardinality.FULL_TABLE:
+                    style.right = "0px";
+                    style.bottom = "0px";
+                    style.top = "0px";
+                    style.left = "0px";
```
giladgray: use numeral `0` instead of string 
https://stylelint.io/user-guide/rules/length-zero-no-unit/

packages/table/test/tableTests.tsx
```diff
@@ -87,6 +87,27 @@ describe("<Table>", () => {
         rowHeaders.forEach((rowHeader) => expectCellLoading(rowHeader, CellType.ROW_HEADER));
     });
 
+    it("Selects all on click of upper-left corner", () => {
+        const renderCell = () => {
+            return <Cell>gg</Cell>;
+        };
```
giladgray: inline. `const renderCell = () => <Cell>gg</Cell>;`

packages/table/src/regions.ts
```diff
@@ -119,6 +119,17 @@ export class Regions {
...
+     *
+     * If both rows and columns are unbounded, then the region covers the
+     * entire table. Therefore, a region like this:
+     *
+     *     {
+     *         rows: null,
+     *         cols: null
```
giladgray: @themadcreator does selection explicitly speak `null`? or does it do `undefined` too?

packages/table/src/table.tsx
giladgray: pretty sure this entire file could use this change, so maybe it's not worth making here just yet.

giladgray: 🤔 hmmm pretty sure this entire file could use this change because React will add `"px"` to `number` values in CSS. 

so maybe it's not worth making here just yet. this is consistent, to be sure.

packages/table/src/regions.ts
gscshoyru: undefined/property not existing works too, indeed that's what I'm doing. Should we document it that way instead?

packages/table/src/table.tsx
gscshoyru: I mean, I kindof wanted to say _why_ I need to do that. But yeah, the comment is kindof useless.

tgreenwatts:
Out of scope but I don't think this looks selected in dark-theme 
image

tgreenwatts:
Also @llorca - how else does a user deselect? This was a large problem in the old pre-blueprint-table pre-react Contour table where users wanted to deselect by clicking again and couldn't

packages/table/src/regions.ts
giladgray: i'm not sure, but that is why i asked. doesn't the code use an empty object `{}`? perhaps we should document that instead.

giladgray:
@themadcreator re-review please?

llorca:
@tgreenwatts just like you'd do it in Excel or Spreadsheet 👍

tgreenwatts:
I think those are spreadsheets not tables, so you can't - in a table you want to be able to deselect still

themadcreator:
Looks great! 10/10

PR #830, opened 2017-03-13
[DRP] Assign right and left together, keep them 1 month apart

### LINKED ISSUE #829: Shortcut not displayed if current month isn't already view
### Bug report

- __Package version(s)__: ^1.6.0
- __Browser and OS versions__: Any

#### Steps to reproduce

1. Go here: http://blueprintjs.com/docs/#components.datetime.daterangepicker
1. Click "Past 3 months", then "Past week"
1. See that no date range is visible in range selector.

#### Actual behavior

Upon selecting "Past week" shortcut, nothing is highlighted. But the value is accurate.

#### Expected behavior

Should jump to current month and show selected range.

omegdadi:

1. During getStateChange, when setting a date range that spans months, left view and right view are kept only 1 month apart. Previous to this fix, the right view would get set to the end month of the range even though the start month and the following month are what gets displayed in the component. This then causes a no-op the following round trip through when selecting a single date range that lies within the previous range's end month.

Behavior with other settings like `contiguousCalendarMonths`, but that is not taken into account throughout the existing code.

#### Screenshots

Selected "Past 3 months"
screen shot 2017-03-12 at 6 20 26 pm

Selected "Past week"
screen shot 2017-03-12 at 6 20 34 pm

### DIFF
--- a/packages/datetime/src/dateRangePicker.tsx
+++ b/packages/datetime/src/dateRangePicker.tsx
@@ -279,7 +279,8 @@ export class DateRangePicker
     public componentWillReceiveProps(nextProps: IDateRangePickerProps) {
         super.componentWillReceiveProps(nextProps);
 
-        const nextState = getStateChange(this.props.value, nextProps.value, this.state, nextProps.contiguousCalendarMonths);
+        const nextState = getStateChange(this.props.value, nextProps.value, this.state,
+            nextProps.contiguousCalendarMonths);
         this.setState(nextState);
     }
 
--- a/packages/datetime/src/dateRangePicker.tsx
+++ b/packages/datetime/src/dateRangePicker.tsx
@@ -619,7 +619,7 @@ export class DateRangePicker
 function getStateChange(value: DateRange,
                         nextValue: DateRange,
                         state: IDateRangePickerState,
-                        contiguousCalendarMonths: Boolean): IDateRangePickerState {
+                        contiguousCalendarMonths: boolean): IDateRangePickerState {
     let returnVal: IDateRangePickerState;
 
     if (value != null && nextValue == null) {
--- a/packages/datetime/src/dateRangePicker.tsx
+++ b/packages/datetime/src/dateRangePicker.tsx
@@ -279,7 +279,7 @@ export class DateRangePicker
     public componentWillReceiveProps(nextProps: IDateRangePickerProps) {
         super.componentWillReceiveProps(nextProps);
 
-        const nextState = getStateChange(this.props.value, nextProps.value, this.state);
+        const nextState = getStateChange(this.props.value, nextProps.value, this.state, nextProps);
         this.setState(nextState);
     }
 
@@ -515,7 +515,7 @@ export class DateRangePicker
 
     private handleNextState(nextValue: DateRange) {
         const { value } = this.state;
-        const nextState = getStateChange(value, nextValue, this.state);
+        const nextState = getStateChange(value, nextValue, this.state, this.props);
 
         if (!this.isControlled) {
             this.setState(nextState);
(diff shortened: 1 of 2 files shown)


palantirtech:
Thanks for your interest in [palantir/blueprint](https://github.com/palantir/blueprint), @omegdadi! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

omegdadi:
Got tests running locally now, please ignore until next update.

omegdadi:
@giladgray This is now up-to-date with a test for the behavior. Thanks for reviewing.

packages/datetime/src/dateRangePicker.tsx
```diff
@@ -691,9 +691,10 @@ function getStateChange(value: DateRange,
             } else {
                 if (!leftView.isSame(nextValueStartMonthAndYear)) {
                     leftView = nextValueStartMonthAndYear;
-                }
-                if (!rightView.isSame(nextValueEndMonthAndYear)) {
+                    rightView = nextValueStartMonthAndYear.getNextMonth();
+                } else if (!rightView.isSame(nextValueEndMonthAndYear)) {
                     rightView = nextValueEndMonthAndYear;
+                    leftView = rightView.getPreviousMonth();
```
leebyp: I'd use the same pattern in line 694 and 697 in setting the "other" so it's obvious what's going on

leebyp:
@omegdadi There's something funky going on with the month views when you click on same shortcut twice (the data is correct though). This is likely related to a `contiguousCalendarMonths` which you mentioned in the PR description, so might be a good simple start on that (and in the tests).

packages/datetime/src/dateRangePicker.tsx
giladgray: to be clear: `leftView = nextValueEndMonthAndYear.getPreviousMonth();`

omegdadi:
@leebyp what do you think if the entire `getStateChange` handler shortcuts if the day/month/year for the dates haven't changed.

omegdadi:
@leebyp Actually, looking closer, my changes wouldn't get reached in the case where the next value has the same start/end date as the current value. [Line 678](https://github.com/omegdadi/blueprint/blob/43063ff997463dac32718f5927cdb1101de81a30/packages/datetime/src/dateRangePicker.tsx#L678) looks like it would handle the same month use case. Can you provide an example of the funky behavior?

leebyp:
@omegdadi i just ran your changes locally and got this:

untitled

leebyp:
The checks for the same start / end date might be more confusing than it's worth because of our checks for `!= null`, but feel free to see how it comes out!

omegdadi:
Got it working with `contiguousCalendarMonths = false`.

packages/datetime/src/dateRangePicker.tsx
```diff
@@ -618,7 +618,8 @@ export class DateRangePicker
 
 function getStateChange(value: DateRange,
                         nextValue: DateRange,
-                        state: IDateRangePickerState): IDateRangePickerState {
+                        state: IDateRangePickerState,
+                        props: IDateRangePickerProps): IDateRangePickerState {
```
leebyp: sweet! one final small nit: can you just pass in `contiguousCalendarMonths` directly rather than the whole prop?

omegdadi:
Sure thing, it's updated.

giladgray:
@omegdadi build failed linting. please fix errors. run `gulp tslint` locally (or, better, configure TSLint plugin for your editor).

omegdadi:
Ugh...sorry about that. Last one hopefully.

giladgray:
@omegdadi close but not quite. still have one line too long. did you run `gulp tslint` locally?

omegdadi:
I did, but the output dumped out in terminal and pushed the error out of scroll history and I missed it.

omegdadi:
Thanks for the feedback

omegdadi:
@giladgray looks like it's passing this time

leebyp:
@omegdadi and in it goes!!!

omegdadi:
@leebyp hooray!

PR #833, opened 2017-03-13
[NumericInput] Support arbitrary precision inferred from smallest step size

cmslewis:
#### Checklist

- [x] Include tests

Currently, the most granular `NumericInput` step size of any kind (minor/major/regular) is 0.1. This PR adds support for arbitrarily small step sizes, adjusting the max precision of the displayed value accordingly (similar in intent to #832). As before, the precision of the displayed value shows the minimum number of decimal places necessary for each value (i.e. `1` displays as `"1"`, not `"1.000"`).

- I copied @giladgray's implemention of `countDecimalPlaces` from #832, will reuse his exact implementation here once that PR merges.
- The `toMaxPrecision` code feels a little wonky. There's probably a simpler way. Thoughts?


giladgray:
@cmslewis i just merged #832 👍

packages/core/src/components/forms/numericInput.tsx
```diff
@@ -524,6 +524,22 @@ export class NumericInput extends AbstractComponent<HTMLInputProps & INumericInp
...
+        let stepMaxPrecision = Utils.countDecimalPlaces(props.stepSize);
+        if (props.minorStepSize != null) {
+            stepMaxPrecision = Math.max(stepMaxPrecision, Utils.countDecimalPlaces(props.minorStepSize));
+        }
+
+        return stepMaxPrecision;
+    }
+
+    private toMaxPrecision(value: number) {
+        let { stepMaxPrecision } = this.state;
```
giladgray: `const`

packages/core/src/components/forms/numericInput.tsx
```diff
@@ -524,6 +524,22 @@ export class NumericInput extends AbstractComponent<HTMLInputProps & INumericInp
...
+        // round the value to have the specified maximum precision (toFixed is the wrong choice,
+        // because it would show trailing zeros in the decimal part out to the specified precision)
+        return +(Math.round(parseFloat(value + "e+" + stepMaxPrecision)) + "e-" + stepMaxPrecision);
```
giladgray: using `parseFloat` and `+` in the same sentence, nice. how about `parseFloat` twice though, for clarity?

packages/core/src/components/forms/numericInput.tsx
```diff
@@ -524,6 +524,22 @@ export class NumericInput extends AbstractComponent<HTMLInputProps & INumericInp
...
+        }
+
+        return stepMaxPrecision;
+    }
+
+    private toMaxPrecision(value: number) {
+        const { stepMaxPrecision } = this.state;
+        // round the value to have the specified maximum precision (toFixed is the wrong choice,
+        // because it would show trailing zeros in the decimal part out to the specified precision)
+        return +(Math.round(parseFloat(value + "e+" + stepMaxPrecision)) + "e-" + stepMaxPrecision);
```
giladgray: using `parseFloat` and `+` in the same sentence, nice. how about `parseFloat` twice though, for clarity?

cmslewis: Ah, yep, makes sense.

packages/core/src/components/forms/numericInput.tsx
```diff
@@ -524,6 +524,23 @@ export class NumericInput extends AbstractComponent<HTMLInputProps & INumericInp
...
+        }
+
+        return stepMaxPrecision;
+    }
+
+    private toMaxPrecision(value: number) {
+        const { stepMaxPrecision } = this.state;
+        // round the value to have the specified maximum precision (toFixed is the wrong choice,
+        // because it would show trailing zeros in the decimal part out to the specified precision)
+        // source: http://stackoverflow.com/a/18358056/5199574
```
giladgray: 👍  thanks!

packages/core/src/components/forms/numericInput.tsx
```diff
@@ -524,6 +524,23 @@ export class NumericInput extends AbstractComponent<HTMLInputProps & INumericInp
...
         return NumericInput.FLOATING_POINT_NUMBER_CHARACTER_REGEX.test(char);
     }
+
+    private getStepMaxPrecision(props: HTMLInputProps & INumericInputProps) {
+        let stepMaxPrecision = Utils.countDecimalPlaces(props.stepSize);
+        if (props.minorStepSize != null) {
+            stepMaxPrecision = Math.max(stepMaxPrecision, Utils.countDecimalPlaces(props.minorStepSize));
+        }
+
+        return stepMaxPrecision;
```
giladgray: what if you did like:

```js
if (minorStepSize) { return countDecimalPlaces(minorStepSize); }
else { return countDecimalPlaces(stepSize); }
```

then we only count maximum once. and the `minorStepSize` should be the smallest one, right? so we can just go off that.

packages/core/src/components/forms/numericInput.tsx
```diff
@@ -524,6 +524,23 @@ export class NumericInput extends AbstractComponent<HTMLInputProps & INumericInp
...
+    }
+
+    private toMaxPrecision(value: number) {
+        const { stepMaxPrecision } = this.state;
+        // round the value to have the specified maximum precision (toFixed is the wrong choice,
+        // because it would show trailing zeros in the decimal part out to the specified precision)
+        // source: http://stackoverflow.com/a/18358056/5199574
+        return parseFloat(Math.round(parseFloat(value + "e+" + stepMaxPrecision)) + "e-" + stepMaxPrecision);
```
giladgray: this math is frightening and possibly expensive, but i get it. also the experience is sweet as the precision adjusts automatically.

packages/core/src/components/forms/numericInput.tsx
cmslewis: Sweet, I like that better. 👍 This was a relic from a few commits ago, when I considered adding a configurable `precision` prop that we'd have to sanitize.

packages/core/src/components/forms/numericInput.tsx
cmslewis: Voila: https://jsperf.com/roundtomaxprecision. Will tweak this line based on these findings.

giladgray: um wow
jsperf

packages/core/src/components/forms/numericInput.tsx
```diff
@@ -526,20 +526,19 @@ export class NumericInput extends AbstractComponent<HTMLInputProps & INumericInp
...
         }
-
-        return stepMaxPrecision;
     }
 
     private toMaxPrecision(value: number) {
-        const { stepMaxPrecision } = this.state;
         // round the value to have the specified maximum precision (toFixed is the wrong choice,
         // because it would show trailing zeros in the decimal part out to the specified precision)
         // source: http://stackoverflow.com/a/18358056/5199574
```
giladgray: update this source link? points to old code

cmslewis: I left it because it still captures the original idea, even though I made syntactic changes. Should I just remove it?

Issue #516, opened 2017-01-19
Fix selection region positioning with no row headers


llorca:
It's clipped right now:
![image](https://cloud.githubusercontent.com/assets/199754/22121706/e51a04d8-de39-11e6-851f-b7ecb92feef0.png)

llorca:
Turns out this is more than just a CSS fix, will need your help @gscshoyru

cmslewis:
Fixed locally! But the fix introduces—or rather _uncovers_—a bug that I've seen in production apps before, wherein a selection region will have a vertically misaligned left border when the row header is turned off. (This is because the left border is showing in a 1px-wide LEFT quadrant whose column headers have height 0, causing the body content—including the selection overlay—to scoot upward).

PR #836, opened 2017-03-14
[Table] Delete special styles for last cells in row

### LINKED ISSUE #628: Right edge of table extends 1px too much
image
image

The rightmost column is misaligned with the header. If the table is smaller than the body content, it creates a double border

cmslewis:

Delete special styles for last cell in row and last cell in table. These styles were presumably present to hide borders appropriately when cells filled the table horizontally, but the styles didn't actually look perfect in that case. Removing them looks generally better all around.

- Are you okay with the After style in each screenshot? (@themadcreator would love your thoughts especially)
- Am I missing any edge cases?

##### Chrome

_Before:_
before3
before2
before1

_After:_
after3
after2
after1

##### Firefox

_Before:_
image
image
image

_After:_
image
image
image

Note: This PR also partly addresses #516:
image

### DIFF
--- a/packages/table/src/cell/_borders.scss
+++ b/packages/table/src/cell/_borders.scss
@@ -81,18 +81,8 @@ for all borders with minimal duplication. See the bottom of this file.
   }
 
   .bp-table-body {
-    .bp-table-last-in-row {
-      box-shadow: inset 0 (-$border-width) 0 $border-color,
-                  $border-width 0 0 $border-color;
-    }
-
     .bp-table-last-in-column {
       box-shadow: inset (-$border-width) 0 0 $border-color,
-                  0 $border-width 0 $border-color;
-    }
-
-    .bp-table-last-in-row.bp-table-last-in-column {
-      box-shadow: $border-width 0 0 $border-color,
                   0 $border-width 0 $border-color;
     }
   }


giladgray:
only weirdness i can find: 
image

looks great otherwise!

cmslewis:
@giladgray Maybe we can add a CSS class when the table cells start to horizontally overflow the table container. All my other ideas have failed me. Thoughts?

cmslewis:
@llorca @gscshoyru regarding #516: I was referring to the fact that before, the blue border was 1px off on the right side; now, it aligns perfectly. Didn't attempt to address the left side (I guess I misunderstood the intent of that issue).

cmslewis:
@llorca - regarding #516, here's what we do on the left border of a selection (blue border is outset) vs. on the bottom border of a selection (blue border is inset). Should we handle both cases in the same way? Won't fix all the `overflow: hidden` issues, but will at least reveal the left border of a selection.

image

cmslewis:
@llorca - I'm slowly going insane. I reverted the commit in this PR to try a different direction, and all the misalignment issues in #628 have gone away, as have the double-border issues introduced by the original change in this PR. Might this issue have been addressed by other recent work?

PR #854, opened 2017-03-16
[Table] Fix blurry table in dialog

### LINKED ISSUE #742: Table has blurry text
image

EDIT: Here's a better snapshot of how egregious this bug is:
image

Possibly related to performance issues

Seen on Chrome on Mac & on Chrome on Windows, not on IE Edge

cmslewis:

`Table`'s `will-change` rules are not playing nicely with `Popover`'s `transition` rules. Consequently, tables appear blurry in dialogs (see the YouTube video linked from #742 for a good illustration of the Chrome bug that's causing all of this).

This PR deletes these `will-change` rules in favor of old-fashioned `tranform: translateZ(0)` rules to remove the blurriness while keeping the hardware-accelerated benefits intact.

I made a `TableDialogExample` locally to verify the changes (but didn't push it to the branch). See the screenshots below for the Before/After.

#### Screenshot

_Before:_
screen shot 2017-03-16 at 12 30 01 pm

_After:_
screen shot 2017-03-16 at 12 30 14 pm

### DIFF
--- a/packages/table/src/table.scss
+++ b/packages/table/src/table.scss
@@ -3,7 +3,7 @@
 // of the license at https://github.com/palantir/blueprint/blob/master/LICENSE
 // and https://github.com/palantir/blueprint/blob/master/PATENTS
 
-@import "../../core/src/common/_mixins";
+@import "../../core/src/common/mixins";
 @import "./common/loading";
 @import "./common/variables";
 @import "./cell/cell";
--- a/packages/table/src/table.scss
+++ b/packages/table/src/table.scss
@@ -3,6 +3,7 @@
 // of the license at https://github.com/palantir/blueprint/blob/master/LICENSE
 // and https://github.com/palantir/blueprint/blob/master/PATENTS
 
+@import "../../core/src/common/_mixins";
 @import "./common/loading";
 @import "./common/variables";
 @import "./cell/cell";
@@ -18,19 +19,9 @@ $row-z-index: $column-z-index + 1 !default;
 $menu-z-index: $row-z-index + 1 !default;
 
 .bp-table-container {
+  @include force-hardware-acceleration();
   display: flex;
   flex-direction: column;
-  // translateZ(0) has no visual effect, but it achieves the following:
-  //   - creates a new stacking context
-  //   - moves this element to a new layer to enable hardware-accelerated animation
-  //
-  // `will-change: transform` does the same thing, but it suffers from a Chrome bug that
-  // makes the Table blurry when transform'd (e.g. when the Table is in a Popover that
-  // animates open). we should move to `will-change` once that bug is fixed.
-  //
-  // further reading: https://dev.opera.com/articles/css-will-change-property/
-  transform: translateZ(0);
-  // will-change: transform;
   background-color: $table-background-color;
   max-width: 100%;
   height: 100%; // IE 11
@@ -134,10 +125,8 @@ $menu-z-index: $row-z-index + 1 !default;
 }
 
 .bp-table-body-scroll-client {
+  @include force-hardware-acceleration();
   position: relative;
-  // see the note above for the reasoning behind this translateZ rule.
-  transform: translateZ(0);
-  // will-change: transform;
   user-select: none;
 }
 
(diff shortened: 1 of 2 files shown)


tgreenwatts:
Question: Why are we removing it entirely instead of just when it's in a dialog?

cmslewis:
@tgreenwatts per @llorca's proposal in https://github.com/palantir/blueprint/issues/742#issuecomment-287135879

llorca:
Two issues I'm seeing so far:
- some tables in the example page now get a weird focus outline after selecting a cell or region
image
- z-indices are now _global_ instead of scoped to the table :/
image

cmslewis:
@llorca - weird, wonder why `will-change` has anything to do with that. I'll read up on it.

cmslewis:
@llorca

> A stacking context is formed, anywhere in the document, by...specifying any attribute above [e.g. `transform`] in `will-change` even if you don't specify values for these attributes directly (See [this post](https://dev.opera.com/articles/css-will-change-property/))

Source: https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Positioning/Understanding_z_index/The_stacking_context

Looks like we need to create a stacking context in a different way.

cmslewis:
@llorca @tgreenwatts fixed with the old-fashioned hardware-acceleration hack! We now (should) have hardware-accelerated table animations AND no blurriness. It's not pretty, but it'll work until Chrome's `will-change `bug is fixed.

llorca:
@cmslewis Nice trick! Forgot about that one, but sounds like it's the right move

packages/table/src/table.scss
```diff
@@ -20,13 +20,23 @@ $menu-z-index: $row-z-index + 1 !default;
...
+  //   - creates a new stacking context
+  //   - moves this element to a new layer to enable hardware-accelerated animation
+  //
+  // `will-change: transform` does the same thing, but it suffers from a Chrome bug that
+  // makes the Table blurry when transform'd (e.g. when the Table is in a Popover that
+  // animates open). we should move to `will-change` once that bug is fixed.
+  //
+  // further reading: https://dev.opera.com/articles/css-will-change-property/
+  transform: translateZ(0);
+  // will-change: transform;
```
giladgray: 😱

packages/table/src/table.scss
```diff
@@ -20,13 +20,23 @@ $menu-z-index: $row-z-index + 1 !default;
 .bp-table-container {
   display: flex;
   flex-direction: column;
+  // translateZ(0) has no visual effect, but it achieves the following:
+  //   - creates a new stacking context
+  //   - moves this element to a new layer to enable hardware-accelerated animation
+  //
+  // `will-change: transform` does the same thing, but it suffers from a Chrome bug that
```
giladgray: can you link to chrome bug?

packages/table/src/table.scss
```diff
@@ -125,8 +135,10 @@ $menu-z-index: $row-z-index + 1 !default;
 
 .bp-table-body-scroll-client {
   position: relative;
+  // see the note above for the reasoning behind this translateZ rule.
+  transform: translateZ(0);
+  // will-change: transform;
```
giladgray: 😱

packages/table/src/table.scss
```diff
@@ -20,13 +20,23 @@ $menu-z-index: $row-z-index + 1 !default;
...
+  // translateZ(0) has no visual effect, but it achieves the following:
+  //   - creates a new stacking context
+  //   - moves this element to a new layer to enable hardware-accelerated animation
+  //
+  // `will-change: transform` does the same thing, but it suffers from a Chrome bug that
+  // makes the Table blurry when transform'd (e.g. when the Table is in a Popover that
+  // animates open). we should move to `will-change` once that bug is fixed.
+  //
+  // further reading: https://dev.opera.com/articles/css-will-change-property/
+  transform: translateZ(0);
```
themadcreator: since this is such a fancy maneuver, i'd like this comment and this css rule to be brought into a mixin called "new-stacking-context" or something.

cmslewis: Happy to add a mixin. Note that this `transform` rule doesn't just create a new stacking context; it also creates a new "containing block" (meaning that fixed-position elements within the block will be transformed as well). I think it'd be clearest to name the mixin `force-hardware-acceleration()`, because that's what it's really meant for.

cmslewis:
@giladgray @themadcreator I made a mixin. Ready for re-review.

Issue #867, opened 2017-03-21
[DateInput] Remove openOnFocus prop and always open on focus


cmslewis:
We removed the calendar icon on `DateRangeInput` before shipping that component. I've heard requests internally to do the same on `DateInput`. Might as well make them work consistently, such that the popover opens on focus without the need for a calendar icon.

PR #868, opened 2017-03-21
[DateInput] Delete calendar icon, add optional rightElement prop

cmslewis:
Got a request in person to add support for hiding the calendar icon in `DateInput`. Seemed reasonable and easy, hence this PR. After initially adding a `showIcon` prop to accomplish this, @llorca and I agreed that deleting the calendar icon altogether would be preferable (this is what we did with `DateRangeInput`).

Related: see #867, which I've assigned to Milestone 2.X due to its API-breaking nature (goal is to remove `openOnFocus` if we go through with deleting the icon element now).

#### Checklist

- [x] Update documentation (auto-updated props table)

- Add `rightElement` prop to `DateInput` to allow customizable right element (particularly: nothing, static icon, interactive icon)

- Does this sound reasonable?
- Is this considered a breaking change or not?
- Is the prop description text okay?

#### Screenshot

_Before_
image

_After_
image

### DIFF
--- a/packages/datetime/src/dateInput.tsx
+++ b/packages/datetime/src/dateInput.tsx
@@ -105,6 +105,12 @@ export interface IDateInputProps extends IDatePickerBaseProps, IProps {
     popoverPosition?: Position;
 
     /**
+     * Whether to show the calendar icon.
+     * @default true
+     */
+    showIcon?: boolean;
+
+    /**
      * The currently selected day. If this prop is provided, the component acts in a controlled manner.
      * To display no date in the input field, pass `null` to the value prop. To display an invalid date error
      * in the input field, pass `new Date(undefined)` to the value prop.
@@ -130,6 +136,7 @@ export class DateInput extends AbstractComponent<IDateInputProps, IDateInputStat
         openOnFocus: true,
         outOfRangeMessage: "Out of range",
         popoverPosition: Position.BOTTOM,
+        showIcon: true,
     };
 
     public displayName = "Blueprint.DateInput";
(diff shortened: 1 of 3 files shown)


llorca:
I'm not in favor of that. I would kill this icon thing entirely, it's pretty useless. What would be useful is leverage composability and let consumers do whatever they want, i.e. exposing `rightElement` from `InputGroup` to `DateInput`. Then, we fill the following use cases:
- no icon
- static icon
- interactive icon

What do you think?

cmslewis:
@llorca Happy to kill the icon. Just wasn't sure if that was considered a breaking change. FWIW I filed #867 to track removing the icon and the `openOnFocus` prop (making open-on-focus the enforced behavior).

cmslewis:
@llorca see latest changes and updated PR description. @giladgray @leebyp - thoughts?

packages/datetime/src/dateInput.tsx
```diff
@@ -167,16 +172,6 @@ export class DateInput extends AbstractComponent<IDateInputProps, IDateInputStat
             "pt-intent-danger": !(this.isMomentValidAndInRange(date) || isMomentNull(date) || dateString === ""),
```
llorca: Unrelated: this class name should be replaced by the corresponding variable

llorca:
We're not removing any capability or interaction, and there's no code update needed, therefore I wouldn't consider this as a breaking change. LGTM 👍

packages/datetime/src/dateInput.tsx
cmslewis: Yeah I spotted that. I think there are a few other places in the codebase too. Can circle back.

PR #897, opened 2017-03-28
[DateRangeInput] [DateRangePicker] Support the same props in DRI as in DRP

cmslewis:
#### Checklist

- [x] Include tests
- [x] Update documentation

Quick PR to ensure users can configure all `DateRangePicker` props from `DateRangeInput`. In particular:

- Add `contiguousCalendarMonths` prop to `DRI`
- Add `shortcuts` prop to `DRI`
- Add unit tests
- Add `contiguousCalendarMonths` toggle to `DRI` example

- This PR was motivated by internal requests.
- Possible to get in for today's release?

### DIFF
--- a/packages/datetime/examples/dateRangeInputExample.tsx
+++ b/packages/datetime/examples/dateRangeInputExample.tsx
@@ -15,6 +15,7 @@ import { FORMATS, FormatSelect } from "./common/formatSelect";
 export interface IDateRangeInputExampleState {
     allowSingleDayRange?: boolean;
     closeOnSelection?: boolean;
+    contiguousCalendarMonths?: boolean;
     disabled?: boolean;
     format?: string;
     selectAllOnFocus?: boolean;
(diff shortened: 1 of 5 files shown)


packages/datetime/examples/dateRangeInputExample.tsx
```diff
@@ -42,6 +47,12 @@ export class DateRangeInputExample extends BaseExample<IDateRangeInputExampleSta
     protected renderOptions() {
         return [
             [
+                <FormatSelect
```
cmslewis: Made this column narrower to make room for longer labels in the toggle column. Felt better on the left side.

leebyp:
@cmslewis i think we chatted about this before, but what was the reason again for not having something like `dateRangePickerProps` and proxying the whole thing across (with some values that cant be overridden)?

cmslewis:
A bad reason: I didn't think to do it. We should do that in 2.0 I think.

PR #906, opened 2017-03-29
[Table] Refactor header menus to divorce click-target size from gradient size

cmslewis:

Table column headers suffer from a problem:
- The gradient effect isn't wide enough for the text to fade out entirely before reaching the popover target (separate issue: the icon doesn't have a background, so text is visible beneath it):
image
- Making the gradient bigger also makes the popover click target bigger—to a silly extent.

This PR divorces the gradient from the popover click target. Now, we can resize the gradient without affecting the size of the click target. Specific changes:

- Elements are now organized like this:
  - `bp-table-th-menu-container` (new)
    - `bp-table-th-menu-container-background` (new; sibling of `bp-table-th-menu`)
    - `bp-table-th-menu` (same thing we had before)
- Click target is now a perfect `30px` x `30px`.
- Gradient is now `10px` wider to ensure the text fades out before reaching the icon.
- Gradient has `pointer-events: none` to not swallow clicks meant for the underlying header cell.

- Implementation look good?
- Visuals look good?
- Interactions feel good?

#### Screenshot

Gradient before:
image

Gradient after:
image

Click target before:
image

Click target after:
image


cmslewis:
Whoops, style regression with interaction bars: 
image

pkwi:
The new gradient and click target are great!

packages/table/src/headers/_headers.scss
```diff
@@ -51,7 +51,20 @@
   }
 }
 
-.bp-table-th-menu {
+.bp-table-th-menu-container {
+  flex-shrink: 0;
```
giladgray: shouldn't need flex properties if you're using `position: absolute` but whatever

packages/table/src/headers/_headers.scss
```diff
@@ -65,18 +78,35 @@
...
+    .pt-dark & {
+      background-image: $dark-menu-selected-background;
+    }
+  }
+}
+
+.bp-table-th-menu {
   cursor: $action-cursor;
-  width: $column-header-min-height + $pt-grid-size;
+  width: $column-header-min-height; // create a larger, more accessible click target
```
giladgray: i don't think comment means much, and will mean even less in a month. it doesn't even look correct when i compare with code on the left, where the width _was_ obviously larger.

packages/table/src/headers/_headers.scss
cmslewis: yeah i wasn't actually sure what this rule was doing (i'm no flex expert), but it was there before, so i left it in the refactor.

packages/table/src/headers/_headers.scss
cmslewis: fair, though in general i like quick little CSS comments explaining the intent behind a rule. regardless of the target size prior to this refactor, this element's purpose is still to create a larger click target than the `16px` icon it wraps, so the comment seemed helpful.

cmslewis:
@llorca @pkwi fixed the visual bugs. ready for re-review.

PR #917, opened 2017-03-30
Documentation Overhaul

themadcreator:
We now intend to publish `@blueprintjs/docs` as its own package. It contains the `<Documentation>` React component that is built to consume and render `Documentalist` output.

The documentation has been restructured to use `Documentalist` to parse all typescript, sass, and markdown across all Blueprint packages then render using this new `@blueprintjs/docs` package.

The Blueprint-specific content (such as the logo, dark theme switcher, etc.) have been moved into `packages/docs-site`.


adidahiya:
we lost date range input? 🤔 

image

packages/docs-site/src/components/clickToCopy.tsx
```diff
@@ -60,7 +57,7 @@ export class ClickToCopy extends React.Component<IClickToCopyProps, IClickToCopy
         return (
             <div
                 {...removeNonHTMLProps(this.props, ["copiedClassName", "value"], true)}
-                className={classNames("docs-clipboard", className, { [copiedClassName]: this.state.hasCopied })}
+                className={classNames("docs-clipboard", className, { [copiedClassName!]: this.state.hasCopied })}
```
cmslewis: whaaaaaat does this syntax mean

packages/docs-site/src/components/clickToCopy.tsx
```diff
@@ -38,8 +36,7 @@ export interface IClickToCopyState {
  *  - `[data-copied-message="<message>"]` will be shown when the element has been copied.
  * The message is reset to default when the user mouses off the element after copying it.
  */
-@PureRender
-export class ClickToCopy extends React.Component<IClickToCopyProps, IClickToCopyState> {
+export class ClickToCopy extends React.PureComponent<IClickToCopyProps, IClickToCopyState> {
```
cmslewis: does this need to happen in this PR?

packages/docs-site/src/components/colorSchemes.tsx
```diff
@@ -68,8 +65,7 @@ export interface IColorSchemeState {
     steps?: number;
 }
 
-@PureRender
-export class ColorScheme extends React.Component<IColorSchemeProps, IColorSchemeState> {
+export class ColorScheme extends React.PureComponent<IColorSchemeProps, IColorSchemeState> {
```
cmslewis: oh boy here we go

packages/docs-site/src/components/clickToCopy.tsx
giladgray: this package is private so it doesn't matter what we do inside. there is no harm in this.

packages/docs-site/src/components/clickToCopy.tsx
giladgray: `strictNullChecks` assert that something is defined (remove `| undefined` basically). necessary because most props are marked optional so they're not required in the JSX.

packages/docs-site/src/components/colorSchemes.tsx
giladgray: where are we going? this package is private. no one will be the wiser.

packages/docs-site/src/components/clickToCopy.tsx
cmslewis: just seemed like making this change now would bloat the number of files affected in this PR unnecessarily. but yeah, no harm in it.

packages/docs-site/src/components/clickToCopy.tsx
cmslewis: ah, so a TS thing. cool, TIL.

PR #951, opened 2017-04-05
Add some class names

ryanmcnamara:
pt-ui-text-large
pt-text-muted


ryanmcnamara:
Is it fine to just add these?

ryanmcnamara:
thanks!
I'll probably do this more as I notice that some are missing. I've stumbled across many missing ones in the past

PR #960, opened 2017-04-06
[Table] Add support for resizing a column's rows based on visible cell contents

### LINKED ISSUE #929: FR: Resize row height for currently visible items automatically on wrap text
Take the rows that are currently visible
Take the max height of a cell within them (for all columns, even outside of view)
Set all rows (including out of view ones) to that height

Basically using the currently visible cells as indicator of what other cells are like

A lot of open questions around how to do this technically, how to expose it in the UI, etc, but it is better than the current resize-wrap workflow

gscshoyru:
Adds a function to the table, that allows you to specify a column, and then resizes all rows by the tallest visible cell in that column.

Since this is meant to be called from outside the table, I'll need to add buttons to the page -- for now, though, you can do formatsTable.resizeRowsBasedOnColumnHeight(columnIndex) in the console to make it resize based on whatever you specified for columnIndex. 

There's some wonky code in here with any's -- does anyone know why the typings here isn't correct?

Can anyone think of a better way to do this? This code pretty strongly relies on the behavior of our truncated format -- it won't work for anything custom that anyone makes.


tgreenwatts:
I updated your description a bit, maybe still not clear?

tgreenwatts:
bug
Spoke to @gscshoyru offline about this bug

packages/table/preview/index.tsx
```diff
@@ -97,24 +97,28 @@ class FormatsTable extends React.Component<{}, {}> {
     private objects = Utils.times(FormatsTable.ROWS, (row: number) => {
         // tslint:disable-next-line:switch-default
         switch (row) {
-            case 1: return "string";
-            case 2: return 3.14;
-            case 5: return undefined;
-            case 6: return null;
-            case 8: return "this is a very long string";
+            case 0: return "Avengers: Age of Ultron (2015)";
```
tgreenwatts: Did you mean to commit this?

gscshoyru: Whoops, thanks, fixed.

llorca:
@gscshoyru Similarly to the screenshot below, can we add a "Resize rows based on selected column height" button on one of the examples?

image

gulp/util/karma-config.js
```diff
@@ -40,8 +40,8 @@ module.exports = function createConfig(project) {
         coverageReporter: {
             check: {
                 each: {
-                    lines: 79,
-                    statements: 79,
+                    lines: 65, // TODO (gcameron): put this back
+                    statements: 65, // TODO (gcameron): put this back
```
cmslewis: :)

packages/table/preview/index.tsx
```diff
@@ -113,8 +113,11 @@ class FormatsTable extends React.Component<{}, {}> {
     private strings = Utils.times(FormatsTable.ROWS, () => ("ABC " + (Math.random() * 10000)));
 
     public render() {
+
```
cmslewis: nit: don't think we need a newline here.

packages/table/preview/index.tsx
```diff
@@ -113,8 +113,11 @@ class FormatsTable extends React.Component<{}, {}> {
...
+        const saveTable = (table: any) => (window as any).formatsTable = table;
```
cmslewis: Do we need to use `any` here?

packages/table/src/locator.ts
```diff
@@ -17,6 +17,8 @@ export interface ILocator {
      */
     getWidestVisibleCellInColumn: (columnIndex: number) => number;
 
+    getTallestVisibleCellInColumn: (columnIndex: number) => number;
```
cmslewis: The goal of this function is rather self-evident, but let's add a description anyway since all the other functions have one.

packages/table/src/locator.ts
```diff
@@ -76,6 +78,27 @@ export class Locator implements ILocator {
         return max;
     }
 
+    public getTallestVisibleCellInColumn(columnIndex: number): number {
+        const cells = this.tableElement.querySelectorAll(`.${Classes.columnCellIndexClass(columnIndex)}`);
+        let max = 0;
+        for (let i = 0; i < cells.length; i++) {
+            const cellValue = (cells.item(i) as any).query(`.${Classes.TABLE_TRUNCATED_VALUE}`);
```
cmslewis: What was the motivation for `any` here?

packages/table/src/locator.ts
```diff
@@ -76,6 +78,27 @@ export class Locator implements ILocator {
...
+            const cellTruncatedFormatText = (cells.item(i) as any).query(`.${Classes.TABLE_TRUNCATED_FORMAT_TEXT}`);
+            let height = 0;
+            if (cellValue != null) {
+                height = cellValue.scrollHeight;
+            } else if (cellTruncatedFormatText != null) {
+                height = cellTruncatedFormatText.scrollHeight;
+            } else {
+                height = (cells.item(i) as any).query(`.${Classes.TABLE_TRUNCATED_TEXT}`).scrollHeight;
+            }
+            if (max < height) {
```
cmslewis: nit: I'd phrase this as `height > max`, because I normally prefer the current thing to be compared to the existing thing, left-to-right.

packages/table/src/table.tsx
```diff
@@ -450,6 +450,18 @@ export class Table extends AbstractComponent<ITableProps, ITableState> {
         );
     }
 
+    public resizeRowsBasedOnColumnHeight(columnIndex: number) {
+        const {locator} = this.state;
```
cmslewis: nit: convention in the rest of the Blueprint codebase is to add spaces within the brackets when deconstructing: `{ locator }`. @adidahiya is there a `tslint` rule or existing issue for this, or should I file a FR?

packages/table/src/table.tsx
```diff
@@ -450,6 +450,18 @@ export class Table extends AbstractComponent<ITableProps, ITableState> {
...
+
+        const tallest = locator.getTallestVisibleCellInColumn(columnIndex);
+        const rowHeights = this.state.rowHeights.slice();
+        for (let row = 0; row < rowHeights.length; row++) {
+            rowHeights[row] = tallest;
+        }
```
cmslewis: More succinct/declarative: `const rowHeights = Array(this.state.rowHeights.length).fill(tallest);`

gulp/util/karma-config.js
themadcreator: booooooo

packages/table/src/locator.ts
themadcreator: if `cells.item(i)` is an HTMLElement, should you use `.querySelector`?

packages/table/src/table.tsx
```diff
@@ -450,6 +450,18 @@ export class Table extends AbstractComponent<ITableProps, ITableState> {
         );
     }
 
+    public resizeRowsBasedOnColumnHeight(columnIndex: number) {
```
themadcreator: we should probably also expose a method for resizing columns since they will surely ask for that next. We already have some of this code anyway, right?

tgreenwatts: Not sure what this would mean for an app - you can already select-all, double click on a column edge and have everything resize

tgreenwatts: Wait, no, that just adjusts all columns to the same size as the one you clicked on

themadcreator: I just mean that for our API where have a method for just one of rows or columns, users usually ask for the other one.

packages/table/preview/index.tsx
gscshoyru: This is just for testing purposes, it'll be replaced with an actual button later, I just wanted eyes on it.

packages/table/src/locator.ts
gscshoyru: For some reason the typings don't seem to think this thing has a "query" attached to it. Even though it does. This is what I wanted to ask about. If there's a better way to do this, please let me know.

gulp/util/karma-config.js
gscshoyru: This is just to get eyes on it -- I'll be adding tests.

themadcreator: okay. consider eyes on =)

cmslewis:
@gscshoyru Can we add the `[Table]` prefix to this PR title? It's becoming a helpful convention for perusing git history on `master`.

packages/table/src/table.tsx
cmslewis: @gscshoyru when glancing at this API, I wasn't really sure what "based on column height" meant ("aren't all the columns the same height / isn't that how tables work?"). Is this the best name for this function?

More verbose but more clear to me: `resizeRowsBasedOnTallestVisibleCellInColumn`

gscshoyru: That would be a bit of a mouthful. But yeah, sense. 

Does anyone have better ideas? Otherwise sure, I'll make the change.

adidahiya:
```ts
resizeRowsBasedOnTallestVisibleCellInColumn(columnIndex: number)
```

lol that's a mouthful. how about

```ts
resizeRows({
  // if absent, will try to resize based on all columns (dangerous? slow perf? maybe warn devs about this)
  columnIndex?: number,
  // ResizeStrategy.TallestVisibleCell, etc.
  strategy: ResizeStrategy,
});
```

themadcreator:
That does look quite nice but why not just function arguments instead of an object?

themadcreator:
also there is really only one strategy

adidahiya:
> why not just function arguments instead of an object

easier to maintain backcompat when using an object. destructuring makes object args pretty simple.

> also there is really only one strategy

for now, yeah. maybe we allow users to define their own in the future? I guess I don't have a great example of this off the top of my head...

adidahiya:
wait but seriously consider a simpler name for the public API.

packages/table/src/table.tsx
```diff
@@ -489,6 +489,15 @@ export class Table extends AbstractComponent<ITableProps, ITableState> {
         );
     }
 
+    public resizeRowsBasedOnTallestVisibleCellInColumn(columnIndex: number) {
+        const { locator } = this.state;
+
+        const tallest = locator.getTallestVisibleCellInColumn(columnIndex);
+        const rowHeights = Array(this.state.rowHeights.length).fill(tallest);
```
giladgray: :tada:

tgreenwatts:
@gscshoyru @cmslewis any word on a "simpler name for the public API" so we can merge this awesomeness?

cmslewis:
The object approach is cool, but I can't think of a great alternative strategy we'd realistically implement. `ResizeStrategy.AllCells` seems like the best candidate, but that seems unfeasibly expensive.

I'm personally fine with the long-winded name, but I admittedly lean verbose in code.

giladgray:
how about

```ts
resizeRowsBasedOnTallestVisibleCellInColumn(columnIndex: number)
// becomes
resizeRowsByTallestCell(columnIndex: number)
```

?

eventually we could add a `visibleOnly` arg that could be disabled to include entire table (theoretically)

adidahiya:
yeah I like @giladgray's proposal.

gscshoyru:
So -- we can't ever actually do this to resize it based on _non_-visible cells, to be clear. We can't size something that isn't drawn to the DOM. 

But yeah, I can rename this that way, will do so :)

tgreenwatts:
:shipit: !!

PR #963, opened 2017-04-07
[Table] Support for column and row reordering

### LINKED ISSUE #213: [Table] drag and drop row re-ordering
Looking to implement drag and drop re-ordering of rows for the table component, but not 100% how to do best do this.

Since the component doesn't contain data, either an interface would have to be provided to explain the transaction that occurred.

Something like
```
onDragDrop: function (originalIndex, newIndex) {}
```

And possibly a prop to enable/disable it (you'd likely have to set the selection mode to rows only?

This way the user could perform the array update via whatever mechanism they need to correctly maintain state for the dataset.

Any thoughts or comments on how best to attack this?

I see there's some drag and drop stuff implemented for selection, so ideally that could be re-purposed?

cmslewis:
#### Checklist

- [x] Include tests
- [x] ~Update documentation~ (will add language in a separate PR if we want it)
- [x] Add new example
- [x] Fix interaction regression w/ `TruncatedPopover`

Behold: column and row reordering in `Table`.

Basic points:
- Reordering works only if selection is also enabled. **If this assumption is bad, let me know.**
- Reordering is performed by mousing down within a selected column header, then dragging. **If reordering is enabled, it is no longer possible to deselect a column by clicking on its header.** LMK how that feels; I don't love it yet.
- You can enable reordering for columns, rows, or both via `Table` props.
- You can reorder a single entity or a region of multiple entities (having `FULL_COLUMNS` or `FULL_ROWS` cardinality).
- You **CANNOT** reorder anything if there are multiple regions selected. It's too much of a pain for a v1 implementation, plus initial feedback suggested that no user will be able to predict what will happen anyway. We can revisit this later if we want.
- The `Table` does not handle the actual reordering! It simply emits `(oldIndex, newIndex, length)` and lets the caller pass in updated data.

- High-level architecture. Does it look reasonable?
- One thing that felt a little gross was the introduction of a "reordering cursor overlay" that displays invisibly over the whole table while you're dragging to reorder. I couldn't think of a better way to guarantee that the cursor would appear as `grabbing` no matter where your mouse moved during the drag. LMK what you think.


llorca:
Can't wait to check it out!

> Reordering works only if selection is also enabled. If this assumption is bad, let me know.

I think this makes sense, otherwise we wouldn't be able to support multi-column/row reordering (the reordering design is based on selection, so it would break). We can always revisit later on if there's a need for reordering without selection.

> Reordering is performed by mousing down within a selected column header, then dragging. If reordering is enabled, it is no longer possible to deselect a column by clicking on its header. LMK how that feels; I don't love it yet.

That's fine, the table shouldn't even allow togglable selection like this. It's not a design pattern, and none of the spreadsheet tools out there work like that.

gulp/util/karma-config.js
```diff
@@ -40,8 +40,8 @@ module.exports = function createConfig(project) {
         coverageReporter: {
             check: {
                 each: {
-                    lines: 79,
-                    statements: 79,
+                    lines: 75, // TODO (clewis): Push this back to 79 before merging :)
+                    statements: 75, // TODO (clewis): Push this back to 79 before merging :)
```
giladgray: push it back to 80 if you can!

packages/table/src/common/classes.ts
```diff
@@ -44,6 +45,8 @@ export const TABLE_OVERLAY_LAYER = "bp-table-overlay-layer";
 export const TABLE_POPOVER_WHITESPACE_NORMAL = "bp-table-popover-whitespace-normal";
 export const TABLE_POPOVER_WHITESPACE_PRE = "bp-table-popover-whitespace-pre";
 export const TABLE_REGION = "bp-table-region";
+export const TABLE_REORDERING = "bp-table-reordering";
+export const TABLE_REORDERING_CURSOR_SHIELD = "bp-table-reordering-cursor-shield";
```
giladgray: does this class need to be here? seems unlikely anyone would want to access this one.

packages/table/src/table.tsx
```diff
@@ -806,9 +831,9 @@ export class Table extends AbstractComponent<ITableProps, ITableState> {
      * intend to redraw the region layer.
      */
     private maybeRenderRegions(getRegionStyle: IRegionStyler) {
-        if (this.isGuidesShowing()) {
-            return undefined;
-        }
+        // if (this.isGuidesShowing()) {
+        //     return undefined;
+        // }
```
giladgray: commented code!

packages/table/src/common/classes.ts
llorca: What's a shield? Is there a more consistent term we could use from the already existing components, such as "indicator" or something?

tgreenwatts:
Which examples are reorderable?

llorca:
@tgreenwatts [Formatted content](https://2458-71939872-gh.circle-artifacts.com/0/home/ubuntu/blueprint/packages/site-docs/dist/index.html#table-js.formatted-content) (PR description):
> For the purposes of testing, I enabled column and row reordering on the [BLUEPRINT_DOCS_URL]/#table-js.formatted-content example, so head there to try out the feature. I'll make a new example and revert all those changes before merging.

adidahiya:
Noticed an interaction regression that's not present in the latest release:

packages/table/examples/tableFormatsExample.tsx
```diff
@@ -77,20 +77,39 @@ const FORMAT_OPTIONS = {
...
+    data?: any[];
+}
+
+export class TableFormatsExample extends BaseExample<ITableFormatExampleState> {
     private data = TIME_ZONES;
     private date = new Date();
 
+    public constructor(props?: any, context?: any) {
+        super(props, context);
+        this.state = {} as ITableFormatExampleState;
```
giladgray: don't use a constructor for this. `public state = ...`

packages/table/examples/tableFormatsExample.tsx
```diff
@@ -77,20 +77,39 @@ const FORMAT_OPTIONS = {
     year: "numeric",
 };
 
-export class TableFormatsExample extends BaseExample<{}> {
+export interface ITableFormatExampleState {
+    children?: any[];
```
giladgray: `children?: React.ReactNode` ?

packages/table/examples/tableFormatsExample.tsx
```diff
@@ -77,20 +77,39 @@ const FORMAT_OPTIONS = {
...
+    data?: any[];
```
giladgray: is this used in this example? seems to go off `this.data` instead.

packages/table/examples/tableFormatsExample.tsx
```diff
@@ -107,4 +126,58 @@ export class TableFormatsExample extends BaseExample<{}> {
...
+    private handleColumnsReordered = (oldIndex: number, newIndex: number, length: number) => {
+        if (oldIndex === newIndex) { return; }
+        const nextChildren = reorderElementsInArray(this.state.children, oldIndex, newIndex, length);
+        this.setState({ children: nextChildren });
+    }
+
+    private handleRowsReordered = (oldIndex: number, newIndex: number, length: number) => {
+        if (oldIndex === newIndex) { return; }
+        this.data = reorderElementsInArray(this.data, oldIndex, newIndex, length);
+        this.forceUpdate();
```
giladgray: why not use state here?

packages/table/src/headers/columnHeader.tsx
```diff
@@ -60,6 +65,12 @@ export interface IColumnHeaderProps extends ISelectableProps, IColumnIndices, IC
     viewportRect: Rect;
 
     /**
+     * Enables/disables the reordering interaction.
+     * @default true
+     */
+    isReorderable?: boolean;
```
giladgray: is it reasonable to prevent specific columns from being reordered? like, does it make sense to support one-off disabling via this prop on `Column`, or is it more logical to prevent _all_ columns from being reordered via a single prop at a higher level?

packages/table/src/headers/columnHeader.tsx
```diff
@@ -205,13 +242,26 @@ export class ColumnHeader extends React.Component<IColumnHeaderProps, {}> {
...
+    private locateDragForSelection = (_event: MouseEvent, coords: ICoordinateData) => {
         const colStart = this.props.locator.convertPointToColumn(coords.activation[0]);
         const colEnd = this.props.locator.convertPointToColumn(coords.current[0]);
         return Regions.column(colStart, colEnd);
     }
+
+    private locateDragForReordering = (_event: MouseEvent, coords: ICoordinateData): number => {
+        let guideIndex = this.props.locator.convertPointToColumnLeftBoundary(coords.current[0]);
+        if (guideIndex < 0) {
+            guideIndex = null;
```
giladgray: `return null` instead of reassigning. 
also, did you mean `undefined`?

packages/table/src/headers/rowHeader.tsx
```diff
@@ -188,18 +209,45 @@ export class RowHeader extends React.Component<IRowHeaderProps, {}> {
...
+    private locateDragForSelection = (_event: MouseEvent, coords: ICoordinateData) => {
         const rowStart = this.props.locator.convertPointToRow(coords.activation[1]);
         const rowEnd = this.props.locator.convertPointToRow(coords.current[1]);
         return Regions.row(rowStart, rowEnd);
     }
+
+    private locateDragForReordering = (_event: MouseEvent, coords: ICoordinateData): number => {
+        let guideIndex = this.props.locator.convertPointToRowTopBoundary(coords.current[1]);
+        if (guideIndex < 0) {
+            guideIndex = null;
```
giladgray: again `return null`, or better `undefined`.

packages/table/src/headers/rowHeader.tsx
```diff
@@ -188,18 +209,45 @@ export class RowHeader extends React.Component<IRowHeaderProps, {}> {
...
+        return (!isReorderable) ? children : (
+            <DragReorderable
+                key={Classes.rowIndexClass(rowIndex)}
+                locateClick={this.locateClick}
+                locateDrag={this.locateDragForReordering}
+                onReorder={onReorder}
+                onReorderPreview={onReorderPreview}
+                onSelection={onSelection}
+                selectedRegions={selectedRegions}
+                toRegion={this.toRegion}
```
giladgray: can you pass `toRegion={Regions.row}` directly???

packages/table/src/interactions/reorderable.tsx
```diff
@@ -0,0 +1,147 @@
...
+ * Copyright 2016 Palantir Technologies, Inc. All rights reserved.
+ * Licensed under the BSD-3 License as modified (the “License”); you may obtain a copy
+ * of the license at https://github.com/palantir/blueprint/blob/master/LICENSE
+ * and https://github.com/palantir/blueprint/blob/master/PATENTS
+ */
+
+import * as PureRender from "pure-render-decorator";
+import * as React from "react";
+import { Utils } from "../common/utils";
+// import { DragEvents } from "../interactions/dragEvents";
```
giladgray: commented code

packages/table/src/interactions/selectable.tsx
```diff
@@ -118,33 +127,43 @@ export class DragSelectable extends React.Component<IDragSelectableProps, {}> {
...
+        const foundIndex = Regions.findMatchingRegion(selectedRegions, region);
         if (foundIndex !== -1) {
-            // If re-clicking on an existing region, we either carefully
-            // remove it if the meta key is used or otherwise we clear the
-            // selection entirely.
             if (DragEvents.isAdditive(event)) {
-                const newSelectedRegions = this.props.selectedRegions.slice();
+                // If re-clicking on an existing region, we either carefully
+                // remove it if the meta key is used or otherwise we clear the
+                // selection entirely.
```
giladgray: i think this comment belongs before the `if`, not after (cuz it also takes about the `else`)

packages/table/src/layers/guides.tsx
```diff
@@ -24,6 +24,11 @@ export interface IGuideLayerProps extends IProps {
 }
 
 export class GuideLayer extends React.Component<IGuideLayerProps, {}> {
+
+    // this should be consistent with the guide width/height specified in CSS.
+    // TODO: find a way to determine this programmatically - maybe after render?
```
giladgray: two options:
1. set the size of the guide element here in JS (and _not_ in CSS, so it's only set in one place), or
2. refactor CSS such that the logic here does not need to account for dimensions of guide in the DOM. like can the offset to center the guide be divorced from its positioning somehow? perhaps by using `transform` instead of negative margin in the CSS?

packages/table/src/locator.ts
```diff
@@ -24,12 +24,26 @@ export interface ILocator {
     convertPointToColumn: (clientX: number) => number;
 
     /**
+     * Locates a column boundary given the client X coordinate. The boundary
+     * is returned as the index of the column appearing to the right of that
+     * boundary. Returns -1 if the coordinate is not over a column.
+     */
+    convertPointToColumnLeftBoundary: (clientX: number) => number;
```
giladgray: i feel like the word "boundary" is wrong here. could this be simply `convertPointToColumnIndex`?

packages/table/src/locator.ts
```diff
@@ -122,8 +156,22 @@ export class Locator implements ILocator {
         return bodyRect.left + this.grid.getCumulativeWidthAt(index);
     }
 
+    private convertCellMidpointToClientX = (index: number) => {
+        const bodyRect = this.getBodyRect();
+        const cumWidth = this.grid.getCumulativeWidthAt(index);
+        const prevCumWidth = (index > 0) ? this.grid.getCumulativeWidthAt(index - 1) : 0;
+        return bodyRect.left + ((cumWidth + prevCumWidth) / 2.0);
```
giladgray: `/ 2` === `/ 2.0` in JS. it's always floating point. i'd say drop the decimal cuz we don't do that elsewhere.

packages/table/src/regions.ts
```diff
@@ -234,6 +236,24 @@ export class Regions {
...
+     * parameter. Returns -1 if no such region is found.
+     */
+    public static findContainingRegion(regions: IRegion[], region: IRegion) {
+        if (regions == null) {
+            return -1;
+        }
+
+        for (let i = 0; i < regions.length; i++) {
+            // containsRegion expects an array of regions as the first param
+            if (Regions.containsRegion([regions[i]], region)) {
```
giladgray: seems kludgy and ripe for refactoring so you can check containment for single regions. possibly a new function that can be reused to check an array?

packages/table/src/regions.ts
```diff
@@ -548,6 +544,42 @@ export class Regions {
         }
     }
 
+    private static maybeFullyOverlapsRegion(regions: IRegion[],
```
giladgray: `maybe` typically denotes _behavior_ that may or may not execute. `maybeRender`, `maybeHandle`.

this is an `is` situation since it returns a boolean.

packages/table/src/table.tsx
```diff
@@ -132,6 +132,10 @@ export interface ITableProps extends IProps, IRowHeights, IColumnWidths {
      */
     isRowHeaderShown?: boolean;
 
+    onColumnsReordered?: (oldIndex: number, newIndex: number, length: number) => void;
+
+    onRowsReordered?: (oldIndex: number, newIndex: number, length: number) => void;
```
giladgray: TODO docs on these two props!

packages/table/examples/tableFormatsExample.tsx
cmslewis: This was quick; I'm going to revert changes to this example and make a new one.

packages/table/src/headers/columnHeader.tsx
cmslewis: Oh gosh, I don't know of a case where a single column would need to disable reordering (unless it were a frozen column, which I don't know if we support yet?). Would prefer to keep it all-or-nothing for now.

packages/table/src/headers/columnHeader.tsx
cmslewis: Fixed

packages/table/src/headers/rowHeader.tsx
cmslewis: Derp, yes. Fixed.

packages/table/src/headers/rowHeader.tsx
cmslewis: Fixed.

packages/table/src/interactions/reorderable.tsx
cmslewis: Fixed

packages/table/src/interactions/selectable.tsx
cmslewis: Done

packages/table/src/locator.ts
cmslewis: @giladgray That naming would be ambiguous unfortunately. See this description of the already existing `convertPointToColumn`:

```
/**
 * Locates a column's index given the client X coordinate. Returns -1 if
 * the coordinate is not over a column.
 */
convertPointToColumn: (clientX: number) => number;
```

`convertPointToColumnLeftBoundary` is simply adding support for an extra column index off to the right (use case: if you're reordering and mousing over the right half of column `N`, this function would return the next greater index, such that a vertical guide would then show on the left boundary of column `N+1`).

Maybe it's better to just add an extra parameter to `convertPointToColumn`? Such a parameter would let us add support for an extra column on the end, rather than returning -1.

packages/table/src/locator.ts
cmslewis: Done

packages/table/src/headers/columnHeaderCell.tsx
```diff
@@ -131,18 +136,19 @@ export class ColumnHeaderCell extends React.Component<IColumnHeaderCellProps, IC
     };
 
     public render() {
-        const { className, isActive, isColumnSelected, loading, resizeHandle, style } = this.props;
+        const { loading } = this.props;
```
gscshoyru: Why the change here?

packages/table/src/regions.ts
cmslewis: @giladgray Yeah, I wasn't sure what to call this. Given that `intervalCompareFn` was parameterized, it's sometimes checking for full overlap and sometimes checking for partial overlap. LMK if you have suggestions. Maybe just `findRegion`?

packages/table/src/interactions/menus/menuContext.ts
```diff
@@ -47,7 +47,7 @@ export class MenuContext implements IMenuContext {
         private selectedRegions: IRegion[],
         private numRows: number,
         private numCols: number) {
-        this.regions = Regions.containsRegion(selectedRegions, target) ? selectedRegions : [ target ];
+        this.regions = Regions.overlapsRegion(selectedRegions, target) ? selectedRegions : [ target ];
```
gscshoyru: Yeah, this is probably a better name for what that does.

packages/table/src/regions.ts
cmslewis: Going with:

```
private static overlapsRegionInternal(regions: IRegion[], query: IRegion, allowPartialOverlap = true) {
    const intervalCompareFn = allowPartialOverlap ? Regions.overlapsRegion : Regions.containsRegion;
    ...
```

packages/table/src/table.tsx
cmslewis: Done

packages/table/src/table.tsx
cmslewis: Fixed

packages/table/src/common/classes.ts
cmslewis: Removed for now, but I guess I generally prefer exporting everything to avoid future requests for random stuff we thought wouldn't be valuable (and it's nice to keep all of our strings in one place.)

packages/table/src/regions.ts
cmslewis: Yeah, this is a little awkward. Happy to do a bigger refactor in `Regions.ts`. Does it make sense to scope that for a separate PR though?

packages/table/src/table.tsx
cmslewis: Done

packages/table/src/common/classes.ts
cmslewis: There are already `TABLE_OVERLAY` and `TABLE_OVERLAY_LAYER` constructs, but this isn't technically implemented as an overlay in the same sense, hence the new "shield" term. Happy to call it an overlay anyway though, if we're all cool with it.

packages/table/src/headers/columnHeaderCell.tsx
cmslewis: The props line got too long when I added my thing, plus elsewhere the Blueprint codebase, we don't deconstruct unless we need the same `this.props.X` variable in more than one place in the same scope. `loading` is the only prop here that's used in more than one place, so I left that and inlined everything else.

packages/table/src/common/classes.ts
cmslewis: The shield is an invisible layer I put over the whole table while the drag-reordering is happening. It's only purpose is to style the cursor as `grabbing` no matter what it hovers over.

packages/table/src/headers/rowHeader.tsx
cmslewis: @giladgray actually, just reran things. Using `Regions.row` directly messes up the `this` binding within it, and changing to fat-arrow notation isn't an option for static functions. Easiest to just leave as is.

packages/table/src/layers/guides.tsx
cmslewis: Fixed

cmslewis:
@giladgray @gscshoyru I met the coverage threshold in `reorderable.tsx`, but I wasn't able to do it in `tableTests.tsx` due to the new reordering tests working in Chrome but not Phantom. :/ Ended up dropping the threshold from 79 to 78, but would prefer not to do that. Will try to find a workaround.

packages/table/src/headers/rowHeader.tsx
cmslewis: (Revisiting: this is probably an argument to refactor `Regions` such that its member functions are declared outside of it.)

packages/table/src/headers/columnHeader.tsx
```diff
@@ -26,7 +27,11 @@ export interface IColumnWidths {
     defaultColumnWidth?: number;
 }
 
-export interface IColumnHeaderProps extends ISelectableProps, IColumnIndices, IColumnWidths, ILockableLayout {
+export interface IColumnHeaderProps extends IColumnIndices,
+                                            IColumnWidths,
+                                            ILockableLayout,
+                                            IReorderableProps,
+                                            ISelectableProps {
```
themadcreator: I don't know what our official styleguide says but i've always preferred:

```
export interface IColumnHeaderProps extends
    IColumnIndices,
    IColumnWidths,
    ILockableLayout,
    IReorderableProps,
    ISelectableProps {
```

packages/table/src/headers/columnHeader.tsx
themadcreator: Sounds good, but then should you move this prop up the stack?

packages/table/src/headers/columnHeader.tsx
```diff
@@ -147,7 +162,7 @@ export class ColumnHeader extends React.Component<IColumnHeaderProps, {}> {
 
         const rect = grid.getColumnRect(columnIndex);
         const handleSizeChanged = (size: number) => {
-            onResizeGuide([rect.left + size + 1]);
+            onResizeGuide([rect.left + size]);
```
themadcreator: What happened to the 1px? i think this was a design-required offset

packages/table/src/headers/columnHeader.tsx
```diff
@@ -168,17 +183,27 @@ export class ColumnHeader extends React.Component<IColumnHeaderProps, {}> {
...
-        return (
+        const children = (
             <DragSelectable
                 allowMultipleSelection={allowMultipleSelection}
+                ignoreSelectedRegionClicks={isColumnReorderable}
                 key={Classes.columnIndexClass(columnIndex)}
                 locateClick={this.locateClick}
-                locateDrag={this.locateDrag}
+                locateDrag={this.locateDragForSelection}
+                selectedRegions={selectedRegions}
```
themadcreator: alpha sort

packages/table/src/headers/columnHeaderCell.tsx
themadcreator: I don't think that's true. I think it's pretty common to destructure all the props you're going to use even if you use them only once.

Also, it's not like this PR is small, so adding a few extra lines in nbd. Also maybe that'll help you avoid lowering the test coverage threshold.

packages/table/src/interactions/dragEvents.ts
```diff
@@ -92,6 +99,9 @@ export class DragEvents {
         if (this.handler.stopPropagation) {
             event.stopPropagation();
         }
+        if (this.handler.stopImmediatePropagation) {
```
themadcreator: I've run into a lot of subtle issues when messing with these. Make sure you comment your usage where you set `stopImmediatePropagation`

packages/table/src/interactions/draggable.tsx
```diff
@@ -138,6 +145,7 @@ const REATTACH_PROPS_KEYS = [
 export class Draggable extends React.Component<IDraggableProps, {}> {
     public static defaultProps = {
         preventDefault: true,
+        stopImmediatePropagation: false,
```
themadcreator: here

packages/table/src/interactions/reorderable.tsx
```diff
@@ -0,0 +1,154 @@
+/**
+ * Copyright 2016 Palantir Technologies, Inc. All rights reserved.
```
themadcreator: 2017

packages/table/src/interactions/reorderable.tsx
```diff
@@ -0,0 +1,154 @@
...
+
+export interface IReorderableProps {
+    /**
+     * A callback that is called while the user is dragging to reorder.
+     *
+     * @param oldIndex the original index of the element or set of elements
+     * @param newIndex the new index of the element or set of elements
+     * @param length the number of contiguous elements that were moved
+     */
+    onReorderPreview: (oldIndex: number, newIndex: number, length: number) => void;
```
themadcreator: i think i called this onResizing for the resize case. Either way, would be nice to be consistent

packages/table/src/interactions/reorderable.tsx
```diff
@@ -0,0 +1,154 @@
...
+        return event.button === 0;
+    }
+
+    private selectedRegionStartIndex: number;
+    private selectedRegionLength: number;
+
+    public render() {
+        const draggableProps = this.getDraggableProps();
+        return (
+            <Draggable {...draggableProps} preventDefault={false}>
```
themadcreator: do we have draggable within draggables now with resizing and reordering?

packages/table/src/interactions/reorderable.tsx
```diff
@@ -0,0 +1,154 @@
...
+        );
+    }
+
+    private getDraggableProps(): IDraggableProps {
+        return this.props.onReorder == null ? {} : {
+            onActivate: this.handleActivate,
+            onDragEnd: this.handleDragEnd,
+            onDragMove: this.handleDragMove,
+        };
+    }
```
themadcreator: there seems to be a bit of duplication between this and `<DragSelectable>` i'm not sure how to de-dupe, but it seems like we could

packages/table/src/interactions/selectable.tsx
```diff
@@ -92,8 +99,10 @@ export class DragSelectable extends React.Component<IDragSelectableProps, {}> {
 
     public render() {
         const draggableProps = this.getDraggableProps();
+        // stop immediate propagation so that the same mousedown gesture won't also trigger
```
themadcreator: helpful

packages/table/src/locator.ts
```diff
@@ -122,8 +156,22 @@ export class Locator implements ILocator {
         return bodyRect.left + this.grid.getCumulativeWidthAt(index);
     }
 
+    private convertCellMidpointToClientX = (index: number) => {
+        const bodyRect = this.getBodyRect();
+        const cumWidth = this.grid.getCumulativeWidthAt(index);
+        const prevCumWidth = (index > 0) ? this.grid.getCumulativeWidthAt(index - 1) : 0;
```
themadcreator: getCumulative width is fairly expensive because it does a linear scan. Instead of doing it twice, you should just get the width of the specific cell from the grid.

packages/table/src/regions.ts
```diff
@@ -548,11 +541,52 @@ export class Regions {
         }
     }
 
+    private static overlapsRegionInternal(regions: IRegion[], query: IRegion, allowPartialOverlap = true) {
```
themadcreator: nit: it's private static, do we need to add "internal" ?

packages/table/src/regions.ts
```diff
@@ -548,11 +541,52 @@ export class Regions {
...
+    }
+
     private static regionsEqual(regionA: IRegion, regionB: IRegion) {
         return Regions.intervalsEqual(regionA.rows, regionB.rows)
             && Regions.intervalsEqual(regionA.cols, regionB.cols);
     }
 
+    private static regionContains(regionA: IRegion, regionB: IRegion) {
+        // containsRegion expects an array of regions as the first param
+        return Regions.overlapsRegionInternal([regionA], regionB, /* allowPartialOverlap */ false);
```
themadcreator: aren't overlaps by definition "partial"?

packages/table/src/table.tsx
```diff
@@ -80,6 +80,12 @@ export interface ITableProps extends IProps, IRowHeights, IColumnWidths {
     getCellClipboardData?: (row: number, col: number) => any;
 
     /**
+     * If `false`, disables reordering of columns.
+     * @default false
+     */
+    isColumnReorderable?: boolean;
```
themadcreator: nvm previous comment

packages/table/src/table.tsx
```diff
@@ -806,10 +849,6 @@ export class Table extends AbstractComponent<ITableProps, ITableState> {
      * intend to redraw the region layer.
      */
     private maybeRenderRegions(getRegionStyle: IRegionStyler) {
-        if (this.isGuidesShowing()) {
-            return undefined;
-        }
```
themadcreator: whytho?

packages/table/src/headers/columnHeader.tsx
cmslewis: @themadcreator do you mean up to a higher-level component? I'd still need some way to let all the column headers know whether or not to allow reordering though, right? That would seem to require a prop here regardless.

packages/table/src/headers/columnHeaderCell.tsx
cmslewis: > Also, it's not like this PR is small, so adding a few extra lines in nbd.

👏

> Also maybe that'll help you avoid lowering the test coverage threshold.

👏👏👏👏👏👏👏👏👏👏👏👏👏👏👏

Regarding the first point, that's just the pattern I've noticed elsewhere in the Blueprint codebase. I don't really mind one way or the other. Seems like the destructuring approach makes sense here though.

gulp/util/karma-config.js
```diff
@@ -40,8 +40,8 @@ module.exports = function createConfig(project) {
         coverageReporter: {
             check: {
                 each: {
-                    lines: 79,
-                    statements: 79,
+                    lines: 78,
+                    statements: 78,
```
cmslewis: 😔 working on it

packages/table/preview/index.tsx
```diff
@@ -531,3 +531,119 @@ ReactDOM.render(
...
+
+    private handleRowsReordered = (oldIndex: number, newIndex: number, length: number) => {
+        if (oldIndex === newIndex) { return; }
+        this.setState({ data: this.reorderElementsInArray(this.state.data, oldIndex, newIndex, length) });
+    }
+
+    private reorderElementsInArray(array: any[], from: number, to: number, length: number) {
+        const before = array.slice(0, from);
+        const within = array.slice(from, from + length);
+        const after = array.slice(from + length);
```
cmslewis: Oh I have no doubt there's a better way to do this. I'll play.

Also @themadcreator @gscshoyru what do you think of exposing this as a static method somewhere? It could take in the params shown here and spit out either an array of old indexes in the new ordering, or simply the reordered array of data. Thoughts?

packages/table/src/headers/columnHeader.tsx
cmslewis: Sounds fine to me!

packages/table/src/headers/columnHeader.tsx
cmslewis: It was causing a mis-alignment between the resize guide and the reorder guide (reorder guide was centered on the line, resize guide was off by one). Things should still look identical given my new CSS changes!

packages/table/src/headers/columnHeader.tsx
cmslewis: 👍

packages/table/src/interactions/reorderable.tsx
cmslewis: 👍

packages/table/src/interactions/reorderable.tsx
cmslewis: Yes, draggables in draggables in draggables:

`<DragReorderable><DragSelectable><Resizable /><DragSelectable><DragReorderable>`

packages/table/src/interactions/reorderable.tsx
cmslewis: Will look into it.

packages/table/src/regions.ts
cmslewis: There's already a function called `overlapsRegion`, and I couldn't think of a better name. Hence my tacking `internal` on the end.

packages/table/src/regions.ts
cmslewis: _At least_ partial in my opinion, which means something could fully overlap another. If you can think of better terminology, do feel free to suggest it! I'm not in love with these names.

packages/table/src/table.tsx
cmslewis: Ha, this is probably an errant change. Will undo if so.

packages/table/preview/index.tsx
themadcreator: Oh i see. Then iterating is fine. And extracting this to a util method would defo make sense

packages/table/src/locator.ts
cmslewis: Aren't the widths cached? `getCumulativeWidth` appears to return a precomputed value.

packages/table/src/table.tsx
cmslewis: K, so I remember now. With this check in place, selected regions don't appear while you're reordering (b/c reordering shows a guide, causing this function to return before drawing the selected region). And then selected regions remain broken even after the reordering completes for some reason.

So will need to remove after all!

gulp/util/karma-config.js
cmslewis: Update: Not sure I'll be able to get around the Phantom issues. May have to leave at 78 or (preferably) blacklist certain files if that's possible.

cmslewis:
@themadcreator @giladgray ready for another CR pass. still working on the `TruncatedPopover` bug that @adidahiya pointed out.

packages/table/src/table.tsx
themadcreator: well "And then selected regions remain broken even after the reordering completes for some reason." worries me that we don't know why. but okay

packages/table/src/locator.ts
themadcreator: ah, okay, then fine

themadcreator:
I expect we need to add some of these new props to some `shouldComponentUpdate`s somewhere

gulp/util/karma-config.js
giladgray: might this be useful?
https://github.com/gotwarlost/istanbul/blob/master/ignoring-code-for-coverage.md

packages/table/src/regions.ts
giladgray: do you know about `/** @internal */` JSDoc tag? things marked suchly will be stripped from the `.d.ts` typings.

giladgray:
worth investing some `@PureRender` usage in table components now?

packages/table/src/interactions/dragEvents.ts
cmslewis: Fixed! Will end up deleting this altogether. Pushing momentarily.

packages/table/src/headers/columnHeader.tsx
```diff
@@ -205,13 +247,24 @@ export class ColumnHeader extends React.Component<IColumnHeaderProps, {}> {
             return null;
         }
 
-        const col = this.props.locator.convertPointToColumn(event.clientX);
-        return Regions.column(col);
+        const columnIndex = this.props.locator.convertPointToColumn(event.clientX);
+
+        return Regions.column(columnIndex);
```
cmslewis: Will revert this cosmetic change.

cmslewis:
@themadcreator @giladgray - yeah, I found some opportunities for performance improvements when debugging the truncated-popover bug. Should we handle that in this PR?

cmslewis:
I'm also seeing a lot of duplicated functionality between `RowHeader` and `ColumnHeader`. We should pull the common code into an `AbstractHeader` component, e.g.

packages/table/preview/index.tsx
```diff
@@ -531,3 +531,77 @@ ReactDOM.render(
...
+    ["E", "Eggplant", "Elk", "Eritrea", "El Paso"],
+].map((array) => ({
+    // tslint:disable:object-literal-sort-keys
+    letter: array[0],
+    fruit: array[1],
+    animal: array[2],
+    country: array[3],
+    city: array[4],
+    // tslint:enable:object-literal-sort-keys
+}));
```
giladgray: i mean you might as well go hard...

```js
([letter, fruit, ...]) => ({ letter, fruit, ... })
```

packages/table/preview/index.tsx
```diff
@@ -531,3 +531,77 @@ ReactDOM.render(
...
+
+class ReorderableTableExample extends React.Component<{}, IReorderableTableExampleState> {
+    public state = {
+        data: REORDERABLE_TABLE_DATA,
+    } as IReorderableTableExampleState;
```
giladgray: always better to declare type up front instead of casting. sometimes you lose type safety when casting, which you don't lose if you type the var first.

`public state: IReorderableTableExampleState = { ... }`

packages/table/src/headers/columnHeader.tsx
```diff
@@ -78,13 +91,35 @@ export interface IColumnHeaderProps extends ISelectableProps, IColumnIndices, IC
...
+export class ColumnHeader extends React.Component<IColumnHeaderProps, IColumnHeaderState> {
     public static defaultProps = {
+        isReorderable: false,
         isResizable: true,
         loading: false,
     };
 
+    public state = {
+        hasSelectionEnded: false,
+    } as IColumnHeaderState;
```
giladgray: again, do it up front. and on one line. 
`public state: IColumnHeaderState = { hasSelectionEnded: false };`

packages/table/src/headers/columnHeader.tsx
```diff
@@ -168,50 +205,98 @@ export class ColumnHeader extends React.Component<IColumnHeaderProps, {}> {
...
+        const guideIndex = this.props.locator.convertPointToColumn(coords.current[0], true);
+        return (guideIndex < 0) ? undefined : guideIndex;
+    }
+
+    private toRegion = (index1: number, index2?: number) => {
+        // can't pass Regions.column directly, because that would break its internal `this` binding.
+        return Regions.column(index1, index2);
+    }
+
+    private isColumnTemporarilyReorderable(isColumnSelected: boolean) {
```
giladgray: what is this for? what do these five conditions mean? code comment please

packages/table/src/headers/columnHeader.tsx
```diff
@@ -168,50 +205,98 @@ export class ColumnHeader extends React.Component<IColumnHeaderProps, {}> {
         });
         const cellLoading = cell.props.loading != null ? cell.props.loading : loading;
         const isColumnSelected = Regions.hasFullColumn(selectedRegions, columnIndex);
-        const cellProps: IColumnHeaderCellProps = { className, isColumnSelected, loading: cellLoading };
+        const isColumnTemporarilyReorderable = this.isColumnTemporarilyReorderable(isColumnSelected);
```
giladgray: does it matter that it's "temporary?" i think "currently" might be a better word, cuz it's determined by props and state. but that's just pure react, so perhaps we don't need such a word at all.

usage doesn't suggest any notion of temporary--you immediately rename it to `isColumnReorderable`, which as a prop is understood to only matter in the now, not next render.

packages/table/src/headers/rowHeader.tsx
```diff
@@ -161,45 +197,93 @@ export class RowHeader extends React.Component<IRowHeaderProps, {}> {
         });
         const cellLoading = cell.props.loading != null ? cell.props.loading : loading;
         const isRowSelected = Regions.hasFullRow(selectedRegions, rowIndex);
-        const cellProps: IRowHeaderCellProps = { className, isRowSelected, loading: cellLoading };
+        const isRowTemporarilyReorderable = this.isRowTemporarilyReorderable(isRowSelected);
```
giladgray: again with the temporary

packages/table/src/interactions/selectable.tsx
```diff
@@ -118,76 +130,105 @@ export class DragSelectable extends React.Component<IDragSelectableProps, {}> {
...
     }
+
+    private shouldIgnoreMouseDown(event: MouseEvent) {
+        return !Utils.isLeftClick(event) || this.props.disabled;
+    }
+
+    private getDragSelectedRegions(event: MouseEvent, coords: ICoordinateData) {
+        let region = (this.props.allowMultipleSelection) ?
+            this.props.locateDrag(event, coords) :
+            this.props.locateClick(event);
```
giladgray:
```
condition
    ? true
    : false;
```

packages/table/src/regions.ts
giladgray: 👎 why can't this just be `overlapsRegion`? the third prop is optional with a default anyway. 
i don't see the need for both.

packages/table/src/regions.ts
giladgray: i don't think we need the prop name comment in typescript. i can easily see it by hovering on function name.

packages/table/preview/index.tsx
cmslewis: Done

packages/table/preview/index.tsx
cmslewis: DOne

packages/table/src/headers/columnHeader.tsx
cmslewis: Done

packages/table/src/headers/columnHeader.tsx
cmslewis: Yeah wasn't sure about this, but I wanted to more strongly disambiguate `isReorderable` from `isColumnReorderable`. It's not necessarily clear at a glance what the difference is.

packages/table/src/headers/columnHeader.tsx
cmslewis: Done

packages/table/src/interactions/selectable.tsx
cmslewis: Oops. Done

packages/table/src/headers/rowHeader.tsx
cmslewis: Done

cmslewis:
@giladgray fixed the stuff in your CR just now.

packages/table/src/headers/columnHeader.tsx
adidahiya: I agree with @themadcreator, this should not be exposed at this level if we don't expect to support it. How about if you make it `@internal`, does that work? If not, you could try to use component `context` to allow `<Table>` to talk to its sub components.

packages/table/src/interactions/reorderable.tsx
```diff
@@ -0,0 +1,157 @@
...
+    onReordering: (oldIndex: number, newIndex: number, length: number) => void;
+
+    /**
+     * A callback that is called when the user is done dragging to reorder.
+     *
+     * @param oldIndex the original index of the element or set of elements
+     * @param newIndex the new index of the element or set of elements
+     * @param length the number of contiguous elements that were moved
+     */
+    onReordered: (oldIndex: number, newIndex: number, length: number) => void;
```
adidahiya: `onReorderEnd` might be more intuitive.

packages/table/src/interactions/reorderable.tsx
```diff
@@ -0,0 +1,157 @@
...
+
+export interface IReorderableProps {
+    /**
+     * A callback that is called while the user is dragging to reorder.
+     *
+     * @param oldIndex the original index of the element or set of elements
+     * @param newIndex the new index of the element or set of elements
+     * @param length the number of contiguous elements that were moved
+     */
+    onReordering: (oldIndex: number, newIndex: number, length: number) => void;
```
adidahiya: what is this useful for?

cmslewis: Updating the position of the vertical and horizontal guides as you drag. See `table.tsx`.

packages/table/src/interactions/reorderable.tsx
cmslewis: Yeah, I couldn't find great precedent for this in the codebase. Slider does `onChange` and `onRelease`, `DragSelectable` now has `onSelection` and `onSelectionEnd` as of a recent commit in this PR—do you think `onReordering` still makes sense if we change `onReordered` to `onReorderEnd`?

packages/table/src/headers/columnHeader.tsx
cmslewis: Haven't used `@internal` before, but looks interesting. I'll take a look.

packages/table/src/interactions/reorderable.tsx
themadcreator: well, wouldn't `onReorder` work? You have `onReordering` while it's going then `onReorder` once it's finished. That kind of matches the other method names

themadcreator:
@cmslewis if I were you, I'd be itching to merge this but I do think that perf regression is pretty bad. Also, it's probably just a matter of checking the stack of `shouldComponentUpdate`s and making sure we're not creating new function instances for props, etc.

Drop some console logs in the render methods and compare before after your changes. If you can't find anything obvious with a concerted effort, we can move on, but I expect users who upgrade to this will not be happy.

themadcreator:
@cmslewis re: `RowHeader` vs. `ColumnHeader`, I also noticed this and other PR reviewers have suggested it as well!

I think I didn't originally do it because there were so many places where we use x/width vs. y/height sprinkled around that it seemed a little annoying to combine. But, If you can reduce those class files by moving at least a few methods into an abstract class, go for it! (not necessary for this PR obvs)

adidahiya:
Yeah I saw the perf issues on @gscshoyru's PR too. Are you sure it's not a difference caused by using production vs. dev builds of React?

Either way, try using [why-did-you-update](https://github.com/garbles/why-did-you-update).

giladgray:
approving since the code change looks legit, but we def need some serious perf work ASAP.

cmslewis:
@themadcreator @adidahiya @giladgray would love to look into perf stuff ASAP. From chatting offline, sounds like our plan is to merge this PR and then address perf immediately in another one. The next Blueprint release would block on the followup PR merging. Sound good?

cmslewis:
Also @giladgray FYI I just pushed a fix for a minor visual regression (hiding selected regions while resizing) and added tests to verify everything works there as expected.

packages/table/test/harness.ts
```diff
@@ -56,6 +56,10 @@ export class ElementHarness {
         this.element = element;
     }
 
+    public exists() {
+        return this.element != null;
+    }
```
themadcreator: 👍

adidahiya:
:shipit:

Issue #953, opened 2017-04-05
[Table] Column resizing doesn't take truncation popover into account


tgreenwatts:
![fandango_fail](https://cloud.githubusercontent.com/assets/24398123/24725980/85c9b052-1a05-11e7-9c77-6e638b6a6e68.gif)

tgreenwatts:
![fail](https://cloud.githubusercontent.com/assets/24398123/24877293/914984e8-1de3-11e7-8ef4-cdc22e1ede88.gif)
You can see that if you expand past the content, then resize, you get the correct width and can see everything. If you make the column smaller and resize, you get the *same width* but with the '...' instead of the content.

PR #1009, opened 2017-04-19
Add cell class to row/column headers so auto-resize works properly

### LINKED ISSUE #952: [Table] Resize columns based only on current column header
Currently if you resize a column (double click on edge) it will resize it to the correct width for all visible content + all header cells. It should only take the current column

Notice in this example the columns I'm resizing get larger instead of smaller - they all in fact will go to ~150px 
resize

gscshoyru:

Added the relevant classes to the column and row headers, so they'll be included if you try to select all the cells in a given column/row. 
Modified the resize logic to only grab cells with that one class, instead of also grabbing all the headers in the entire table.

### DIFF
--- a/packages/table/src/headers/columnHeader.tsx
+++ b/packages/table/src/headers/columnHeader.tsx
@@ -209,9 +209,9 @@ export class ColumnHeader extends React.Component<IColumnHeaderProps, IColumnHea
         };
 
         const cell = cellRenderer(columnIndex);
-        const className = classNames(cell.props.className, extremaClasses, {
+        const className = classNames(extremaClasses, {
             [Classes.TABLE_DRAGGABLE]: (onSelection != null),
-        }, Classes.columnCellIndexClass(columnIndex));
+        }, Classes.columnCellIndexClass(columnIndex), cell.props.className);
         const cellLoading = cell.props.loading != null ? cell.props.loading : loading;
         const isColumnSelected = Regions.hasFullColumn(selectedRegions, columnIndex);
         const isColumnCurrentlyReorderable = this.isColumnCurrentlyReorderable(isColumnSelected);
--- a/packages/table/src/headers/columnHeader.tsx
+++ b/packages/table/src/headers/columnHeader.tsx
@@ -211,7 +211,7 @@ export class ColumnHeader extends React.Component<IColumnHeaderProps, IColumnHea
         const cell = cellRenderer(columnIndex);
         const className = classNames(cell.props.className, extremaClasses, {
             [Classes.TABLE_DRAGGABLE]: (onSelection != null),
-        });
+        }, Classes.columnCellIndexClass(columnIndex));
         const cellLoading = cell.props.loading != null ? cell.props.loading : loading;
         const isColumnSelected = Regions.hasFullColumn(selectedRegions, columnIndex);
         const isColumnCurrentlyReorderable = this.isColumnCurrentlyReorderable(isColumnSelected);
--- a/packages/table/src/locator.ts
+++ b/packages/table/src/locator.ts
@@ -70,11 +70,7 @@ export class Locator implements ILocator {
     }
 
     public getWidestVisibleCellInColumn(columnIndex: number): number {
-        const cellClasses = [
-            `.${Classes.columnCellIndexClass(columnIndex)}`,
-            `.${Classes.TABLE_COLUMN_NAME}`,
-        ];
-        const cells = this.tableElement.querySelectorAll(cellClasses.join(", "));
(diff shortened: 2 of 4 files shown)


adidahiya:
devtest looks great except for this one:

gscshoyru:
Hmm.

That's a separate (unrelated) problem that may take more work in a different way. The problem should have already existed, if I'm wrong do let me know. But essentially -- the way the resize works (presently) is by measuring the length of the text content. The issue here is that there's a lot of text content in these cells, and there's some newlines in there -- so the measured text content isn't going to line up with the way it's going to end up rendering. 

The issue here is that determining "how it's going to end up rendering" is hard, with custom cell renderers and custom headers. But we should still try and file it as a separate issue -- there's probably _something_ we can do.

cmslewis:
Other issues with column resizing in general, observed in the developer preview:

- Auto-resizing is off by a few pixels here:


- Auto-resizing leaves this custom column header a little wider than it needs to be:

packages/table/src/headers/columnHeader.tsx
```diff
@@ -211,7 +211,7 @@ export class ColumnHeader extends React.Component<IColumnHeaderProps, IColumnHea
         const cell = cellRenderer(columnIndex);
         const className = classNames(cell.props.className, extremaClasses, {
```
giladgray: can you move `props.className` to be the last arg? this allows the user to override any of our computed props.

packages/table/src/locator.ts
```diff
@@ -70,11 +70,7 @@ export class Locator implements ILocator {
     }
 
     public getWidestVisibleCellInColumn(columnIndex: number): number {
-        const cellClasses = [
-            `.${Classes.columnCellIndexClass(columnIndex)}`,
-            `.${Classes.TABLE_COLUMN_NAME}`,
-        ];
-        const cells = this.tableElement.querySelectorAll(cellClasses.join(", "));
+        const cells = this.tableElement.querySelectorAll(`.${Classes.columnCellIndexClass(columnIndex)}`);
```
giladgray: 🛑 there's a tool for this!

```js
this.tableElement.getElementsByClassName(Classes.columnCellIndexClass(columnIndex))
```

packages/table/test/columnTests.tsx
```diff
@@ -75,13 +76,16 @@ describe("Column", () => {
         expectCellLoading(columnHeaders[1], CellType.COLUMN_HEADER);
         expectCellLoading(columnHeaders[2], CellType.COLUMN_HEADER, false);
 
-        const col0cells = table.element.queryAll(`.${Classes.columnCellIndexClass(0)}`);
+        const col0CellsClass = classNames(Classes.columnCellIndexClass(0), Classes.TABLE_CELL);
+        const col0cells = table.element.queryAll(col0CellsClass);
```
giladgray: 🛑 these tests are not actually passing because this always returns an empty array cuz! 
you're querying for `"a b"`, not `".a.b"`. use `getElementsByClassName("a b")`, or refactor to generate CSS selectors instead of HTML `class` attribute values.

same applies for changes below in this file.

gscshoyru:
@cmslewis 

First one is already tracked here: https://github.com/palantir/blueprint/issues/953

Second one: same as Adi's comment. It's measuring all the text -- including the "sup" at the top and the "hey dawg" at the bottom. It's expanding it to be the total of those two -- that's why it's off.

packages/table/test/columnTests.tsx
gscshoyru: We should probably do a length check on the number of cells, in this test, then. Whoops, though, good catch. Will fix and add length check.

packages/table/src/locator.ts
giladgray: why not make this change?

giladgray: ah you changed the one below but missed this instance

gscshoyru: Ah, sorry, should have said. 

Because the thing returned by getElementsByClassName is not iterable using forEach, essentially. It's not an array -- I'd have to use a normal for loop to iterate over it, which seems silly.

gscshoyru: ...huh, I didn't realize I was already doing it differently below. Weird. Consistency is probably more important here, I'll make it a for loop.

gscshoyru: ...it's already a for loop in this one too. I could have sworn I made this change somewhere, it didn't like it, and so I changed it back. Maybe I'm going crazy.

tgreenwatts:
Ping on this @gscshoyru

PR #1023, opened 2017-04-20
[Table] Auto-scroll viewport when moving focused cell with arrow keys 

### LINKED ISSUE #915: Scroll table to currently focused selection
For example:
User clicks on a cell in the table and use the arrow keys to move to the right until off the screen
Expected: The table scrolls
Actual: The table does not scroll, focus cell is out of view

Follow up to #718
One piece of #298

cmslewis:
#### Checklist

- [x] Include tests
- [x] Update documentation

- Auto-scroll the table body when you move the focused cell around with arrow keys.
- Fix existing errors that were being thrown when arrow-navigating when `enableFocus={false}`.
- Add unit tests for existing focus-cell arrow navigation behavior.
- Add unit tests for new auto-scroll behavior.
- 🎉 Bumped the coverage threshold back up to `80%`.

- When moving the focused cell `left` or `up`, its border would get clipped slightly if you were to auto-scroll to its top/left edge. Subtracting `1px` solves the issue.
- Added special logic to handle the following edge case that popped up earlier on:
    1. Say you have a focused cell wider than the viewport.
    1. Now keyboard-navigate `up` and `down` a bunch.
    1. Observe that the viewport see-saws between aligning to the left edge of the focused cell, then the right edge, then the left edge, then the right edge, and so on.
    - **Instead of the above:** the special casing simply favors aligning to the _left_ edge of oversized cells always. And in the other dimension, we favor aligning to the _top_ edge when going left and right with an overly tall focused cell.  
- Are there any other weird edge cases you can find?

#### Screenshot

Basic behavior:

Handling overly wide cells:

Handling overly tall cells:

Handling cells that are both overly wide and overly tall:

### DIFF
--- a/gulp/util/karma-config.js
+++ b/gulp/util/karma-config.js
@@ -40,8 +40,8 @@ module.exports = function createConfig(project) {
         coverageReporter: {
             check: {
                 each: {
-                    lines: 79,
-                    statements: 79,
+                    lines: 80,
+                    statements: 80,
                 },
             },
             includeAllSources: true,
@@ -54,8 +54,8 @@ module.exports = function createConfig(project) {
                 { type: "text" },
             ],
             watermarks: {
-                lines: [79, 90],
-                statements: [79, 90],
+                lines: [80, 90],
+                statements: [80, 90],
             },
         },
         files: filesToInclude,
--- a/packages/table/preview/index.tsx
+++ b/packages/table/preview/index.tsx
@@ -286,6 +286,7 @@ ReactDOM.render(
     getTableComponent(200, 100 * 1000, {} , {
         fillBodyWithGhostCells: true,
         selectionModes: SelectionModes.ALL,
+        isRowResizable: true,
     }),
     document.getElementById("table-big"),
 );
--- a/packages/table/preview/index.tsx
+++ b/packages/table/preview/index.tsx
@@ -286,7 +286,6 @@ ReactDOM.render(
     getTableComponent(200, 100 * 1000, {} , {
         fillBodyWithGhostCells: true,
         selectionModes: SelectionModes.ALL,
-        isRowResizable: true,
     }),
     document.getElementById("table-big"),
 );
--- a/packages/table/src/table.tsx
+++ b/packages/table/src/table.tsx
@@ -1152,7 +1152,6 @@ export class Table extends AbstractComponent<ITableProps, ITableState> {
     // no good way to call arrow-key keyboard events from tests
     /* istanbul ignore next */
     private handleFocusMove = (e: KeyboardEvent, direction: "up" | "down" | "left" | "right") => {
-        console.log("handleFocusMove", direction, this.state.focusedCell);
         e.preventDefault();
         e.stopPropagation();
 
@@ -1182,9 +1181,6 @@ export class Table extends AbstractComponent<ITableProps, ITableState> {
                 break;
         }
 
-        console.log("  newFocusedCell", newFocusedCell);
-        console.log("  grid", grid.numRows, grid.numCols);
-
         if (newFocusedCell.row < 0 || newFocusedCell.row >= grid.numRows ||
             newFocusedCell.col < 0 || newFocusedCell.col >= grid.numCols) {
             return;
@@ -1192,7 +1188,6 @@ export class Table extends AbstractComponent<ITableProps, ITableState> {
 
         // change selection to match new focus cell location
         const newSelectionRegions = [Regions.cell(newFocusedCell.row, newFocusedCell.col)];
-        console.log("newSelectionRegions", newSelectionRegions);
         this.handleSelection(newSelectionRegions);
         this.handleFocus(newFocusedCell);
 
@@ -1231,28 +1226,53 @@ export class Table extends AbstractComponent<ITableProps, ITableState> {
         const isFocusedCellWiderThanViewport = focusedCellWidth > viewportRect.width;
         const isFocusedCellTallerThanViewport = focusedCellHeight > viewportRect.height;
 
+        let nextScrollTop = viewportRect.top;
+        let nextScrollLeft = viewportRect.left;
+
         // keep the top end of an overly tall focused cell in view when moving left and right
         // (without this OR check, the body seesaws to fit the top end, then the bottom end, etc.)
         if (focusedCellBounds.top < viewportBounds.top || isFocusedCellTallerThanViewport) {
             // scroll up (minus one pixel to avoid clipping the focused-cell border)
-            this.bodyElement.scrollTop = Math.max(0, focusedCellBounds.top - 1);
+            nextScrollTop = Math.max(0, focusedCellBounds.top - 1);
         } else if (focusedCellBounds.bottom > viewportBounds.bottom) {
             // scroll down
-            this.bodyElement.scrollTop = viewportBounds.top + (focusedCellBounds.bottom - viewportBounds.bottom);
+            nextScrollTop = viewportBounds.top + (focusedCellBounds.bottom - viewportBounds.bottom);
         }
 
         // keep the left end of an overly wide focused cell in view when moving up and down
         if (focusedCellBounds.left < viewportBounds.left || isFocusedCellWiderThanViewport) {
             // scroll left (again minus one additional pixel)
-            this.bodyElement.scrollLeft = Math.max(0, focusedCellBounds.left - 1);
+            nextScrollLeft = Math.max(0, focusedCellBounds.left - 1);
         } else if (focusedCellBounds.right > viewportBounds.right) {
             // scroll right
-            this.bodyElement.scrollLeft = viewportBounds.left + (focusedCellBounds.right - viewportBounds.right);
+            nextScrollLeft = viewportBounds.left + (focusedCellBounds.right - viewportBounds.right);
+        }
+
+        const didScrollTopChange = nextScrollTop !== viewportRect.top;
+        const didScrollLeftChange = nextScrollLeft !== viewportRect.left;
+
+        if (didScrollTopChange || didScrollLeftChange) {
+            // we need to modify the body element explicitly for the viewport to shift
+            if (didScrollTopChange) {
+                this.bodyElement.scrollTop = nextScrollTop;
+            }
+            if (didScrollLeftChange) {
+                this.bodyElement.scrollLeft = nextScrollLeft;
+            }
+
+            // and we also need to update the viewport explicitly so that unit tests can grab the
+            // new offsets (in PhantomJS, bodyElement has all properties zeroed out)
+            const nextViewportRect = new Rect(
+                nextScrollLeft,
+                nextScrollTop,
+                viewportRect.width,
+                viewportRect.height,
+            );
+            this.setState({ viewportRect: nextViewportRect });
         }
     }
 
     private handleFocus = (focusedCell: IFocusedCellCoordinates) => {
-        console.log("handleFocus", this.props.enableFocus);
         if (!this.props.enableFocus) {
             // don't set focus state if focus is not allowed
             return;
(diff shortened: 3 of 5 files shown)


packages/table/preview/index.tsx
```diff
@@ -286,6 +286,7 @@ ReactDOM.render(
     getTableComponent(200, 100 * 1000, {} , {
         fillBodyWithGhostCells: true,
         selectionModes: SelectionModes.ALL,
+        isRowResizable: true,
```
cmslewis: Do we want to keep this? (It was helpful for this PR.) Or do we want to wait until we build one Table example to rule them all.

packages/table/src/table.tsx
```diff
@@ -1185,6 +1190,86 @@ export class Table extends AbstractComponent<ITableProps, ITableState> {
         const newSelectionRegions = [Regions.cell(newFocusedCell.row, newFocusedCell.col)];
         this.handleSelection(newSelectionRegions);
         this.handleFocus(newFocusedCell);
+
+        // keep the focused cell in view
+        this.scrollBodyToFocusedCell(newFocusedCell);
+    }
+
+    private scrollBodyToFocusedCell = (focusedCell: IFocusedCellCoordinates) => {
+        if (focusedCell == null) {
```
cmslewis: Whoops, this check is redunant with [Line 1159](https://github.com/palantir/blueprint/pull/1023/files#diff-6bc89efba539cc29af0dae9b17c57f75R1159) now. Will remove.

packages/table/src/table.tsx
```diff
@@ -1185,6 +1190,86 @@ export class Table extends AbstractComponent<ITableProps, ITableState> {
...
+            // this can happen if we have a selectedRegionTransform in play
+            return;
+        }
+
+        const { row, col } = focusedCell;
+        const { viewportRect } = this.state;
+
+        // sort keys in normal CSS position order (per the trusty TRBL/"trouble" acronym)
```
cmslewis: This doesn't really matter, but it makes me happier this way.

packages/table/test/tableTests.tsx
```diff
@@ -383,6 +387,149 @@ describe("<Table>", () => {
...
+    describe("Focused cell", () => {
+        let onFocus: Sinon.SinonSpy;
+
+        const NUM_ROWS = 3;
+        const NUM_COLS = 3;
+
+        // center the initial focus cell
+        const DEFAULT_FOCUSED_CELL_COORDS = { row: 1, col: 1 } as IFocusedCellCoordinates;
+
+        // Enzyme appears to render our Table at 60px high x 400px wide. make all rows and columns
```
cmslewis: A dangerous assumption. How can I specify viewport size explicitly? Something like http://phantomjs.org/api/webpage/property/viewport-size.html?

packages/table/test/tableTests.tsx
```diff
@@ -383,6 +387,149 @@ describe("<Table>", () => {
...
+
+                component.simulate("keyDown", keyEventConfig);
+                expect(component.state("viewportRect").top).to.equal(EXPECTED_TOP_OFFSET);
+                component.simulate("keyDown", keyEventConfig);
+                expect(component.state("viewportRect").top).to.equal(EXPECTED_TOP_OFFSET);
+            });
+
+            it("keeps left edge of oversized focus cell in view when moving up and down", () => {
+                // subtract one pixel to avoid clipping the focus cell
+
```
cmslewis: Whoops, malformatted. Will fix.

packages/table/test/tableTests.tsx
```diff
@@ -383,6 +387,149 @@ describe("<Table>", () => {
...
+                keyCode,
+                shiftKey,
+                preventDefault: () => { /* Empty */ },
+                stopPropagation: () => { /* Empty */ },
+                target: (component as any).getNode(), // `getNode` is a real Enzyme method, just not in the typings?
+                which: keyCode,
+            };
+            return {
+                eventConfig,
+                nativeEvent: eventConfig,
```
cmslewis: @themadcreator here's the keyboard testing magic.

packages/table/test/tableTests.tsx
```diff
@@ -383,6 +387,149 @@ describe("<Table>", () => {
...
+                                                    expectedOffset: number) {
+                it(key, () => {
+                    const { component } = mountTable();
+                    component.simulate("keyDown", createKeyEventConfig(component, key, keyCode));
+                    expect(component.state("viewportRect")[attrToCheck]).to.equal(expectedOffset);
+                });
+            }
+        });
+
+
```
cmslewis: Whoops, extra newline. Will delete.

packages/table/src/table.tsx
cmslewis: Done

packages/table/test/tableTests.tsx
cmslewis: Done

packages/table/test/tableTests.tsx
cmslewis: Done

packages/table/src/table.tsx
```diff
@@ -1185,6 +1191,81 @@ export class Table extends AbstractComponent<ITableProps, ITableState> {
...
+        let nextScrollLeft = viewportRect.left;
+
+        // keep the top end of an overly tall focused cell in view when moving left and right
+        // (without this OR check, the body seesaws to fit the top end, then the bottom end, etc.)
+        if (focusedCellBounds.top < viewportBounds.top || isFocusedCellTallerThanViewport) {
+            // scroll up (minus one pixel to avoid clipping the focused-cell border)
+            nextScrollTop = Math.max(0, focusedCellBounds.top - 1);
+        } else if (focusedCellBounds.bottom > viewportBounds.bottom) {
+            // scroll down
+            nextScrollTop = viewportBounds.top + (focusedCellBounds.bottom - viewportBounds.bottom);
```
giladgray: this looks like `focusedCellBounds.bottom - viewportBounds.height`? unless i'm doing the math wrong?

`height = bottom - top` so `top - bottom = -height`

packages/table/src/table.tsx
```diff
@@ -1185,6 +1191,81 @@ export class Table extends AbstractComponent<ITableProps, ITableState> {
...
+        }
+
+        // keep the left end of an overly wide focused cell in view when moving up and down
+        if (focusedCellBounds.left < viewportBounds.left || isFocusedCellWiderThanViewport) {
+            // scroll left (again minus one additional pixel)
+            nextScrollLeft = Math.max(0, focusedCellBounds.left - 1);
+        } else if (focusedCellBounds.right > viewportBounds.right) {
+            // scroll right
+            nextScrollLeft = viewportBounds.left + (focusedCellBounds.right - viewportBounds.right);
```
giladgray: same here but with `right - width`

packages/table/src/table.tsx
```diff
@@ -1185,6 +1191,81 @@ export class Table extends AbstractComponent<ITableProps, ITableState> {
...
+            // we need to modify the body element explicitly for the viewport to shift
+            if (didScrollTopChange) {
+                this.bodyElement.scrollTop = nextScrollTop;
+            }
+            if (didScrollLeftChange) {
+                this.bodyElement.scrollLeft = nextScrollLeft;
+            }
+
+            // and we also need to update the viewport explicitly so that unit tests can grab the
+            // new offsets (in PhantomJS, bodyElement has all properties zeroed out)
```
giladgray: wait this can't be _just_ for unit tests; you rely on `viewportRect` for the scrolling logic in this very function

packages/table/test/tableTests.tsx
```diff
@@ -383,6 +387,148 @@ describe("<Table>", () => {
...
+                                      key: string,
+                                      keyCode: number,
+                                      shiftKey = false) {
+            const eventConfig = {
+                key,
+                keyCode,
+                shiftKey,
+                preventDefault: () => { /* Empty */ },
+                stopPropagation: () => { /* Empty */ },
+                target: (component as any).getNode(), // `getNode` is a real Enzyme method, just not in the typings?
```
giladgray: we should probably update all our `@types`...

packages/table/src/table.tsx
cmslewis: That makes sense. I originally thought of this as `initialPosition` + `delta` and had a separate `scrollDelta` variable, but your approach is better when it's all inlined like this.

packages/table/src/table.tsx
cmslewis: K

gulp/util/karma-config.js
```diff
@@ -40,8 +40,8 @@ module.exports = function createConfig(project) {
         coverageReporter: {
             check: {
                 each: {
-                    lines: 79,
-                    statements: 79,
+                    lines: 80,
```
gscshoyru: Yay!

packages/table/src/table.tsx
```diff
@@ -1156,6 +1156,12 @@ export class Table extends AbstractComponent<ITableProps, ITableState> {
         e.stopPropagation();
 
         const { focusedCell } = this.state;
+        if (focusedCell == null) {
```
gscshoyru: Hm, what interaction made this necessary? We should look into how we want to support this, if we do. Which we may not.

packages/table/test/tableTests.tsx
```diff
@@ -383,6 +387,148 @@ describe("<Table>", () => {
...
+        // make these values arbitrarily bigger than the table bounds
+        const OVERSIZED_ROW_HEIGHT = 10000;
+        const OVERSIZED_COL_WIDTH = 10000;
+
+        beforeEach(() => {
+            onFocus = sinon.spy();
+        });
+
+        describe("moves a focus cell with arrow keys", () => {
+            runFocusCellMoveTest("up", Keys.ARROW_UP, { row: 0, col: 1 });
```
gscshoyru: You got arrow keys to work in the tests? How?

packages/table/test/tableTests.tsx
gscshoyru: Yeah, would like explanation of this at some point if possible. What.

packages/table/src/table.tsx
cmslewis: Yeah, interactions worked before I added this explicit `setState`, hence the comment. But including the `setState` does seem to be the right approach.

packages/table/test/tableTests.tsx
cmslewis: Agree. In this PR?

packages/table/src/table.tsx
cmslewis: Try keyboard-arrowing around in the `selectedRegionTransform` example on `/table/preview`. That's where the errors were showing up for me.

cmslewis: Try arrowing around in the `selectedRegionTransform` example.

image

packages/table/test/tableTests.tsx
cmslewis: See below.

packages/table/test/tableTests.tsx
cmslewis: `HotkeysTarget` passes `event.nativeEvent` to the component's `onKeyDown`, so you have to explicitly override that object here (while also setting the top-level `key`, `which`, etc., because those are referenced here and there.)

packages/table/test/tableTests.tsx
giladgray: heck no

packages/table/test/tableTests.tsx
giladgray: `nativeEvent` is a `React.SyntheticEvent` thing. this makes 💯  sense.

cmslewis:
@giladgray @gscshoyru @themadcreator ready for re-review

packages/table/src/table.tsx
giladgray: indeed it does 👍

packages/table/test/tableTests.tsx
giladgray: filed #1031 to track

packages/table/src/table.tsx
```diff
@@ -1185,6 +1191,81 @@ export class Table extends AbstractComponent<ITableProps, ITableState> {
...
+
+        // keep the focused cell in view
+        this.scrollBodyToFocusedCell(newFocusedCell);
+    }
+
+    private scrollBodyToFocusedCell = (focusedCell: IFocusedCellCoordinates) => {
+        const { row, col } = focusedCell;
+        const { viewportRect } = this.state;
+
+        // sort keys in normal CSS position order (per the trusty TRBL/"trouble" acronym)
```
giladgray: "trouble" or "blurt" 😆

PR #1027, opened 2017-04-21
[Table] Fix row border for ghost cells

### LINKED ISSUE #972: Ghost and loading cells styling doesn't divider between row headers and row contents
Seen on Chrome & Table 1.10
With row headers, but without any actual rows, ghost cells don't show the divider between the row headers and the row contents

image

Related context: https://pl.ntr/eI

tgreenwatts:
#### Fixes #972 

NOTE: Left the example with no rows temporarily, before merging I will put it back! 

Pretty much making this up, so please tell me if I'm doing it wrong

### DIFF
--- a/packages/table/preview/index.tsx
+++ b/packages/table/preview/index.tsx
@@ -250,7 +250,7 @@ ReactDOM.render(
 );
 
 ReactDOM.render(
-    getTableComponent(0, 0, {}, {
+    getTableComponent(2, 2, {}, {
         fillBodyWithGhostCells: true,
         selectionModes: SelectionModes.ALL,
     }),
--- a/packages/table/preview/index.tsx
+++ b/packages/table/preview/index.tsx
@@ -250,7 +250,7 @@ ReactDOM.render(
 );
 
 ReactDOM.render(
-    getTableComponent(2, 2, {}, {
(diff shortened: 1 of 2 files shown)


packages/table/preview/index.tsx
```diff
@@ -250,7 +250,7 @@ ReactDOM.render(
 );
 
 ReactDOM.render(
-    getTableComponent(2, 2, {}, {
+    getTableComponent(0, 0, {}, {
```
tgreenwatts: I will put this back before merging, left for testing

llorca:
Right idea! We gotta make sure we test all situations this time. I only see a couple of issues so far, one with the last row headers:
image

One with an empty ghost table (Safari only, but it wasn't broken before so it's a regression):
image

tgreenwatts:
Updated! I think it works this time, @cmslewis and I could not repro the Safari issue

tgreenwatts:
Reset the example so you can see with both a small table with ghost cells & an empty table with ghost cells

cmslewis:
A lot of our tables don't render on our preview page, because the maximum stack size is exceeded. But the few tables that did appear looked good.

tgreenwatts:
@llorca can I get a review here?

PR #1030, opened 2017-04-24
[Table] Delete @PureRenders on TruncatedFormat and JSONFormat

### LINKED ISSUE #1022: [Table] Truncated icons persist in cells even when text isn't truncated
To reproduce:
On the Formatted Content example, expand the date column to where it does not need the truncation ellipsis
Expected: Truncation ellipsis goes away
Actual: Ellipsis stays
Now scroll down and come back
Expected: Scrolling changes nothing
Actual: Now the ellipsis is gone!

ellipsis

This is a Regression

cmslewis:
#### Checklist

- [ ] Include tests

Delete @PureRender on `TruncatedFormat` and `JSONFormat` so that cell content updates properly when cells are resized.

This is a quick fix that may increase the number of unnecessary re-renders. Would be better to think of a not-weird way to inform cell content that a cell parent's width has updated.

Some weird ways that come to mind:
- Passing a callback that a parent fires for a child to respond to is backward and weird.
- Passing a prop value like `shouldUpdate`/`didCellSizeChange` is goofy and weird.
- `context`? Have never done much with that, but from what I've read, even this use case seems ill-fitted for that.
- Or, keeping the API more straightforward (but bloated): `cellWidth`/`cellHeight` props.

### DIFF
--- a/packages/table/src/cell/formats/jsonFormat.tsx
+++ b/packages/table/src/cell/formats/jsonFormat.tsx
@@ -6,7 +6,6 @@
  */
 
 import * as classNames from "classnames";
-import * as PureRender from "pure-render-decorator";
 import * as React from "react";
 import * as Classes from "../../common/classes";
 import { ITruncatedFormatProps, TruncatedFormat, TruncatedPopoverMode } from "./truncatedFormat";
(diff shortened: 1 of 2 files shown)


cmslewis:
@gscshoyru I'm definitely open to exploring how best to pass width/height awareness to children.

PR #1046, opened 2017-04-27
[DateRangeInput] Fix controlled-popover issues

cmslewis:
#### Checklist

- [x] Include tests
  - [x] Invokes `popoverProps.onInteraction` when start input focuses (maybe we should set `isOpen=true` and invoke `onInteraction` only if popover is not yet open?)
  - [x] Invokes `popoverProps.onInteraction` when end input focuses (same question as above)
  - [x] Invokes `popoverProps.onInteraction` when date range is fully selected
  - [x] Does not invoke `popoverProps.onInteraction` when date range is fully selected but `closeOnSelection={false}`
  - [x] Popover opens right away if `popoverProps.isOpen={true}` (should we also auto-focus in the start field, or just let that happen automatically as soon as the user hovers over a date?)
  - [x] Popover never closes if `popoverProps.isOpen={true}`
  - [x] Popover never opens if `popoverProps.isOpen={false}`
- [ ] Revert example changes once we've tested new behavior

- Explicitly invoke `popoverProps.onInteraction` whenever we update `isOpen` in `dRI.tsx`.
- Add an `isOpen()` helper that returns `popoverProps.isOpen` if it exists (instead of this.state.isOpen`).

- Should we bother updating `this.state.isOpen` at all if the `popoverProps.isOpen` is defined? Currently, I'm updating state but simply ignoring it if the popover is controlled.
- I'll revert my `DateRangeInputExample` changes before merging.

### DIFF
--- a/packages/datetime/examples/dateRangeInputExample.tsx
+++ b/packages/datetime/examples/dateRangeInputExample.tsx
@@ -19,6 +19,7 @@ export interface IDateRangeInputExampleState {
     disabled?: boolean;
     format?: string;
     selectAllOnFocus?: boolean;
+    popoverIsOpen?: boolean;
 }
 
 export class DateRangeInputExample extends BaseExample<IDateRangeInputExampleState> {
@@ -28,6 +29,7 @@ export class DateRangeInputExample extends BaseExample<IDateRangeInputExampleSta
         contiguousCalendarMonths: true,
         disabled: false,
         format: FORMATS[0],
+        popoverIsOpen: false,
         selectAllOnFocus: false,
     };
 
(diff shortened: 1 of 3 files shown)


cmslewis:
@giladgray I've been thinking more about this. Even if we expose `open`/`close` instance methods in a future PR (which seems like the easier approach for devs who want to control the popover momentarily), it seems like we'd still want to invoke `onInteraction` in all the correct places and respect `popoverIsOpen` if provided. Maybe we just don't advertise it as the recommended approach?

cmslewis:
Closing for now in favor of the uncontrolled/instance-method approach. If people are clamoring for controlled-popover usage, we can revisit.

Issue #505, opened 2017-01-17
Scrolling the table body when selecting columns


adidahiya:
Sub-issue of #298 (but I didn't want to add it to the already pretty long list...)

### Enhancement request

- __Package version(s)__: core 1.6, table 1.3
- __Browser and OS versions__: OS X 10.11.6, Chrome 57

#### Steps to reproduce

![2017-01-17 12 30 15](https://cloud.githubusercontent.com/assets/723999/22032099/08613c5c-dcb1-11e6-99d8-036af673427a.gif)

#### Expected behavior

When creating a drag selection across table columns and I want to keep selecting beyond the visible viewport, I want the table body to scroll along with my mouse cursor. At the moment, this works pretty inconsistently, as shown in the above GIF. Once I scroll downwards a little, horizontal scrolling gets activated too, but the interaction still isn't very smooth.

Relatedly, it'd be nice to have a table option (probably in dropdown in the top left) that lets me select all columns / cells at once.

tgreenwatts:
Related: https://github.com/palantir/blueprint/issues/950

cmslewis:
This works as of 1.22.0! :D

Issue #957, opened 2017-04-06
[Table] Can't scroll on header cells


tgreenwatts:
The table won't scroll if you're over the header cells - row or column. 
Unintuitive to users and feels broken

tgreenwatts:
Side effect of this: If there are no rows in a table, users can't see all their column headers

Example workflow:
User filters their table based on a value
No rows match
User wants to see what the exact name of that column was again, oops now they can't

PR #1066, opened 2017-05-08
[Table] Partially fix issues with scroll while drag-selecting

### LINKED ISSUE #950: Multi-row selection cut off during scrolling
When selecting more than what's visible at a time by scrolling during selection, the selection is cut off at the bottom to what's currently visible, see gif:
bluprint-table-bug

cmslewis:
### ❗  DO NOT MERGE (let's merge #1114 first while I investigate auto-scrolling more) 
---

#### Checklist

- [x] Include tests

- Fix issues with manual scrolling while a drag-selection is in progress: 
    - Mousing down on a **row** header and scrolling anchors the selection to that original row.
    - Mousing down on a **column** header and scrolling anchors the selection to that original column.
    - Mousing down on a **cell** and scrolling anchors the selection that original cell.
- Mousing down on a **cell** and dragging past the edge of the table will auto-scroll the table for you, expanding the selection in the process.

NOTE: I'm not addressing #505 here after all, because we observed some buggy behavior that I couldn't immediate figure out how to solve: clicking and column header and dragging to scroll the table would stop expanding the selection after ~10 columns or so and prevent you from adjusting the drag-selection further. Leaving #505 open to track that.

- Is it bad to not return -1 from the affected `Locator` functions anymore? I couldn't see a case when we'd really need that behavior anymore.

#### Screenshot

Scrolling while drag-selecting cells - works great:


llorca:
Summary from the discussion offline: we're reinventing click+drag scrolling, even though it's already supported natively in the browser. We have handlers on headers that prevent native scrolling from triggering—we should find the root cause and investigate leveraging native scroll as much as possible. I'm pretty sure it's related to this other issue about not being able to scroll when hovering headers: https://github.com/palantir/blueprint/issues/957

Aside from that, there are some nice improvements in this PR that can be merged separately such as selection with scroll at the same time correctly updating the selection region.

gulp/util/karma-config.js
```diff
@@ -40,8 +40,8 @@ module.exports = function createConfig(project) {
         coverageReporter: {
             check: {
                 each: {
-                    lines: 79,
-                    statements: 79,
+                    lines: 0, // TODO: Undo before merging.
```
gscshoyru: Don't forget :)

packages/table/src/headers/columnHeader.tsx
```diff
@@ -271,6 +275,7 @@ export class ColumnHeader extends React.Component<IColumnHeaderProps, IColumnHea
     }
 
     private handleDragSelectableSelectionEnd = () => {
+        this.activationCol = null;
```
gscshoyru: I realize this is probably good practice and we should keep it -- just asking from a code understanding perspective -- is this necessary? You're setting this on start of drag every time, I'm just making sure that if you _wanted_ to remove this line you could.

packages/table/src/locator.ts
```diff
@@ -24,24 +24,22 @@ export interface ILocator {
...
-     * the coordinate is not over a row.
+     * Locates a row's index given the client Y coordinate.
      * If `useMidpoint` is `true`, returns the index of the row whose top
      * edge is closest, splitting on the midpoint of each row.
      */
     convertPointToRow: (clientY: number, useMidpoint?: boolean) => number;
 
     /**
      * Locates a cell's row and column index given the client X
-     * coordinate. Returns -1 if the coordinate is not over a table cell.
```
gscshoyru: Why?

gscshoyru:
@llorca I'm not sure what you're saying re: clicking and dragging being native to browsers is true. It's only true insofar as you're highlighting things. I'm pretty sure we had to reinvent click and drag with slate as well when moving widgets; it won't work natively if you override drag behavior (as we must)

packages/table/src/headers/columnHeader.tsx
cmslewis: No, not strictly necessary. :)

packages/table/src/locator.ts
cmslewis: If you _don't_ return -1 from here—and instead return the next index beyond the index currently in view—then the viewport will move to out-of-view indices as desired. If you return -1, the viewport doesn't update, making the whole drag-to-scroll feature stop working.

llorca:
@gscshoyru check out this GIF, first scroll try is what this PR is about, the scrolling stops intermittently and doesn't accelerate the further I move the mouse. The second one (when I start from column BV), is a much better experience, it's the native interaction. Only problem is that it doesn't extend the selection as you can see:

llorca:
Just checked this out again. The GIF from my previous is still accurate, the current scrolling experience in the body is subpar -- feels like a bandaid: mouse needs to move at all times, no inertia 👎 

There seems to be a more pressing concern though, it looks like we broke something around column selection. Check out the new GIF attached: I can do all sorts of cool selection, but when I select a column, selection is frozen and I can't create a new region anymore.

cmslewis:
@llorca will look into it

cmslewis:
@llorca @gscshoyru I pulled out the direct fix to #950 into its own PR. Let's merge that first while I look more into the auto-scrolling behavior.

cmslewis:
This is addressed by `Shift + Click`. Will pick this particular feature up later.

PR #1071, opened 2017-05-09
[Table] Prototype: Shift+click or shift+click/drag to expand selection

### LINKED ISSUE #823: Shift selection mode for table regions
This would mitigate #505 because you could make one selection, scroll over to the end of your range, and then make another selection in the <kbd>Shift</kbd> mode of interaction.

Part of #298, but I'd like to track it separately.

cmslewis:
#### Checklist

- [x] Include tests

~Quick work-in-progress PR to get feedback on the interaction of Shift+Clicking to expand a selection.~

- Can now shift+click or shift+drag to expand a selection.
- Add unit test

Caveats:
- This PR does not yet leverage the focus cell; a polished version would need to consider that.
- The focus cell moves to the appended region when you expand the selection (erroneous, but not a big deal for now).

Ignoring the code and the lack of focus-cell consideration at the moment, how does the basic interaction feel?

~Should I extend the `DragSelectable` API to include knowledge of the focus cell for this PR? Or should I wait.~ **Decision:** we need to support shift+click when `enableFocus={false}` anyway, so might as well merge this PR as is.

### DIFF
--- a/gulp/util/karma-config.js
+++ b/gulp/util/karma-config.js
@@ -40,8 +40,8 @@ module.exports = function createConfig(project) {
         coverageReporter: {
             check: {
                 each: {
-                    lines: 80,
-                    statements: 80,
+                    lines: 0,
+                    statements: 0,
                 },
             },
             includeAllSources: true,
@@ -54,8 +54,8 @@ module.exports = function createConfig(project) {
                 { type: "text" },
             ],
             watermarks: {
-                lines: [80, 90],
-                statements: [80, 90],
+                lines: [0, 90],
+                statements: [0, 90],
             },
         },
         files: filesToInclude,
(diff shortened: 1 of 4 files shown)


tgreenwatts:
@giladgray for thoughts on doing this without the logic of where a focus cell is 
@themadcreator too
Basically we won't get all the extra things of "if you shift click over here, but you're focused here, select this region instead"

adidahiya:
I'm totally cool with shipping this as a first pass for shift-click selection, even when focus cells are available. I understand that it doesn't work quite like Excel, but the caveat noted here seems like kind of an edge case. If i'm trying to fix / adjust my selection when using focus cells, I'm more likely going to start the whole selection again rather than doing another shift-click, like so:

llorca:
Cool with that too, great improvement. We can do a big pass on all focus cell interactions later

cmslewis:
@giladgray @llorca @gscshoyru thoughts on this? Don't have any changes requested at present. Would love to merge to move on to the focus-cell case.

llorca:
@cmslewis I'm happy with this basic shift+click, shift+click/drag interaction

cmslewis:
🚢

PR #1100, opened 2017-05-15
[Table] Fixed the row resize for truncated non-word-wrapped

gscshoyru:
#### Changes proposed in this pull request

The truncated non-word-wrapped cells were taking up the
whole cell and so the size calculation wasn't working properly.

Now it should work properly.

This is pretty straightforward :)


cmslewis:
Can you add a screenshot/GIF even if it's a quick change?

gscshoyru:
Yeah, seems fine in contour when I make the css change, and it seems to be measuring everything correctly.

PR #1102, opened 2017-05-16
[Table] Use width of popover dots when calculating whether to show it

### LINKED ISSUE #953: [Table] Column resizing doesn't take truncation popover into account

gscshoyru:

Take the width into account when measuring if we should show or hide the popover handle

This is... pretty straightforward. I had a much more complicated way of doing this, but realized we might as well just do it the easy way first.

#### Screenshot
may-16-2017 11-23-45

### DIFF
--- a/packages/table/src/cell/formats/truncatedFormat.tsx
+++ b/packages/table/src/cell/formats/truncatedFormat.tsx
@@ -12,6 +12,10 @@ import * as React from "react";
 
 import * as Classes from "../../common/classes";
 
+// amount in pixels that the content div width changes when truncated vs when
+// not truncated. Note: could be modified by styles
+const CONTENT_DIV_WIDTH_DELTA = 25;
+
 export enum TruncatedPopoverMode {
     ALWAYS,
     NEVER,
@@ -78,7 +82,6 @@ export class TruncatedFormat extends React.Component<ITruncatedFormatProps, ITru
     public state: ITruncatedFormatState = { isTruncated: false };
 
     private contentDiv: HTMLDivElement;
-    private popoverComponent: Popover;
 
     public render() {
         const { children, detectTruncation, preformatted, truncateLength, truncationSuffix } = this.props;
@@ -110,7 +113,6 @@ export class TruncatedFormat extends React.Component<ITruncatedFormatProps, ITru
                 <div className={className}>
                     <div className={Classes.TABLE_TRUNCATED_VALUE} ref={this.handleContentDivRef}>{cellContent}</div>
                     <Popover
-                        ref={this.handlePopoverComponentRef}
                         className={Classes.TABLE_TRUNCATED_POPOVER_TARGET}
                         tetherOptions={{ constraints }}
                         content={popoverContent}
@@ -137,8 +139,6 @@ export class TruncatedFormat extends React.Component<ITruncatedFormatProps, ITru
 
     private handleContentDivRef = (ref: HTMLDivElement) => this.contentDiv = ref;
 
-    private handlePopoverComponentRef = (ref: Popover) => this.popoverComponent = ref;
-
     private shouldShowPopover(content: string) {
         const { detectTruncation, showPopover, truncateLength } = this.props;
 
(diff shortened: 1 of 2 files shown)


packages/table/src/cell/formats/truncatedFormat.tsx
```diff
@@ -5,6 +5,8 @@
  * and https://github.com/palantir/blueprint/blob/master/PATENTS
  */
 
+const POPOVER_HANDLE_WIDTH = 25;
```
cmslewis: I'd love to rename this `ASSUMED_ POPOVER_HANDLE_WIDTH` to make it abundantly clear that this is dependent on CSS styling. Would also be helpful to add a comment in the CSS mentioning that this value here needs to be kept in sync with any style changes (and maybe add a comment here too?)

gscshoyru: Rename is doable, comments are doable... I'm just not sure how useful comments will be given that they're not going to be surfaced in any way anyone will be able to see.

llorca:
Is there a way to compute the width of the target instead of hard coding it?

tgreenwatts:
Wait I tried it out with my movie table and it didn't work - why?
didntwork

gscshoyru:
...thanks for the catch, @tgreenwatts . 

The issue here is that for cells with wrapped text, the offending word has gone and wrapped to the next line, and so it's not making the div want to be wider (because it doesn't need to be wider because the word is on the next line). So this works fine for non-wrapped text, but not for wrapped. 

I don't know that there's a good way around this issue. Does anyone have a good idea?

cmslewis:
Uhhhh, hmm. Find the tallest cell in the column, put its contents into a temporary div that you create elsewhere in the DOM, expand the width of that DIV incrementally until the height reduces to `< 2 * line-height` (?), then use that width as your column width?

gscshoyru:
@tgreenwatts I think that may need to be a separate issue. This fix here is still a good one, and fixes some of the cases, but making it work in the case you're specifying is a much much harder problem.

tgreenwatts:
Cool so merge this one and file a follow up issue?

packages/table/src/cell/formats/truncatedFormat.tsx
```diff
@@ -157,8 +161,11 @@ export class TruncatedFormat extends React.Component<ITruncatedFormatProps, ITru
             return;
         }
 
+        // if the popover handle exists, take it into account
+        const popoverHandleAdjustmentFactor = this.state.isTruncated ? this.popoverComponent.state.targetWidth : 0;
```
giladgray: clever to reuse the already-computed state, but also somewhat sketchy to rely on state field from component instance...

giladgray:
measuring in temporary divs presents a consistency problem: how do you know that it measures the same in this div and in the table? what if there's custom CSS you can't predict? 🤢

gscshoyru:
"Measuring in temporary divs" what do you mean by that? This isn't measuring in temporary divs; it's measuring in the actual table.

giladgray:
see @cmslewis's comment above

gscshoyru:
Oh, responding to him, not the PR. My bad.

packages/table/src/cell/formats/truncatedFormat.tsx
gscshoyru: Yeah I wasn't sure what else to do here. The other option was to modify popover so I could pass a callback that would pass up the ref (what react more or less says to do in their docs) but I didn't want to modify the popover in a table PR if at all possible. I can do that instead though.

giladgray: @gscshoyru refactor to use `targetElement`?

gscshoyru:
@giladgray @cmslewis -- so from doing this re-write, I discovered that the width of the popover target has literally nothing to do with it -- it's just from the styling that sets the `right` of the div with the content. 

I'm going to go back to the fixed-number-and-document model we had to begin with, sorry for the confusion :(

packages/table/src/cell/formats/truncatedFormat.tsx
```diff
@@ -12,6 +12,10 @@ import * as React from "react";
 
 import * as Classes from "../../common/classes";
 
+// amount in pixels that the content div width changes when truncated vs when
+// not truncated. Note: could be modified by styles
+const CONTENT_DIV_WIDTH_DELTA = 25;
```
giladgray: ideally there'd be a prop with this as default value so users could customize if absolutely necessary, but i suppose that could come later?

llorca: Well ideally we don't need to hardcode such things, a refactor is needed in the future

giladgray: i'm not so sure about that... @gscshoyru explains above that this value is basically the size of the ellipsis at our current font size. i'm asking for a prop along the lines of `Popover` `arrowSize` so you can adjust the JS value to match your actual CSS.

PR #1124, opened 2017-05-18
[Forms] Add support for large file upload input

### LINKED ISSUE #1112: Allow large file input boxes
Basically to do this would mean adding something along the lines of
```css
.pt-file-upload-input {
  &.pt-large {
    &:after {
      height: $pt-input-height-large;
      line-height: $pt-button-height-large;
    }
  }
}
```

I don't have time to actually add it and PR right now, but if this is still here when I have free time, I'll attempt it.

llorca:
#### Fixes #1112 

image

image


packages/core/src/components/forms/_file-upload.scss
```diff
@@ -16,12 +16,16 @@ Markup:
 </label>
 
 :disabled - Disabled
+.pt-large - Larger size
 .pt-fill - Take up full width of parent element
 
 Styleguide pt-file-upload
 */
 
+$file-upload-input-selector: ".pt-file-upload-input";
```
giladgray: this is not a pattern we follow elsewhere... just repeat the selector instead of pulling it out to a string.

giladgray: you could do simply `.pt-large &` inside the original styles for `.pt-file-upload-input` to avoid repeating the selector. this is how buttons work--any large parent (not just `.pt-button-group`) will make nested buttons large.

llorca:
That's fixed:
image

PR #1139, opened 2017-05-23
[DateInput] Fixes confusing typo in prop deprecation notice

lucasray:

The deprecation notice says `popoverProps` is deprecated, this is not true. Especially confusing since we then tell you to use `popoverProps.position`.


adidahiya:
thanks @lucasray!

PR #1154, opened 2017-05-26
[Table] Create a playground where you can toggle all features in one place

cmslewis:

- Out with the old `perf.html` spreadsheet example.
- In with the new `Table` Playground. (The term "playground" is a little overloaded now; feel free to suggest something else, though note that string doesn't appear in this PR.)

- I've discovered so many `Table` bugs and/or weirdnesses from this already!
- I want to merge this quickly and add to it iteratively in future PRs (TODO: text truncation, `resizeByTallest...`, reorganize the table preview pages).
- Play with it at `[table docs url]/perf.html`.
- Happy to polish styles in future PRs if we need to.

#### Screenshot

screen shot 2017-05-26 at 1 06 01 pm

image

screen shot 2017-05-26 at 1 06 09 pm

### DIFF
--- a/packages/table/preview/perf.html
+++ b/packages/table/preview/perf.html
@@ -14,6 +14,8 @@
 
       margin: 0;
       padding: 0;
+
+      background-color: #F5F8FA;
     }
 
     body.pt-dark {
@@ -29,7 +31,6 @@
     }
 
     .container {
-      background: #F5F8FA;
       display: flex;
       overflow: hidden;
       height: 100%;
@@ -38,7 +39,7 @@
     .table {
       flex: 1;
       order: 2;
-      z-index: 1;
+      z-index: 2;
     }
 
     .table.is-inline {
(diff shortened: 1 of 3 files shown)


packages/table/preview/perf.html
```diff
@@ -6,25 +6,131 @@
   <link rel="stylesheet" type="text/css" href="dist/perf.css">
   <style>
     body {
-      padding: 20px 20px;
+      position: absolute;
```
cmslewis: Time to improve the build setup for this preview page methinks (future PR).

packages/table/preview/perf.html
```diff
@@ -6,25 +6,131 @@
...
+    .sidebar label:last-child {
+      margin-bottom: 0;
+    }
+
+    label.tbl-select-label {
+      margin-top: -3px;
+      margin-bottom: 7px;
+    }
+
+    .tbl-select-label .pt-select {
```
cmslewis: `tbl-` is my little custom namespace.

packages/table/preview/perf.tsx
```diff
@@ -183,9 +363,101 @@ class MutableTable extends React.Component<{}, IMutableTableState> {
...
+    }
 }
 
 ReactDOM.render(
     <MutableTable/>,
-    document.querySelector(".table"),
+    document.querySelector("#page-content"),
 );
+
+// TODO: Pull these from @blueprintjs/docs
```
cmslewis: Happy to do this in this PR. Wasn't able to wrangle it after a few seconds of poking, so I copied+pasted to unblock myself.

adidahiya:
- unchecking "multiple selections" disables drag-to-select-a-region. not sure if this is intentional
- unchecking "focus cell" disables the feature but keeps around the old focussed cell

overall looks good! can we throw in a focus style manager "only show outline on tabs" somewhere?

tgreenwatts:
Insert with random text takes a shockingly long time, but other than that looks great! Excited to use this going forward

tgreenwatts:
The focus cell doesn't move around :(

cmslewis:
@adidahiya:

1. According to the current implementation, "multiple" = more than one thing. "drag-selection" = more than one cell, hence `Table` disables it. i don't necessarily agree with this logic though; intuitively, i'd expect `allowMultipleSelections={false}` to simply disable cmd + select another region.
2. Those is an issue with `componentWillReceiveProps` in `Table`, not in the example code.

cmslewis:
@adidahiya I added a focus-style control.

image

cmslewis:
Merging because this only affects internal Table preview, and I have other stuff I want to do with it today.

PR #1162, opened 2017-05-30
[Table] Eliminate excessive selection re-renders

cmslewis:

- Fix bugs with boolean-checking and array deep-comparison in `Utils.deepCompareKeys`
- Add a custom `shouldComponentUpdate` to `Table` to deeply compare `state.selectedRegions`
    - Eliminates unnecessary re-renders when selection is disabled (before: every `mouseup` and `mousedown` with selection disabled would trigger a re-render of every cell in view, because `selectedRegions` was being continually overwritten with a new empty-array instance). 
    - Eliminates re-render on `mouseup` if one just happened on `mousedown`.
- Continues to shallowly compare every other key in `props` and `state`

### DIFF
--- a/packages/table/src/common/utils.ts
+++ b/packages/table/src/common/utils.ts
@@ -235,11 +235,7 @@ export const Utils = {
             return false;
         } else if (Array.isArray(objA) || Array.isArray(objB)) {
             return Utils.arraysEqual(objA, objB, Utils.deepCompareKeys);
-        } else if (typeof objA === "string" || typeof objB === "string" ) {
-            return objA === objB;
-        } else if (typeof objA === "number" || typeof objB === "number" ) {
-            return objA === objB;
-        } else if (typeof objA === "boolean" || typeof objB === "boolean" ) {
+        } else if (_isSimplePrimitiveType(objA) || _isSimplePrimitiveType(objB)) {
             return objA === objB;
         } else if (keys != null) {
             return _deepCompareKeys(objA, objB, keys);
@@ -415,4 +411,10 @@ function _deepCompareKeys(objA: any, objB: any, keys: string[]): boolean {
         return objA.hasOwnProperty(key) === objB.hasOwnProperty(key)
             && Utils.deepCompareKeys(objA[key], objB[key]);
     });
+}
+
+function _isSimplePrimitiveType(value: any) {
+    return typeof value === "number"
+        || typeof value === "string"
+        || typeof value === "boolean";
 }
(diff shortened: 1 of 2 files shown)


packages/table/src/table.tsx
```diff
@@ -338,6 +336,50 @@ export class Table extends AbstractComponent<ITableProps, ITableState> {
         selectionModes: SelectionModes.ALL,
     };
 
+    private static SHALLOWLY_COMPARABLE_PROP_KEYS = [
```
gscshoyru: Is there a better way to do this than this? Are there going to be issues in the future as we add more props and they're not in this list? I know there's no way to pull the list of items out of the type because the type isn't real... but this seems like a not-so-great solution. Wouldn't it be better to specify the ones we _don't_ want to do this with?

packages/table/src/common/utils.ts
```diff
@@ -234,11 +234,13 @@ export const Utils = {
...
             return false;
         } else if (Array.isArray(objA) || Array.isArray(objB)) {
-            return Utils.arraysEqual(objA, objB);
+            return Utils.arraysEqual(objA, objB, Utils.deepCompareKeys);
         } else if (typeof objA === "string" || typeof objB === "string" ) {
             return objA === objB;
         } else if (typeof objA === "number" || typeof objB === "number" ) {
             return objA === objB;
+        } else if (typeof objA === "boolean" || typeof objB === "boolean" ) {
+            return objA === objB;
```
giladgray: ok so we now have three identical statements here, if either object is a primitive type. refactor perhaps?

packages/table/src/table.tsx
giladgray: this makes me want to be able to provide a blacklist instead of a whitelist. blacklist would contain one thing: `["selectedRegions"]`.

perhaps we should change the `keys` arg of `shallowCompareKeys` to
`{ exclude: string[] } | { include: string[] }` to accept blacklist _or_ whitelist.

packages/table/src/common/utils.ts
cmslewis: haha, okay

packages/table/src/table.tsx
cmslewis: Will play.

cmslewis: @giladgray @gscshoyru I have @giladgray's suggestion working, but it requires changes all over the codebase. Can I push them up in a separate PR? Can open immediately after this one merges.

giladgray: @cmslewis sounds fine

adidahiya:
this looks buggy. it caused the big table example to break:

cmslewis:
@adidahiya I've observed that behavior for a few weeks now. Don't think it's related to this PR.

PR #1201, opened 2017-06-07
[Table] Batch Child Rendering

themadcreator:
Allows react components to add or remove a limited number of children at a time to reduce to DOM reconciliation time. 

This removes jank while scrolling, but introduces a "scanline" render style for table cells. This is preferable to slow, janky frame rates.

Includes unit test for `Batcher`.

Batch renders are invoked using a polyfill for `requestIdleCallback` that has these semantics:

> Invokes the provided callback on the next available frame after the stack frame is empty.
> 
> At most one callback per frame is invoked, and the callback may be delayed multiple frames until the page is idle.


packages/table/src/common/batcher.ts
```diff
@@ -0,0 +1,182 @@
...
+ * Licensed under the BSD-3 License as modified (the “License”); you may obtain a copy
+ * of the license at https://github.com/palantir/blueprint/blob/master/LICENSE
+ * and https://github.com/palantir/blueprint/blob/master/PATENTS
+ */
+
+import { requestIdleCallback } from "./requestIdleCallback";
+
+export interface IIndex<T> {
+    [key: string]: T;
+}
```
giladgray: `Dict<string, T>` 😱

packages/table/src/common/batcher.ts
```diff
@@ -0,0 +1,182 @@
...
+
+export type SimpleStringifyable = string | number | null;
```
giladgray: `undefined` too?

packages/table/src/common/batcher.ts
```diff
@@ -0,0 +1,182 @@
...
+/**
+ * This class helps batch updates to large lists.
+ *
+ * For example, if your react component has many children, updating them all at
+ * once may cause jank when reconciling the DOM. This class helps you update
+ * only a few per frame.
+ *
+ * A typical usage would be:
+ *
+ * ```typescript
```
giladgray: "```tsx", and remove initial blank line

packages/table/src/common/batcher.ts
```diff
@@ -0,0 +1,182 @@
...
+ *
+ * public renderChildren = (allChildrenKeys: string[]) => {
+ *
+ *     batcher.startNewBatch();
+ *
+ *     allChildrenKeys.forEach((prop1, index) => {
+ *         batcher.addArgsToBatch(prop1, "prop2", index);
```
giladgray: `prop1, "prop2"` 1, 2 intentional?

packages/table/src/common/batcher.ts
```diff
@@ -0,0 +1,182 @@
...
+        this.batchArgs = {};
+    }
+
+    /**
+     * Stores the variadic arguments to be later batched together.
+     *
+     * The arguments must be simple stringifyable objects.
+     */
+    public addArgsToBatch(...args: SimpleStringifyable[]) {
+        this.batchArgs[args.join("-")] = args;
```
giladgray: if you must `join-hash`, at least use a less common character, such as `|` or `😱 `

packages/table/src/common/batcher.ts
```diff
@@ -0,0 +1,182 @@
...
+    }
+
+    /**
+     * Compares the set of "batch" arguments to the "current" set. Creates any
+     * new objects using the callback as a factory. Removes old objects.
+     *
+     * Arguments are in the "current" set but were not part of the last "batch"
```
giladgray: Arguments _that_ are in

packages/table/src/common/batcher.ts
```diff
@@ -0,0 +1,182 @@
...
+     * set are considered candidates for removal. Similarly, Arguments that are
+     * part of the "batch" set but not the "current" set are candidates for
+     * addition.
+     *
+     * The number of objects added and removed maybe limited with the `..Limit`
```
giladgray: _may be_ limited

giladgray: ok comments in this file need some general editing, i'm not going to call out each typo here :book:

packages/table/src/headers/header.tsx
```diff
@@ -213,13 +215,22 @@ export interface IHeaderState {
     hasSelectionEnded?: boolean;
 }
 
+const RESET_CELL_KEYS_BLACKLIST = [
+    "endIndex",
+    "shallowlyComparablePropKeysList",
+    "startIndex",
+    "style",
+    "viewportRect",
+] as Array<keyof IInternalHeaderProps>;
```
giladgray: declare type up front instead of casting. `const X: Array<...> = [...]`

packages/table/src/headers/header.tsx
```diff
@@ -213,13 +215,22 @@ export interface IHeaderState {
...
+
 @PureRender
 export class Header extends React.Component<IInternalHeaderProps, IHeaderState> {
     public state: IHeaderState = {
         hasSelectionEnded: false,
     };
 
     protected activationIndex: number;
+    private batcher = new Batcher<React.ReactElement<any>>();
```
giladgray: use `JSX.Element` instead--it's equivalent and shorter

giladgray: use `JSX.Element` instead--it's equivalent and shorter

packages/table/src/headers/header.tsx
```diff
@@ -233,7 +244,13 @@ export class Header extends React.Component<IInternalHeaderProps, IHeaderState>
         }
     }
 
-    public componentWillReceiveProps(nextProps?: IHeaderProps) {
+    public componentWillReceiveProps(nextProps?: IInternalHeaderProps) {
+        const resetKeysBlacklist = { exclude: RESET_CELL_KEYS_BLACKLIST };
+        const resetBatcher = !Utils.shallowCompareKeys(this.props, nextProps, resetKeysBlacklist);
```
giladgray: `shouldResetBatcher`?

packages/table/src/headers/header.tsx
adidahiya: we should expand https://palantir.github.io/tslint/rules/no-object-literal-type-assertion/ to cover this kind of thing

themadcreator:
Performance comparison:

Prior to these changes:
https://2997-71939872-gh.circle-artifacts.com/0/home/ubuntu/blueprint/packages/table/preview/index.html
image

With these changes:
https://3003-71939872-gh.circle-artifacts.com/0/home/ubuntu/blueprint/packages/table/preview/index.html
image

packages/table/src/common/batcher.ts
```diff
@@ -0,0 +1,220 @@
...
+        const toRemove = this.setKeysOperation(this.currentObjects, this.batchArgs, "difference", removeOldLimit);
+        toRemove.forEach((key) => delete this.currentObjects[key]);
+
+        // remove ALL old object not in batch
+        const toRemoveOld = this.setKeysOperation(this.oldObjects, this.batchArgs, "difference", -1);
+        toRemoveOld.forEach((key) => delete this.oldObjects[key]);
+
+        // copy ALL old objects into current objects if not defined
+        let oldObjectsUsed = 0;
+        const toShallowCopy = this.setKeysOperation(this.batchArgs, this.oldObjects, "intersect", -1);
```
cmslewis: nit: Also, the arg names here are a little opaque at first glance. I'd probably rework the name to include a noun. Maybe `keysToShallowCopy`?

packages/table/src/common/batcher.ts
```diff
@@ -0,0 +1,220 @@
...
+    public removeOldAddNew(
+            callback: (...args: any[]) => T,
+            addNewLimit = Batcher.DEFAULT_ADD_LIMIT,
+            removeOldLimit = Batcher.DEFAULT_REMOVE_LIMIT,
+            updateLimit = Batcher.DEFAULT_UPDATE_LIMIT,
+    ) {
+
+        // remove old
+        const toRemove = this.setKeysOperation(this.currentObjects, this.batchArgs, "difference", removeOldLimit);
+        toRemove.forEach((key) => delete this.currentObjects[key]);
```
cmslewis: Is there any perf benefit to pulling this anonymous function out into a persistent `private` function?

packages/table/src/common/batcher.ts
```diff
@@ -0,0 +1,220 @@
...
+
+        // remove ALL old object not in batch
+        const toRemoveOld = this.setKeysOperation(this.oldObjects, this.batchArgs, "difference", -1);
+        toRemoveOld.forEach((key) => delete this.oldObjects[key]);
```
cmslewis: Same thought RE: persistent `private` function.

packages/table/src/common/batcher.ts
```diff
@@ -0,0 +1,220 @@
...
+
+        // copy ALL old objects into current objects if not defined
+        let oldObjectsUsed = 0;
+        const toShallowCopy = this.setKeysOperation(this.batchArgs, this.oldObjects, "intersect", -1);
+        toShallowCopy.forEach((key) => {
+            if (this.currentObjects[key] == null) {
```
cmslewis: Same thought RE: persistent `private` function. Though in this case you'd have to either do a `map` with a side effect (to get the `oldObjectsUsed` at the end) or have two functions and do two separate passes through the `toShallowCopy` keys (one pass to count the `oldObjectsUsed`, another to make the changes).

packages/table/src/common/batcher.ts
```diff
@@ -0,0 +1,220 @@
...
+     */
+    public isDone() {
+        return this.done;
+    }
+
+    /**
+     * Returns all the objects in the "current" set.
+     */
+    public getList(): T[] {
+        return Object.keys(this.currentObjects).map((key) => this.currentObjects[key]);
```
cmslewis: Same thought RE: persistent `private` function.

packages/table/src/common/requestIdleCallback.ts
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    // finally, invoke the callback. exceptions will be propagated
+    if (callback) {
+        callback();
+    }
+};
+
+// check for window since we might be in a headless server environment
+if (typeof window !== "undefined") {
```
cmslewis: 👏

packages/table/src/common/requestIdleCallback.ts
```diff
@@ -0,0 +1,91 @@
...
+    if (IDLE_STATE.callbacks.length > 0) {
+        callback = IDLE_STATE.callbacks.shift();
+    }
+
+    if (IDLE_STATE.callbacks.length > 0) {
+        triggerIdleFrame();
+    }
+
+    // finally, invoke the callback. exceptions will be propagated
+    if (callback) {
```
cmslewis: Can we use `Utils.safeInvoke` from `@blueprintjs/core`? May not be worth the cross-package dependency though.

packages/table/src/common/requestIdleCallback.ts
```diff
@@ -0,0 +1,91 @@
...
+     *
+     * Then, we call to `postMessage` to invoke the `handleIdle` method only
+     * once the current stack frame is empty.
+     *
+     * With this approach, the idle callback will be invoked at most once per
+     * frame and only after the stack frame is empty.
+     */
+    requestAnimationFrame(() => {
+        requestAnimationFrame(() => {
+            postMessage(MESSAGE_EVENT_DATA, "*");
```
cmslewis: Putting the BILL in BrILLiant. 👏 👏 👏

packages/table/src/common/requestIdleCallback.ts
```diff
@@ -0,0 +1,91 @@
...
+};
+
+/**
+ * Invokes the provided callback on the next available frame after the stack
+ * frame is empty.
+ *
+ * At most one callback per frame is invoked, and the callback may be delayed
+ * multiple frames until the page is idle.
+ *
+ * TODO: return a token from this method that allows you to cancel the callback
```
cmslewis: Can we track this TODO in a Github issue if it's not going to get done immediately?

packages/table/src/headers/header.tsx
```diff
@@ -234,14 +240,30 @@ export class Header extends React.Component<IInternalHeaderProps, IHeaderState>
...
             this.setState({ hasSelectionEnded: false });
         }
     }
 
+    public componentWillUpdate(nextProps?: IInternalHeaderProps, nextState?: IHeaderState) {
+        const resetKeysBlacklist = { exclude: RESET_CELL_KEYS_BLACKLIST };
+        let shouldResetBatcher = !Utils.shallowCompareKeys(this.props, nextProps, resetKeysBlacklist);
+        // const stateDiff = Utils.shallowCompareKeys(this.state, nextState);
+        // //tslint:disable-next-line
+        // console.log(stateDiff, shouldResetBatcher, this.state, nextState);
```
cmslewis: remove commented code

packages/table/src/tableBody.tsx
```diff
@@ -116,15 +141,24 @@ export class TableBody extends React.Component<ITableBodyProps, {}> {
...
+                this.batcher.addArgsToBatch(rowIndex, columnIndex);
             }
         }
+        this.batcher.removeOldAddNew((row: number, col: number) => {
+            const extremaClasses = grid.getExtremaClasses(row, col, rowIndexEnd, columnIndexEnd);
+            const isGhost = grid.isGhostIndex(row, col);
+            return this.renderCell(row, col, extremaClasses, isGhost);
+        });
+        if (!this.batcher.isDone()) {
+            this.batcher.idleCallback(() => this.forceUpdate());
```
cmslewis: To avoid instantiating a new anonymous function on each `render`, can we just do `this.batcher.idleCallback(this.forceUpdate)`?

packages/table/src/tableBody.tsx
```diff
@@ -146,6 +180,7 @@ export class TableBody extends React.Component<ITableBodyProps, {}> {
                 </ContextMenuTargetWrapper>
             </DragSelectable>
         );
+
```
cmslewis: nit: don't need this newline

packages/table/src/common/batcher.ts
themadcreator: 👍

packages/table/src/common/requestIdleCallback.ts
themadcreator: 👍

packages/table/src/headers/header.tsx
themadcreator: oh whoops

adidahiya:
strangely, the big table stops rendering at 838,861 rows. this seems to be a problem on `master` and #1209 too though, so outside the scope of this PR.

image

themadcreator:
fascinating! I believe this is because the styles are somehow being set pixel values in scientific notation:

image

themadcreator:
Well:
https://stackoverflow.com/questions/38905916/content-in-google-chrome-larger-than-16777216-px-not-being-rendered
https://bugs.chromium.org/p/chromium/issues/detail?id=401762

themadcreator:
Furthermore:
px-limit

packages/table/src/common/batcher.ts
```diff
@@ -0,0 +1,220 @@
...
+            removeOldLimit = Batcher.DEFAULT_REMOVE_LIMIT,
+            updateLimit = Batcher.DEFAULT_UPDATE_LIMIT,
+    ) {
+
+        // remove old
+        const toRemove = this.setKeysOperation(this.currentObjects, this.batchArgs, "difference", removeOldLimit);
+        toRemove.forEach((key) => delete this.currentObjects[key]);
+
+        // remove ALL old object not in batch
+        const toRemoveOld = this.setKeysOperation(this.oldObjects, this.batchArgs, "difference", -1);
```
themadcreator: i disagree. I think it's more natural (or at least more common) to use a number than a `null` value

packages/table/src/common/batcher.ts
themadcreator: 👍

packages/table/src/common/batcher.ts
themadcreator: oh man, i'm not sure this is better. I would rather convert these to inline for loops than extract private functions--the method would be much harder to understand if these operations were distributed in a bunch of methods. thoughts?

packages/table/src/common/batcher.ts
themadcreator: 👍

packages/table/src/common/requestIdleCallback.ts
themadcreator: https://github.com/palantir/blueprint/issues/1223

packages/table/src/tableBody.tsx
themadcreator: the `forceUpdate` method not bound to `this` and `this.forceUpdate.bind(this)` is equivalent;

packages/table/src/common/batcher.ts
```diff
@@ -0,0 +1,220 @@
...
+        this.batchArgs = {};
+    }
+
+    /**
+     * Stores the variadic arguments to be later batched together.
+     *
+     * The arguments must be simple stringifyable objects.
+     */
+    public addArgsToBatch(...args: SimpleStringifyable[]) {
+        this.batchArgs[args.join("|")] = args;
```
themadcreator: 👍

packages/table/src/common/batcher.ts
cmslewis: Is there a noticeable perf improvement with the for-loop or private-function approach? If not, can leave as is.

PR #1206, opened 2017-06-08
[Table] Add custom shouldComponentUpdates in Header and HeaderCell

cmslewis:

- Add more specific `shouldComponentUpdate`s in `*Header`s and `*HeaderCell`s.
- Delete `HeaderCell`'s wonky internal `shallowlyComparablePropKeys` prop

This is part of constellation of upcoming perf-related PRs; it doesn't have much effect on its own, but it will in concert with other changes (which are all implemented on `cl/table-perf` if you want to peek at that).

### DIFF
--- a/packages/table/src/cell/editableCell.tsx
+++ b/packages/table/src/cell/editableCell.tsx
@@ -5,13 +5,13 @@
  * and https://github.com/palantir/blueprint/blob/master/PATENTS
  */
 
-import * as PureRender from "pure-render-decorator";
 import * as React from "react";
 import * as ReactDOM from "react-dom";
 
-import { EditableText, Utils } from "@blueprintjs/core";
+import { EditableText, Utils as CoreUtils } from "@blueprintjs/core";
 
 import * as Classes from "../common/classes";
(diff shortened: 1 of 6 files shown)


Lojka59:
i got one problem with this issue.

1. my editable cell contains value = "";
2. i try to input letter "s", it is stored in "dirtyValue"
3. onChange event happens -> 
4. my outside code checks for permitted inputs (it got "s" from onChange)
5. as letters are not permitted, it sets value back to "";
6. my editable cell get newProps (newValue = "")
7. editable cell check previous value, wich is "" and compare to new value, that is changed to "" already.
8. they are equal, and editable cell dosn't change dirty or saved value (wich is "s");

so in this situation i set value in controlled mode, but it failed to set, as it was same, as previous value. (savedValue from cell.state show me "s", but value is "")

cmslewis:
@Lojka59 `EditableCell` controlled mode has been wonky for awhile, but your issue might be fixed in our next release by this PR: https://github.com/palantir/blueprint/pull/1793

Issue #1214, opened 2017-06-09
[Table] Investigate how to put shouldComponentUpdate into Header directly 


cmslewis:
### Bug report

- __Package version(s)__: v1.19.0

#### Description

Currently, we have the same custom `shouldComponentUpdate` code in both `ColumnHeader` and `RowHeader` (as of #1215). The preferred approach is to put that code in `Header#shouldComponentUpdate`, but that causes tests to fail. Need to figure out why.

cmslewis:
Done in #1524

PR #1215, opened 2017-06-09
[Table] Fix shouldComponentUpdate issues in *Header's

cmslewis:

In #1206, I added a custom `shouldComponentUpdate` to the `Header` component, but I accidentally left a `@PureRender` decorator on it as well, which leads to a runtime warning that we didn't catch before merging that PR. Unfortunately, while `pure-render-decorator` warns you that it can't add its default `shouldComponentUpdate` because a `shouldComponentUpdate` is already defined, _it overwrites yours anyway_. Removing the decorator to get rid of this warning caused tests to fail.

This PR fixes those tests by moving the `Header`'s custom `shouldComponentUpdate` code a level up into `ColumnHeader` and `RowHeader`. I'm still not clear why that fixes things, but I opened an issue (#1214) to track.

Exporting a custom function like this is awkward, but is it reasonable for now? I want to unblock @themadcreator so that we can merge his batching PR (#1201).

### DIFF
--- a/packages/table/src/headers/header.tsx
+++ b/packages/table/src/headers/header.tsx
@@ -383,6 +383,9 @@ export class Header extends React.Component<IInternalHeaderProps, IHeaderState> 
  * temporary solution, we at least want to avoid writing the same code in two places, so we expose
  * this utility for each higher-level component to leverage in their own respective
  * shouldComponentUpdate functions.
+ *
+ * (See: https://github.com/palantir/blueprint/issues/1214)
+ *
  * @param props - the current props
  * @param nextProps - the next props
  */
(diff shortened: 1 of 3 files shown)


packages/table/src/headers/header.tsx
```diff
@@ -390,8 +374,33 @@ export class Header extends React.Component<IInternalHeaderProps, IHeaderState>
             // both selection and reordering behavior.
             && selectedRegions.length === 1;
     }
+}
 
-    private isSelectedRegionRelevant = (selectedRegion: IRegion) => {
-        return Regions.getRegionCardinality(selectedRegion) === this.props.fullRegionCardinality;
+/**
+ * Due to reasons we're still investigating, our tests fail if we put a common, custom
```
themadcreator: this comment will probably be out of date as soon as monday. We're probably not going to continue investigating.

I suspect the reason those particular tests were failing is because the interactions to do re-ordering were adjusting state or props and causing some updates. The subsequent classname selector lookups wasn't able to find the right element. Or perhaps, a chain of new objects were created and the state was out of sync.

Either way, we should probably don't want to imply that we intend to change this in the future.

PR #1210, opened 2017-06-08
[Table] Introduce Cell/HeaderCell index props and pass them to callbacks

cmslewis:
#### Checklist

- [x] Include tests

#### Changes proposed in this pull request:

- Add optional new `index` prop to `HeaderCell` (can be passed to callbacks to avoid having to re-instantiate anonymous functions in parent components on each re-render).
- Add optional new `rowIndex` and `columnIndex` props to `Cell` (same reasoning).
- Pass `index` callback to `ColumnHeaderCell`'s `renderName` callback.
- Pass `rowIndex` and `columnIndex` to `EditableCell` callbacks.
- Refactor `preview/index.tsx` to take advantage of these new `*index` props

#### Reviewers should focus on

Perf gains! This approach lets you provide a persistent function instance as a callback to the various `*Cell` components, causing shallow equality checks on that function instance to pass now. This avoids lots of unnecessary re-renders.

PR #1218, opened 2017-06-10
[Table] Introduce HeaderCell#renderMenu prop, deprecate HeaderCell#menu prop

cmslewis:
#### Checklist

- [x] Include tests

The `HeaderCell#menu` prop we have now is bad, because `<MenuItem>`s typically require callbacks that are instantiated as inline anonymous functions so that they can have access to a row/column index.

This PR deprecates `HeaderCell#menu` and introduces `HeaderCell#renderMenu` instead, a callback that receives a row/column index parameter and expects a JSX element in return. This works in concert with the `HeaderCell#index` props introduced in #1210.

The major benefit here is that `shouldComponentUpdate` checks pass, because `renderMenu` can be a single, persistent function instance; whereas `menu` was typically reinstantiated on each re-render.

Does this API change sound good?

### DIFF
--- a/packages/table/src/headers/columnHeaderCell.tsx
+++ b/packages/table/src/headers/columnHeaderCell.tsx
@@ -8,9 +8,17 @@
 import * as classNames from "classnames";
 import * as React from "react";
 
-import { Classes as CoreClasses, IProps, Popover, Position, Utils as CoreUtils } from "@blueprintjs/core";
+import {
+    AbstractComponent,
+    Classes as CoreClasses,
+    IProps,
+    Popover,
+    Position,
+    Utils as CoreUtils,
+} from "@blueprintjs/core";
 
 import * as Classes from "../common/classes";
+import * as Errors from "../common/errors";
 import { LoadableContent } from "../common/loadableContent";
 import { HeaderCell, IHeaderCellProps } from "./headerCell";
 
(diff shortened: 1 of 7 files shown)


packages/table/src/headers/columnHeaderCell.tsx
```diff
@@ -179,12 +196,17 @@ export class ColumnHeaderCell extends React.Component<IColumnHeaderCellProps, {}
             to: "window",
         }];
 
+        // prefer renderMenu if it's defined
+        const content = CoreUtils.isFunction(renderMenu)
+            ? renderMenu(index)
+            : menu;
```
themadcreator: wait, so we're actually supporting both? Just deprecating and console.warn-ing?

packages/table/src/headers/headerCell.tsx
```diff
@@ -8,7 +8,7 @@
 import * as classNames from "classnames";
 import * as React from "react";
 
-import { Classes as CoreClasses, ContextMenuTarget, IProps } from "@blueprintjs/core";
+import { Classes as CoreClasses, ContextMenuTarget, IProps, Utils as CoreUtils } from "@blueprintjs/core";
```
themadcreator: might be multiline time for this import

packages/table/src/headers/columnHeaderCell.tsx
cmslewis: That's the plan, yes. Does that sound okay?

packages/table/src/headers/headerCell.tsx
cmslewis: I normally just do that when the linter tells me to.

PR #1222, opened 2017-06-12
[Table] Add new utils: getShallowUnequalKeyValues and getDeepUnequalKeyValues

cmslewis:
#### Checklist

- [x] Include tests

Proposing to add `getShallowUnequalKeyValues` and `getDeepUnequalKeyValues` to our `Table` utils. These functions are useful primarily for debugging `shouldComponentUpdate`; I kept reimplementing the same helpers to do just this, so I figured I'd open a PR to add them to the codebase permanently.

- Is it worth having these around permanently? I think so.
- If you agree, then I'll add tests. Not before. :)

### DIFF
--- a/packages/table/src/common/utils.ts
+++ b/packages/table/src/common/utils.ts
@@ -259,6 +259,32 @@ export const Utils = {
     },
 
     /**
+     * Returns a descriptive object for each key whose values are shallowly unequal between two
+     * provided objects. Useful for debugging shouldComponentUpdate.
+     */
+    getShallowUnequalKeyValues<T extends object>(objA: T, objB: T, keys?: IKeyBlacklist<T> | IKeyWhitelist<T>) {
+        const filteredKeys = _filterKeys(objA, objB, keys == null ? { exclude: [] } : keys);
+        return _getUnequalKeyValues(
+            objA,
+            objB,
+            filteredKeys,
+            (a, b, key) => Utils.shallowCompareKeys(a, b, { include: [key] }));
+    },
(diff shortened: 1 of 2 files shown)


packages/table/src/common/utils.ts
```diff
@@ -259,6 +259,32 @@ export const Utils = {
...
+            objB,
+            filteredKeys,
+            (a, b, key) => Utils.shallowCompareKeys(a, b, { include: [key] }));
+    },
+
+    /**
+     * Returns a descriptive object for each key whose values are deeply unequal between two
+     * provided objects. Useful for debugging shouldComponentUpdate.
+     */
+    getDeepUnequalKeyValues<T extends object>(objA: T, objB: T, keys?: Array<keyof T>) {
```
themadcreator: that is some fancy typing, sir

packages/table/src/common/utils.ts
```diff
@@ -259,6 +259,32 @@ export const Utils = {
     },
 
     /**
+     * Returns a descriptive object for each key whose values are shallowly unequal between two
+     * provided objects. Useful for debugging shouldComponentUpdate.
```
themadcreator: I'm fine with adding some handy utility methods for debugging to the codebase. It does add some bytes to the download, but so long as we don't go crazy it wont be a big deal.

Do you think we should unit test these things? (that does coverage say?)

cmslewis: Wrote some tests.

PR #1225, opened 2017-06-12
[Table] Final cleanups from perf work

cmslewis:
Some final, tiny cleanups from last week's `Table` perf work:
- Move from anonymous functions to private functions in `ColumnHeaderCell`.
- Use fat arrow instead of `.bind(this)` in `index.tsx`
- Provide indices to cells when editing is disabled on Table preview.


packages/table/preview/index.tsx
```diff
@@ -181,17 +181,17 @@ class MutableTable extends React.Component<{}, IMutableTableState> {
     // Renderers
     // =========
 
-    public renderColumns() {
+    private renderColumns() {
```
cmslewis: Not sure why this was `public`; it's not used anywhere.

PR #1250, opened 2017-06-19
[Table] Add column reorder handle when interaction bar enabled

### LINKED ISSUE #1242: [Table] Add reorder handle to column header cells when reordering enabled
**Problem**
When tasked with reordering a single column in the `Table`, users don't realize that you have to click-select the column first, _then_ reorder it via click-dragging.

**Proposal**
Add drag-reorder handles to `ColumnHeaderCell`s that, when click-dragged, always triggering reordering regardless of the column's selected state.

cmslewis:
#### Checklist

- [x] Include tests
- [x] Update documentation

Add reorder handles to column headers when `useInteractionBar={true}`.

- When reorder handles are enabled, only a `mousedown` _within_ the handle will trigger reordering.
- When reorder handles are _not_ enabled, a `mousedown` in the entire cell will trigger reordering, as before.
- Clicking any reorder handle within a selection is equivalent to clicking the left-most reorder handle in the selection (the whole block moves together).
- Clicking a reorder handle outside of the current selection will move the selection to that column on `mousedown`.

Also, update docs to include a new "Reorderable content" section.

How does the general architecture look? Feels slightly messy, partly because I opted not to create a separate `ReorderHandle` component, since this logic is used in only one place (`ResizeHandle`, on the other hand, is used for both row and column headers.) May be passable though.

#### Screenshot

reorder-handle

image


packages/core/src/common/utils.ts
```diff
@@ -94,3 +94,10 @@ export function throttleEvent(target: EventTarget, eventName: string, newEventNa
     target.addEventListener(eventName, func);
     return func;
 };
+
+/**
+ * Returns true if at least one item in the array satifies the predicate.
+ */
+export function some(a: any[] = [], predicate: (item: any) => boolean) {
```
gscshoyru: Isn't this a thing that already exists: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some?v=control ?

packages/table/src/interactions/reorderable.tsx
```diff
@@ -112,22 +112,27 @@ export class DragReorderable extends React.Component<IDragReorderable, {}> {
...
+            const selectedRegion = selectedRegions[selectedRegionIndex];
+            if (Regions.getRegionCardinality(selectedRegion) !== cardinality) {
+                // ignore FULL_TABLE selections
+                return false;
+            }
+
+            // cache for easy access later in the lifecycle
+            const selectedInterval = isRowHeader ? selectedRegion.rows : selectedRegion.cols;
+            this.selectedRegionStartIndex = selectedInterval[0];
+            // add 1 to correct for the fencepost
```
gscshoyru: Nitpick -- can you be a little bit more descriptive here? A "because selected interval is inclusive" or "includes its endpoints" or something would be ok; just looking for a why-the-fencepost.

packages/core/src/common/utils.ts
cmslewis: -.-

cmslewis: -.-

packages/table/src/interactions/reorderable.tsx
cmslewis: Yep

packages/table/src/headers/headerCell.tsx
```diff
@@ -73,6 +77,7 @@ export interface IHeaderCellProps extends IProps {
 export interface IInternalHeaderCellProps extends IHeaderCellProps {
     /**
      * Specifies if the cell is reorderable.
+     * @deprecated since 1.21.0; pass `isReorderable` to `ColumnHeader` or `RowHeader` instead
```
cmslewis: Geoff what do you think about this? I moved the logic for adding the `TABLE_HEADER_REORDERABLE` class from here into `Header`, so this prop serves no purpose here anymore. See: https://github.com/palantir/blueprint/pull/1250/files#diff-34e8a9161a519e81e1d774377ef4c28cR323

gscshoyru: Seems fine

PR #1256, opened 2017-06-20
[Labs] Propagate event correctly when handling empty list

k-simons:
#### Checklist

- [x] Include tests
- [ ] Update documentation

Right now if you have a class that instantiates a raw queryList and passes in an empty filtered list and undefined as the active state, then on `componentDidUpdate` it will call update and pass an active item of undefined. If you have a parent class that listens to `onActiveItemChange` and then sets its state, the parent class will be re-rendered. Then the queryList will will be re-rendered, and so on and so forth. This will cause an infinite re-rendering loop and a stackover-flow error. You can emulate this behavior if you remove @pure-render from select.tsx and remove:

```
        if (filteredItems.length === 0) {
            return noResults;
        }
```

so the query list is always rendered.

Then if you filter to an empty list in select.tsx, you will get a stack-overflow error

#### Screenshot

@giladgray This fixes an edge case of propagating an incorrect change event

### DIFF
--- a/packages/labs/src/query-list/queryList.tsx
+++ b/packages/labs/src/query-list/queryList.tsx
@@ -186,7 +186,9 @@ export class QueryList<T> extends React.Component<IQueryListProps<T>, IQueryList
             this.shouldCheckActiveItemInViewport = false;
         }
         // reset active item (in the same step) if it's no longer valid
-        if (this.getActiveIndex() < 0) {
+        // Also don't fire the event if the active item is already undefined and there is nothing to pick
+        if (this.getActiveIndex() < 0 &&
+            (this.state.filteredItems.length !== 0 || this.props.activeItem !== undefined)) {
             Utils.safeInvoke(this.props.onActiveItemChange, this.state.filteredItems[0]);
         }
     }
--- a/packages/labs/test/queryListTests.tsx
+++ b/packages/labs/test/queryListTests.tsx
@@ -6,7 +6,7 @@
  */
 
 import { assert } from "chai";
-import { shallow } from "enzyme";
+import { mount, shallow } from "enzyme";
 import * as React from "react";
 
 import { Film, TOP_100_FILMS } from "../examples/data";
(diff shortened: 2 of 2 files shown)


tgreenwatts:
Gilad is out of town but @gscshoyru will take a look

gscshoyru:
@k-simons what exactly is the edge case here? What was the bad behavior, before?

packages/labs/src/query-list/queryList.tsx
```diff
@@ -217,6 +217,12 @@ export class QueryList<T> extends React.Component<IQueryListProps<T>, IQueryList
         }
     }
 
+    private handlingEmptyListCorrectly() {
+        const emptyFilteredList = this.state.filteredItems.length === 0;
+        const activeItemIsUndefined = this.props.activeItem === undefined;
+        return emptyFilteredList && activeItemIsUndefined;
```
gscshoyru: I think this honestly might make more sense to inline, above, since it's just in one place, and the function name is not particularly helpful. Plus the logic is harder to follow like this, I think.

llorca:
@k-simons Can you please fill up the description and provide details about the edge case you hit, before/after behaviors, etc? Thanks!

packages/labs/src/query-list/queryList.tsx
k-simons: @gscshoyru inline into `componentDidUpdate` ?

k-simons:
@gscshoyru @llorca apologies, gilad and I had talked about this a little Friday. Updated with context

packages/labs/src/query-list/queryList.tsx
gscshoyru: Yeah, in the if statement condition. The fact that I have to demorgan's law on this in my head while going back and forth between two lines makes this somewhat difficult to parse.

k-simons:
@gscshoyru fixed

packages/labs/src/query-list/queryList.tsx
```diff
@@ -186,7 +186,7 @@ export class QueryList<T> extends React.Component<IQueryListProps<T>, IQueryList
             this.shouldCheckActiveItemInViewport = false;
         }
         // reset active item (in the same step) if it's no longer valid
-        if (this.getActiveIndex() < 0) {
+        if (this.getActiveIndex() < 0 && (this.state.filteredItems.length !== 0 || this.props.activeItem !== undefined)) {
```
gscshoyru: Much easier to understand. 

It might be prudent to add a bit more to the above comment explaining what this is doing (essentially, don't fire the event if there's nothing to pick and the active item is already undefined) but it's probably fine either way.

k-simons:
@gscshoyru updated comment. Should be good to merge

k-simons:
fixed lint 🙃

packages/labs/src/query-list/queryList.tsx
```diff
@@ -186,7 +186,9 @@ export class QueryList<T> extends React.Component<IQueryListProps<T>, IQueryList
             this.shouldCheckActiveItemInViewport = false;
         }
         // reset active item (in the same step) if it's no longer valid
-        if (this.getActiveIndex() < 0) {
+        // Also don't fire the event if the active item is already undefined and there is nothing to pick
+        if (this.getActiveIndex() < 0 &&
+            (this.state.filteredItems.length !== 0 || this.props.activeItem !== undefined)) {
```
cmslewis: Should we expand this check to account for `null` values too? (i.e. `this.props.activeItem != null`)

k-simons: @cmslewis sure will work on a unit test. I don't think we should add null. Mostly because if this.state.filteredItems is empty, then this.state.filteredItems[0] is undefined. In that case there actually is a difference to props.selectedItem and the event (null vs undefined)

k-simons:
@cmslewis tests added

cmslewis:
Thanks, @k-simons!

Issue #1262, opened 2017-06-21
[Table] Resolve differences between <EditableCell> and <EditableName>


cmslewis:
### Bug report

- __Package version(s)__: v1.20.0

### Changes proposed

There are some differences between `EditableCell`  and `EditableName` that we should resolve:
- Should emit the `columnIndex` as a second parameter in `onCancel`, `onChange`, and `onConfirm` callbacks
- Should fully control the value so that updating `props.name` actually renders the new name. (See: #1257, #1234)

gscshoyru:
Note: so far as I can tell, doing the first part here is going to require cloning the results of renderName the same way we do for table cells, so it knows what index it is. Is this alright? I'm somewhat loathe to do more cloning, 'cause I know it tends to impact perf.

Issue #1261, opened 2017-06-21
[Table] Editing body cell too quickly after focus is buggy  


cmslewis:
(Note: This is happening on a branch, so maybe it's due to something I did locally. Capturing anyway.)

### Bug report

- __Package version(s)__: v1.20.0
- __Browser and OS versions__: Chrome 59 on OS X Sierra

#### Steps to reproduce

1. Double click an editable body cell in the table.
1. Immediately start typing. 

#### Actual behavior

Your typing is partially or fully replaced by the original cell value. However, if you wait a moment before typing, everything is fine.

![edit-bug](https://user-images.githubusercontent.com/443450/27403068-e9e99264-567d-11e7-8b5b-0e33f534b1f2.gif)

#### Expected behavior

Your typing is honored no matter how soon you do it after focus.

Issue #1259, opened 2017-06-21
[Table] Scrolling quickly from high-numbered to low-numbered rows ruins alignments


cmslewis:
### Bug report

- __Package version(s)__: v1.20.0
- __Browser and OS versions__: Chrome 59 on Mac OS X Sierra

#### Steps to reproduce

1. Scroll quickly from a high-numbered, multi-digit row to a low-numbered, fewer-digit row (by clicking and dragging the scroll bar pill, e.g.).

#### Actual behavior

Column header alignment gets out of sync until you scroll again:
![scroll-bug](https://user-images.githubusercontent.com/443450/27402039-25d938e6-567a-11e7-8c21-b403a1d88dad.gif)

#### Expected behavior

Column headers remain aligned with cells.

tgreenwatts:
Is this still true @cmslewis ?

cmslewis:
@tgreenwatts yep

Issue #1258, opened 2017-06-21
[Table] Focus cell does not move after reordering


cmslewis:
### Bug report

- __Package version(s)__: v1.20.0
- __Browser and OS versions__: Chrome 59 on Mac OS X Sierra

#### Steps to reproduce

1. Enable focus cell and row reordering.
2. Click a row to select it.
3. Reorder it elsewhere.

#### Actual behavior

Focus cell does not move with the selection after reordering is complete.

![reorder-focus-bug](https://user-images.githubusercontent.com/443450/27401462-2b433586-5678-11e7-9773-5959308694b1.gif)
 

#### Expected behavior

Focus cell should move with the newly reordered selection.

PR #1263, opened 2017-06-21
[Table] Make reordering work in the table example

cmslewis:

Column and row reordering now work in the Table example (had to totally change the data store backing the example, hence the new `DenseGridMutableStore`.

I discovered a bunch of bugs in this work, including:
- #1257 
- #1258 
- #1259 
- #1261 
- #1262

#### Screenshot

reordering-example


packages/table/preview/index.tsx
```diff
@@ -97,10 +97,11 @@ const COLUMN_COUNTS = [
...
     0,
     1,
-    10,
+    5,
+    20,
     100,
-    10000,
-    1000000,
+    1000,
+    100000,
```
cmslewis: Rendering `1,000,000` cells now throws `Maximum Call Stack Exceeded` errors because I'm saving the rows in memory to retain their order, which I wasn't doing before. So I got rid of that option for now.

tgreenwatts:
This doesn't have the reorder handle icon in the interaction bar?

PR #1267, opened 2017-06-22
[Core] Adds additional arg to Tag onRemove handler

lucasray:
#### Checklist

- [X] Include tests
- [ ] Update documentation

Adds a 2nd arg to the `Tag`'s `onRemove` handler. This a non-breaking change, as passing functions that accept only 1 argument will still typecheck. This is useful in order to properly handle the `onRemove` callback without creating a lambda or additional React component, which may have perf consequences.


packages/core/src/components/tag/tag.tsx
```diff
@@ -26,7 +26,7 @@ export interface ITagProps extends IProps, IIntentProps, React.HTMLAttributes<Ta
      * Click handler for remove button.
      * Button will only be rendered if this prop is defined.
```
cmslewis: Can we add a line to the description explaining what the `props` parameter is?

e.g. `Receives the props for the tag that was removed.`

cmslewis: or maybe rename to `tagProps`

packages/core/src/components/tag/tag.tsx
```diff
@@ -50,6 +50,13 @@ export class Tag extends React.Component<ITagProps, {}> {
             </span>
         );
     }
+
+    private onRemoveClick = (e: React.MouseEvent<HTMLButtonElement>) => {
+        const { onRemove } = this.props;
+        if (onRemove) {
+            onRemove(e, this.props);
```
llorca: Is `safeInvoke` necessary here?

packages/core/src/components/tag/tag.tsx
lucasray: sure, I'll rename to `tagProps`, at which point I think it's fairly clear what props are being passed, but happy to also add to the description if you'd like

packages/core/src/components/tag/tag.tsx
lucasray: since `onRemove` is effectively `Function | undefined`, the typechecker should guarantee this to be safe. this callback only actually occurs if `isFunction(onRemove)` from line 42, though that isn't expressed in the types here. what does `safeInvoke` provide that the typechecker cannot? happy to rewrite this whole function as:

```typescript
private onRemoveClick = (e: React.MouseEvent<HTMLButtonElement>) => {
  const { onRemove } = this.props;
  safeInvoke(onRemove, e, this.props);
}
```

if preferred, just seems slightly more complicated for no gain.

packages/core/src/components/tag/tag.tsx
cmslewis: `tagProps` alone is fine to me. 👍

packages/core/src/components/tag/tag.tsx
cmslewis: The benefit is more for succinctness than anything. Note this whole thing becomes one line with `safeInvoke`:
```tsx
Utils.safeInvoke(this.props.onRemove, e, this.props);
```

lucasray: @cmslewis okay, do you all prefer this style? I slightly prefer the existing (clear what is happening without prior knowledge of what `safeInvoke` does, very slightly more performant since it doesn't add another call), but fine with changing it

lucasray:
<h3>Fixes lint</h3>

Preview: <strong><a href='https://circleci.com/api/v1/project/lucasray/blueprint/3/artifacts/0//home/ubuntu/blueprint/packages/site-docs/dist/index.html' target='_blank'>documentation</a></strong>
Coverage: <a href='https://circleci.com/api/v1/project/lucasray/blueprint/3/artifacts/0//home/ubuntu/blueprint/packages/core//coverage/PhantomJS%202.1.1%20%28Linux%200.0.0%29/index.html' target='_blank'>core</a> | <a href='https://circleci.com/api/v1/project/lucasray/blueprint/3/artifacts/0//home/ubuntu/blueprintpackages/datetime//coverage/PhantomJS%202.1.1%20%28Linux%200.0.0%29/index.html' target='_blank'>datetime</a>

packages/core/src/components/tag/tag.tsx
cmslewis: Yeah, it's BP convention at this point, but we don't need to block on it.

PR #1268, opened 2017-06-22
[Table] Add onVisibleCellsChange prop

### LINKED ISSUE #1260: [Table] Add callback for viewport changes to public API
Users who want to use the loading state as a standin for loaded data need to know what data to load based on the visible cells in the viewport.

We currently use the `viewportRect` and `Grid` to update the visible cell indices internally, but we should also expose these values in a callback so users can kick off a data load.

I propose a prop on `<Table>` like this:
```tsx
onVisibleCellsChange?: (rowStartIndex: number, rowEndIndex: number, colStartIndex: number, colEndIndex: number) => void;
```

cmslewis:
#### Checklist
- [x] Include tests

Add callback prop to `Table` with the following API:

```tsx
onVisibleCellsChange?: (rowIndices: IRowIndices, columnIndices: IColumnIndices) => void;
```

This prop allows consumers to know which cells to lazily load next.

**NOTE:** I _was_ exposing the following API before, but the long params were causing the function type definition to break the docs layout by exceeding the default width of the page.

```tsx
onVisibleCellsChange?: (rowStartIndex: number, rowEndIndex: number, colStartIndex: number, colEndIndex: number) => void;
```
 
Invoked whenever the viewport changes as a result of manual or programmatic scrolling.

Which API do you prefer? Is the broken docs layout a good enough reason to go with the two-parameter approach, or should we expose four parameters? Another consideration is that the two-parameter approach will force us to expose `IRowIndices` and `IColumnIndices` to consumers; these are currently only used internally.


packages/core/src/common/utils.ts
```diff
@@ -31,6 +31,13 @@ export function safeInvoke<A, B, C, R>(
     arg2: B,
     arg3: C,
 ): R;
+export function safeInvoke<A, B, C, D, R>(
```
themadcreator: oh man

packages/table/src/table.tsx
```diff
@@ -177,6 +177,14 @@ export interface ITableProps extends IProps, IRowHeights, IColumnWidths {
     onCopy?: (success: boolean) => void;
 
     /**
+     * A callback called when the visible cell indices change in the table.
+     */
+    onVisibleCellsChange?: (rowStartIndex: number,
```
themadcreator: this might be hard to read in the docs. alternative is to define a type alias for this function, but we don't currently document those =(

cmslewis:
@themadcreator I'll avoid merging until I address that.

packages/table/src/table.tsx
cmslewis: This actually messes up the formatting by making the page overly wide. :/ What's the best workaround? Leave as is, or maybe:

```tsx
onVisibleCellsChange?: (rowIndices: IRowIndices, columnIndices: IColumnIndices) => void;`
```

?

cmslewis:
@themadcreator I updated the PR to invoke the function whenever the grid invalidates (by changing num cols, num rows, row widths, or row heights).

packages/table/src/table.tsx
```diff
@@ -1588,6 +1594,17 @@ export class Table extends AbstractComponent<ITableProps, ITableState> {
         return loadingOptions.indexOf(loadingOption) >= 0;
     }
 
+    private updateViewportRect = (nextViewportRect: Rect) => {
```
themadcreator: you know what react needs? state change listeners. BRING BACK BACKBONE

packages/table/test/tableTests.tsx
```diff
@@ -144,6 +145,38 @@ describe("<Table>", () => {
         expect(table.state.rowHeights[0]).to.equal(MAX_HEIGHT);
     });
 
+    it("Invokes onVisibleCellsChange on mount", () => {
```
themadcreator: 👍

packages/table/preview/index.tsx
```diff
@@ -644,6 +646,13 @@ class MutableTable extends React.Component<{}, IMutableTableState> {
         this.maybeLogCallback(`[onCopy] success = ${success}`);
     }
 
+    private onVisibleCellsChange = (rowIndices: IRowIndices, columnIndices: IColumnIndices) => {
+        const { rowIndexStart, rowIndexEnd } = rowIndices;
+        const { columnIndexStart, columnIndexEnd } = columnIndices;
+        this.maybeLogCallback(`[onVisibleCellsChange] rowIndexStart=${rowIndexStart} rowIndexEnd=${rowIndexEnd}` +
```
themadcreator: you can also use multiline strings in backticks, but i think it includes the newline

packages/table/test/tableTests.tsx
```diff
@@ -833,6 +882,7 @@ describe("<Table>", () => {
 
             const viewportRect = table.state("viewportRect");
             expect(viewportRect.left).to.equal((NUM_COLS * UPDATED_COL_WIDTH) - viewportRect.width);
+            expect(onVisibleCellsChange.callCount).to.be.at.least(3);
```
themadcreator: why is this not precise?

themadcreator: is it inconsistent? If so, that seems like an issue. If it's consistent, just do an equals check. Tighten those bounds!

packages/table/preview/index.tsx
cmslewis: Yeah, it includes both the newline and the indentation:
image

packages/table/test/tableTests.tsx
cmslewis: I didn't make it precise, because the callback is invoked kind of all over the place, and I simply didn't know off the top of my head how many times to expect it to be called. I'll just count the calls and tighten the bounds here though.

PR #1275, opened 2017-06-23
[Labs] New MultiSelect component

### LINKED ISSUE #118: Component request: Typeahead / Autocomplete

llorca:
#### Checklist

- [ ] Include tests
- [ ] Update documentation

Final piece of #118. The new `MultiSelect<T>` component is similar to `Select<T>`. It composes `QueryList<T>` and `TagInput` together. Usage is very similar to `Select<T>`.

- Having a hard time defining default props. I keep getting this error: `Static members cannot reference class type parameters`, so for now I had to resort to `componentDidMount`.
- I have a bunch of private functions to handle list manipulation for `selectedItems`. I'm far behind on algorithms and data structures, what can be done better here?
- Spent a bunch of time on all the micro-interactions, I'm happy with the current boolean props and keyboard navigation. Am I missing anything?
- Note that in the future, I'd like to pull out the single/multi selection behavior, either into its own component or baked into `QueryList<T>`. There's too much duplicated code across `Select<T>` and `MultiSelect<T>` at the moment.

#### Screenshot


packages/labs/src/multi-select/multiSelect.tsx
```diff
@@ -0,0 +1,299 @@
...
+                onInteraction={this.handlePopoverInteraction}
+                popoverClassName={classNames("pt-multi-select-popover", popoverProps.popoverClassName)}
+                popoverDidOpen={this.handlePopoverDidOpen}
+            >
+                <div
+                    onKeyDown={this.handleKeyDown}
+                    onKeyUp={this.state.isOpen && handleKeyUp}
+                >
+                    <TagInput
+                        inputProps={defaultInputProps}
```
giladgray: oof this'll get clobbered if the user defines even one custom `inputProp`... gonna have to spread `tagInputProps.inputProps` into `defaultProps`.

packages/labs/src/multi-select/multiSelect.tsx
```diff
@@ -0,0 +1,299 @@
...
+    private queryList: QueryList<T>;
+    private queryListKeyDown: React.EventHandler<React.KeyboardEvent<HTMLElement>>;
+    private refHandlers = {
+        input: (ref: HTMLInputElement) => this.input = ref,
+        queryList: (ref: QueryList<T>) => this.queryList = ref,
+    };
+
+    public componentDidMount() {
+        // can't use defaultProps because "static members
+        // cannot reference class type parameters"
```
giladgray: i'm sure you _can_ use defaultProps, but you probably don't need to. safe to remove this comment as it doesn't really make sense on its own (def makes sense in PR description tho). 

think two months from now--will we even know what this means? why it matters?

packages/labs/src/multi-select/multiSelect.tsx
```diff
@@ -0,0 +1,299 @@
...
+        this.setState({ selectedItems: this.props.selectedItems || [] });
```
giladgray: going to need to update state based on prop in `componentWillReceiveProps` too (you've only done the initial render, not all subsequent renders).

but actually, i would kill the state field entirely and make it only controlled via props.

packages/labs/src/multi-select/multiSelect.tsx
```diff
@@ -0,0 +1,299 @@
...
+     * Whether the item is selected.
+     */
+    isSelected?: boolean;
+}
+
+export interface IMultiSelectState<T> {
+    activeItem?: T;
+    isOpen?: boolean;
+    query?: string;
+    selectedItems?: T[];
```
giladgray: i would argue for removing this from state entirely and supporting _only_ controlled usage. the other state fields are about internal interactions, but `selectedItems` is the big deal and the user will want to track it anyway. this'll make usage and testing much simpler and more reliable.

note how `Select<T>` doesn't even have a prop for its value, just `onItemSelect` event handler for user to receive updates and manage their own state.

packages/labs/src/multi-select/multiSelect.tsx
```diff
@@ -0,0 +1,299 @@
...
+        this.setState({ selectedItems: this.state.selectedItems.filter((_, i) => i !== index) });
+    }
+
+    private handleTagRemove = (_item: string, index: number) => {
+        this.deselectItem(index);
+
+        Utils.safeInvoke(this.props.tagInputProps.onRemove, _item, index);
+    }
+
+    private handleItemSelect = (item: T, _event: React.SyntheticEvent<HTMLElement>) => {
```
giladgray: no need for `_` prefix cuz it's used on the last line. same with `handleTagRemove` above.

packages/labs/src/multi-select/multiSelect.tsx
```diff
@@ -0,0 +1,299 @@
...
+    private selectItem(item: T) {
+        const { openOnKeyDown, resetOnSelect } = this.props;
+        let nextState = { selectedItems: [...this.state.selectedItems, item] };
+
+        if (openOnKeyDown) {
+            nextState = { ...nextState, isOpen: false };
+        }
+        if (resetOnSelect) {
+            nextState = { ...nextState, query: "" };
+        }
```
giladgray: this should be doable all at once, rather than creating up to three objects in one go. also you lose all type safety with your approach cuz it's just an `object`. 

something like:

```js
this.setState({
   selectedItems: ...
   isOpen: openOnKeyDown ? false : this.state.isOpen,
   query: resetOnSelect ? "" : this.state.query
});
```

packages/labs/src/multi-select/multiSelect.tsx
```diff
@@ -0,0 +1,299 @@
...
+    private handleTagRemove = (_item: string, index: number) => {
+        this.deselectItem(index);
+
+        Utils.safeInvoke(this.props.tagInputProps.onRemove, _item, index);
+    }
+
+    private handleItemSelect = (item: T, _event: React.SyntheticEvent<HTMLElement>) => {
+        // check whether the query string is in the item list
+        if (item !== undefined && this.state.activeItem !== undefined) {
+            if (!this.isItemSelected(item)) {
```
giladgray: how about getting the `selectedIndex` once and using it here and on line 233? then you don't need `isItemSelected` at all (it's a simple `>= 0` check anyway).

packages/labs/src/multi-select/multiSelect.tsx
```diff
@@ -0,0 +1,299 @@
...
+
+    private handlePopoverInteraction = (nextOpenState: boolean) => requestAnimationFrame(() => {
+        // deferring to rAF to get properly updated activeElement
+        const { popoverProps = {} } = this.props;
+
+        if (this.input != null && this.input !== document.activeElement) {
+            // the input is no longer focused so we can close the popover
+            let nextState = { isOpen: false };
+
+            if (this.props.resetOnSelect) {
```
giladgray: hmmm this logic happens at least three times. i'm sure we can come up with a helpful util to handle it only once, but i don't have the mental space for that right now 🌴

packages/labs/src/multi-select/multiSelect.tsx
```diff
@@ -0,0 +1,299 @@
...
+        Utils.safeInvoke(popoverProps.popoverDidOpen);
+    }
+
+    private handleActiveItemChange = (activeItem: T) => {
+        this.setState({ activeItem });
+    }
+
+    private handleKeyDown = (event: React.KeyboardEvent<HTMLElement>) => {
+        if (event.which === Keys.ESCAPE
+            || event.which === Keys.TAB) {
```
giladgray: one line, def not long enough for two. avoid awkward alignment.

packages/labs/src/multi-select/multiSelect.tsx
```diff
@@ -0,0 +1,299 @@
...
+            let nextState = { isOpen: false };
+
+            if (this.props.resetOnSelect) {
+                nextState = { ...nextState, query: "" };
+            }
+
+            this.setState(nextState);
+        } else if (!(event.which === Keys.ARROW_LEFT
+           || event.which === Keys.ARROW_RIGHT
+           || event.which === Keys.BACKSPACE)) {
```
giladgray: deconstructing `which` might let this all fit on one line...

`which === Keys.ARROW_LEFT || which === Keys.ARROW_RIGHT || which === Keys.BACKSPACE`

giladgray: _but actually_ `TagInput` calls `preventDefault()` when those keys are triggered... maybe you can just check `event.isDefaultPrevented` instead of redefining this logic?

packages/labs/src/multi-select/multiSelect.tsx
```diff
@@ -0,0 +1,299 @@
...
+           // these three keys help with navigating tags in the tag input,
+           // when triggered the popover should not open
+           this.setState({ isOpen: true });
+        }
+
+        if (this.queryListKeyDown) {
+            Utils.safeInvoke(this.queryListKeyDown, event);
```
giladgray: _oh snap_ i see what's going on here. `listProps.handleKeyDown` is only accessible inside the callback. ok, c'est la vie 🌴

packages/labs/src/multi-select/multiSelect.tsx
```diff
@@ -0,0 +1,299 @@
...
+    openOnKeyDown?: boolean;
+
+    /** Props to spread to `Popover`. Note that `content` cannot be changed. */
+    popoverProps?: Partial<IPopoverProps>;
+
+    /**
+     * Whether the filtering state should be reset to initial when an item is selected
+     * (immediately before `onItemSelect` is invoked), or when the popover closes.
+     * The query will become the empty string and the first item will be made active.
+     * @default true
```
giladgray: i think this should default to `false`. it's quite jarring when it's on.

packages/labs/src/multi-select/multiSelect.tsx
giladgray: perhaps echo `TagInput`'s API: `onItemAdd`, `onItemRemove`.

pkwi:
@llorca there is some text overlapping in the droppdown:

screen shot 2017-06-26 at 2 35 39 pm

llorca:
@pkwi https://github.com/palantir/blueprint/pull/1136

packages/labs/src/multi-select/multiSelect.tsx
llorca: That's the problem I'm hitting:
image

I'd still like to be able to set default props for `openOnKeyDown` and `resetOnSelect`

packages/labs/src/multi-select/multiSelect.tsx
llorca: Nah, check it out again. Makes more sense to reset the query after selection, because you're then ready to search again without having to manually clear the field. Select2 and react-select also have this default

packages/labs/src/multi-select/multiSelect.tsx
llorca: That's what I had before, I wasn't sure if setting a state to its current value was bad practice or not so I went for `if`s and spreads. I'll put this back

packages/labs/src/multi-select/multiSelect.tsx
llorca: Good catch, seems to be working!

packages/labs/src/multi-select/multiSelect.tsx
llorca: Let me know if you think of anything 🥇

packages/labs/src/multi-select/multiSelect.tsx
llorca: ✅

packages/labs/src/multi-select/multiSelect.tsx
llorca: It's fully controlled now, let me know what you think

packages/labs/src/multi-select/multiSelect.tsx
llorca: Nevermind, `event.isDefaultPrevented` didn't work out so I had to put back those 3 keys

cmslewis:
A couple interaction questions:

1. Mouthful: when you `keydown` on the <kbd>Enter</kbd> key, then press the <kbd>Up/Down</kbd> arrows, then `keyup` on the <kbd>Enter</kbd> key; the _newly focused_ item's selection state is toggled. This is jarring if you rapidly press <kbd>Up/Down</kbd>, <kbd>Enter</kbd>, <kbd>Up/Down</kbd>, <kbd>Enter</kbd>, <kbd>Up/Down</kbd>, <kbd>Enter</kbd>. In that case, I'd expect every item in succession to get selected, but there can be little race conditions that break that assumption if I `keypress` <kbd>Up/Down</kbd> _before_ I `keyup` on <kbd>Enter</kbd>:

multiselect-keypresses

2. If I navigate through the list using arrow keys, I'd expect <kbd>Space</kbd> to toggle the selected state of the focused item, but it actually just types a `' '` (space character) into the ghosted input. Is there any value in "reclaiming" the space bar to toggle selection if and only if the input is currently empty?

packages/labs/src/multi-select/multiSelect.tsx
```diff
@@ -0,0 +1,228 @@
+/*
```
giladgray: i think both these files belong in the `select/` directory as they share interfaces and concepts

llorca: ✅

llorca:
@cmslewis 
1. Nice catch! Unfortunately, this is tough to handle because our `Button`s react to `onKeyUp`. I can nicely implement what you suggest for `MultiSelect`, but it'd then break `Select` usages that contain a button as the target. This functionality comes from `QueryList`. Gilad commenting on `keyUp` vs `keyDown` for `enter`, related to `Select`:
> using keyup for enter to play nice with Button's keyboard clicking. if we were to process enter on keydown, then Button would click itself on keyup and the popover would re-open out of our control :(.

2. Also a good thought! I played with it but wasn't able to get a nice UX: the space character would appear in the field, then the item would get selected, and finally the field would clear. I'm not sure how to prevent the space character from showing up when the input is focused. FWIW, my hunch is that this interaction will be more confusing than helpful as we break the consistency of a focused input receiving typed characters.

packages/labs/src/common/classes.ts
```diff
@@ -5,6 +5,7 @@
  * and https://github.com/palantir/blueprint/blob/master/PATENTS
  */
 
+export const MULTISELECT = "pt-multiselect";
```
adidahiya: this should be pt-multi-select

packages/labs/src/components/select/multiSelect.tsx
```diff
@@ -0,0 +1,256 @@
...
+    public render() {
+        // omit props specific to this component, spread the rest.
+        const {
+            itemRenderer,
+            noResults,
+            openOnKeyDown,
+            popoverProps,
+            resetOnSelect,
+            tagInputProps,
+            ...props,
```
adidahiya: nit: would prefer naming this `restProps` for clarity through verbosity

packages/labs/src/components/select/multiSelect.tsx
```diff
@@ -0,0 +1,256 @@
...
+            onActiveItemChange={this.handleActiveItemChange}
+            onItemSelect={this.handleItemSelect}
+            query={this.state.query}
+            ref={this.refHandlers.queryList}
+            renderer={this.renderQueryList}
+        />;
+    }
+
+    private renderQueryList = (listProps: IQueryListRendererProps<T>) => {
+        const { tagInputProps = {}, popoverProps = {} } = this.props;
```
adidahiya: minor: set these to `{}` with `public static defaultProps` instead of as destructuring defaults

packages/labs/src/components/select/multiSelect.tsx
```diff
@@ -0,0 +1,256 @@
...
+        const { handleKeyDown, handleKeyUp, query } = listProps;
+        const defaultInputProps: HTMLInputProps = {
+            placeholder: "Search...",
+            ...this.props.tagInputProps.inputProps,
```
adidahiya: this looks like it should be `tagInputProps.inputProps`, otherwise, this could throw a null pointer exception

packages/labs/src/components/select/multiSelect.tsx
```diff
@@ -0,0 +1,256 @@
...
+            // tslint:disable-next-line:object-literal-sort-keys
```
adidahiya: looks like tslint is outdated (I believe we fixed this bug). but this is fine for this PR

packages/labs/src/components/select/multiSelect.tsx
```diff
@@ -0,0 +1,256 @@
...
+                popoverWillOpen={this.handlePopoverWillOpen}
+            >
+                <div
+                    onKeyDown={this.handleKeyDown}
+                    onKeyUp={this.state.isOpen && handleKeyUp}
+                >
+                    <TagInput
+                        inputProps={defaultInputProps}
+                        {...tagInputProps}
+                        className={classNames("pt-multi-select", tagInputProps.className)}
```
adidahiya: `Classes.MULTI_SELECT`

packages/labs/src/components/select/multiSelect.tsx
```diff
@@ -0,0 +1,256 @@
...
+            <Popover
+                autoFocus={false}
+                canEscapeKeyClose={true}
+                enforceFocus={false}
+                isOpen={this.state.isOpen}
+                position={Position.BOTTOM_LEFT}
+                {...popoverProps}
+                className={classNames(listProps.className, popoverProps.className)}
+                onInteraction={this.handlePopoverInteraction}
+                popoverClassName={classNames("pt-multi-select-popover", popoverProps.popoverClassName)}
```
adidahiya: `${Classes.MULTI_SELECT}-popover`

packages/labs/src/components/select/multiSelect.tsx
```diff
@@ -0,0 +1,256 @@
...
+        const defaultInputProps: HTMLInputProps = {
+            placeholder: "Search...",
+            ...this.props.tagInputProps.inputProps,
+            // tslint:disable-next-line:object-literal-sort-keys
+            onChange: this.handleQueryChange,
+            ref: this.refHandlers.input,
+            value: query,
+        };
+
+        this.queryListKeyDown = handleKeyDown;
```
adidahiya: a little strange to see a class member being assigned in a render method... what's going on here?

packages/labs/src/components/select/multiSelect.tsx
```diff
@@ -0,0 +1,256 @@
...
+
+    private handleQueryChange = (event: React.FormEvent<HTMLInputElement>) => {
+        const query = event.currentTarget.value;
+        this.setState({ query, isOpen: !(query.length === 0 && this.props.openOnKeyDown) });
+    }
+
+    private handleItemSelect = (item: T, event: React.SyntheticEvent<HTMLElement>) => {
+        this.input.focus();
+        // make sure the query is valid by checking if activeItem is defined
+        if (this.state.activeItem) {
```
adidahiya: please compare against null (`this.state.activeItem != null`), don't implicitly coerce to boolean

packages/labs/src/components/select/multiSelect.tsx
```diff
@@ -0,0 +1,256 @@
...
+            isActive: item === activeItem,
+        }));
+    }
+
+    private handleQueryChange = (event: React.FormEvent<HTMLInputElement>) => {
+        const query = event.currentTarget.value;
+        this.setState({ query, isOpen: !(query.length === 0 && this.props.openOnKeyDown) });
+    }
+
+    private handleItemSelect = (item: T, event: React.SyntheticEvent<HTMLElement>) => {
```
adidahiya: don't name events `event` (annoyingly, it conflicts with `window.event`), just go with `e` here

packages/labs/src/components/select/multiSelect.tsx
```diff
@@ -0,0 +1,256 @@
...
+                    query: "",
+                });
+            }
+            Utils.safeInvoke(this.props.onItemSelect, item, event);
+        }
+    }
+
+    private handlePopoverInteraction = (nextOpenState: boolean) => requestAnimationFrame(() => {
+        // deferring to rAF to get properly updated activeElement
+        const { popoverProps = {}, resetOnSelect } = this.props;
```
adidahiya: again use `defaultProps` so that you're not setting the default in multiple places

packages/labs/src/components/select/multiSelect.tsx
```diff
@@ -0,0 +1,256 @@
...
+        const { resetOnSelect } = this.props;
+
+        if (which === Keys.ESCAPE || which === Keys.TAB) {
+            this.setState({
+                activeItem: resetOnSelect ? this.props.items[0] : this.state.activeItem,
+                isOpen: false,
+                query: resetOnSelect ? "" : this.state.query,
+            });
+        } else if (!(which === Keys.BACKSPACE || which === Keys.ARROW_LEFT || which === Keys.ARROW_RIGHT)) {
+          this.setState({ isOpen: true });
```
adidahiya: expected 4-space indentation

packages/labs/src/components/select/multiSelect.tsx
llorca: That `handleKeyDown` is given to my `renderQueryList` function by the `renderer` prop of `QueryList`, therefore I don't have access to it outside of this function even though I need it. Had to store it in a class member, any better option?

packages/labs/src/components/select/multiSelect.tsx
llorca: Yeah that's what I wanted, but hit a road block. We went over this earlier in this thread. That's the problem I'm hitting:
image

I also want to be able to set default props for `openOnKeyDown` and `resetOnSelect`

packages/labs/src/components/select/multiSelect.tsx
llorca: Addressed above

packages/labs/src/components/select/multiSelect.tsx
llorca: How does this even happen

packages/labs/src/components/select/multiSelect.tsx
adidahiya: ah, I can see that `Select` has this code issue as well. we can address them both outside this PR.

packages/labs/src/components/select/multiSelect.tsx
adidahiya: meh yeah this is kind of hairy. would prefer to make an exception to the jsx lambda rule here and refactor `this.handleKeyDown` into a getter:

```ts
    <div
        onKeyDown={this.getTargetKeyDownHandler(listProps.handleKeyDown)}
        onKeyUp={this.state.isOpen && handleKeyUp}
    >

...

    private getTargetKeyDownHandler(handleQueryListKeyDown: React.EventHandler<React.KeyboardEvent<HTMLElement>) {
        return (e: React.KeyboardEvent<HTMLElement>) => {
            const { which } = e;
            const { resetOnSelect } = this.props;

            ...

            if (this.state.isOpen) {
                Utils.safeInvoke(handleQueryListKeyDown, e);
            }
        };
    }
```

llorca: Nice

PR #1282, opened 2017-06-26
[Table] Improve Utils.times(...) to work with huge numbers 

### LINKED ISSUE #1279: [Table] Large number of rows causes stack overflow error
### Bug report

- __Package version(s)__: v1.17.0
- __Browser and OS versions__: Electron  on macOS Sierra 

#### Steps to reproduce

1. Set large number of rows on Table: 151152 in my case.

#### Actual behavior

Runtime error:
utils.js:43 Uncaught RangeError: Maximum call stack size exceeded
    at Object.times (utils.js:43)
    at new Table (table.js:407)
    at react-dom.js:4756
    at measureLifeCyclePerf (react-dom.js:4537)
    at ReactCompositeComponentWrapper._constructComponentWithoutOwner (react-dom.js:4755)
    at ReactCompositeComponentWrapper._constructComponent (react-dom.js:4741)
    at ReactCompositeComponentWrapper.mountComponent (react-dom.js:4649)
    at Object.mountComponent (react-dom.js:11551)
    at ReactCompositeComponentWrapper.performInitialMount (react-dom.js:4832)
    at ReactCompositeComponentWrapper.mountComponent (react-dom.js:4719)

#### Expected behavior

Component should be able to handle large numRows count.
If I replace "times" function with a simple loop it works fine.
```
Utils.times = (n: number, callback: (i: number) => any): any[]=>{
    let result = Array(n);
    for(let i = 0;i < n;i++){
        result[i] = callback(i);
    }
    return result;
}
```

cmslewis:
#### Checklist

- [x] Include tests

Use a basic for loop in `Utils.times(...)` instead of a `.map(...)`. This lets the function handle much bigger numbers without filling up the call stack.

Poke around the table example to make sure everything still works.

### DIFF
--- a/packages/table/src/common/utils.ts
+++ b/packages/table/src/common/utils.ts
@@ -53,7 +53,11 @@ export const Utils = {
      * is the return value. Similar to _.times
      */
     times<T>(n: number, callback: (i: number) => T): T[] {
-        return Array.apply(null, Array(n)).map((_none: any, index: number) => callback(index));
+        const result: T[] = Array(n);
+        for (let index = 0; index < n; index++) {
+            result[index] = callback(index);
+        }
+        return result;
     },
 
     /**
--- a/packages/table/test/utilsTests.ts
+++ b/packages/table/test/utilsTests.ts
@@ -67,7 +67,7 @@ describe("Utils", () => {
         });
     });
 
-    describe("times", () => {
+    describe.only("times", () => {
         it("returns empty array for 0", () => {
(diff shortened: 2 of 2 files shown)


cmslewis:
@llorca the example works with 1,000,000 rows now, but due to the data store saving data in memory to allow for reordering, it still falls over when you have that many rows PLUS lots of columns. I'll leave the example alone for this PR.

PR #1295, opened 2017-06-28
[Table] Frozen Rows/Cols - Part 1: 4 Scrollable Quadrants

cmslewis:
#### Checklist

- [x] ~Include tests~  (I `it.skip`'d all the failing tests for now. Will fix them in future PRs.)

This is the first in a series of upcoming PRs that will implement frozen columns and frozen rows (note that this PR targets a feature branch). This PR is very rough and ugly. See the "Development process" section below for reasoning why.

The essential change here is to introduce a concept of "quadrants" in the `Table`. The current table body will now become the "main quadrant," with three additional quadrants layered on top of it:
- __Left quadrant:__ contains row headers and frozen columns
- __Top quadrant:__ contains column headers and frozen rows
- __Top-left quadrant:__ contains the headers and cells shared between the frozen columns and the frozen rows

table-quadrants

The crucial UX goals here are:
- Make the top quadrant horizontally scrollable, but immobile during vertical scrolls.
- Make the left quadrant vertically scrollable, but immobile during horizontal scrolls.
- Listen for bidirectional scroll events in all of these quadrants.
- Sync scroll distances between all of the quadrants.

Thus, scrolling horizontally or vertically within any quadrant—even the fixed top-left quadrant—triggers scrolling in all other quadrants simultaneously.

- __Development process:__ It's going to be prohibitive to keep every piece of existing table functionality working as I build this new feature, so my strategy is to comment out (not delete!) everything except the most basic behaviors, then rebuild and re-integrate all of the existing features on top of this new four-quadrant structure. I don't _think_ this will require any breaking changes in the end.

- __Performance:__ There's noticeable jank/misalignment between quadrants when scrolling any one of them. I'm hoping that we can find ways to streamline this.

#### Screenshot

table-frozen-demo


cmslewis:
@giladgray @themadcreator @gscshoyru @llorca @tgreenwatts etc. - I'd like to add `it.skip`s to all of the failing tests on the first few PRs here. Trying to keep all of them working at each stage is probably going to require an unsustainable amount of code change in each PR. Sound okay?

packages/table/src/headers/columnHeader.tsx
```diff
@@ -103,18 +103,13 @@ export class ColumnHeader extends React.Component<IColumnHeaderProps, {}> {
     }
 
     private wrapCells = (cells: Array<React.ReactElement<any>>) => {
-        const { grid, viewportRect, columnIndexStart } = this.props;
+        const { grid } = this.props;
```
cmslewis: `ColumnHeader` simply scrolls with the regular body cells now; it doesn't need to do fancy `transform`s anymore.

packages/table/src/table.scss
```diff
@@ -163,3 +163,70 @@ $menu-z-index: $row-z-index + 1 !default;
   height: $pt-grid-size * 10;
   overflow: scroll;
 }
+
+// TEMPORARY
```
cmslewis: I guess most of these styles are actually not temporary; it's just the background colors that are.

packages/table/src/table.tsx
```diff
@@ -477,29 +503,108 @@ export class Table extends AbstractComponent<ITableProps, ITableState> {
...
         return (
             <div
                 className={classes}
                 ref={this.setRootTableRef}
                 onScroll={this.handleRootScroll}
             >
-                 <div className={Classes.TABLE_TOP_CONTAINER}>
-                    {isRowHeaderShown ? this.renderMenu() : undefined}
-                    {this.renderColumnHeader()}
+                <div className="bp-table-quadrant-main" style={mainQuadrantStyles} ref={this.setMainQuadrantRef}>
```
cmslewis: Pull these class strings into `table/**/classes.ts`.

packages/table/src/table.tsx
```diff
@@ -477,29 +503,108 @@ export class Table extends AbstractComponent<ITableProps, ITableState> {
...
+                        <div className={Classes.TABLE_BOTTOM_CONTAINER}>
+                            {this.renderRowHeader()}
+                            {this.renderBody(null, null, 0, numFrozenColumns, numFrozenColumns, null)}
+                        </div>
+                    </div>
                 </div>
-                <div className={Classes.TABLE_BOTTOM_CONTAINER}>
-                    {isRowHeaderShown ? this.renderRowHeader() : undefined}
-                    {this.renderBody()}
+                <div className="bp-table-quadrant-top-left" style={baseStyles} ref={this.setTopLeftQuadrantRef}>
```
cmslewis: These are essentially four instances of the same pattern:

```tsx
<div className style ref>
    <div className="bp-table-quadrant-scroll-container" ref onWheel onScroll />
        <div className={Classes.TABLE_TOP_CONTAINER}>
            {this.renderMenu(ref)} // or maybe a menu placeholder for underlying quadrants? we just need one menu element to be visible...
            {this.renderColumnHeader(numFrozenColumns?)} // or maybe a placeholder for underlying quadrants?; can simplify these args
        </div>
        <div className={Classes.TABLE_BOTTOM_CONTAINER}>
            {this.renderRowHeader(numFrozenRows?)} // or a placeholder; can simplify these args
            {this.renderBody(numFrozenRows?, numFrozenColumns?)} // can simplify these args
        </div>
    </div>
</div>
```

Let's pull it out into a `TableQuadrant` component.

packages/table/src/table.tsx
```diff
@@ -555,7 +660,8 @@ export class Table extends AbstractComponent<ITableProps, ITableState> {
             }
         });
 
-        this.syncMenuWidth();
+        this.syncMenuElementWidths();
+        this.syncQuadrantSizes();
```
cmslewis: Since these functions are always called back-to-back and _must_ be invoked in this order, let's combine them into one function.

packages/table/src/table.tsx
```diff
@@ -679,29 +786,122 @@ export class Table extends AbstractComponent<ITableProps, ITableState> {
...
+        const nextScrollLeft = this.mainQuadrantScrollElement.scrollLeft + event.deltaX;
+
+        this.mainQuadrantScrollElement.scrollTop = nextScrollTop;
+        this.mainQuadrantScrollElement.scrollLeft = nextScrollLeft;
+
+        this.leftQuadrantScrollElement.scrollTop = nextScrollTop;
+        this.topQuadrantScrollElement.scrollLeft = nextScrollLeft;
+
+        this.handleBodyScroll(event);
+    }
```
cmslewis: Look for a way to bundle these into a new `TableQuadrant` component.

packages/table/src/table.tsx
```diff
@@ -679,29 +786,122 @@ export class Table extends AbstractComponent<ITableProps, ITableState> {
...
+        const {
+            mainQuadrantMenuElement,
+            mainQuadrantScrollElement,
+            topQuadrantElement,
+            leftQuadrantElement,
+            topLeftQuadrantElement,
+        } = this;
+
+        const frozenColumnsCumulativeWidth = this.grid.getCumulativeWidthAt(this.props.numFrozenColumns - 1);
+        const frozenRowsCumulativeHeight = this.grid.getCumulativeHeightAt(this.props.numFrozenRows - 1);
```
cmslewis: This `- 1` is a _little_ unintuitive at first glance. Maybe create two variables `maxFrozenRowIndex` and `maxFrozenColumnIndex`?

packages/table/src/table.tsx
```diff
@@ -785,7 +985,7 @@ export class Table extends AbstractComponent<ITableProps, ITableState> {
         }
     }
 
-    private renderColumnHeader() {
+    private renderColumnHeader(columnIndexStart?: number, columnIndexEnd?: number) {
```
cmslewis: Refactor to one parameter and rename to `numFrozenColumns?: number`. If that parameter is defined, then set `columnIndexStart = 0` and `columnIndexEnd = numFrozenColumns - 1`.

packages/table/src/table.tsx
```diff
@@ -835,7 +1036,7 @@ export class Table extends AbstractComponent<ITableProps, ITableState> {
         );
     }
 
-    private renderRowHeader() {
+    private renderRowHeader(rowIndexStart?: number, rowIndexEnd?: number) {
```
cmslewis: Same.

packages/table/src/table.tsx
```diff
@@ -898,7 +1101,14 @@ export class Table extends AbstractComponent<ITableProps, ITableState> {
...
     }
 
-    private renderBody() {
+    private renderBody(
+        rowIndexStart?: number,
+        rowIndexEnd?: number,
+        columnIndexStart?: number,
+        columnIndexEnd?: number,
+        numFrozenColumns?: number,
+        numFrozenRows?: number,
```
cmslewis: Rework these parameters. Don't need all of them.

packages/table/src/table.tsx
```diff
@@ -907,33 +1117,37 @@ export class Table extends AbstractComponent<ITableProps, ITableState> {
...
-        const noVerticalScroll = fillBodyWithGhostCells &&
-            grid.isGhostIndex(rowIndices.rowIndexEnd, 0) &&
-            viewportRect != null && viewportRect.top === 0 ||
-            this.hasLoadingOption(loadingOptions, TableLoadingOption.ROW_HEADERS);
-        const noHorizontalScroll = fillBodyWithGhostCells &&
-            grid.isGhostIndex(0, columnIndices.columnIndexEnd) &&
-            viewportRect != null && viewportRect.left === 0 ||
-            this.hasLoadingOption(loadingOptions, TableLoadingOption.COLUMN_HEADERS);
+
+        // const noVerticalScroll = fillBodyWithGhostCells &&
```
cmslewis: Add `TODO: FROZEN` comment above each of these commented lines/blocks. That can serve as the `grep`-able cue that something needs to be revisited before merging the feature branch back into `master`.

packages/table/src/table.tsx
```diff
@@ -946,20 +1160,25 @@ export class Table extends AbstractComponent<ITableProps, ITableState> {
...
                         viewportRect={viewportRect}
-                        {...rowIndices}
-                        {...columnIndices}
-                    />
 
-                    {this.maybeRenderRegions(this.styleBodyRegion)}
+                        columnIndexStart={columnIndexStart == null ? columnIndices.columnIndexStart : columnIndexStart}
+                        columnIndexEnd={columnIndexEnd == null ? columnIndices.columnIndexEnd : columnIndexEnd}
+                        rowIndexStart={rowIndexStart == null ? rowIndices.rowIndexStart : rowIndexStart}
+                        rowIndexEnd={rowIndexEnd == null ? rowIndices.rowIndexEnd : rowIndexEnd}
```
cmslewis: Simplify this logic as we'll do in `ColumnHeader`.

packages/table/src/tableBody.tsx
```diff
@@ -46,6 +46,16 @@ export interface ITableBodyProps extends ISelectableProps, IRowIndices, IColumnI
     locator: ILocator;
 
     /**
+     * The number of columns to freeze to the left side of the table, counting from the leftmost column.
+     */
```
cmslewis: Add an explanation of how these props are used by the component.

packages/table/src/headers/columnHeader.tsx
themadcreator: how does this affect the computation of the viewport rect and which cells are visible?

packages/table/src/headers/columnHeader.tsx
```diff
@@ -103,18 +103,13 @@ export class ColumnHeader extends React.Component<IColumnHeaderProps, {}> {
...
 
         // always set width so that the layout can push out the element unless it overflows.
         const style: React.CSSProperties = {
             width: `${grid.getRect().width}px`,
         };
 
-        // use CSS translation to offset the cells
-        if (viewportRect != null) {
```
themadcreator: do we even use viewportRect anymore? can it be removed as a prop?

packages/table/src/quadrants/tableQuadrant.tsx
```diff
@@ -0,0 +1,161 @@
...
+                        </div>
+                    </div>
+                </div>
+            </div>
+        );
+    }
+
+    protected validateProps(nextProps: ITableQuadrantProps) {
+        const { quadrantType } = nextProps;
+        if (nextProps.onScroll != null && quadrantType !== QuadrantType.MAIN) {
```
themadcreator: I think you could actually typecheck this by specifying a base props object and child interfaces like so:

```
interface IBaseQuadProps {
   type: QuadrantType;
   onScroll?: /* optional */;
}

interface IMainQuadProps extends IBaseQuadProps {
   type: QuadrantType.MAIN;
   onScroll: /* NOT optional */;
}

interface IOtherQuadProps  extends IBaseQuadProps {
   type: QuadrantType.LEFT | QuadrantType:TOP | ... ;
   onScroll: undefined; // <-- not sure about this line, but you get the idea
}
```

packages/table/src/quadrants/tableQuadrant.tsx
```diff
@@ -0,0 +1,161 @@
...
+            console.warn(Errors.QUADRANT_ON_SCROLL_UNNECESSARILY_DEFINED);
+        }
+        if (nextProps.onWheel && quadrantType === QuadrantType.MAIN) {
+            console.warn(Errors.QUADRANT_ON_WHEEL_UNNECESSARILY_DEFINED);
+        }
+    }
+
+    private getQuadrantCssClass() {
+        const { quadrantType } = this.props;
+        if (quadrantType === QuadrantType.MAIN) {
```
themadcreator: switch case, please

cmslewis: Done 👍

packages/table/src/quadrants/tableQuadrant.tsx
cmslewis: Interesting. Would this require two new components too?

`<MainQuadrant {...mainQuadrantProps} />`
`<NonMainQuadrant {...topQuadrantProps} />`
`<NonMainQuadrant {...leftQuadrantProps} />`
`<NonMainQuadrant {...topLeftQuadrantProps} />`

packages/table/src/table.tsx
themadcreator: Maybe we should add a circleCI script to grep for `// NOCOMMIT` (after the demo comment is generated)

themadcreator:
Looking at the interaction, it doesn't seem right. Consider this gif:
scroll-wut

Why am i allowed to scroll when there are ghost cells? I should be able to scroll the content when there is nothing worth scrolling. Also, why am I only able to scroll 2 rows worth?

Also, cell selection doesn't seem to work.

cmslewis:
@themadcreator I intentionally disabled/broke a bunch of features that would have required too many LOC to fix in this PR, including selection/reordering/etc. I can fix those here if we want, but it seems more manageable to fix that stuff in future PRs targeting this same feature branch.

The goal of this PR is mostly to get feedback on the implementation, directionally. In followup PRs, I'll fix all of the issues you describe.

themadcreator:
Well, I don't know how I feel about that approach. The last time we tried this, there was a performance regression in a table PR that you said you'd fix before release. That didn't happen and we released with the big perf regression--only recently addressed when we had that sprint. I think we should only merge this code if it doesn't cause feature regressions. Thoughts?

themadcreator:
Alternatively if you want smaller PRs, you can create branches BASED on this one, then merge them one by one to this, then finally back to master

cmslewis:
> Alternatively if you want smaller PRs, you can create branches BASED on this one, then merge them one by one to this, then finally back to master

Yep, that's what I'm doing here, targeting `feature/table-frozen-rows-cols` so that I can keep PRs small without endangering `master`.

cmslewis:
Merging into a feature branch (not `master`) with the intention of fixing lingering brokenness in future PRs targeting this same branch.

PR #1319, opened 2017-07-05
[Menu] Add ulRef prop

llorca:
#### Checklist

- [ ] Include tests
- [ ] Update documentation

Add `ulRef` prop to `Menu`, to avoid breaking out of `Menu` usage when needing a `ref` to the underlying `ul`.
```tsx
<ul className={Classes.MENU} ref={...}>...</ul>
```
becomes
```tsx
<Menu ulRef={...}>...</Menu>
```


packages/core/src/components/menu/menu.tsx
```diff
@@ -11,15 +11,23 @@ import * as React from "react";
 import * as Classes from "../../common/classes";
 import { IProps } from "../../common/props";
 
-// we need some this interface to show up in docs
-// tslint:disable-next-line no-empty-interface
-export interface IMenuProps extends IProps {}
+export interface IMenuProps extends IProps {
+    /** Ref handler that receives the HTML `<ul>` element backing this component. */
+    ulRef?: (ref: HTMLUListElement) => any;
```
llorca: `ulRef`, `wrapperRef` or simply `ref`?

adidahiya: It shouldn't be simply `ref`, because that's used to get a handle on the `Menu` instance itself instead of the DOM element. `ulRef` or `uListRef` sounds good

adidahiya:
merge conflict

PR #1325, opened 2017-07-06
[Labs] New Omnibox component

llorca:
#### Checklist

- [ ] Include tests
- [ ] Update documentation

Fully controlled `Omnibox<T>` component, generic just like `Select<T>` and `MultiSelect<T>`. Built with `Overlay` + `QueryList<T>`.

Any thoughts on a better name?

#### Screenshot


tgreenwatts:
@michael-yx-wu for SA!

llorca:
Added `allowInInput` to the hotkey, feels so nice

michael-yx-wu:
Does this allow for categories? You'll also want to allow for custom preview content much like how spotlight does this on macOS as well as escape key behavior to clear text / close depending on whether there is text entered.

llorca:
@michael-yx-wu the docs example is using a `Menu`, but you can render the list however you want to thanks to `QueryList`: http://blueprintjs.com/docs/#labs.renderer-api
We provide keyboard interactions as well as a `resectOnSelect` prop to configure query persistence.

michael-yx-wu:
Can you change the fade in animation to be faster? It looks nice, but it's not as functional as something that appears instantly or closer to instantly.

llorca:
It's totally functional: as soon as you trigger the box, you can *instantly* type in a query--can't get faster than that.

packages/labs/src/components/omnibox/omnibox.tsx
```diff
@@ -0,0 +1,191 @@
...
+    /**
+     * Props to spread to `InputGroup`. All props are supported except `ref` (use `inputRef` instead).
+     * If you want to control the filter input, you can pass `value` and `onChange` here
+     * to override `Select`'s own behavior.
+     */
+    inputProps?: IInputGroupProps & HTMLInputProps;
+
+    /**
+     * Toggles the visibility of the omnibox.
+     * This prop is required because the component is controlled.
```
giladgray: doesn't look required here! remove the `?` or remove the comment.

packages/labs/src/components/omnibox/omnibox.tsx
```diff
@@ -0,0 +1,191 @@
...
+     */
+    isOpen?: boolean;
+
+    onClose?: (event?: React.SyntheticEvent<HTMLElement>) => void;
```
giladgray: add docs

packages/labs/src/labs.md
```diff
@@ -91,8 +91,25 @@ const renderMenuItem = ({ handleClick, item: film, isActive }: ISelectItemRender
...
+
+@reactExample MultiSelectExample
+
+@interface IMultiSelectProps
+
+@interface ISelectItemRendererProps
+
 @## Omnibox
 
+`Omnibox<T>` is a Spotlight-style typeahead component composing `Overlay` and `QueryList<T>`. Usage is similar to `Select<T>`: provide your items as well as a predicate to customize the filtering algorithm. The component is fully controlled via the `isOpen` prop, which means you can decide exactly how to trigger the component—we use a both a button and a keyboard shortcut to showcase the example below.
```
giladgray: changes in [braces]:

> `Omnibox<T>` is a [OS X] Spotlight-style typeahead component composing `Overlay` and `QueryList<T>`. Usage is similar to `Select<T>`: provide your items [and] a predicate to customize the filtering algorithm. The component is fully controlled via the `isOpen` prop, which means you can decide exactly how to trigger the component—[the following example responds to a button and a hotkey].

packages/labs/src/components/omnibox/omnibox.tsx
```diff
@@ -0,0 +1,197 @@
...
+     * A callback that is invoked when user interaction causes the overlay to close, such as
+     * clicking on the overlay or pressing the `esc` key (if enabled).
+     * Receives the event from the user's interaction, if there was an event (generally either a
+     * mouse or key event). Note that, since this component is controlled by the `isOpen` prop, it
+     * will not actually close itself until that prop becomes `false`.
+     */
+    onClose?: (event?: React.SyntheticEvent<HTMLElement>) => void;
+
+    /** Props to spread to `Overlay`. Note that `content` cannot be changed. */
+    overlayProps?: Partial<IOverlayProps>;
```
cmslewis: Make the type `Partial<IOverlayProps> & object` for strict type-checking. (Otherwise, you can pass in stuff like `{ overlayProps: "not an object" }`, and it won't complain!

packages/labs/src/components/omnibox/omnibox.tsx
```diff
@@ -0,0 +1,197 @@
...
+    resetOnSelect?: boolean;
+}
+
+export interface IOmniboxState<T> extends IOverlayableProps, IBackdropProps {
+    activeItem?: T;
+    query?: string;
+}
+
+@PureRender
+export class Omnibox<T> extends AbstractComponent<IOmniboxProps<T>, IOmniboxState<T>> {
```
cmslewis: `AbstractComponent` doesn't seem to be providing any value here above the basic `React.Component`. Only use `AbstractComponent` if you'll implement `validateProps` or if you'll leverage the `setTimeout`/`clearTimeouts` behavior.

packages/labs/src/components/omnibox/omnibox.tsx
```diff
@@ -0,0 +1,197 @@
...
+        const { query } = this.state;
+        const { ref, ...htmlInputProps } = inputProps;
+        const { handleKeyDown, handleKeyUp } = listProps;
+
+        return (
+            <Overlay
+                hasBackdrop={true}
+                {...overlayProps}
+                isOpen={isOpen}
+                className={classNames(overlayProps.className, `${Classes.OMNIBOX}-overlay`)}
```
cmslewis: I'd prefer to just write out the full string `pt-omnibox-overlay` (either here or in `Classes.ts`) to ensure that it's grep-able.

packages/labs/src/components/omnibox/omnibox.tsx
```diff
@@ -0,0 +1,197 @@
...
+                onClose={this.handleOverlayClose}
+            >
+                <div
+                    className={classNames(listProps.className, Classes.OMNIBOX)}
+                    onKeyDown={isOpen && query.length > 0 && handleKeyDown}
```
cmslewis: The intention here isn't clear to me as written. Are you trying to do a ternary?

`(isOpen && query.length > 0) ? handleKeyDown : undefined`

packages/labs/src/components/omnibox/omnibox.tsx
```diff
@@ -0,0 +1,197 @@
...
+                    onKeyUp={isOpen && query.length > 0 && handleKeyUp}
```
cmslewis: Same.

packages/labs/src/components/omnibox/omnibox.tsx
```diff
@@ -0,0 +1,197 @@
...
+    private renderItems({ activeItem, filteredItems, handleItemSelect }: IQueryListRendererProps<T>) {
+        const { itemRenderer, noResults } = this.props;
+        if (filteredItems.length === 0) {
+            return noResults;
+        }
+        return filteredItems.map((item, index) => itemRenderer({
+            index,
+            item,
+            handleClick: (e) => handleItemSelect(item, e),
+            isActive: item === activeItem,
```
cmslewis: Just want to confirm that `===` is a sufficient equality check for all use cases. Will consumers ever need/want to customize this check?

packages/labs/src/components/omnibox/omnibox.tsx
```diff
@@ -0,0 +1,197 @@
...
+export interface IOmniboxProps<T> extends IListItemsProps<T> {
+    /**
+     * Custom renderer for an item in the dropdown list. Receives a boolean indicating whether
+     * this item is active (selected by keyboard arrows) and an `onClick` event handler that
+     * should be attached to the returned element.
+     */
+    itemRenderer: (itemProps: ISelectItemRendererProps<T>) => JSX.Element;
+
+    /** React child to render when filtering items returns zero results. */
+    noResults?: string | JSX.Element;
```
llorca: If you don't mind, I can do this in a follow up PR across `Omnibox`, `Select` and `MultiSelect`

packages/labs/src/components/omnibox/omnibox.tsx
llorca: Snap, I'm sure we do this wrong in other places too. What about `inputProps` and `popoverProps`?

packages/labs/src/components/omnibox/omnibox.tsx
llorca: Gotcha, can do this as a follow up across other Labs components

packages/labs/src/components/omnibox/omnibox.tsx
llorca: I'm trying to inline this:
```
if (isOpen && query.length > 0) {
    return handleKeyDown;
}
```
I'd prefer cleaning this up across the Labs components as a follow up.

Subsidiary question, is this syntax ever okay? Like with less chaining, i.e. `isOpen && handleKeyDown`

packages/labs/src/components/omnibox/omnibox.tsx
llorca: Good question, I think it's best for us to keep control of that because it basically represents the custom "focus" behavior that the keyboard shortcuts rely on (up, down, return). I don't see a need to customize this now, we can update this function if we ever need to!

packages/labs/src/components/omnibox/omnibox.tsx
cmslewis: > Subsidiary question, is this syntax ever okay?

It may work, but I don't find it particularly clear. I like everything in a boolean expression to be an actual boolean check.

packages/labs/src/components/omnibox/omnibox.tsx
cmslewis: AFAIK, anywhere you use `Partial`, you'll want the `... & object` piece too.

packages/labs/src/components/omnibox/omnibox.tsx
llorca: So apparently it's a JavaScript syntax thing, and the React guys seem to be okay with it since they suggest it right next to proper ternaries: https://facebook.github.io/react/docs/conditional-rendering.html#inline-if-with-logical--operator

llorca: Will probably end up changing this line to a proper ternary by throwing `undefined` at the else statement. 🎩  @giladgray

cmslewis:
Some of the issues I raised are prevalent in both this and other files, so we'll merge this now and fix those issues in all Labs components immediately thereafter.

packages/labs/src/components/omnibox/omnibox.tsx
giladgray: `===` is def not sufficient, this has come up in chat a few times. 
easy way to break it is to create new objects in each render (rather than reusing existing ones), which is a very common JS pattern.

we'll need to add an `isEqual` prop with a default `===` implementation so folks can define their own equality checks.

PR #1333, opened 2017-07-10
Disable the `<RoundSize>` component wrapper

themadcreator:
Previously, a bug in chrome broke scrolling in the table.

As demonstrated in this fiddle, the bug has been fixed.
https://jsfiddle.net/2rmz7p1d/5/

The RoundSize workaround has non-zero performance impact,
so we should disable it now that chrome is fixed.

### DIFF
--- a/packages/table/src/headers/rowHeader.tsx
+++ b/packages/table/src/headers/rowHeader.tsx
@@ -10,7 +10,6 @@ import * as React from "react";
 
 import * as Classes from "../common/classes";
 import { IRowIndices } from "../common/grid";
-import { RoundSize } from "../common/roundSize";
 import { IClientCoordinates } from "../interactions/draggable";
 import { IIndexedResizeCallback } from "../interactions/resizable";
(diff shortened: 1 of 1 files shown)


packages/table/src/headers/rowHeader.tsx
```diff
@@ -10,7 +10,6 @@ import * as React from "react";
 
 import * as Classes from "../common/classes";
 import { IRowIndices } from "../common/grid";
-import { RoundSize } from "../common/roundSize";
```
cmslewis: @themadcreator Can we delete the `RoundSize` file too?

themadcreator: It's currently exported from common, so that's technically a breaking change

cmslewis: 👍

cmslewis: Tracking in https://github.com/palantir/blueprint/issues/1334

PR #1336, opened 2017-07-11
[Form group] Better pt-control support

### LINKED ISSUE #1305: Switches in form groups have a margin-bottom
### Bug report

Latest, Chrome

#### Steps to reproduce

1. Put a switch inside a form group
2. Put some helper text below the switch
3. Boom, white space

#### Actual behavior

Lots of white space.

#### Expected behavior

Not so much white space.

llorca:

- Proper `min-height` for `.pt-control` (16px normal, 20px large)
- Align `.pt-form-group` items to their baseline
- Tweak `.pt-control` vertical margin within `.pt-form-group`

#### Screenshot

Before:
screen shot 2017-07-11 at 4 15 18 pm

After:
screen shot 2017-07-11 at 4 15 26 pm

### DIFF
--- a/packages/core/src/components/forms/_form-group.scss
+++ b/packages/core/src/components/forms/_form-group.scss
@@ -34,42 +34,67 @@ Markup:
   </div>
 </div>
 <div class="pt-form-group">
-  <label class="pt-label" for="example-form-group-switch-a">
-    Label A
+  <label class="pt-label" for="example-form-group-switch-c">
+    Label C
     <span class="pt-text-muted">(required)</span>
   </label>
   <div class="pt-form-content">
     <label class="pt-control pt-switch">
-      <input id="example-form-group-switch-a" type="checkbox" />
+      <input id="example-form-group-switch-c" type="checkbox" />
       <span class="pt-control-indicator"></span>
-      Switch A
+      Switch C
     </label>
     <div class="pt-form-helper-text">Helper text with details / user feedback</div>
   </div>
 </div>
-<div class="pt-form-group pt-inline pt-large pt-disabled">
-  <label class="pt-label" for="example-form-group-input-c">
-    Label C
+<div class="pt-form-group pt-inline">
+  <label class="pt-label" for="example-form-group-input-d">
+    Label D
     <span class="pt-text-muted">(optional)</span>
   </label>
   <div class="pt-form-content">
-    <div class="pt-input-group pt-large pt-disabled">
+    <div class="pt-input-group">
       <span class="pt-icon pt-icon-calendar"></span>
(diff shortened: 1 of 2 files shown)


cmslewis:
Verified in different browsers. Looks good!

llorca:
Hehe, did you notice that, for inline form groups, the label text and input text (not switch) are off by 2px? I could not find any elegant fix

cmslewis:
@llorca should we fix that before merging this? Figure we might as well throw 30 minutes at the problem now and hope for a resolution that obviates the need for future follow-up.

llorca:
@cmslewis I would love to fix it and would be willing to do it before merging this, I just couldn't figure out a good solution (i.e. w/o nasty 2px margins here and there). If you have 10mins, mind taking a look? Happy to pair with you!

llorca:
Thanks to @cmslewis, we no longer have alignment issues 🎉

packages/core/src/components/forms/_form-group.scss
```diff
@@ -86,6 +119,12 @@ Styleguide pt-form-group
       margin: 0 $pt-grid-size 0 0;
       line-height: $pt-input-height;
     }
+
+    // not ideal but baseline doesn't align text perfectly
+    // so we have to align it to the label manually
+    .pt-control {
+      margin-top: $pt-grid-size * 1.2;
```
giladgray: ❌ absolutely not. try removing `.pt-large` from the example and it breaks...
image

gonna have to do this in a smarter way. perhaps make the height equal to `$input-height[-large]` (same as label)? will take some work to get this right...

packages/core/src/components/forms/_form-group.scss
```diff
@@ -46,19 +60,38 @@ Markup:
...
+<div class="pt-form-group pt-inline pt-large pt-disabled">
+  <label class="pt-label" for="example-form-group-switch-b">
+    Label B
+    <span class="pt-text-muted">(required)</span>
+  </label>
+  <div class="pt-form-content">
+    <label class="pt-control pt-switch">
+      <input id="example-form-group-switch-b" type="checkbox" />
+      <span class="pt-control-indicator"></span>
+      Switch B
```
giladgray: use the correct next letter please. A B C D E

cmslewis:
Ah yeah, @llorca, definitely add `margin-top: 1.2 * $pt-grid-size` only for `.pt-large`. The margin in the default case should be smaller (looking specifically at the `pt-switch`.

packages/core/src/components/forms/_form-group.scss
llorca: That didn't work, the `.pt-control` markup doesn't vertically centers. I went back to `baseline`, it's literally the right tool for the job, I suspect the alignment problems are browser issues, specifically regarding text baseline _within_ an input. I can live with that if you guys can too, check out the new preview

PR #1343, opened 2017-07-13
[Table] Frozen Rows/Cols - Part 2: Style Fixes & Example Controls

cmslewis:
#### Checklist

- [x] ~Include tests~ (will start doing this in part 3)

- Can now configure the number of frozen rows and columns from the example page.
    - Num. frozen rows: `[0, 1, 2, 5, 20, 100, 1000]`
    - Num. frozen columns: `[0, 1, 2, 5, 20, 100, 1000]`
- Headers are now fixed AND __scrollable__:
      
    - Row headers are now contained in the `LEFT` quadrant, leaving them fixed during scrolling.
    - Column headers are now contained in the `TOP` quadrant, leaving them fixed during scrolling.
    - The top-left "menu" element is now contained in the `TOP_LEFT` quadrant, leaving it fixed and above everything else.
- Toggling `isRowHeaderShown` properly resizes and repositions quadrants.

- Make sure the quadrants resize properly when:
    - Changing `Num. frozen columns`
    - Changing `Num. frozen rows`
    - Changing `Num. columns`
    - Changing `Num. rows`
    - Toggling row headers

- Known bugs that I'll fix in future PRs:
    - There currently is __NO__ handling for the case when `numFrozenColumns > numColumns` or `numFrozenRows > numRows`.
    - ~Header-border styles are broken if `numFrozen{Rows,Columns} === 0` and you scroll from within the headers~ __FIXED__
    - Selection still broken
    - Resizing still broken
    - Reordering still broken
    - Others?

### DIFF
--- a/packages/table/src/quadrants/tableQuadrant.tsx
+++ b/packages/table/src/quadrants/tableQuadrant.tsx
@@ -43,6 +43,12 @@ export interface ITableQuadrantProps extends IProps {
     bodyRef?: React.Ref<HTMLElement>;
 
     /**
+     * If `false`, hides the row headers and settings menu.
+     * @default true
+     */
+    isRowHeaderShown?: boolean;
+
+    /**
      * An optional callback invoked when the quadrant is scrolled via the scrollbar OR the trackpad/mouse wheel.
      * This callback really only makes sense for the MAIN quadrant, because that's the only quadrant whose
      * scrollbar is visible. Other quadrants should simply provide an `onWheel` callback.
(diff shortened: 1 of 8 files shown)


packages/table/src/quadrants/tableQuadrant.tsx
```diff
@@ -99,14 +105,23 @@ export interface ITableQuadrantProps extends IProps {
 }
 
 export class TableQuadrant extends AbstractComponent<ITableQuadrantProps, {}> {
+    // we want the user to explicitly pass a quadrantType. define defaultProps as a Partial to avoid
+    // declaring that and other required props here.
+    public static defaultProps: Partial<ITableQuadrantProps> & object = {
+        isRowHeaderShown: true,
```
themadcreator: since table quadrants can contain row headers and column headers, should we have both here?

themadcreator:
row size gets out of sync when you scroll because the farther you scroll down, the numbers require more space and make the labels wider:
quadrant-sync

packages/table/src/quadrants/tableQuadrant.tsx
cmslewis: I would, but there's no `isColumnHeaderShown` prop on `Table`, so we'd just need `isRowHeaderShown` here I believe.

cmslewis:
Eek, @themadcreator I'm actually seeing really weird scroll-jumping as you scroll down.

cmslewis:
Seems like this happens precisely when you scroll down far enough to trigger additional rows to render beyond the initial bleed region.

cmslewis:
@themadcreator the batcher is doing some funky stuff here (essentially, it adds two bleed cells after the current block, then removes them and adds two before the current block, scrolling the content in the process). Would love to chat with you today to dig in a bit:

packages/table/src/table.tsx
```diff
@@ -670,6 +682,22 @@ export class Table extends AbstractComponent<ITableProps, ITableState> {
         });
     }
 
+    // Quadrant refs
+    // =============
+
+    private generateQuadrantRefHandlers(quadrantType: QuadrantType): IQuadrantRefHandlers {
```
gscshoyru: This is pretty funky. Can we do this either iteratively over the four items, or... something else? We're sortof in the wrong place between re-useable code and not, here.

cmslewis:
@themadcreator I fixed the scroll issues! Can you take another look?

packages/table/src/table.tsx
cmslewis: Done 👍

cmslewis:
:shipit:

PR #1344, opened 2017-07-13
[TruncatedFormat] Fix popover size issues leading to infinite loops

### LINKED ISSUE #1335: Stack Overflows in table at 90% zoom
### Bug report

- __Package version(s)__: 1.16
- __Browser and OS versions__: Mac, Chrome 

#### Steps to reproduce
At 90% zoom seeing stack overflows on certain tables 

Context: https://pl.ntr/fG

gscshoyru:

Add a bit of buffer space to the calculation -- if there's a fractional amount of pixels off, we err on the side of not showing the popover, to deal with inconsistencies when at different zoom levels. 

Also only change the popover visibility when set to WHEN_TRUNCATED.

### DIFF
--- a/packages/table/src/cell/formats/truncatedFormat.tsx
+++ b/packages/table/src/cell/formats/truncatedFormat.tsx
@@ -157,12 +157,15 @@ export class TruncatedFormat extends React.Component<ITruncatedFormatProps, ITru
     }
 
     private setTruncationState() {
-        if (!this.props.detectTruncation) {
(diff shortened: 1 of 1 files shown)


tgreenwatts:
@cmslewis

cmslewis:
@gscshoyru can you post a GIF of Before/After demonstrating the problem, or is it difficult to repro?

gscshoyru:
@cmslewis it's not easy to repro. And the before causes chrome to freeze.

tgreenwatts:
Let's :shipit: !

PR #1347, opened 2017-07-14
Add max_line_length for editor configs

llorca:
Useful for tools editor side like prettier


.gitignore
```diff
@@ -16,7 +16,7 @@ lerna-debug.log
 .gradle/
 .node/
```
llorca: Side question: are these two lines still useful?

PR #1357, opened 2017-07-18
[Table] Show menu icon when menu is open in column header

### LINKED ISSUE #1297: Menu icon should always show when menu popover is open
### Bug report

- __Package version(s)__: `@blueprintjs/table` 1.18.0
- __Browser and OS versions__: macOS Sierra, latest Chrome

#### Steps to reproduce

1. Hover a column header
2. Click on menu icon to see menu popover
3. Hover away

#### Actual behavior

Menu is open but icon is gone:
image

#### Expected behavior

Icon always show when the menu is open.

llorca:
#### Screenshot

screen shot 2017-07-18 at 12 17 25 pm
screen shot 2017-07-18 at 12 13 43 pm

### DIFF
--- a/packages/table/src/headers/columnHeaderCell.tsx
+++ b/packages/table/src/headers/columnHeaderCell.tsx
@@ -136,7 +136,7 @@ export class ColumnHeaderCell extends AbstractComponent<IColumnHeaderCellProps, 
     }
 
     protected validateProps(nextProps: IColumnHeaderCellProps) {
-        if (nextProps != null && nextProps.menu != null) {
+        if (nextProps.menu != null) {
             // throw this warning from the publicly exported, higher-order *HeaderCell components
             // rather than HeaderCell, so consumers know exactly which components are receiving the
             // offending prop
--- a/packages/table/src/headers/columnHeaderCell.tsx
+++ b/packages/table/src/headers/columnHeaderCell.tsx
@@ -73,11 +73,15 @@ export interface IColumnHeaderCellProps extends IHeaderCellProps, IColumnNamePro
     menuIconName?: string;
 }
 
-export function HorizontalCellDivider(): JSX.Element {
-    return <div className={Classes.TABLE_HORIZONTAL_CELL_DIVIDER}/>;
+export interface IColumnHeaderCellState {
+    isActive?: boolean;
 }
 
-export class ColumnHeaderCell extends AbstractComponent<IColumnHeaderCellProps, {}> {
+export function HorizontalCellDivider(): JSX.Element {
+    return <div className={Classes.TABLE_HORIZONTAL_CELL_DIVIDER} />;
+}
+
+export class ColumnHeaderCell extends AbstractComponent<IColumnHeaderCellProps, IColumnHeaderCellState> {
     public static defaultProps: IColumnHeaderCellProps = {
         isActive: false,
         menuIconName: "chevron-down",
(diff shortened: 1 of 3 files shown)


cmslewis:
I noticed a style bug. Let's change the background color to `$header-selected-background-color` by default and `$header-hover-background-color` on hover.

packages/table/src/headers/columnHeaderCell.tsx
```diff
@@ -98,6 +102,13 @@ export class ColumnHeaderCell extends AbstractComponent<IColumnHeaderCellProps,
             || target.classList.contains(Classes.TABLE_HEADER_CONTENT);
     }
 
+    public constructor() {
+        super();
+        this.state = {
+            isActive: false,
+        };
+    }
```
giladgray: ❌ don't define constructor for static state, just do `public state = { isActive: false };` and be done with it. constructor is only needed if initial state depends on props.

giladgray: also this constructor implementation is incorrect: must accept `(props, context?)` params. 
but really just remove it.

packages/table/src/headers/columnHeaderCell.tsx
```diff
@@ -128,7 +139,7 @@ export class ColumnHeaderCell extends AbstractComponent<IColumnHeaderCellProps,
     }
 
     protected validateProps(nextProps: IColumnHeaderCellProps) {
-        if (nextProps.menu != null) {
+        if (nextProps != null && nextProps.menu != null) {
```
giladgray: why add this check? nextProps should never be null; there are always props.

packages/table/src/headers/columnHeaderCell.tsx
```diff
@@ -198,14 +209,17 @@ export class ColumnHeaderCell extends AbstractComponent<IColumnHeaderCellProps,
             pin: true,
             to: "window",
         }];
+        const classes = classNames(Classes.TABLE_TH_MENU_CONTAINER, {
+            [Classes.TABLE_TH_MENU_OPEN]: this.state.isActive,
```
giladgray: when does `this.state.isActive` change? i don't see any `setState`s in this PR...

packages/table/src/headers/columnHeaderCell.tsx
llorca: It was undefined (with console error) and this check fixed it. I'll investigate more tomorrow and get back to you

packages/table/src/headers/columnHeaderCell.tsx
llorca: The component was _setting_ its state already, even though there was not state _defined_--check the very bottom of the file outside of this PR. Oversight from previous work on this file?

packages/table/src/headers/columnHeaderCell.tsx
llorca: Will do.

packages/table/src/headers/columnHeaderCell.tsx
giladgray: oh wow ok 🆒

packages/table/src/headers/columnHeaderCell.tsx
llorca: Works now, might have been because of the constructor? I dunno.

llorca:
Fixed the visual bugs (nice catch @cmslewis, so much better now) and addressed comments 👍

llorca:
@giladgray ready for review!

PR #1359, opened 2017-07-18
[1.23.1] Apply form group baseline to inline form only

llorca:


adidahiya:
This doesn't look right; we don't want the icon changes and we shouldn't be seeing multiple package version bumps. Can you try this please:

- branch off the commit where we tagged 1.23.0. call that branch `release/1.23.1`.
- apply your baseline change https://github.com/palantir/blueprint/pull/1359/commits/33e4028c7d1246d78754482dc24b5a43f319cdbc

PR #1390, opened 2017-07-26
[Table] Frozen Rows/Cols - Part 6: Refactor quadrant logic

cmslewis:

- __Changed__ `Table` quadrant-sync logic now lives in a new component, `TableQuadrantStack`. 

How does this refactor look? Are any of the constructs or patterns weird? Any typos etc?

### DIFF
--- a/packages/table/src/quadrants/tableQuadrantStack.tsx
+++ b/packages/table/src/quadrants/tableQuadrantStack.tsx
@@ -140,13 +140,13 @@ export class TableQuadrantStack extends AbstractComponent<ITableQuadrantProps, {
     private wasMainQuadrantScrollChangedFromOtherOnWheelCallback = false;
 
     public componentDidMount() {
-        this.emitMainQuadrantRefs();
+        this.emitRefs();
         this.syncQuadrantSizes();
         this.syncQuadrantMenuElementWidths();
     }
 
     public componentDidUpdate() {
-        this.emitMainQuadrantRefs();
+        this.emitRefs();
         this.syncQuadrantSizes();
         this.syncQuadrantMenuElementWidths();
     }
@@ -351,7 +351,7 @@ export class TableQuadrantStack extends AbstractComponent<ITableQuadrantProps, {
     // Emitters
     // ========
 
-    private emitMainQuadrantRefs() {
+    private emitRefs() {
         CoreUtils.safeInvoke(this.props.columnHeaderRef, this.quadrantRefs[QuadrantType.MAIN].columnHeader);
         CoreUtils.safeInvoke(this.props.quadrantRef, this.quadrantRefs[QuadrantType.MAIN].quadrant);
         CoreUtils.safeInvoke(this.props.rowHeaderRef, this.quadrantRefs[QuadrantType.MAIN].rowHeader);
--- a/packages/table/src/quadrants/tableQuadrantStack.tsx
+++ b/packages/table/src/quadrants/tableQuadrantStack.tsx
@@ -152,13 +152,7 @@ export class TableQuadrantStack extends AbstractComponent<ITableQuadrantProps, {
     }
 
     public render() {
-        const {
-            grid,
-            isRowHeaderShown,
-            renderBody,
-            renderColumnHeader,
-        } = this.props;
-
+        const { grid, isRowHeaderShown, renderBody } = this.props;
         return (
             <div>
                 <TableQuadrant
@@ -452,18 +446,5 @@ export class TableQuadrantStack extends AbstractComponent<ITableQuadrantProps, {
         // this child element dictates the width of all row-header cells
         const elementToResize = rowHeaderElement.querySelector(selector) as HTMLElement;
         elementToResize.style.width = `${width}px`;
-    }
-
-    // Helpers
-    // =======
-
-    private findColumnHeader(quadrantType: QuadrantType) {
-        const quadrantElement = this.quadrantRefs[quadrantType].quadrant;
-        return quadrantElement.querySelector(`.${Classes.TABLE_COLUMN_HEADERS}`) as HTMLElement;
-    }
-
-    private findRowHeader(quadrantType: QuadrantType) {
-        const quadrantElement = this.quadrantRefs[quadrantType].quadrant;
-        return quadrantElement.querySelector(`.${Classes.TABLE_ROW_HEADERS}`) as HTMLElement;
     }
 }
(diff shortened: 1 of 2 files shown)


packages/table/src/quadrants/tableQuadrantStack.tsx
```diff
@@ -0,0 +1,438 @@
...
+}
+
+export class TableQuadrantStack extends AbstractComponent<ITableQuadrantProps, {}> {
+    // we want the user to explicitly pass a quadrantType. define defaultProps as a Partial to avoid
+    // declaring that and other required props here.
+    public static defaultProps: Partial<ITableQuadrantProps> & object = {
+        isHorizontalScrollDisabled: false,
+        isRowHeaderShown: true,
+        isVerticalScrollDisabled: false,
+    };
```
cmslewis: Wasn't quite sure how to set default flag values without relying on the `Partial` construct here. (If I don't use `Partial`, then I have to supply defined values for all of the required props.)

packages/table/src/quadrants/tableQuadrantStack.tsx
```diff
@@ -0,0 +1,438 @@
...
+    private generateQuadrantRefHandlers(quadrantType: QuadrantType): IQuadrantRefHandlers {
+        const reducer = (agg: IQuadrantRefHandlers, key: keyof IQuadrantRefHandlers) => {
+            agg[key] = (ref: HTMLElement) => {
+                this.quadrantRefs[quadrantType][key] = ref;
+                if (quadrantType === QuadrantType.MAIN) {
+                    if (key === "quadrant") {
+                        CoreUtils.safeInvoke(this.props.quadrantRef, ref);
+                    } else if (key === "scrollContainer") {
+                        CoreUtils.safeInvoke(this.props.scrollContainerRef, ref);
+                    }
```
cmslewis: This area feels weird too. I want to keep a private `ref` to each of these things but _also_ emit the `ref` via the provided props. Is there a better way to do this while also keeping this generator function around? Do we need the generator function?

packages/table/src/quadrants/tableQuadrantStack.tsx
```diff
@@ -0,0 +1,438 @@
...
+            : BORDER_WIDTH_CORRECTION;
+        const topQuadrantGridContentHeight = numFrozenRows > 0
+            ? grid.getCumulativeHeightAt(numFrozenRows - 1)
+            : BORDER_WIDTH_CORRECTION;
+
+        // all menus are the same size, so arbitrarily use the one from the main quadrant.
+        // assumes that the menu element width has already been sync'd after the last render
+
+        const rowHeader = this.findRowHeader(QuadrantType.MAIN);
+        const columnHeader = this.findColumnHeader(QuadrantType.MAIN);
```
cmslewis: Is it too expensive to continually `find` these elements whenever we need to sync quadrant sizes? Doesn't happen too often, so I figured the minor performance hit wouldn't add up in any meaningful way.

packages/table/src/quadrants/tableQuadrantStack.tsx
giladgray: yeah `Partial` is legit, but I don't feel a strong need for the `& object` cuz you assign an object to it immediately -- that guard is more useful for props to enforce a base type

packages/table/src/quadrants/tableQuadrantStack.tsx
```diff
@@ -0,0 +1,438 @@
...
+        isHorizontalScrollDisabled: false,
+        isRowHeaderShown: true,
+        isVerticalScrollDisabled: false,
+    };
+
+    private quadrantRefs = {
+        [QuadrantType.MAIN]: {} as IQuadrantRefs,
+        [QuadrantType.TOP]: {} as IQuadrantRefs,
+        [QuadrantType.LEFT]: {} as IQuadrantRefs,
+        [QuadrantType.TOP_LEFT]: {} as IQuadrantRefs,
```
giladgray: TSlint@5 is gonna ban all these literal casts. can you do this some other way?

```ts
private quadrantRefs: Record<QuadrantType, IQuadrantRefs> = { ... }
```

packages/table/src/quadrants/tableQuadrantStack.tsx
giladgray: depends how often this component updates... if it's every scroll event then yes that's too expensive. if not, then we're probably okay. worth gathering some data.

packages/table/src/quadrants/tableQuadrantStack.tsx
cmslewis: Done

packages/table/src/quadrants/tableQuadrantStack.tsx
cmslewis: Refactored to use a ref.

packages/table/src/quadrants/tableQuadrantStack.tsx
cmslewis: `QuadrantType` needs to be a string: 
image

And can't do an interface, because having `QuadrantType` keys yields an error: `A computed property name in an interface must directly refer to a built-in symbol`.

packages/table/src/table.tsx
```diff
@@ -540,56 +526,28 @@ export class Table extends AbstractComponent<ITableProps, ITableState> {
                 ref={this.setRootTableRef}
                 onScroll={this.handleRootScroll}
             >
-                <TableQuadrant
+                <TableQuadrantStack
```
gscshoyru: Is there a better name than this? I can't think of one, but this is a little silly.

packages/table/src/quadrants/tableQuadrantStack.tsx
cmslewis: Maybe string enums to the rescue?

packages/table/src/table.tsx
cmslewis: I don't think it's so bad—it's an ordered stack of quadrants! Other options:

`TableQuadrants`
`TableQuadrant...Chimney`

image

packages/table/src/quadrants/tableQuadrantStack.tsx
cmslewis: Nope:

```tsx
export enum QuadrantType {
    /**
     * The main quadrant beneath any frozen rows or columns.
     */
    MAIN = "MAIN",

    /**
     * The top quadrant, containing column headers and frozen rows.
     */
    TOP = "TOP",

    /**
     * The left quadrant, containing row headers and frozen columns.
     */
    LEFT = "LEFT",

    /**
     * The top-left quadrant, containing the headers and cells common to both the frozen columns and
     * frozen rows.
     */
    TOP_LEFT = "TOP_LEFT",
}
```

```tsx
[ts]
Type '{ [x: string]: {}; }' is not assignable to type 'Record<QuadrantType, IQuadrantRefMap<HTMLElement>>'.
  Property 'MAIN' is missing in type '{ [x: string]: {}; }'.
-------------------
import QuadrantType
```

cmslewis: Gonna merge this for now and figure something out later. Maybe I'll just use string literals for `QuadrantType`.

PR #1391, opened 2017-07-26
[Button] Remove small button icon restriction

### LINKED ISSUE #1153: pt-small hides icons for buttons with no content
Here is a button group I have:
image

Here is what it looks like if I add `pt-small` to the container with class `pt-button-group`:
image

Perhaps the icon should not be hidden if there is no text?

llorca:
#### Fixes #1153 

Getting too many requests about this. Two solutions:
1. we lift the restriction
1. we create a whitelist of icons that are okay inside small buttons

This PR does the former. I really dislike the latter, and it seems to make sense to push this responsibility to consumers anyway.

### DIFF
--- a/packages/core/src/components/button/_button.scss
+++ b/packages/core/src/components/button/_button.scss
@@ -204,12 +204,6 @@ Styleguide pt-button
   &.pt-small,
   .pt-small & {
     @include pt-button-small();
-
-    // small buttons are too small to contain any Blueprint icons
-    &[class*="pt-icon-"]::before,
-    #{$icon-classes} {
-      display: none;
-    }
   }
 
   // ensure icon button with no text is a perfect square


cmslewis:
Anything we can do to get the icon vertically centered by default in small buttons?

image

llorca:
@cmslewis not with the current architecture. Fixable in 3 lines:
```css.sass
display: inline-flex;
align-items: center;
justify-contents: center;
```
Probably better as a separate PR since moving to flex buttons is a significant change. What do you guys think?

giladgray:
@llorca FF does not support `display: flex` on `<button>` tags, so that is not an actual solution. see internal PR 963 from July 2016.

giladgray:
can we adjust padding a little bit? cuz the button doesn't look centered relative to the text (aside from position within button itself)

llorca:
@giladgray I just tested flex positioning inside buttons on Firefox, works fine. Looks like it was fixed in version 52: https://bugzilla.mozilla.org/show_bug.cgi?id=984869#c37

Text is aligned within the button, and on the same baseline as text outside of it:
image

It's really the icon that's misaligned and needs to be adjusted, unfortunately adding some padding does not work.

giladgray:
nice good find with the bug fix!

can the centering be done with margin? anything easy we can do? if not then w/e

llorca:
I tried `margin`, `padding` and all the classic stuff, didn't work. Looks like it's either whatever or move to `inline-flex` 🙃

pkwi:
Agree with @llorca @cmslewis here. Let's simply lift the restriction and give users the choice.

PR #1392, opened 2017-07-27
[Button] Refactor button gradients

llorca:

- Remove `linear-gradient-with-fallback` mixin and rely on straight CSS
  - Good browser support: http://caniuse.com/#search=linear-gradient)
- Gradients on buttons are `!default` variables that can be customized like all the other ones

### DIFF
--- a/packages/core/src/components/button/_common.scss
+++ b/packages/core/src/components/button/_common.scss
@@ -7,8 +7,6 @@
 @import "../../common/variables";
 @import "../progress/common";
 
-$white-transparent: rgba($white, 0) !default;
-
 $button-border-width: 1px !default;
 $button-padding: 0 $pt-grid-size !default;
 $button-padding-small: 0 ($pt-grid-size * 0.7) !default;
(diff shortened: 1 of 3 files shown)


packages/core/src/components/button/_common.scss
```diff
@@ -178,12 +172,9 @@ $button-intents: (
...
 
   &:hover {
-    @include linear-gradient-with-fallback(
-      rgba($white, 0.1),
-      $white-transparent,
-      $hover-color
-    );
     box-shadow: $button-intent-box-shadow;
+    background-color: $hover-color;
+    background-image: $button-intent-gradient;
```
giladgray: do you need to repeat same `background-image` in pseudo states like `:hover` (compare to line 165)? 
in devtools, it seems safe to remove.

this comment applies several times throughout the changed files, i'm just gonna leave this one comment for them all.

llorca: Yeah, looks fine

pkwi:
I'm not sure about the new flatter buttons. Any chance we can reproduce the old gradient style?

llorca:
@pkwi that's just a refactor, not a change. Buttons didn't get flatter :)

llorca:
With 70% opacity for `$button-gradient`, here's what we get:
screen shot 2017-07-27 at 12 42 23 pm

Compare to before:
screen shot 2017-07-27 at 12 42 18 pm

Seems pretty legit, gonna push a commit now and wait for @pkwi confirmation tomorrow.

pkwi:
80% looks good to me.

giladgray:
menus and inputs don't have `background` gradients, they have `box-shadow`s. so we're 🆒

omeid:
A bit late to the party, but perhaps the `$button-gradient`, `$button-intent-gradient`, and `$dark-button-gradient` could be called `-background` instead of `-gradient`?

giladgray:
`-gradient` is accurate because it is used as `background-image` (one of many `background` sub-properties) and the value is actually a gradient, not any of the other valid background values.

We name our sass variables for their type: `-color`, `-size`, etc. So `-gradient` is consistent, but `-background` is not.

omeid:
Right, that is a good point! I overlooked that there was also `background-color`, perhaps `-background-image`? to along with `-button-background-color`? It just that `gradient` describes the content of the variable (the details about the gradient) instead of _what_, which is a background.

PR #1416, opened 2017-08-07
[Table] Frozen Rows/Cols - Part 7: Scroll Performance

cmslewis:

- __Fixed__ `Table` quadrants now stay (almost) perfectly in sync as you scroll, and scrolling is fluid!

    - I say "almost," because there's a known, momentary scroll-sync bug when you wheel with force off the right side of the `MAIN` quadrant:

 

- How snazzy throttling via `requestAnimationFrame` can be.
- The usual CR stuff.

### DIFF
--- a/packages/core/src/common/utils.ts
+++ b/packages/core/src/common/utils.ts
@@ -91,18 +91,15 @@ export function countDecimalPlaces(num: number) {
  * @see https://developer.mozilla.org/en-US/docs/Web/Events/scroll
  */
 export function throttleEvent(target: EventTarget, eventName: string, newEventName: string) {
-    let running = false;
-    /* istanbul ignore next: borrowed directly from MDN */
-    const func = (event: Event) => {
-        if (running) { return; }
-        running = true;
-        requestAnimationFrame(() => {
+    const throttledFunc = throttleHelper(
+        undefined,
+        undefined,
+        (event: Event) => {
             target.dispatchEvent(new CustomEvent(newEventName, event));
-            running = false;
-        });
-    };
-    target.addEventListener(eventName, func);
-    return func;
+        },
+    );
+    target.addEventListener(eventName, throttledFunc);
+    return throttledFunc;
 };
 
 export interface IThrottledReactEventOptions {
@@ -118,24 +115,43 @@ export function throttleReactEventCallback(
     callback: (event: React.SyntheticEvent<any>, ...otherArgs: any[]) => any,
     options: IThrottledReactEventOptions = {},
 ) {
+    const throttledFunc = throttleHelper(
+        (event2: React.SyntheticEvent<any>) => {
+            if (options.preventDefault) {
+                event2.preventDefault();
+            }
+        },
+        (event2: React.SyntheticEvent<any>) => {
+            // prevent React from reclaiming the event object before we reference it
+            event2.persist();
+        },
+        (event2: React.SyntheticEvent<any>, ...otherArgs2: any[]) => {
+            callback(event2, ...otherArgs2);
+        },
+    );
+    return throttledFunc;
+}
+
+type ThrottleHelperCallback = (...args: any[]) => void;
+
+function throttleHelper(
+    onBeforeIsRunningCheck: ThrottleHelperCallback,
+    onAfterIsRunningCheck: ThrottleHelperCallback,
+    onAnimationFrameRequested: ThrottleHelperCallback,
+) {
     let isRunning = false;
-    const func = (event2: React.SyntheticEvent<any>, ...otherArgs2: any[]) => {
-        // need to preventDefault synchronously and without regard to the rAF timeline,
-        // if we want it to work
-        if (options.preventDefault) {
-            event2.preventDefault();
-        }
+    const func = (...args: any[]) => {
+        safeInvoke(onBeforeIsRunningCheck, ...args);
 
         if (isRunning) {
             return;
         }
         isRunning = true;
 
-        // prevent React from reclaiming the event object before we reference it
-        event2.persist();
+        safeInvoke(onAfterIsRunningCheck, ...args);
 
         requestAnimationFrame(() => {
-            callback(event2, ...otherArgs2);
+            safeInvoke(onAnimationFrameRequested, ...args);
             isRunning = false;
         });
     };
--- a/packages/core/src/common/utils.ts
+++ b/packages/core/src/common/utils.ts
@@ -105,6 +105,10 @@ export function throttleEvent(target: EventTarget, eventName: string, newEventNa
     return func;
 };
 
+export interface IThrottledReactEventOptions {
+    preventDefault?: boolean;
+};
+
 /**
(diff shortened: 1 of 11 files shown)


packages/table/src/table.tsx
```diff
@@ -790,16 +790,24 @@ export class Table extends AbstractComponent<ITableProps, ITableState> {
...
             ? Math.max(0, tableBottom - viewportRect.height)
             : viewportRect.top;
 
-        const nextScrollLeft = (tableRight < viewportRect.left + viewportRect.width)
+        let nextScrollLeft = (tableRight < viewportRect.left + viewportRect.width)
             // scroll the last column into view
             ? Math.max(0, tableRight - viewportRect.width)
             : viewportRect.left;
 
+        // reset the scroll position when scrolling is disabled
```
cmslewis: Gah, meant to revert this area of changes. This doesn't actually work (was trying to scroll to `(0,0)` if the scroll _became_ disabled (by reducing row/col sizes or by reducing the number of either). 

I'll investigate how to do this properly now, actually.

llorca:
🔥 🔥 🔥

packages/core/src/common/utils.ts
```diff
@@ -104,3 +104,36 @@ export function throttleEvent(target: EventTarget, eventName: string, newEventNa
...
 };
+
+/**
+ * Throttle a callback by wrapping it in a `requestAnimationFrame` call. Returns the throttled
+ * function.
+ * @see https://www.html5rocks.com/en/tutorials/speed/animations/
+ */
+export function throttleReactEventCallback(
+    callback: (event: React.SyntheticEvent<any>, ...otherArgs: any[]) => any,
+    preventDefault?: boolean,
```
cmslewis: Another option here is `{ preventDefault?: boolean }?: IThrottledReactEventOptions`. That lets us explicitly write `{ preventDefault: true }` when invoking this function (right now, I'm adding an inline comment like `/* preventDefault */ true`.

packages/table/test/tableTests.tsx
```diff
@@ -329,52 +329,64 @@ describe("<Table>", () => {
...
                 // scrollTop/scrollLeft the way it would in practice, so we need to tweak those
                 // explicitly before triggering.
                 mainScrollContainer.scrollLeft = SCROLL_OFFSET_X;
                 mainScrollContainer.scrollTop = SCROLL_OFFSET_Y;
                 TestUtils.Simulate.scroll(mainScrollContainer);
 
-                assertScrollPositionEquals(topScrollContainer, SCROLL_OFFSET_X, 0);
-                assertScrollPositionEquals(leftScrollContainer, 0, SCROLL_OFFSET_Y);
-                assertScrollPositionEquals(topLeftScrollContainer, 0, 0);
+                requestAnimationFrame(() => {
```
cmslewis: Had to get a little cute here putting the tests themselves in `rAF`, to give the scroll logic a chance to execute.

packages/core/src/common/utils.ts
adidahiya: yeah I like the object options approach better.

cmslewis: It also makes the lines the _exact_ same length: 🤤

image

packages/core/src/common/utils.ts
```diff
@@ -104,3 +104,40 @@ export function throttleEvent(target: EventTarget, eventName: string, newEventNa
...
+        // prevent React from reclaiming the event object before we reference it
+        event2.persist();
+
+        requestAnimationFrame(() => {
+            callback(event2, ...otherArgs2);
+            isRunning = false;
+        });
+    };
+    return func;
+}
```
giladgray: hmmm would be cool for the two throttle util functions here to share _some_ (if not most) of their code...

packages/table/test/tableTests.tsx
```diff
@@ -329,55 +329,81 @@ describe("<Table>", () => {
...
+                });
             });
         });
 
+        // Clock management
+        // ================
+
+        function delayToNextAnimationFrame(callback: () => void) {
+            // two rAFs are required to force the callback to execute on the *next* frame, allowing
+            // any initial rAF'd code to execute
```
giladgray: or one `setTimeout`.... why not use the built-in method that guarantees next-frame-execution?

giladgray: and perhaps call it simply `delayToNextFrame`?

packages/core/src/common/utils.ts
cmslewis: Think I managed it:

```tsx

type ThrottleHelperCallback = (...args: any[]) => void;

function throttleHelper(
    onBeforeIsRunningCheck: ThrottleHelperCallback,
    onAfterIsRunningCheck: ThrottleHelperCallback,
    onAnimationFrameRequested: ThrottleHelperCallback,
) {
    let isRunning = false;
    const func = (...args: any[]) => {
        safeInvoke(onBeforeIsRunningCheck, ...args);

        if (isRunning) {
            return;
        }
        isRunning = true;

        safeInvoke(onAfterIsRunningCheck, ...args);

        requestAnimationFrame(() => {
            safeInvoke(onAnimationFrameRequested, ...args);
            isRunning = false;
        });
    };
    return func;
}
```

packages/table/test/tableTests.tsx
cmslewis: Done

packages/core/src/common/utils.ts
```diff
@@ -86,21 +86,81 @@ export function countDecimalPlaces(num: number) {
...
+ * Throttle a callback by wrapping it in a `requestAnimationFrame` call. Returns the throttled
+ * function.
+ * @see https://www.html5rocks.com/en/tutorials/speed/animations/
+ */
+export function throttleReactEventCallback(
+    callback: (event: React.SyntheticEvent<any>, ...otherArgs: any[]) => any,
+    options: IThrottledReactEventOptions = {},
+) {
+    const throttledFunc = throttleHelper(
+        (event2: React.SyntheticEvent<any>) => {
```
gscshoyru: Why are all the things in here named with 2's?

adidahiya:
this scrollbar position is unfortunate, I wonder if we can do something about it in another PR. people might not realize that the scroll bar handle is hidden. I also think it's fine to ship even with this defect though.

image

cmslewis:
@adidahiya fortunately, it's okay when scroll bars are visible (i.e. when you have a mouse plugged in):

image

not sure how to make it the `TOP` quadrant be flush with the right edge _and_ have the `MAIN` quadrant scrollbar visible on top though. :(

packages/core/src/common/utils.ts
cmslewis: There's already an `(event: React.SyntheticEvent<any>, ...otherArgs: any[]) => any` specified in the callback type. Felt this would be clearer.

adidahiya:
oh that's good news! I'm no longer worried about it then

PR #1439, opened 2017-08-11
[Table] Frozen Rows/Cols - Part 8: Fix tests

cmslewis:
#### Checklist

- [x] Include tests

- Fix tests that I broke and `skip`'d in past PRs.
    - In particular, be sure to `delayToNextFrame` when testing anything to do with scrolling, which is now throttled via `requestAnimationFrame`.
    - Also, test the `MAIN` quadrant's scroll container by default.

- ~HOW DO I FIX THE BROKEN `TruncatedFormat` test? Can't figure it out.~ Fixed with `position: relative;`. :(
    - https://github.com/palantir/blueprint/blob/cl/table-frozen-rows-cols-8-fix-tests/packages/table/test/formatsTests.tsx#L55

### DIFF
--- a/packages/table/test/formatsTests.tsx
+++ b/packages/table/test/formatsTests.tsx
@@ -51,8 +51,10 @@ describe("Formats", () => {
             expect(comp.find(`.${Classes.TABLE_TRUNCATED_POPOVER_TARGET}`).element).to.exist;
         });
 
-        // TODO: FROZEN
-        it.skip("can automatically truncate and show popover when truncated and word wrapped", () => {
+        // This test was super flaky. It started failing without clear cause when the Table Frozen
+        // Columns/Rows changes merged, even though nothing about the TruncatedFormat component
+        // changed. Adding the position: relative rule fixes it, but more investigation is needed.
+        it("can automatically truncate and show popover when truncated and word wrapped", () => {
             const str = `
                 We are going to die, and that makes us the lucky ones. Most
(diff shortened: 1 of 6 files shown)


cmslewis:
Ughhh and now I'm seeing the same test flaking out on `master`:
image

packages/table/test/formatsTests.tsx
```diff
@@ -69,7 +71,7 @@ describe("Formats", () => {
                 majority have never stirred?
             `;
 
-            const style = { height: "200px" };
+            const style = { height: "200px", position: "relative" };
```
themadcreator: i dont entirely understand why this would have broken or why this change fixes it but it's possible some deeply nested SCSS rule became invalid with the new DOM hierarchy.

if so, we should add this css rule to the TABLE_TRUNCATED_TEXT class instead of just the style on the test

cmslewis: > if so, we should add this css rule to the TABLE_TRUNCATED_TEXT class instead of just the style on the test

that's what i was wondering. lemme test it out locally.

cmslewis: NOPE. That would be bad.



@themadcreator would you prefer I find the exact reason, or just leave the fix here, merge this PR, and revisit? I'm leaning toward the former, but if I add `position: relative`, quite a few downstream CSS fixes are required too. Might be out of scope for now.

PR #1458, opened 2017-08-16
[Icon] Fix pt-icon-bank-account className

### LINKED ISSUE #1455: Fix typo in pt-icon-bank-account icon
#### Actual behavior

Typo in the name of the icon in the documentation. It says `pt-icon-banl-account`. 

It is correct in svg name of the icon in assets folder.

screenshot 2017-08-16 13 38 10

#### Expected behavior

`pt-icon-bank-account`

llorca:
#### Fixes #1455

(Technically a breaking change, but we think it'd be silly to wait for 2.0 for such a small correction)

### DIFF
--- a/packages/core/resources/icons/icons.json
+++ b/packages/core/resources/icons/icons.json
@@ -2773,7 +2773,7 @@
   },
   {
     "name": "Bank account",
-    "className": "pt-icon-banl-account",
+    "className": "pt-icon-bank-account",
     "tags": "money, finance, funds",
     "group": "miscellaneous",
     "content": "\\e76f"


cmslewis:
@llorca wait, does this mean people had to write stuff like `<a class="pt-button pt-icon-banl-account">` before now? If so, isn't this PR effectively breaking, even though it's a correction?

llorca:
Yeah that's probably true, but I doubt anybody would've used it without calling out the error. In any case, that fix seems fine to me -- don't think it'd be appropriate to wait for 2.0 for such a thing?
image

cmslewis:
Yeah waiting to 2.0 seems like overkill. Just wanted to call it out.

cmslewis:
Let's get one more opinion from @adidahiya or @giladgray before we merge.

PR #1463, opened 2017-08-17
Prepare Release v1.25.1

cmslewis:
## `@blueprintjs/core 1.25.1`
- __Fixed__ Fix user-land SCSS compilation error (`File to import not found or unreadable: common/icons.`)


cmslewis:
@adidahiya how do you want to go about resolving merge conflicts here? Would it be easier to just include the yarn stuff in this patch release?

adidahiya:
you shouldn't be merging this release into `master`; you should just cut from `release/1.25.x`

PR #1456, opened 2017-08-16
[Core] v1.25.0 regression fix for user-land sass compilation

bbenezech:
Without this I get Sass error `File to import not found or unreadable: common/icons.`

You may want to push a patch release for this, this is a regression from previous behaviour.

PR #1467, opened 2017-08-18
Prepare release 1.25.1

adidahiya:
### 🐛  Bug fixes

- #1456 Fix Sass compilation for users who were importing Blueprint's `.scss` files
- #1458 Fixed `pt-icon-bank-account` class name (note that this is a very minor breaking change&mdash;previously this had a typo and was called `pt-icon-banl-account`).


adidahiya:
CI failure looks like a test flake.
On a related note, we should migrate from phantomjs to https://github.com/graphcool/chromeless.

adidahiya:
docs preview lgtm.

PR #1479, opened 2017-08-21
[Table] Fix Table sizing when shown in flex parent

### LINKED ISSUE #1472: [Table] As of 1.1.22 Table no longer expands to fit contents when inside a flex container
On Blueprint 1.1.18 the table can grow to the height of it's contents, when inside a flex container.

On Blueprint 1.1.22 the table does not expand to fit it's contents when inside a flex container, and is instead always about 30px high.

cmslewis:
#### Checklist

- [x] Include tests

- BEFORE: All quadrants were `position: absolute`, meaning the surrounding table container had no idea how big to make itself, so it collapsed down to its `min-height` of `60px` in containers that didn't have a fixed size.  

- AFTER: Keep the `MAIN`-quadrant content in the document flow to ensure the table sizes itself appropriately.

- CSS nuances. Am I using `flex` appropriately?
- I've found the _Features (Legacy)_ page in the table preview helpful for testing this!
- Please test the dev preview thoroughly to ensure nothing looks broken! (e.g. scrollbars should still not be visible in non`MAIN` quadrants)

### DIFF
--- a/packages/table/src/quadrants/_quadrants.scss
+++ b/packages/table/src/quadrants/_quadrants.scss
@@ -94,8 +94,10 @@ $table-quadrant-scroll-container-overflow: 20px;
 
   .bp-table-quadrant-scroll-container {
     position: absolute;
+    top: 0;
     right: -$table-quadrant-scroll-container-overflow;
-    height: 100%;
+    bottom: 0;
+    height: auto;
     overflow-x: hidden;
   }
 }
--- a/packages/table/src/quadrants/_quadrants.scss
+++ b/packages/table/src/quadrants/_quadrants.scss
@@ -12,6 +12,11 @@ $table-quadrant-z-index-top-left: $table-quadrant-z-index-left + 1;
 // overflow within their parents; it should be large enough to hide a scrollbar of typical width.
 $table-quadrant-scroll-container-overflow: 20px;
 
+.bp-table-quadrant-stack {
+  display: flex;
+  height: 100%;
+}
+
 .bp-table-quadrant {
   position: absolute;
   top: 0;
@@ -26,7 +31,8 @@ $table-quadrant-scroll-container-overflow: 20px;
 
 .bp-table-quadrant-scroll-container {
   @include force-hardware-acceleration();
-  position: absolute;
+  flex: 1 1 100%;
+  position: relative;
   top: 0;
   right: 0;
   bottom: 0;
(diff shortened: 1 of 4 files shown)


giladgray:
@cmslewis failed linting and scrolling unit tests... need a preview to properly review this.

packages/table/src/quadrants/_quadrants.scss
```diff
@@ -51,9 +56,22 @@ $table-quadrant-scroll-container-overflow: 20px;
 }
 
 .bp-table-quadrant-main {
-  right: 0;
-  bottom: 0;
+  // use `relative` to ensure the table can size itself according to the MAIN quadrant's contents alone.
+  // other quadrants should remain `absolute` to ensure they're taken out of the document flow.
```
themadcreator: 👍 this is usually my go-to solution for cutting my components out of the layout flow

packages/table/src/quadrants/_quadrants.scss
```diff
@@ -51,9 +56,22 @@ $table-quadrant-scroll-container-overflow: 20px;
...
+  position: relative;
+  top: auto;
+  left: auto;
   z-index: $table-quadrant-z-index-main;
+  // the MAIN quadrant should fill the whole table so that ghost cells will be visible if enabled
+  width: 100%;
+  height: 100%;
```
themadcreator: `height: 100%` can be tricky sometimes. just want to make sure this works correctly.

```html
<div>
    <div style="height: 100%">everything is cool</div>
    <div>what now?</div>
</div>
```

 this should be okay since we control the siblings of this div, but just something to keep an eye on.

packages/table/src/quadrants/_quadrants.scss
```diff
@@ -75,7 +93,9 @@ $table-quadrant-scroll-container-overflow: 20px;
   z-index: $table-quadrant-z-index-left;
 
   .bp-table-quadrant-scroll-container {
+    position: absolute;
     right: -$table-quadrant-scroll-container-overflow;
+    height: 100%;
```
themadcreator: i prefer to use `top: 0; bottom: 0;` with `position: absolute`

packages/table/test/tableTests.tsx
```diff
@@ -562,13 +566,15 @@ describe("<Table>", () => {
...
 
             const tableComponent = ReactDOM.render(
-                <Table
-                    numRows={LARGE_NUM_ROWS}
-                    numFrozenColumns={NUM_FROZEN_COLUMNS}
-                    numFrozenRows={NUM_FROZEN_ROWS}
-                >
-                    {renderColumns({}, LARGE_NUM_COLUMNS)}
-                </Table>,
+                <div style={{ height: CONTAINER_HEIGHT, width: CONTAINER_WIDTH }}>
```
themadcreator: so, what happens if users DON'T put the table in an explicitly sized div? can we never use normal flow?

cmslewis: That was the point of this PR actually: to fix that.

BEFORE (buggy):
image


AFTER (fluid height):
image

packages/table/src/quadrants/_quadrants.scss
cmslewis: Yep, I was cognizant of this. We're good as far as I can tell.

packages/table/src/quadrants/_quadrants.scss
```diff
@@ -94,8 +94,10 @@ $table-quadrant-scroll-container-overflow: 20px;
 
   .bp-table-quadrant-scroll-container {
     position: absolute;
+    top: 0;
     right: -$table-quadrant-scroll-container-overflow;
-    height: 100%;
+    bottom: 0;
+    height: auto;
```
themadcreator: not sure this is entirely necessary, but won't hurt

packages/table/test/tableTests.tsx
themadcreator: oh, excellent.

packages/table/src/quadrants/_quadrants.scss
cmslewis: Whenever I specify both `top` and `bottom`, I like to take `height` out of the picture so the two schemes don't compete unpredictably.

PR #1488, opened 2017-08-22
Prepare release 1.25.2

cmslewis:
### 🐛  Bug fixes

- #1479 Fix `Table` sizing when shown in fluid-height container


cmslewis:
*Reviewers:* the base branch is messed up at present. Gonna try to get the `release/1.25.x` in a better state (or recreate it, since we never really used it last time).

PR #1483, opened 2017-08-22
[Table] Fix for user-land sass compilation

tyscorp:
Similar issue to #1456, but when importing the table's scss.

Without this I get `File to import not found or unreadable: ../../core/src/common/mixins.` and `File to import not found or unreadable: ./common/variables.`

PR #1489, opened 2017-08-22
Prepare release 1.25.2

cmslewis:
### 🐛  Bug fixes

- #1479 Fix `Table` sizing when shown in fluid-height container
- #1483 Fix `Table` SCSS compilation errors

### 🐛  Bug fixes

- #1478 Fix `normalize.css` version on docs site


llorca:
We should include this too, it was also a regression: https://github.com/palantir/blueprint/pull/1478

cmslewis:
@giladgray @adidahiya any idea why there are so many deletions in those icon files? not immediately evident to me why that happened when I ran the `prepRelease` script.

adidahiya:
yes we should include the fix for the docs site.

adidahiya:
would be great to include #1483 as well, since that's table related

adidahiya:
you actually don't need to rebuild the docs site if you cherry-pick that commit ^

Issue #1407, opened 2017-08-04
[Labs / Select] Allow value container to be an input that filters items


sushain97:
Currently, labs' `<Select>` does not display the selected value itself, the docs opt to display the value in a `<Button>` which is nice. However, for a consistent UX when both `<MultiSelect>` and `<Select>` are being used in tandem, it would be perfect if the selected value could be displayed inside an `<InputGroup>` which supports filtering rather than filtering through a popover. Of course, using an `<InputGroup>` instead of  `<Button>` to hold the selected value is possible with a couple hacks to make the focus less janky. This however results in a double input field UI-quirk:

![image](https://user-images.githubusercontent.com/4211838/28949899-e6235728-788d-11e7-9d3a-f722080a3a83.png)

I tried to disable filtering and make the `<InputGroup>` that holds the selected value control the query and filter the items. However, this is gnarly to accomplish since it requires controlling a bunch of keyboard events e.g. moving the cursor to the start of the input, hiding the selected value during input, and manually controlling the popover. I could implement it manually using `<QueryList>` but that seems sub-optimal and this seems like a reasonable enough use case/argument for visual consistency to warrant inclusion in the "core".

I might be able to send a PR for some progress on this if this makes sense as a core feature (let me know).

Prior Art: https://jedwatson.github.io/react-select/

cc: @llorca thanks for all your help!

dmackerman:
+1 for this one.

cmslewis:
This was always on the radar for us since tackling the typeahead work over the past couple months, and it's now released as the new `Suggest` component. Closing!

PR #1499, opened 2017-08-24
[Labs] Suggest component

llorca:
#### Checklist

- [x] Include tests
- [x] Update documentation

`Suggest` component composes `Popover`, `InputGroup` and `QueryList`

#### Screenshot

### DIFF
--- a/packages/labs/src/components/select/suggest.tsx
+++ b/packages/labs/src/components/select/suggest.tsx
@@ -31,6 +31,7 @@ import * as Classes from "../../common/classes";
 export interface ISuggestProps<T> extends IListItemsProps<T> {
     /**
      * Whether the popover should close after selecting an item.
+     * @default true
      */
     closeOnSelect?: boolean;
 
--- a/packages/labs/src/components/select/suggest.tsx
+++ b/packages/labs/src/components/select/suggest.tsx
@@ -122,8 +122,8 @@ export class Suggest<T> extends React.Component<ISuggestProps<T>, ISuggestState<
         const { ref, ...htmlInputProps } = inputProps;
         const { handleKeyDown, handleKeyUp } = listProps;
         const inputValue: string = isTyping
-            ? query : selectedItem
-                ? inputValueRenderer(selectedItem) : "";
+            ? query
+            : (selectedItem ? inputValueRenderer(selectedItem) : "");
 
         return (
             <Popover
--- a/packages/labs/src/components/select/suggest.tsx
+++ b/packages/labs/src/components/select/suggest.tsx
@@ -160,19 +160,16 @@ export class Suggest<T> extends React.Component<ISuggestProps<T>, ISuggestState<
                 popoverDidOpen={this.handlePopoverDidOpen}
                 popoverWillClose={this.handlePopoverWillClose}
             >
-                <div
+                <InputGroup
+                    placeholder="Search..."
+                    value={inputValue}
+                    {...htmlInputProps}
+                    inputRef={this.refHandlers.input}
+                    onChange={this.handleQueryChange}
+                    onFocus={this.handleInputFocus}
                     onKeyDown={this.getTargetKeyDownHandler(handleKeyDown)}
-                    onKeyUp={this.state.isOpen ? handleKeyUp : undefined}
-                >
-                    <InputGroup
-                        placeholder="Search..."
-                        value={inputValue}
-                        {...htmlInputProps}
-                        inputRef={this.refHandlers.input}
-                        onChange={this.handleQueryChange}
-                        onFocus={this.handleInputFocus}
-                    />
-                </div>
+                    onKeyUp={this.getTargetKeyUpHandler(handleKeyUp)}
+                />
                 <div onKeyDown={handleKeyDown} onKeyUp={handleKeyUp}>
                     <Menu ulRef={listProps.itemsParentRef}>
                         {this.renderItems(listProps)}
@@ -294,6 +291,10 @@ export class Suggest<T> extends React.Component<ISuggestProps<T>, ISuggestState<
         return (e: React.KeyboardEvent<HTMLElement>) => {
             const { which } = e;
             const { isTyping, selectedItem } = this.state;
+            const {
+                inputProps = this.DEFAULT_PROPS.inputProps,
+                openOnKeyDown = this.DEFAULT_PROPS.openOnKeyDown,
+            } = this.props;
 
             if (which === Keys.ESCAPE || which === Keys.TAB) {
                 this.input.blur();
@@ -301,13 +302,31 @@ export class Suggest<T> extends React.Component<ISuggestProps<T>, ISuggestState<
                     isOpen: false,
                     selectedItem: isTyping ? undefined : selectedItem,
                 });
-            } else if (!(which === Keys.BACKSPACE || which === Keys.ARROW_LEFT || which === Keys.ARROW_RIGHT)) {
+            } else if (openOnKeyDown
+                && which !== Keys.BACKSPACE
+                && which !== Keys.ARROW_LEFT
+                && which !== Keys.ARROW_RIGHT
+            ) {
                 this.setState({ isOpen: true });
             }
 
             if (this.state.isOpen) {
                 Utils.safeInvoke(handleQueryListKeyDown, e);
             }
+
+            Utils.safeInvoke(inputProps.onKeyDown, e);
+        };
+    }
+
+    private getTargetKeyUpHandler = (
+        handleQueryListKeyDown: React.EventHandler<React.KeyboardEvent<HTMLElement>>,
+    ) => {
+        return (e: React.KeyboardEvent<HTMLElement>) => {
+            const { inputProps = this.DEFAULT_PROPS.inputProps } = this.props;
+            if (this.state.isOpen) {
+                Utils.safeInvoke(handleQueryListKeyDown, e);
+            }
+            Utils.safeInvoke(inputProps.onKeyUp, e);
         };
     }
 }
--- a/packages/labs/src/components/select/suggest.tsx
+++ b/packages/labs/src/components/select/suggest.tsx
@@ -1,0 +1,265 @@
+/*
+ * Copyright 2017 Palantir Technologies, Inc. All rights reserved.
+ * Licensed under the BSD-3 License as modified (the “License”); you may obtain a copy
+ * of the license at https://github.com/palantir/blueprint/blob/master/LICENSE
+ * and https://github.com/palantir/blueprint/blob/master/PATENTS
+ */
+
+import * as classNames from "classnames";
+import * as PureRender from "pure-render-decorator";
+import * as React from "react";
+
+import {
+    HTMLInputProps,
+    IInputGroupProps,
+    InputGroup,
+    IPopoverProps,
+    Keys,
+    Menu,
+    Popover,
+    Position,
+    Utils,
+} from "@blueprintjs/core";
+import {
+    IListItemsProps,
+    IQueryListRendererProps,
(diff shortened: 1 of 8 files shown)


packages/labs/src/components/select/suggest.tsx
```diff
@@ -0,0 +1,265 @@
...
+
+    private renderQueryList = (listProps: IQueryListRendererProps<T>) => {
+        // not using defaultProps cuz they're hard to type with generics (can't use <T> on static members)
+        const { inputValueRenderer, inputProps = {}, popoverProps = {} } = this.props;
+        const { isTyping, selectedItem, query } = this.state;
+        const { ref, ...htmlInputProps } = inputProps;
+        const { handleKeyDown, handleKeyUp } = listProps;
+        const inputValue: string = isTyping
+            ? query : selectedItem
+                ? inputValueRenderer(selectedItem) : "";
```
giladgray: alignment to clarify double-ternary
```
const inputValue: string = isTyping
    ? query 
    : (selectedItem ? inputValueRenderer(selectedItem) : "");
```

packages/labs/src/components/select/suggest.tsx
```diff
@@ -0,0 +1,265 @@
...
+                >
+                    <InputGroup
+                        placeholder="Search..."
+                        value={inputValue}
+                        {...htmlInputProps}
+                        inputRef={this.refHandlers.input}
+                        onChange={this.handleQueryChange}
+                    />
+                </div>
+                <div onKeyDown={handleKeyDown} onKeyUp={handleKeyUp}>
```
giladgray: why can't these handlers go on `Menu`? shouldn't _need_ an extra wrapper element.

packages/labs/src/components/select/suggest.tsx
```diff
@@ -0,0 +1,265 @@
...
+                className={classNames(listProps.className, popoverProps.className)}
+                onInteraction={this.handlePopoverInteraction}
+                popoverClassName={classNames(Classes.SELECT_POPOVER, popoverProps.popoverClassName)}
+                popoverDidOpen={this.handlePopoverDidOpen}
+                popoverWillClose={this.handlePopoverWillClose}
+            >
+                <div
+                    onKeyDown={this.getTargetKeyDownHandler(handleKeyDown)}
+                    onKeyUp={this.state.isOpen ? handleKeyUp : undefined}
+                    onClick={this.handleInputClick}
```
giladgray: can these handlers go on `InputGroup`? you'll need to invoke corresponding `inputProps.on*` in each handler.

packages/labs/src/components/select/suggest.tsx
```diff
@@ -0,0 +1,265 @@
...
+        if (this.input == null) {
+            return;
+        }
+
+        if (this.input !== document.activeElement) {
+            // the input is no longer focused so we can close the popover
+            this.input.blur();
+            this.setState({ isOpen: false });
+        } else {
+            this.setState({ isOpen: closeOnSelect ? nextOpenState : true });
```
giladgray: prefer `else if (closeOnSelect) { setState( true ) }` instead of the no-op `nextOpenState` in ternary

packages/labs/src/components/select/suggest.tsx
```diff
@@ -0,0 +1,265 @@
...
+    private handlePopoverInteraction = (nextOpenState: boolean) => requestAnimationFrame(() => {
+        const { closeOnSelect, popoverProps = {} } = this.props;
+
+        if (this.input == null) {
+            return;
+        }
+
+        if (this.input !== document.activeElement) {
+            // the input is no longer focused so we can close the popover
+            this.input.blur();
```
giladgray: why blur if we already know it's not focused?

packages/labs/src/components/select/suggest.tsx
```diff
@@ -0,0 +1,265 @@
...
+            this.queryList.scrollActiveItemIntoView();
+        }
+
+        Utils.safeInvoke(popoverProps.popoverDidOpen);
+    }
+
+    private handlePopoverWillClose = () => {
+        const { popoverProps = {} } = this.props;
+
+        // resetting the query when the popover close
```
giladgray: no `resetOnClose`?

packages/labs/src/components/select/suggest.tsx
llorca: I'll try and report back

packages/labs/src/components/select/suggest.tsx
llorca: We can have `Menu` support `ul` HTML props in a separate PR (and then refactor all the labs components to use that)

packages/labs/src/components/select/suggest.tsx
giladgray: ✅ wait, i think i get it and this code is actually ok. `nextOpenState` is param so it needs to make it into state somehow, which my suggestion does not allow.

packages/labs/src/components/select/suggest.tsx
llorca: We always reset on close here. If we don't, you can end up with the filtered list desynced from what's shown in the input (because we don't show `query` in the input at all times)

giladgray: gotcha, cool. please update comment to clarify _why_ it always resets.

packages/labs/src/components/select/suggest.tsx
llorca: I'm not sure what you mean by no-op, but your suggestion is not equivalent to what I have. 

This logic is messed up anyway, I'm gonna handle `closeOnSelect` stuff where it belongs: `handleItemSelect`

packages/labs/src/components/select/suggest.tsx
llorca: Good question, it's because `handleItemSelect` always refocuses the input. Now it only does with `closeOnSelect={false}` ✅

dmackerman:
Neat. Will this support `InputGroup` props so that icons can be applied and such?

llorca:
@dmackerman you bet

llorca:
Not sure why [this commit](https://github.com/palantir/blueprint/pull/1499/commits/cfcd57a600295db67b55a72b639a3b8e4fcc4f2c) broke the build... The change is literally one word 🤔
```diff
- `Suggest` behaves similarly to `Select`, except it renders a text input as Popover target instead of arbitrary children.
+ `Suggest` behaves similarly to `Select`, except it renders a text input as the `Popover` target instead of arbitrary children.
```

giladgray:
yeah not sure about the build... looks suspiciously like a test flake cuz it's about focus. 😕

llorca:
Flaking test:
```
  <Overlay>
    Focus management
      ✖ returns focus to overlay if enforceFocus=true
        Chrome 60.0.3112 (Mac OS X 10.12.6)
      Error: Uncaught AssertionError: expected <button></button> to not equal <button></button> (test/index.ts:4311)
```
@giladgray how can I fix that? The underlying `Popover` has `enforceFocus={false}` and this test is about `enforceFocus={true}`:
https://github.com/palantir/blueprint/blob/al/suggest2/packages/labs/src/components/select/suggest.tsx#L138

llorca:
Well that test actually fails on `master` too...

cmslewis:
@llorca the "Open popover on key down" option doesn't appear to work:

packages/labs/src/components/select/suggest.tsx
```diff
@@ -248,17 +248,26 @@ export class Suggest<T> extends React.Component<ISuggestProps<T>, ISuggestState<
 
     private handlePopoverWillClose = () => {
         const { popoverProps = {} } = this.props;
+        const { selectedItem } = this.state;
 
         // reset the query when the popover close, make sure that the list
         // isn't filtered on when the popover opens next
-        this.setState({ query: "" });
+        this.setState({
+            activeItem: selectedItem ? selectedItem : this.props.items[0],
```
cmslewis: nit: Prefer explicit `selectedItem != null`. I like to avoid loose falsy checks.

packages/labs/src/components/select/suggest.tsx
```diff
@@ -0,0 +1,296 @@
...
+        // omit props specific to this component, spread the rest.
+        const {
+            itemRenderer,
+            inputProps,
+            noResults,
+            popoverProps,
+            ...restProps,
+        } = this.props;
+
+        return <this.TypedQueryList
```
cmslewis: nit: typical format for us is:

```tsx
return (
    // indented component
);
```

packages/labs/src/components/select/suggest.tsx
```diff
@@ -0,0 +1,296 @@
...
+        }
+    }
+
+    private renderQueryList = (listProps: IQueryListRendererProps<T>) => {
+        // not using defaultProps cuz they're hard to type with generics (can't use <T> on static members)
+        const { inputValueRenderer, inputProps = {}, popoverProps = {} } = this.props;
+        const { isTyping, selectedItem, query } = this.state;
+        const { ref, ...htmlInputProps } = inputProps;
+        const { handleKeyDown, handleKeyUp } = listProps;
+        const inputValue: string = isTyping
```
cmslewis: Can we do away with the explicit `: string` type here?

packages/labs/src/components/select/suggest.tsx
```diff
@@ -0,0 +1,296 @@
...
+            ? query
+            : (selectedItem ? inputValueRenderer(selectedItem) : "");
```
cmslewis: Same

packages/labs/src/components/select/suggest.tsx
```diff
@@ -0,0 +1,296 @@
...
+        if (this.input != null) {
+            // wait until the input is properly focused to select the text inside of it
+            requestAnimationFrame(() => this.input.setSelectionRange(0, this.input.value.length));
+        }
+    }
+
+    private handleInputFocus = (event: React.SyntheticEvent<HTMLInputElement>) => {
+        const { openOnKeyDown, inputProps = {} } = this.props;
+
+        this.selectText();
```
cmslewis: Worth adding a `selectAllOnFocus: boolean` prop like we do for `NumericInput`?

packages/labs/src/components/select/suggest.tsx
```diff
@@ -0,0 +1,296 @@
...
+
+    private handleActiveItemChange = (activeItem: T) => this.setState({ activeItem });
+
+    private handleItemSelect = (item: T, event: React.SyntheticEvent<HTMLElement>) => {
+        const { closeOnSelect } = this.props;
+        let nextOpenState: boolean;
+
+        if (!closeOnSelect) {
+            this.input.focus();
+            this.selectText();
```
cmslewis: Is this line necessary even though `focus`'ing will trigger `handleInputFocus` and subsequently invoke `this.selectText()`? (Per the `requestAnimationFrame` comment above, I'm assuming focus doesn't happen synchronously here?)

cmslewis:
@giladgray etc. I'm gonna merge this and then do an audit to make sure controlled mode works as expected before we release. Might even...

(•_•)
( •_•)>⌐■-■
(⌐■_■) ...write some tests.

packages/labs/src/components/select/suggest.tsx
cmslewis: Done

cmslewis:
Brought test coverage up to 90%.

sushain97:
If I understand this component correctly, it fulfills https://github.com/palantir/blueprint/issues/1407 (please feel free to close the issue if I'm right). 
@mfedderly, probably worth switching to this to easily get rid of the double input box problem.

Issue #1374, opened 2017-07-22
Props to control batching behavior


llorca:
Follow up from https://github.com/palantir/blueprint/issues/1368

Current UX is okay for the average table. For scenarios where a table is as simple as text only, consumers should be able to render more cells per frame. For cases where the table is large and contains complex cells, we may want to decrease the number of cells per frame.

llorca:
From @cmslewis:
> Another idea is, instead of a prop `useBatchRendering?: boolean;`, to add a `renderOptimizationMode?: RenderOptimizationMode;` prop (working title) that accepts values like:
> - `RenderOptimizationMode.BATCH`: render as many new cells as possible per frame, and no more (this is what we currently have, and this scheme also has the strongest guarantee against bringing the UI to a crawl)
> - `RenderOptimizationMode.DEBOUNCE`: debounce scrolling, hiding cells on scroll start and rendering cells all at once after the scroll event finishes (this has a risk of locking the UI if the cells are small/numerous/complex/have truncated content, e.g.)
> - `RenderOptimizationMode.NONE`: no optimizations, just render everything (useful for smaller tables, but can easily bring the UI to a crawl on larger tables)
> 
> (Maybe also a `RenderOptimizationMode.THROTTLE` for throttling scrolling).

tgreenwatts:
Need to get at least ability to turn off batching by 9/1, for consumers who are creating PDFs of tables

tgreenwatts:
Going to close this, can open a new one with other modes and ideas

PR #1503, opened 2017-08-25
[Table] Add renderMode prop

cmslewis:
#### Checklist

- [x] Include tests
- [x] Update documentation (contained in new prop's description)

- 💎 __NEW__ `Table` now supports a `renderMode?: RenderMode` prop:
    - `RenderMode.BATCH` _(default)_: renders cells in batches to improve performance
    - `RenderMode.NONE`: renders cells synchronously all at once
- Added a new "Batch rendering" switch to the `Table` example:
image

I designed the prop as a `RenderMode` enum instead of a `useBatchRendering` boolean to allow us to support additional rendering modes in the future. See: https://github.com/palantir/blueprint/issues/1374#issuecomment-318497948, pasted here for your convenience:

> Another idea is, instead of a prop `useBatchRendering?: boolean;`, to add a `renderOptimizationMode?: RenderOptimizationMode;` prop (working title) that accepts values like:
> - `RenderOptimizationMode.BATCH`: render as many new cells as possible per frame, and no more (this is what we currently have, and this scheme also has the strongest guarantee against bringing the UI to a crawl)
> - `RenderOptimizationMode.DEBOUNCE`: debounce scrolling, hiding cells on scroll start and rendering cells all at once after the scroll event finishes (this has a risk of locking the UI if the cells are small/numerous/complex/have truncated content, e.g.)
> - `RenderOptimizationMode.NONE`: no optimizations, just render everything (useful for smaller tables, but can easily bring the UI to a crawl on larger tables)
> 
> (Maybe also a `RenderOptimizationMode.THROTTLE` for throttling scrolling).

### DIFF
--- a/packages/table/preview/index.tsx
+++ b/packages/table/preview/index.tsx
@@ -42,7 +42,7 @@ ReactDOM.render(<Nav selected="perf" />, document.getElementById("nav"));
 
 import { IFocusedCellCoordinates } from "../src/common/cell";
 import { IColumnIndices, IRowIndices } from "../src/common/grid";
-import { RenderOptimizationMode } from "../src/common/renderOptimizationMode";
+import { RenderMode } from "../src/common/renderMode";
 import { IRegion } from "../src/regions";
 import { DenseGridMutableStore } from "./denseGridMutableStore";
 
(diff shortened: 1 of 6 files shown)


packages/table/preview/index.tsx
```diff
@@ -42,6 +42,7 @@ ReactDOM.render(<Nav selected="perf" />, document.getElementById("nav"));
 
 import { IFocusedCellCoordinates } from "../src/common/cell";
 import { IColumnIndices, IRowIndices } from "../src/common/grid";
+import { RenderOptimizationMode } from "../src/common/renderOptimizationMode";
```
gscshoyru: Can we just call it render mode? And mention that we do optimizations in the docs? That way we can get rid of the extra mouthful.

cmslewis: Sure

PR #1505, opened 2017-08-26
[Table] Add onCompleteRender prop

### LINKED ISSUE #1495: resizeByTallest doesn't work on mount because batch loader
>I think what may actually be happening is that Table doesn't finish rendering its cells before TableBoard.componentDidUpdate is called. So, table.resizeRowsByTallest is called with incomplete information (i.e. it doesn't know what the tallest cell is) and thus doesn't resize properly. It behaves as expected once all cells are rendered, though. So, I think Table's batch rendering is actually the root cause here.

>One suggestion is to add a onCompleteRender or something to that effect so we only resize once we know we are ready. Thoughts?

cmslewis:
#### Checklist

- [x] Include tests
- [x] Update documentation (done automatically with new prop description)

- 💎 __NEW__ `Table` now offers an `onCompleteRender` prop, invoked on mount/update when all cells in view have finished rendering.
- 🐛 __FIX__ `Table` no longer fires `onVisibleCellsChange` redundantly when the viewport hasn't actually changed.

- The callback is invoked when the `MAIN` quadrant's cells have finished loading, given that the `MAIN` quadrant is a superset of all other quadrants (`TOP`, `LEFT`, `TOP_LEFT`).
    - Since each quadrant's `TableBody` has a separate `Batcher`, it's _possible_ to have a race condition wherein the `MAIN` quadrant cells finish rendering before the cells in other, smaller quadrants. One solution there would involve using a single batcher instance passed into the `TableBody` from the parent `Table` to properly queue cell rendering across all quadrants; we could also make the quadrants smarter by not double-rendering cells located underneath other quadrants. None of that is implemented in this PR, though, so we'll live with the potential for the race condition for now. **What do you think?**
- This PR invokes `onCompleteRender` on mount _and_ on update. But looking at the console logs in the table example, I'm not convinced that invoking this callback on update is actually helpful (for instance, the callback is invoked multiple times over the course of a single scroll interaction). Invoking `onCompleteRender` only on mount would still solve #1495. **What do you think?**

#### Screenshot

on-complete-render


packages/table/src/table.tsx
```diff
@@ -452,8 +457,8 @@ export class Table extends AbstractComponent<ITableProps, ITableState> {
 
         return !Utils.shallowCompareKeys(this.props, nextProps, propKeysBlacklist)
             || !Utils.shallowCompareKeys(this.state, nextState, stateKeysBlacklist)
-            || !Utils.deepCompareKeys(this.props, nextProps, ["selectedRegions"])
-            || !Utils.deepCompareKeys(this.state, nextState, ["selectedRegions"]);
+            || !Utils.deepCompareKeys(this.props, nextProps, Table.SHALLOW_COMPARE_PROP_KEYS_BLACKLIST)
```
cmslewis: These lists happen to be complementary, so I'm using the shallow-compare blacklists for the deep-compare whitelists. Is that too clever? Should I create separate `Table.DEEP_COMPARE_{PROP,STATE}_KEYS_WHITELIST`s for the sake of robustness? (I could still set those to equal the shallow-compare blacklist arrays, with a comment to clarify why that's okay.)

packages/table/src/table.tsx
```diff
@@ -1736,7 +1757,16 @@ export class Table extends AbstractComponent<ITableProps, ITableState> {
...
-        this.invokeOnVisibleCellsChangeCallback(nextViewportRect);
+
+        const { viewportRect } = this.state;
+
+        const didViewportChange =
+            (viewportRect != null && !viewportRect.equals(nextViewportRect))
+            || (viewportRect == null && nextViewportRect != null);
+
+        if (didViewportChange) {
+            this.invokeOnVisibleCellsChangeCallback(nextViewportRect);
```
cmslewis: Noticed we were invoking this even when the `viewportRect` values hadn't actually changed. This and the `shouldComponentUpdate` change address that double-invocation issue.

packages/table/src/table.tsx
gscshoyru: I'm not sure what you mean by this, so... um, possibly a comment anyways? What are you doing that's odd, exactly?

cmslewis: Eh I think I was overthinking it. Fine as is IMO.

PR #1516, opened 2017-08-30
[Table] Export `RenderMode` from @blueprintjs/table

cmslewis:

- 🐛 __FIXED__ `@blueprintjs/table` now exports the `RenderMode` enum for use with the `renderMode` prop.


cmslewis:
@giladgray just added some...creative...unit tests. Thoughts on those?

packages/table/test/indexTests.ts
```diff
@@ -0,0 +1,108 @@
...
+import * as tableExports from "../src";
+
+describe("@blueprintjs/table", () => {
+    interface IKnownKeysMap {
+        [key: string]: boolean;
+    }
+
+    const knownKeys: IKnownKeysMap = {};
+
+    describe("Exports all expected values", () => {
```
adidahiya: not really a fan of these tests; they're tautological and changes to exports in index files should be caught in code review

cmslewis: Yeah, if there were some automated way to guarantee that every type and value mentioned in the docs were exported, that'd be nice; but with this approach here, you necessarily have to remember to write a test in order to guarantee that the thing you care about is exported, at which point you'll already have remembered that you need to export the thing.

PR #1519, opened 2017-08-31
[Table] A slightly different tactic to hopefully fully popover overflow

### LINKED ISSUE #1459: Table stack overflows at 90% zoom
Related: https://github.com/palantir/blueprint/pull/1344 

Today the table overflows but continues to render, with a pause each time it overflows
https://pl.ntr/fX

gscshoyru:

Modify the popover behavior so that it doesn't go away if the two things are equal, because of issues with pixel-perfectness at different zoom levels.

This unfortunately means that double-clicking to resize a column won't always work. I really can't figure out a way to resolve this and also not have infinite loops at 90% zoom level, though.

### DIFF
--- a/packages/table/src/cell/formats/truncatedFormat.tsx
+++ b/packages/table/src/cell/formats/truncatedFormat.tsx
@@ -157,14 +157,16 @@ export class TruncatedFormat extends React.Component<ITruncatedFormatProps, ITru
         }
 
         // if the popover handle exists, take it into account
-        let popoverHandleAdjustmentFactor = this.state.isTruncated ? CONTENT_DIV_WIDTH_DELTA : 0;
-        // add a slight bit of buffer space where we don't show the popover, to deal with cases
-        // where everything isn't pixel perfect
-        popoverHandleAdjustmentFactor += .5;
+        const popoverHandleAdjustmentFactor = this.state.isTruncated ? CONTENT_DIV_WIDTH_DELTA : 0;
 
-        const isTruncated = this.contentDiv !== undefined &&
+        let isTruncated = this.contentDiv !== undefined &&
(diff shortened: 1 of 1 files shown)


tgreenwatts:
>This unfortunately means that double-clicking to resize a column won't always work
This seems bad. When would it not work? Is it only at a certain zoom level? 

Does this overflow issue only happen at 90% or other percentages too? 

Also @philcchen for passive SA on this

gscshoyru:
It happens regardless of zoom level. It's just a pixel worth of difference, but, a pixel is all you need sometimes. There needs to be some buffer so it doesn't rapidly switch back and forth under certain circumstances.

tgreenwatts:
What does it mean that double-click-to-resize won't always work? Under what conditions won't it work, and what will it do instead?

gscshoyru:
If it starts smaller than (i.e. starts with a ... popover on the side) and you double click, then depending on weirdness with pixels it's possible the largest one in the column will keep the popover. 

This was already true sometimes, I think, though. 

The only way to work around this is maybe make double-click about a pixel wider than it "needs" to be? Not sure how you feel about that.

tgreenwatts:
Yea I think that happens today already, can probably address in a separate PR - I think estimating is our best bet. Cool will see what @cmslewis thinks

cmslewis:
> it's possible the largest one in the column will keep the popover

This doesn't seem like a blocker to me, particularly because (IIRC) I've already observed this behavior of late.

gscshoyru:
Suggestions make sense, I'll implement and take a look shortly :)

gscshoyru:
Refactored more or less as suggested, but shouldComponentUpdate doesn't work because it keeps the cells from re-measuring themselves when the columns are resized. Will look into ways to make that better right now, to decrease the total number of measures.

cmslewis:
Subsuming this work into #1550

PR #1521, opened 2017-08-31
Prepare release table-1.24.1

cmslewis:

### 🐛  Bug fixes
- `@blueprintjs/table` now exports `RenderMode` enum (for use in `Table`'s `renderMode?: RenderMode` prop).


cmslewis:
@Reviewers - FYI I also included the "Skip flaky test" commit (dd90bfa, #1514), but no need to mention that, right?

gscshoyru:
I don't think you need to mention that :P

tgreenwatts:
Version is missing in the commit description

PR #1523, opened 2017-09-02
[Table] Select the target cell on right-click

### LINKED ISSUE #1187: Table selection/context menu bugginess
### Bug report

- __Browser and OS versions__: Mac Sierra

#### Steps to reproduce

There are actually 2 things happening here, the first, and more important one for me, is that it doesn't seem like the Table component handles context menu clicks in the usual way: 
asdfasdfasd

As you can see (on the docs page itself), when I right click above in your example table, the selected region doesn't change. I believe this is because the table `onSelection` prop does not get fired for right-clicks. I am running into this same issue with my own Table as well but am not sure quite how to solve it. Any help would be appreciated.

The second, smaller issue , is that for some reason opening the context menu causes the whole page to shift slightly. You can see this in the video as well if you look closely. I also find this to be the case when implementing my own blueprint tables. Does anyone know why this is happening?

Thanks very much!

cmslewis:
#### Checklist

- [x] Include tests

- 🐛 __Fixed__ `Table` now selects the target cell on right-click if the cell is not already contained in a selection.

### DIFF
--- a/packages/table/src/tableBody.tsx
+++ b/packages/table/src/tableBody.tsx
@@ -21,7 +21,7 @@ import { ICoordinateData } from "./interactions/draggable";
 import { IContextMenuRenderer, MenuContext } from "./interactions/menus";
 import { DragSelectable, ISelectableProps } from "./interactions/selectable";
 import { ILocator } from "./locator";
-import { Regions } from "./regions";
+import { Regions, IRegion } from "./regions";
 
 export interface ITableBodyProps extends ISelectableProps, IRowIndices, IColumnIndices, IProps {
     /**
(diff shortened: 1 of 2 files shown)


packages/table/src/tableBody.tsx
```diff
@@ -215,8 +215,24 @@ export class TableBody extends React.Component<ITableBodyProps, {}> {
             return undefined;
         }
 
-        const target = this.locateClick(e.nativeEvent as MouseEvent);
-        return renderBodyContextMenu(new MenuContext(target, selectedRegions, grid.numRows, grid.numCols));
+        const targetRegion = this.locateClick(e.nativeEvent as MouseEvent);
+        const { numRows, numCols } = grid;
+
+        const menuContext = new MenuContext(targetRegion, selectedRegions, numRows, numCols);
```
gscshoyru: Should we not be fixing the selectedRegions first here? Before we render the menu?

cmslewis: Sure, fixed

llorca:
Doesn't seem to work on my end.

Right click behavior:
rightclick

Ctrl click behavior:
ctrlclick

Expected behavior:
- Right/ctrl click on a cell/column/row selects said cell/column/row. It shouldn't append to any current selection but create a new one instead

cmslewis:
@llorca Try with the new build once it finishes. The upcoming version works fine and passes tests locally FWIW.

llorca:
Yup, right-click behavior now works!

Note that ctrl-click behavior is still broken, still behaving as the GIF above.

cmslewis:
@llorca @gscshoyru I'm going to fix the cmd+click thing once #1524 merges. I have an issue tracking the follow-up already.

PR #1537, opened 2017-09-08
[Labs] Add disabled prop to Select

### LINKED ISSUE #1436: [labs/select] Provide mechanism for disabling Select handlers
I'm trying to disable a Select input... use case is that one select should be filled out before the second one. There are two pieces here - I can disable the underlying button/child element for the visual side of things but need a way to disable the handlers on the `Select`  component.

screen recording 2017-08-09 at 05 57 pm

brieb:
#### Checklist

- [x] Include tests
- [x] Update documentation

- Add a `disabled` prop that disables the internal popover if set

### DIFF
--- a/packages/labs/examples/selectExample.tsx
+++ b/packages/labs/examples/selectExample.tsx
@@ -21,11 +21,13 @@ export interface ISelectExampleState {
     minimal?: boolean;
     resetOnClose?: boolean;
     resetOnSelect?: boolean;
+    disabled?: boolean;
 }
 
 export class SelectExample extends BaseExample<ISelectExampleState> {
 
     public state: ISelectExampleState = {
+        disabled: false,
         film: TOP_100_FILMS[0],
         filterable: true,
         minimal: false,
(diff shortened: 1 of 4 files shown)


palantirtech:
Thanks for your interest in [palantir/blueprint](https://github.com/palantir/blueprint), @brieb! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

cmslewis:
@brieb Behavior feels good. 👍 Looking at the code now.

packages/labs/examples/selectExample.tsx
```diff
@@ -53,6 +57,7 @@ export class SelectExample extends BaseExample<ISelectExampleState> {
                 <Button
                     rightIconName="caret-down"
                     text={film ? film.title : "(No selection)"}
+                    disabled={disabled}
```
cmslewis: The need for a `disabled` prop on both the `Select` and the child could be a sticking point for consumers.  @llorca - can you come up with messaging to communicate the need for both? I'm thinking it could be a `pt-primary` info callout or something, probably just below the **Javascript API** heading. Initial idea for language:

> ### Disabling a Select
> Because a `Select` can accept arbitrary content for the trigger, disabling the component requires passing `disabled={true}` (or equivalent) to both the `Select` and its child, separately.

packages/labs/src/components/select/select.tsx
```diff
@@ -40,6 +40,13 @@ export interface ISelectProps<T> extends IListItemsProps<T> {
      */
     itemRenderer: (itemProps: ISelectItemRendererProps<T>) => JSX.Element;
 
+    /**
+     * Whether the component is non-interactive.
+     * Note that you'll also need to disable the component's children, if appropriate.
```
cmslewis: Ah, this might be enough. Thoughts @llorca?

llorca: Seems fine to me

packages/labs/examples/selectExample.tsx
llorca: Nice, let's make it a primary callout. I'd put it in between the initial docs and example, since that's where the target is mentioned in the first place:
image

Small edit:
> and its target child

brieb:
Went ahead and added the suggested documentation, for clarity.

image

cmslewis:
@brieb yep looks good to me. Thoughts @llorca?

cmslewis:
Thanks @brieb! 🎩

PR #1566, opened 2017-09-13
Add prettier using tslint-plugin-prettier

llorca:
To do:
- [ ] Run codebase through prettier


cmslewis:
Looks good so far. Just merged a bunch of big PRs btw, so now may be a good time to bite the bullet and do this.

adidahiya:
This is strange... something must be wrong with the linting build task because this should've triggered a bunch of formatting lint failures with the new rule we just enabled

Issue #890, opened 2017-03-27
Timezone support for time components


adidahiya:
- DateTimeInput (#748)
- TimePicker
- DateTimePicker

UselessPickles:
Is there any update on this? It would be awesome if the Date/Time Input components would operate in terms of moment instead of Date objects so that we could take advantage of timezone support in moment-timezone. Working in terms of moment would also make i18n easier, thanks to moment's locale support, and locale-based format patterns.

giladgray:
@UselessPickles we're actually in the process of considering dropping the `moment` dependency entirely: see #1952 (and issue #317). there's also issue #1852, proposing using ISO _strings_ rather than `Date` objects, because ISO strings _do_ include precise timezone information.

adidahiya:
I think https://github.com/palantir/blueprint/issues/1852 might be a better approach so that we're not tied to moment? If you agree I'll close this issue in favor of that one.

UselessPickles:
If working in terms of ISO strings, then I suppose it would be simple enough to create a wrapper component that works in terms of moment and a timezone ID if I want to use moment in my code.

kaiyoma:
+1 to moving forward with some kind of resolution here. I'm currently evaluating Blueprint v2 and was disheartened to find out today that the various date and time pickers don't support timezones. I'm building an app that supports switching back and forth between UTC and the user's local timezone, so our date/time picker needs to support that as well.

kaiyoma:
Being able to pass in an ISO string or Moment object would be the most flexible/complete solution, as you'd be able to show the correct date/time in any arbitrary timezone. That could be useful to some others, whereas my needs are simply to flip back and forth between the local timezone and UTC.

Right now, we're using `react-datetime` in our project:

https://github.com/YouCanBookMe/react-datetime

You can see that their API supports a `utc` prop. If you guys had a similar prop, that would satisfy my needs. Given that there are already date/time pickers out there that support UTC vs. local time, I see no reason to write custom code to do the timezone shift manually. (That kind of code is prone to bugs or logic errors anyway.) Given the choice, I'm going to stick with `react-datetime` because it already does what I need.

tinopetros:
Curious if there are any updates on this? Timezone support within the components listed above would be really helpful. I am interested in taking a look but would want to take a direction that makes sense. Seems like moment is still used, would you recommend continuing that way or looking into using ISO strings instead?

jayarjo:
[...]

PR #1568, opened 2017-09-14
[Labs] Timezone Picker

brieb:
#### Fixes part of #890

Out of scope for this PR: Incorporating with the other datetime components

#### Checklist

- [x] Include tests
- [x] Update documentation

Timezone select component


packages/labs/src/components/timezone-select/timezoneSelect.tsx
```diff
@@ -0,0 +1,509 @@
...
+    };
+
+    private timezones: ITimezoneItem[];
+    private timezoneToQueryCandidates: { [timezone: string]: string[] };
+    private initialTimezones: ITimezoneItem[];
+
+    constructor(props: ITimezoneSelectProps, context?: any) {
+        super(props, context);
+
+        const date = props.date || new Date();
```
giladgray: deconstruct with default value. more type-safe than `||` pattern.
```
const { date = new Date() } = props;
```

packages/labs/src/components/timezone-select/timezoneSelect.tsx
```diff
@@ -0,0 +1,509 @@
...
+        this.initialTimezones = getInitialTimezoneItems(date, props.showLocalTimezone);
+    }
+
+    public render() {
+        const { className, disabled, popoverProps } = this.props;
+        const { query } = this.state;
+
+        const finalPopoverProps: Partial<IPopoverProps> & object = {
+            ...popoverProps,
+            popoverClassName: classNames(popoverProps.popoverClassName, Classes.TIMEZONE_SELECT_POPOVER),
```
giladgray: nit: prefer putting className prop last in arguments to `classNames()`

packages/labs/src/components/timezone-select/timezoneSelect.tsx
```diff
@@ -0,0 +1,509 @@
...
+                disabled={disabled}
+                onQueryChange={this.handleQueryChange}
+            >
+                {this.renderTarget()}
+            </TypedSelect>
+        );
+    }
+
+    public componentWillReceiveProps(nextProps: ITimezoneSelectProps) {
+        const nextDate = nextProps.date || new Date();
```
giladgray: deconstruct again

packages/labs/src/components/timezone-select/timezoneSelect.tsx
```diff
@@ -0,0 +1,509 @@
...
+                text={displayValue || defaultDisplayValue || placeholder}
+                rightIconName="caret-down"
+                disabled={disabled}
+            />
+        );
+    }
+
+    private updateTimezones(date: Date): void {
+        const timezones = getTimezoneItems(date);
+        this.timezones = timezones;
```
giladgray: ```this.timezones = getTimezoneItems(date);```

why two lines?

packages/labs/src/components/timezone-select/timezoneSelect.tsx
```diff
@@ -0,0 +1,509 @@
...
+}
+
+function getPopulousTimezoneItems(date: Date): ITimezoneItem[] {
+    const timezones = moment.tz.names()
+        .filter((timezone) => (
+            // Filter out noisy timezones
+            // See https://github.com/moment/moment-timezone/issues/227
+            /\//.test(timezone) &&
+            !/Etc\//.test(timezone)
+        ));
```
giladgray: nit: two lines instead of 6
```js
// Filter out noisy timezones. See https://github.com/moment/moment-timezone/issues/227
.filter((timezone) => /\//.test(timezone) && !/Etc\//.test(timezone))
```

packages/labs/src/components/timezone-select/timezoneSelect.tsx
```diff
@@ -0,0 +1,509 @@
...
+}
+
+function getTimezoneItems(date: Date): ITimezoneItem[] {
+    return moment.tz.names().map((timezone) => toTimezoneItem(timezone, date));
+}
+
+function getInitialTimezoneItems(date: Date, showLocalTimezone: boolean): ITimezoneItem[] {
+    const populous = getPopulousTimezoneItems(date);
+    const local = showLocalTimezone ? getLocalTimezoneItem(date) : undefined;
+    return local ? [local, ...populous] : populous;
```
giladgray:
```js
return showLocalTimezone 
    ? [getLocalTimezoneItem(date) ...populous]
    : populous;
```

packages/labs/src/components/timezone-select/timezoneSelect.tsx
```diff
@@ -0,0 +1,509 @@
...
+        }
+        Utils.safeInvoke(this.props.onChange, timezone.timezone);
+    }
+
+    private handleQueryChange = (query: string) => {
+        this.setState({ query });
+    }
+}
+
+function getTimezoneItems(date: Date): ITimezoneItem[] {
```
adidahiya: can you move all these static helpers to a separate file to break it up?

packages/labs/src/components/timezone-select/timezoneSelect.tsx
```diff
@@ -0,0 +1,509 @@
...
+    return {
+        iconName: undefined,
+        key: timezone,
+        label: offsetAsString,
+        text: timezone + (abbreviation ? ` (${abbreviation})` : ""),
+        timezone,
+    };
+}
+
+interface ITimezoneMetadata {
```
adidahiya: same here, I would put this in another file, `timezoneSelectTypes.ts` or `timezoneSelectInterfaces.ts`

packages/labs/src/components/timezone-select/timezoneSelect.tsx
giladgray: let's pull these out to a `timezoneUtils` file in the same directory.

__OR__ a bunch of small utility files, one per function (some of these funcs are pretty complex)

packages/labs/src/components/timezone-select/timezoneSelect.tsx
```diff
@@ -0,0 +1,509 @@
...
+     * Props to spread to `Popover`. Note that `content` cannot be changed.
+     */
+    popoverProps?: Partial<IPopoverProps> & object;
+}
+
+export type ITimezoneSelectTargetRenderer = (targetProps: ITimezoneSelectTargetProps) => JSX.Element | null;
+
+export interface ITimezoneSelectTargetProps {
+    /** The currently selected timezone. */
+    value: string | undefined;
```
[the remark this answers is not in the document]
adidahiya: might as well add it now though, to ease the transition later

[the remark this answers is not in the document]
giladgray: yeah i agree, just calling out that it doesn't mean anything inside blueprint

packages/labs/examples/timezoneSelectExtendedExample.tsx
```diff
@@ -0,0 +1,232 @@
...
+    disabled?: boolean;
+    showLocalTimezone?: boolean;
+    useDefault?: boolean;
+    defaultToUserTimezoneGuess?: boolean;
+    useCustomRenderer?: boolean;
+}
+
+const EXAMPLE_DEFAULT_TIMEZONE = "Pacific/Honolulu";
+
+export class TimezoneSelectExtendedExample extends BaseExample<ITimezoneSelectExtendedExampleState> {
```
giladgray: i'm not sure this example is valuable. it doesn't show much beyond the basic example. would be fine moving props switches to basic example and killing this (can always add later).

packages/labs/examples/timezoneSelectExtendedExample.tsx
```diff
@@ -0,0 +1,232 @@
...
+                    >
+                        <Icon iconName="time" />
+                        <span style={{ marginLeft: 10 }}>{selection}</span>
+                    </Tag>
+                </div>
+            </div>
+        );
+    }
+
+    protected renderOptions() {
```
giladgray: please move all simple prop switches to the basic example, to demonstrate that they're built-in to the component (not a result of "extended" behavior).

llorca:
❤️ 

Gonna look at the design closely, in the meantime, regarding the naming:
- "time zone" is actually composed of two words, it'd be nice to reflect that appropriately
- "time zone select" is pretty appropriate cause it's indeed a select control, but I'm kinda leaning towards "time zone picker", to match the naming of the other `datetime` package components (which ultimately this component belongs to)

packages/labs/src/components/timezone-select/_timezone-select.scss
```diff
@@ -0,0 +1,20 @@
...
+ * and https://github.com/palantir/blueprint/blob/master/PATENTS
+ */
+
+@import "~@blueprintjs/core/src/common/variables";
+@import "~@blueprintjs/core/src/components/forms/common";
+@import "~@blueprintjs/core/src/components/tag/common";
+
+.pt-timezone-select-target {
+  &.pt-timezone-select-target-placeholder {
+    color: $pt-text-color-muted;
```
brieb: is there a good way to indicate that a button is in a placeholder state? i did this, but didn't feel good about it.

packages/labs/src/components/timezone-select/timezone-select.md
```diff
@@ -0,0 +1,70 @@
...
+See [Moment Timezone's documentation](https://momentjs.com/timezone/docs/#/using-timezones/guessing-user-timezone/)
+for more information.
+
+@## Basic example
+
+@reactExample TimezoneSelectBasicExample
+
+@## Basic usage
+
+```tsx
```
llorca: Maybe remove the "Basic usage" heading and move the code block down to the "JavaScript API" section? What do you think @giladgray? Also wondering if most of the technical description from above can be moved to the "JavaScript API" section

giladgray: yes to all points!

packages/labs/src/components/timezone-select/timezoneSelect.tsx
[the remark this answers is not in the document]
brieb: what's the status of #325? would it make sense to try again now?

brieb:
@llorca 
> "time zone" is actually composed of two words, it'd be nice to reflect that appropriately

Chose "timezone" because that's what was on the moment timezone [website](https://momentjs.com/timezone/). [Appears](https://english.stackexchange.com/questions/3934/time-zone-vs-timezone) to be the case that both spellings are valid, "timezone" being the favored spelling in the US. Looked at what some java libraries use and they went with two words ([java.util](https://docs.oracle.com/javase/7/docs/api/java/util/TimeZone.html), [joda](http://www.joda.org/joda-time/apidocs/org/joda/time/DateTimeZone.html). Will change to `TimeZoneSelect`.

llorca:
Synced with Brie offline on design things 💯

packages/labs/src/components/timezone-select/timezoneSelect.tsx
```diff
@@ -0,0 +1,505 @@
...
+    Select,
+} from "..";
+import * as Classes from "../../common/classes";
+
+export interface ITimezoneSelectProps extends IProps {
+    /**
+     * The currently selected timezone (IANA time zone identifier).
+     * If this prop is provided, the component acts in a controlled manner.
+     */
+    value?: string;
```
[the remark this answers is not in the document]
brieb: What about semantic grouping and priority? Should a less important, but lexicographically lesser property come before a more important one? Is this how things should be blueprint-wide? I've found some violations. Why is this valuable?

packages/labs/src/components/timezone-select/timezoneSelect.tsx
brieb: How do you feel about the `||` pattern in react components?
```tsx
           <Button
                className={classes}
                text={displayValue || defaultDisplayValue || placeholder}
                rightIconName="caret-down"
                disabled={disabled}
            />
```

packages/labs/src/components/timezone-select/timezoneSelect.tsx
brieb: `timezones` is used on the next line
so, the ordering dependency felt clearer this way
otherwise, if you swap the two lines, its a bug

packages/labs/src/components/timezone-select/timezoneSelect.tsx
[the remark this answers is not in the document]
giladgray: listen, i enjoy a well-semantically-organized grouping as much as the next guy (if not more!), but with the sheer number of components in blueprint, it comes down to an effort calculation: can you develop and maintain a proper semantic grouping for each component in the library? is it worth it?

i've found that it's a lot less headache to just alphabetize all the props and use all that extra time to build cool shit.

packages/labs/src/components/timezone-select/timezoneSelect.tsx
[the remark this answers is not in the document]
giladgray: it won't be any easier now! we should look into it for 2.0, a few of us should be able to churn through the migration in a day or two.

packages/labs/src/components/timezone-select/timezoneSelect.tsx
giladgray: i'm intrigued by the conciseness but it's honestly quite unsettling: classic abuse of a js language "feature," which we've tended to avoid in this codebase.

packages/labs/src/components/timezone-select/timezoneSelect.tsx
giladgray: fair enough

packages/labs/src/components/timezone-select/_timezone-select.scss
brieb: talked with @llorca about this, and we thought the best thing to do was to create a separate pr that adds "pt-placeholder" modifier to buttons that would have this effect - to mimic `input`'s `::placeholder`.

brieb:
Actually, looks like the JavaScript date api uses "timezone", not "time zone".
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/getTimezoneOffset

Might leave as timezone then, since there's some precedent for it in the JS community. What do you think?

llorca:
Docs looking great! Couple more nits from latest preview:
- Can we showcase the example with a regular button instead of primary? Wouldn't want folks to get caught up with this opinionated default, might be a bit confusing.
- Let's tweak the wording of the search input placeholder to be something like "Search for timezones..." instead of "Filter..."
- (maybe out of scope) what do you think about an "all" value for target display format, that would show something like `name (abbreviation offset)` i.e. `America/Los_Angeles (PDT -07:00)`?

packages/labs/test/timezonePickerTests.tsx
```diff
@@ -0,0 +1,303 @@
...
+            <TimezonePicker popoverProps={{ isOpen: true, inline: true }} showLocalTimezone={true} />,
+        );
+        const select = findSelect(timezonePicker);
+        const items = select.prop("items");
+        assert.isTrue(items.length > 0);
+        const firstItem = items[0];
+        assert.strictEqual(firstItem.timezone, moment.tz.guess());
+    });
+
+    it("does not render local timezone at top of item list if show local timezone is disabled", () => {
```
giladgray: use prop name in test name: `if showLocalTimezone=false` 
this comment applies widely to this file.

packages/labs/src/components/timezone-picker/timezone-picker.md
```diff
@@ -0,0 +1,63 @@
+@# TimezonePicker
+
+`TimezonePicker` allows the user to select from a list of timezones.
+
+@reactExample TimezonePickerExample
+
+@## JavaScript API
+
+[Moment Timezone](http://momentjs.com/timezone/) is used internally
+for the list of available timezones and timezone metadata.
```
giladgray: this line belongs at the end of the section, not the beginning: it's an implementation detail, not relevant to usage.

packages/labs/examples/timezonePickerExample.tsx
```diff
@@ -29,20 +29,16 @@ export class TimezonePickerExample extends BaseExample<ITimezonePickerExampleSta
         timezone: "",
     };
 
-    private handleDisabledChange = handleBooleanChange((disabled) => this.setState({ disabled }));
-    private handleShowLocalTimezoneChange = handleBooleanChange((showLocalTimezone) =>
-        this.setState({ showLocalTimezone }));
+    private handleDisabledChange = handleBooleanChange(disabled => this.setState({ disabled }));
+    private handleShowLocalTimezoneChange = handleBooleanChange(showLocalTimezone =>
```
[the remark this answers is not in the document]
adidahiya: no, this is the prettier standard

cmslewis:
On filter, can we scroll the matched item to the top of the list? This is what's happening now:


Also, `EST` doesn't suggest `EST`, but rather `EST5EDT`. I'd expect `fuzz-aldrin` to return the former, no?

packages/labs/src/components/timezone-picker/timezoneMetadata.ts
```diff
@@ -23,9 +23,7 @@ export function getTimezoneMetadata(timezone: string, date: Date): ITimezoneMeta
     const offsetAsString = zonedDate.format("Z");
     const abbreviation = getAbbreviation(timezone, timestamp);
     // TODO: amend moment-timezone typings to include the population field
```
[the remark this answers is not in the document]
brieb: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/19947

packages/labs/src/components/timezone-picker/timezonePicker.tsx
```diff
@@ -21,14 +21,11 @@ import {
     MenuItem,
     Utils,
 } from "@blueprintjs/core";
-import {
-    ISelectItemRendererProps,
-    Select,
-} from "..";
+import { ISelectItemRendererProps, Select } from "..";
```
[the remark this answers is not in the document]
brieb: See some other instances of this
https://github.com/palantir/blueprint/blob/master/packages/labs/src/components/select/select.tsx#L25
https://github.com/palantir/blueprint/blob/master/packages/labs/src/components/select/suggest.tsx#L23
We should fix those too.

brieb:
@cmslewis Ya, it appears that select has the behavior that if an earlier query matched an item (est5edt in your case), that value stays selected even when a longer query has a different first item. I would expect the first item slot to remain active instead of jumping half-way down the list. We should distinguish between whether the user actually activated an item vs us just activating the first item by default.

packages/labs/src/components/select/select.tsx
```diff
@@ -155,6 +160,13 @@ export class Select<T> extends React.Component<ISelectProps<T>, ISelectState<T>>
         );
     }
 
+    public componentWillReceiveProps(nextProps: ISelectProps<T>) {
+        const { inputProps: nextInputProps = {} } = nextProps;
+        if (nextInputProps.value !== undefined && this.state.query !== nextInputProps.value) {
+            this.setState({ query: nextInputProps.value });
```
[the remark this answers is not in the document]
brieb: Going to start on this in a separate PR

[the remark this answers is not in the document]
brieb: https://github.com/palantir/blueprint/pull/1597

packages/labs/src/components/timezone-select/_timezone-select.scss
brieb: https://github.com/palantir/blueprint/pull/1599

Issue #1573, opened 2017-09-17
[Table] Scrolling is sticky at horizontal and vertical ends of table


sighrobot:
### Bug report

- __Package version(s)__: core 1.28.0, table 1.25.0
- __Browser and OS versions__: latest Chrome, FF, Safari on MacOS 10.12

#### Steps to reproduce

1. Go to [CI build of table 1.25.0](https://3946-71939872-gh.circle-artifacts.com/0/home/ubuntu/blueprint/packages/table/preview/index.html)
1. Scroll all the way to the bottom of the demo table
1. Try to scroll up just one row (using trackpad)

#### Actual behavior

❌ In Chrome only, vertical scroll bar does not align with the bottom of the table when scrolled all the way to the last row; similarly, horizontal scroll bar does not align with the far edge of the table when scrolled all the way to the right
❌ In all browsers, scrolling up from the bottom (or leftward from the right edge) is "sticky" (once you reach a certain amount of scroll inertia, the table will scroll normally):

![bp-scroll-bad](https://user-images.githubusercontent.com/4038920/30523352-eabb5b12-9bad-11e7-8b43-5ff4ce861824.gif)

#### Expected behavior

✅ Scrolling up from the bottom (or toward the first column from the last column) should be smooth, as seen in this example of scrolling down from the top row:

![bp-scroll-good](https://user-images.githubusercontent.com/4038920/30523347-e11753f4-9bad-11e7-846d-d1c3491e5873.gif)

cmslewis:
@sighrobot I've noticed this on my end too. Thanks for filing. Definitely a peculiar bug.

sighrobot:
Updated my original description to reflect that this is not a horizontal-only issue; rather it affects scrolling at the farther edges of both axes

cmslewis:
Yep, tracking both.

PR #1578, opened 2017-09-19
[Table] Improve scroll behavior in TableQuadrantStack

### LINKED ISSUE #1259: [Table] Scrolling quickly from high-numbered to low-numbered rows ruins alignments

cmslewis:
#### Checklist

- [x] Include tests

- 🔄 __CHANGED__ `Table` now debounces quadrant-view sync'ing until ~`250ms`~ `500ms` after scroll has stopped.
    - This means we're doing a _lot_ less DOM accessing per `scroll`/`wheel` event. Now, the only thing we do _during_ the scroll is update the `scroll{Top,Left}` of each quadrant to keep them in sync. This still results in reflows, but we're doing objectively less work per frame now.
    - What now goes into this "quadrant-view sync'ing" on scroll _end_ you ask?
        - `Row-header width`: adjusting to fit its widest content (e.g. `1000` is wider than `10`)
        - `Menu-element size`: adjusting to fit the latest row- and column-header sizes.
        - `LEFT/TOP_LEFT quadrant widths`: adjusting to fit the latest row-header width.
        - `TOP/TOP_LEFT quadrant heights`: adjusting to fit the latest column-header height.
        - `LEFT quadrant bottom offset`: adjusting to clear the bottom scrollbar if it exists
        - `TOP quadrant right offset`: adjusting to clear the right scrollbar if it exists
        - [Here](https://github.com/palantir/blueprint/pull/1578/files#diff-72daeac06c3d7ebe4c4ee3cc435e0f2eR614) is the function in the code.
    - ^ All of that used to happen on _every_ `scroll`/`wheel` event. Now it happens only when scroll has stopped.
    - The debounce delay can be disabled for easier unit testing.
- 🐛  __FIXED__ As a result of the above work, scrolling quickly (e.g. clicking and dragging the scroll bar itself) no longer yields row-header width bugs (see: #1259).

Under the hood:
- Add `ScrollUtils.measureScrollBarThickness(element: HTMLElement, direction: "vertical" | "horizontal"`
    - Write unit tests for it.
- Refactor logic in `TableQuadrantStack` to avoid mixing DOM reads and writes
    - Write unit tests for the whole component + fix little bugs I found as a result.
    - Add a caching layer (`TableQuadrantStackCache`) to avoid as many DOM lookups as possible.
- Add unit tests for `TableQuadrant`.
    - Trying to build confidence in our under-the-hood APIs by testing each module in isolation.

- Don't freak out. Most of the LOC changes are in test files.
- @llorca asked about animating the row header to its new width. ~That's seems tricky at a glance, it would be harder to test, and it would add more processing work into the picture. Would rather keep things lean and mean for now, then add that kind of additional processing work once general performance is more solid.~ It's doable with simple CSS, but it's pretty jittery on big tables.
- There is a race condition: the row-header width won't be updated properly if the cells haven't finished rendering before the "desired" measurement it done. `500ms` seems like a long enough wait in most cases though.

#### Screenshot

The major visual change here is that row-headers now rejigger to their new "desired" size when scrolling has finished.

This can be subtle when the header has to expand by only a pixel or so:

Or it can be quite noticeable when the destination number is much wider:

 
This behavior is inspired by Excel, which does the exact same thing:


themadcreator:
The table scrolling gets stuck at the bottom. If I scroll up slowly it doesn't move, but if i scroll quickly, it becomes unstuck.


table-scroll-stuck

tgreenwatts:
How much does this change perf? Can we gets some ###s ?

packages/table/src/common/internal/scrollUtils.ts
```diff
@@ -59,6 +63,23 @@ export function getScrollPositionForRegion(
...
+    const isHorizontal = direction === "horizontal";
+
+    // measure the *height* of horizontal scroll bars.
+    // measure the *width* of vertical scroll bars.
+    const offsetSize = isHorizontal ? element.offsetHeight : element.offsetWidth;
+    const clientSize = isHorizontal ? element.clientHeight : element.clientWidth;
+
+    // offset size includes the scroll bar. client size does not.
+    // the difference gives the width of the scroll bar.
+    return offsetSize - clientSize;
```
giladgray: how about (not including comments, which you should def keep):

```js
return (direction === "horizontal")
    ? element.offsetHeight - element.clientHeight
    : element.offsetWidth - element.clientWidth;
```

packages/table/src/locator.ts
```diff
@@ -51,14 +51,33 @@ export class Locator implements ILocator {
...
+    // the scrollable element that wraps the cell container.
+    private scrollContainerElement: HTMLElement;
+
+    // the element containing all body cells in the grid (excluding headers).
+    private cellContainerElement: HTMLElement;
+
+    public constructor(
+        tableElement: HTMLElement,
+        scrollContainerElement: HTMLElement,
+        cellContainerElement: HTMLElement,
```
giladgray: you can declare `private` _and_ document these properties in signature...

```js
public constructor(
    /** Documentation */
    private property: Type;
) {
```

packages/table/src/locator.ts
```diff
@@ -51,14 +51,33 @@ export class Locator implements ILocator {
...
+    private numFrozenColumns: number;
+
+    // the root table element within which a click is deemed valid and relevant.
+    private tableElement: HTMLElement;
+
+    // the scrollable element that wraps the cell container.
+    private scrollContainerElement: HTMLElement;
+
+    // the element containing all body cells in the grid (excluding headers).
+    private cellContainerElement: HTMLElement;
```
giladgray: why declare these here instead of in constructor which uses 80% fewer lines (1 instead of 5)? I'd understand if you were applying `/** */` comments, tho turns out you can still do that in constructor (see below).

packages/table/src/quadrants/tableQuadrant.tsx
```diff
@@ -99,15 +101,15 @@ export interface ITableQuadrantProps extends IProps {
      * A callback that renders either all of or just frozen sections of the table body.
      */
     renderBody: (
-        quadrantType: QuadrantType,
+        quadrantType?: QuadrantType,
         showFrozenRowsOnly?: boolean,
         showFrozenColumnsOnly?: boolean,
```
giladgray: oof this is messy. why are all the arguments optional? would be more accurate to declare the type as an override accepting either "3 defined params" or "0 params"

packages/table/src/quadrants/tableQuadrantStack.tsx
```diff
@@ -190,46 +228,58 @@ export class TableQuadrantStack extends AbstractComponent<ITableQuadrantStackPro
...
-        this.syncQuadrantSizes();
-        this.syncQuadrantMenuElementWidths();
-        CoreUtils.safeInvoke(this.props.columnHeaderRef, this.findColumnHeader(QuadrantType.MAIN));
-        CoreUtils.safeInvoke(this.props.rowHeaderRef, this.findRowHeader(QuadrantType.MAIN));
+    public componentDidUpdate(prevProps: ITableQuadrantStackProps) {
+        // sync'ing quadrant views triggers expensive reflows, so we only call
+        // it when layout-affecting props change.
+        if (
+            this.props.numFrozenColumns !== prevProps.numFrozenColumns ||
+            this.props.numFrozenRows !== prevProps.numFrozenRows ||
```
giladgray: style: I prefer putting the boolean operator at the _beginning_ of the line so it's immediately recognizable as a continuation of the previous line.

```js
if (
    A
    || B
    || C
    || D
) {
```

packages/table/src/quadrants/tableQuadrantStack.tsx
```diff
@@ -265,7 +315,7 @@ export class TableQuadrantStack extends AbstractComponent<ITableQuadrantStackPro
                 <TableQuadrant
                     grid={grid}
                     isRowHeaderShown={isRowHeaderShown}
-                    onWheel={this.throttledHandleWheel}
+                    onWheel={throttleScrolling ? this.throttledHandleWheel : this.handleWheel}
```
giladgray: _definitely_ add a variable for this next to `onMainQuadrantScroll`! 
I see the same ternary statement repeated four times.

packages/table/src/quadrants/tableQuadrantStackCache.ts
```diff
@@ -0,0 +1,87 @@
...
+
+    public setColumnHeaderHeight(height: number) {
+        this.cachedColumnHeaderHeight = height;
+    }
+
+    public setRowHeaderWidth(width: number) {
+        this.cachedRowHeaderWidth = width;
+    }
+
+    public setQuadrantScrollOffset(quadrantType: QuadrantType, scrollKey: "scrollLeft" | "scrollTop", offset: number) {
```
giladgray: `scrollKey: keyof IScrollOffsetMap` ?

packages/table/src/quadrants/tableQuadrantStack.tsx
cmslewis: Me too, but `prettier` doesn't. :(

packages/table/src/quadrants/tableQuadrantStackCache.ts
cmslewis: Sure, can reuse that. 👍

packages/table/src/quadrants/tableQuadrantStack.tsx
cmslewis: Ah whoops.

packages/table/src/locator.ts
cmslewis: Okey doke.

cmslewis:
@themadcreator yep, that's an unrelated (well, related, but existing) issue already documented in #1573.

packages/table/src/quadrants/tableQuadrant.tsx
cmslewis: Couldn't figure out how to do this syntactically. Leaving as is unless someone can show me how!

cmslewis:
@giladgray @themadcreator @llorca @gscshoyru ready for re-review

packages/table/src/quadrants/tableQuadrant.tsx
gscshoyru: ...you could probably define both as types and then do typea | typeb? What's wrong with that? Other than ugly as all get out.

cmslewis: Tried that; didn't compile.

packages/table/src/headers/_headers.scss
```diff
@@ -261,6 +261,9 @@ $dark-selectable-header-menu-selected-hover-background:
 
 .bp-table-row-headers .bp-table-header {
   min-width: $row-header-min-width;
+  // prevent text from leaking out of the cell during scroll, before the header
+  // width is updated on scroll end.
+  overflow: hidden;
```
cmslewis: `#ThingsToWatchForAfterThisMerges`

 Do people put `inline` popovers or tooltips in their row headers? If so, this might break them. Fortunately, the context menu is _not_ `inline`, so that case is fine.

cmslewis: Actually, these cells were already styled with `Classes.TABLE_TRUNCATED_TEXT`, which includes `overflow: hidden;`, so this is fine.

cmslewis:
:shipit:

cmslewis:
@llorca I left the width-animation in.

llorca:
@cmslewis ha! did you leave it at `$pt-transition-duration`? I had better results earlier with `$pt-transition-duration * 10`, but all in all still felt pretty choppy

packages/table/src/headers/_headers.scss
llorca: I haven't seen `inline` popovers/tooltips in the wild, except with the legacy react-select stuff. Should be safe

PR #1579, opened 2017-09-19
[Labs/Select] Use Popover2 instead of Popover in Select

### LINKED ISSUE #1512: [Labs/Select] Change Popover by Popover2
I would like to have Popover2 instead of Popover in the Select component on Lab, with the option `placement="bottom-start"`. That way it will handle automatically positioning of the popover, right now if the select component is on the right of the screen the popover will appear cut on the right.

brieb:


llorca:
🚫  Known bug appearing again: when you click on the button and the popover opens, the page scrolls all the way back up. @giladgray knows a lot about this

There's also a problem with the spacing between the minimal popover and its target:
image
It should be 0 in any direction, but it turns out even the `Popover2` docs has problems with `left` placement (1px spacing instead of 0):
- bad:
image
- good:
image

acronoah:
I need this for stretching the select in my use case.  Was hoping `popoverProps={{placement:'bottom-stretch'}}` would work, but no dice as it's the old popover...

brieb:
@giladgray Fixed the scroll issue in a separate PR: https://github.com/palantir/blueprint/pull/1598

packages/labs/src/components/select/select.tsx
```diff
@@ -290,6 +297,13 @@ export class Select<T> extends React.Component<ISelectProps<T>, ISelectState<T>>
             this.list.scrollActiveItemIntoView();
         }
 
+        requestAnimationFrame(() => {
+            const { inputProps = {} } = this.props;
+            if (inputProps.autoFocus !== false && this.input != null) {
```
cmslewis: FYI I had to mull over this logic in my first glance. could use a quick comment like `// autofocus by default` I think.

packages/labs/test/selectTests.tsx
```diff
@@ -5,14 +5,14 @@
...
-import { Classes, InputGroup, Popover } from "@blueprintjs/core";
+import { Classes, InputGroup } from "@blueprintjs/core";
 import { assert } from "chai";
 import * as classNames from "classnames";
 import { mount } from "enzyme";
 import * as React from "react";
 
 import { Film, TOP_100_FILMS } from "../examples/data";
-import { ISelectItemRendererProps, ISelectProps, Select } from "../src/index";
+import { ISelectItemRendererProps, ISelectProps, Popover2, Select } from "../src/index";
```
cmslewis: nit: Import from the source file, not from `index`.

brieb: That's how we do it in the other tests. Is index okay for tests?

brieb:
So, I tried looking into the off-by-one-px issue, and I narrowed it down to a bit...

Could be something to do with pixel boundaries (if you resize the page slowly you can see it snap in and out of place).

sep-27-2017 00-02-34

And/or the scrollbar might have something to do with it...

Before:
screen shot 2017-09-26 at 10 27 29 pm

After unconditionally adding [`document.body.classList.add(Classes.OVERLAY_OPEN);`](https://github.com/palantir/blueprint/blob/5626890d117b4c21164eb932980c4f76e847e6aa/packages/core/src/components/overlay/overlay.tsx#L295) which applies 
```
body.pt-overlay-open {
  overflow: hidden;
}
```
screen shot 2017-09-26 at 10 28 03 pm

Easier to see with a border around the overlay (as it's actually a shadow)

Before:
screen shot 2017-09-26 at 10 29 30 pm

After:
screen shot 2017-09-26 at 10 30 01 pm

Not sure what the best solution is here.

Can we file a separate issue for this?
Is this considered blocking?

llorca:
I don't seen any urgency with merging this PR, and consider this visual regression blocking. Since we don't have much firepower on core as of now, I'd rather wait until we can address the remaining `Popover2` issues before starting to integrate it in a bunch of places.

The left alignment you're showing is one issue that can unfortunately be observed across other placements in `Popover2`... In your recent screenshots, I'm also seeing a second issue where there's 4px space between the target and the popover, which isn't expected either.

cmslewis:
@llorca I'm wondering if we should go ahead and merge this with the visual regression present after all. I think of it as whack-a-mole where the mole we whack with this PR is much bigger than the mole than pops up again. Thoughts?

packages/labs/test/selectTests.tsx
giladgray: index is fine for tests, I guess, cuz it's how we've always done it. not a big deal.

packages/labs/examples/selectExample.tsx
```diff
@@ -62,7 +62,7 @@ export class SelectExample extends BaseExample<ISelectExampleState> {
                 itemRenderer={this.renderFilm}
                 noResults={<MenuItem disabled={true} text="No results." />}
                 onItemSelect={this.handleValueChange}
-                popoverProps={{ popoverClassName: minimal ? Classes.MINIMAL : "" }}
+                popoverProps={{ minimal }}
```
giladgray: check it out: this is the fix for that pesky 4px gap! as in, we already had a fix in this new `minimal` prop (instead of the className-only approach on the left), just had to use it properly. 

cc @brieb

brieb: Oh nice -- thanks! :)

cmslewis:
Still seeing a vertical edge misalignment based on page width:

giladgray:
@cmslewis I don't think that's an actual issue... happens with `Popover2` itself (tho harder to see). pretty sure it's Popper.js "doing the right thing" and only setting integer offsets (whereas scrolling happily supports sub-pixels).

cmslewis:
I said in an earlier comment that I'm okay with it: https://github.com/palantir/blueprint/pull/1579#issuecomment-335229598

so I'm happy to merge in light of the other improvements here.

PR #1584, opened 2017-09-20
[Table] Fix DragSelectable context-menu behavior on Ctrl + Click

### LINKED ISSUE #1549: [Table] Fix cmd+click context-menu selection behavior
### Bug report

- __Package version(s)__: v1.27.0

### Description

Right-clicking a cell and cmd+clicking a cell should behave in exactly the same way. They don't right now. See: https://github.com/palantir/blueprint/pull/1523#issuecomment-328594232.

Labeling as P1 so I return to this quickly.

cmslewis:
#### Checklist

- [x] ~Include tests~ (turns out Ctrl + Click doesn't trigger onContextMenu in Phantom 🙁) 

- 🐛 __FIXED__ `DragSelectable` now responds to <kbd>Ctrl + Click</kbd> the same way it does to right-click (vis-a-vis triggering context menus).

- Should we try to handle <kbd>Ctrl + Click</kbd> events from `ContextMenuTarget` directly? That way no one ever has to worry about them behaving differently (in theory).

### DIFF
--- a/packages/table/src/interactions/selectable.tsx
+++ b/packages/table/src/interactions/selectable.tsx
@@ -217,8 +217,13 @@ export class DragSelectable extends React.Component<IDragSelectableProps, {}> {
     private shouldIgnoreMouseDown(event: MouseEvent) {
         const { ignoredSelectors = [] } = this.props;
         const element = event.target as HTMLElement;
+
+        const isLeftClick = Utils.isLeftClick(event);
+        const isContextMenuTrigger = isLeftClick && event.ctrlKey;
+
         return (
-            !Utils.isLeftClick(event) ||
+            !isLeftClick ||
+            isContextMenuTrigger ||
             this.props.disabled ||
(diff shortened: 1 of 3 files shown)


llorca:
> Should we try to handle <kbd>Ctrl + Click</kbd> events from `ContextMenuTarget` directly? That way no one ever has to worry about them behaving differently (in theory).
Yes, most definitely, but it's strange that it's not already the case. It's a system thing (read: OS), which in theory we don't need to worry about. Maybe there's a bug to fix instead?

cmslewis:
@llorca - it might not be possible to do cleanly, actually. The browser _does_ trigger `onContextMenu` on <kbd>Ctrl + Click</kbd>, so I don't want to add a bunch of explicit logic in the `ContextMenuTarget` to make that happen doubly so. Then again, without that logic, the unit tests for <kbd>Ctrl + Click</kbd> won't pass.

PR #1586, opened 2017-09-20
[File Upload Input] add height to file upload, when using .pt-large class

Landish:

Hi, first of all, thanks for having such a great library  🤗

Second of all, when working on the File upload element, I noticed that, it might be missing `height`, when using `.pt-large` CSS class.

Not sure if this is a bug, or it was left out initially, but I felt, I had to PR.  

#### Screenshot
before

after

### DIFF
--- a/packages/core/src/components/forms/_file-upload.scss
+++ b/packages/core/src/components/forms/_file-upload.scss
@@ -63,6 +63,10 @@ $file-upload-button-padding-large: ($pt-input-height-large - $pt-button-height) 
   &.pt-fill {
     width: 100%;
   }
+
+  &.pt-large {
+    height: $pt-input-height-large;
+  }
 }
 
 .pt-file-upload-input {


palantirtech:
Thanks for your interest in [palantir/blueprint](https://github.com/palantir/blueprint), @Landish! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

packages/core/src/components/forms/_file-upload.scss
```diff
@@ -63,6 +63,10 @@ $file-upload-button-padding-large: ($pt-input-height-large - $pt-button-height)
   &.pt-fill {
     width: 100%;
   }
+
+  &.pt-large {
```
cmslewis: What about the `.pt-large &` case? See [L103](https://github.com/palantir/blueprint/pull/1586/files#diff-fa25bb68536fc9097a2548cf91d320e1R103).

giladgray: @cmslewis not sure that's relevant actually... this line applies to a different element than L103.

cmslewis: Right, but when the `pt-file-upload` element is made to be large via cascading styles (by applying `pt-large` to a parent), then `&.pt-large` styles won't increase this element's height in that case. Should it?

I linked the L103 as an example of the selector I think we should use if so.

llorca: I just tested it and it looks like we should @cmslewis (cascaded `.pt-large` changes the visual size but not the actual size):
screen shot 2017-09-20 at 3 41 33 pm

cmslewis: @Landish - per @llorca's investigation above, can you change the selector to `.pt-large &`?

Landish: @cmslewis 

So, you're saying to just change selector and everything should work? 

That selector compiles to this code:

screen shot 2017-09-21 at 14 05 36

Just confirm it and I'll update the PR.

cmslewis: Yep, that's correct, reasoning being that `pt-file-upload` is one of the several components that can receive Large styling in two ways: 

- by adding `pt-large` directly to it (`&.pt-large`)
- by adding `pt-large` to an ancestor (`.pt-large &`)

We should fix the height in both cases.

llorca:
Closing in favor of https://github.com/palantir/blueprint/pull/1600

Issue #1108, opened 2017-05-17
Refactor Table API to only require a single top-level component


adidahiya:
Right now, tables must be constructed using an API like this:

```ts
<Table>
     <Column .../>
     <Column .../>
     <Column .../>
</Table>
```

This works fine for static tables like the ones in the docs site and examples page, but it's not great for applications that have dynamic numbers of columns. The compiler doesn't do a good job of type inference when it comes to React children. I can't even do something like this:

```ts
const columns = someArray.map((e) => {
    return <Column .../>; // assume the right props are provided here
});

return <Table>{columns}</Table>;
```

This fails with:

![image](https://cloud.githubusercontent.com/assets/723999/26166378/4f9139de-3b01-11e7-9534-8a80b9854ec5.png)

Another thing this refactor opens up is easier API support for merged cells / headers / etc. [Why I wrote React-Table ...](https://medium.com/@tannerlinsley/why-i-wrote-react-table-and-the-problems-it-has-solved-for-nozzle-others-445c4e93d4a8) is a good read for understanding these API decisions.

Discussed offline with @gscshoyru and @cmslewis

crispamares:
FYI I've just updated a project from Typescript 2.8 to Typescript 3.5 and this error rises even with multiple fixed columns, (it works when passing just 1 column):

```jsx
<Table
	className="vertical-align"
	defaultRowHeight={32}
	defaultColumnWidth={224}
	enableColumnResizing={false}
	enableRowHeader={false}
	numRows={data.length}
>
	<Column cellRenderer={this.renderCell} name={keyColumnName} />
	<Column cellRenderer={this.renderCell} name={valueColumnName} />
</Table>
```

getdir:
P2 for 4.5 years. woot

PR #1587, opened 2017-09-20
[Table] Validation improvements

### LINKED ISSUE #1371: [Table] Table silently ignores `rowHeights` prop if it has the wrong length
```
    /**
     * Takes in one full array of values and one sparse array of the same
     * length and type. Returns a copy of the `defaults` array, where each
     * value is replaced with the corresponding non-null value at the same
     * index in `sparseOverrides`.
     *
     * @param defaults - the full array of default values
     * @param sparseOverrides - the sparse array of override values
     */
    assignSparseValues: function (defaults, sparseOverrides) {
        if (sparseOverrides == null || defaults.length !== sparseOverrides.length) {
            return defaults;
        }
        defaults = defaults.slice();
        for (var i = 0; i < defaults.length; i++) {
            var override = sparseOverrides[i];
            if (override != null) {
                defaults[i] = override;
            }
        }
        return defaults;
    },
```

It should throw an error, otherwise it can be confusing to debug why it isn't working.

cmslewis:
#### Checklist

- [x] Include tests

- 🐛  __FIXED__ `Table` now validates props on mount _and_ on update.
- 🔄 __CHANGED__ `Table` now validates props as follows (assuming each prop involved is defined):
    - Errors:
        - `numRows < 0`
        - `numFrozenRows < 0`
        - `numFrozenColumns < 0`
        - `numRows !== rowHeights.length`
        - `numColumns !== columnWidths.length`
        - `non-<Column> child` (*)
    - Warnings:
        - `numFrozenRows > numRows` (changed the message to mention that we'll clamp for you)
        - `numFrozenColumns > numColumns` (same)

- How does the error-message language look?
- (*) Is this the right choice? Per my code comment, we're currently just printing a warning, but the Table would eventually throw a convoluted error from deep inside when processing non-`<Column>` children, so I figure we might as well just throw a clear error at the outset.

#### Screenshot

```
Validation
  on mount
    errors
      ✔ throws an error if numRows < 0
      ✔ throws an error if numFrozenRows < 0
      ✔ throws an error if numFrozenColumns < 0
      ✔ throws an error if rowHeights.length !== numRows
      ✔ throws an error if columnWidths.length !== number of <Column>s
      ✔ throws an error if a non-<Column> child is provided
    warnings
      ✔ should print a warning when numFrozenRows > numRows
      ✔ should print a warning when numFrozenColumns > num <Column>s
  on update
    errors
      ✔ on numRows update, throws an error if numRows < 0
      ✔ on numFrozenRows update, throws an error if numFrozenRows < 0
      ✔ on numFrozenColumns update, throws an error if numFrozenColumns < 0
      ✔ on numRows update, throws an error if rowHeights.length !== numRows
      ✔ on rowHeights update, throws an error if rowHeights.length !== numRows
      ✔ on children update, throws an error if columnWidths.length !== number of <Column>s
      ✔ on columnWidths update, throws an error if columnWidths.length !== number of <Column>s
      ✔ on children update, throws an error if a non-<Column> child is provided
    warnings
      ✔ on numFrozenRows update, should print a warning when numFrozenRows > numRows
      ✔ on numRows update, should print a warning when numFrozenRows > numRows
      ✔ on numFrozenColumns update, should print a warning when numFrozenColumns > num <Column>s
      ✔ on children update, should print a warning when numFrozenColumns > num <Column>s
      ✔ should print a warning when numFrozenColumns > num <Column>s
```


packages/table/src/table.tsx
```diff
@@ -524,6 +524,8 @@ export class Table extends AbstractComponent<ITableProps, ITableState> {
     }
 
     public componentWillReceiveProps(nextProps: ITableProps) {
+        super.componentWillReceiveProps(nextProps);
```
cmslewis: #whoops

packages/table/src/table.tsx
```diff
@@ -685,29 +687,49 @@ export class Table extends AbstractComponent<ITableProps, ITableState> {
...
         React.Children.forEach(children, (child: React.ReactElement<any>) => {
             // save as a variable so that union type narrowing works
             const childType = child.type;
 
-            if (typeof childType === "string") {
-                console.warn(Errors.TABLE_NON_COLUMN_CHILDREN_WARNING);
+            // the second part of this conditional will never be true, but it
+            // informs the TS compiler that we won't be invoking
+            // childType.prototype on a "string" element.
+            if (typeof child === "string" || typeof childType === "string") {
```
cmslewis: Janky, but it was throwing an error before I added the `typeof child === "string"` check. Turns out `typeof childType === undefined` for strings, so the check wasn't working as expected before.

packages/table/src/table.tsx
```diff
@@ -685,29 +687,49 @@ export class Table extends AbstractComponent<ITableProps, ITableState> {
...
+                throw new Error(Errors.TABLE_NON_COLUMN_CHILDREN_WARNING);
             } else {
                 const isColumn = childType.prototype === Column.prototype || Column.prototype.isPrototypeOf(childType);
                 if (!isColumn) {
-                    console.warn(Errors.TABLE_NON_COLUMN_CHILDREN_WARNING);
+                    throw new Error(Errors.TABLE_NON_COLUMN_CHILDREN_WARNING);
```
cmslewis: Is there ever a case where we'd want to let people wrap child `<Column>`s with a thing? If not, would prefer this to be an `Error`.

packages/table/src/table.tsx
gscshoyru: Womp

packages/table/src/table.tsx
gscshoyru: ...so it's a... typeguard? That's... weird. I am confused as to how this could possibly be working. But... ok?

packages/table/src/table.tsx
gscshoyru: I think you're probably right, @adidahiya might know if the other ever comes up? Not sure.

packages/table/src/table.tsx
cmslewis: `typeof childType === "string"` is the typeguard, yeah. But it never returns `true`. @giladgray, do you know a way out of this comical quagmire?

cmslewis:
Adding @adidahiya for CR because you've used the `Table` in the wild.

packages/table/src/table.tsx
adidahiya: do we really need to validate the react children? what does this buy us? what's the harm in being lenient? please keep https://github.com/palantir/blueprint/issues/1108 in mind; enforcing constraints on React children in typescript usually gets in the way more often than it helps.

cmslewis: `table.tsx` accesses `props.children` expecting it to be an array of `<Column>s` (e.g. in `Table.getColumnProps`). If that accessing fails, the error message is somewhat convoluted; throwing a clear error from `validateProps` would be more helpful.

Only reservation is that maybe there are cases when people would want to compose a basic `<Column>` into something else they'd want to provide as a direct child of `Table`? In that case, a duck-typing approach would seem preferable (i.e. as long as the child in question supports the `Column` API, we don't care).

I've never heard of anyone doing the above customization, so I'm including the error check as a result of my first point. ^ Thoughts?

adidahiya: for now I'm fine with enforcing `Column` to be a direct child of `Table` -- as you said we haven't seen the wrapping component use case come up. makes sense to have a better error message and halt execution since the table won't render correctly. let's get to #1108 soon for 2.0.

cmslewis: 👍

PR #1592, opened 2017-09-21
[Table] Fix onCompleteRender for RenderMode.BATCH

cmslewis:
#### Checklist

- [x] Include tests

- 🐛 __FIXED__ `Table` now fires `onCompleteRender` as expected when using `RenderMode.BATCH`.

Not sure how this broke originally. :/ I tried and failed to add a test for a multi-frame batch render, so I added/updated a couple simple tests for now.

### DIFF
--- a/packages/table/src/tableBody.tsx
+++ b/packages/table/src/tableBody.tsx
@@ -133,7 +133,6 @@ export class TableBody extends React.Component<ITableBodyProps, {}> {
 
     private activationCell: ICellCoordinates;
     private batcher = new Batcher<JSX.Element>();
-    private isRenderingBatchedCells = false;
 
     public componentDidMount() {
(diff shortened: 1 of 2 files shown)


packages/table/src/tableBody.tsx
```diff
@@ -133,7 +133,6 @@ export class TableBody extends React.Component<ITableBodyProps, {}> {
 
     private activationCell: ICellCoordinates;
     private batcher = new Batcher<JSX.Element>();
-    private isRenderingBatchedCells = false;
```
cmslewis: This variable used to seem relevant, but it wasn't (and it was never being set to `true` anyway, whoops). When `renderMode === RenderMode.BATCH`, you can guarantee the batched rendering will trigger; so all you care about from there is whether it's done yet.

PR #1594, opened 2017-09-21
[Table] Make editable cells only show up when they're being edited

gscshoyru:
Editable cells now act like "normal" cells under most circumstances, only acting like editable ones when in edit-mode. This allows us to do... less of the hacky things we were doing that made interactions break when editable cells existed. Now, you can click on cells and use keyboard shortcuts and other microinteractions, even if those cells are editable!

I cleaned up a bunch of unnecessary stuff, or what looked to be it. If I got rid of something important, let me know.


llorca:
- Double click to edit is quite inconsistent:

- Hitting a character while a cell is focused should bring said cell into editable mode.

- Should hitting `esc` after typing some text cancel text insertion? It doesn't at the moment, and FWIW Excel does. To be clear, hitting `esc` in Excel resets the cell to its previous state, whether there was text or not.

- When the focus cell is enabled, we should preserve `enter`/`shift+enter`/`tab`/`shift+tab` interaction while editing cells.

tgreenwatts:
I managed to select one focus and edit the other! Just after clicking around a lot, so maybe not an issue (I think this is focus rather than selected)
image

llorca:
^ On that note, I'm fairly confident that a cell that's being edited has to be focused. The desync here doesn't make sense

gscshoyru:
...to be clear, folks @llorca specifically -- I was _starting_ to do all of that and realized that I should get the basic idea out there first. This PR is not intended to make editable cells function the way they should with everything else, yet. All this PR does is make it so that editable cells _stop_ blocking all keyboard shortcuts when they exist. Doing all those other things is a bigger project and needs doing, but the thing blocking _all_ of them, for the longest time, was this little nightmare of browser focus stuff. I don't think those other things belong in this PR, I think they belong in later PRs. Does that make sense?

packages/table/test/editableCellTests.tsx
```diff
@@ -26,7 +26,7 @@ describe("<EditableCell>", () => {
 
     it("renders", () => {
         const elem = harness.mount(<EditableCell value="test-value-5000" />);
-        expect(elem.find(".pt-editable-content").text()).to.equal("test-value-5000");
+        expect(elem.find(".bp-table-truncated-text").text()).to.equal("test-value-5000");
```
cmslewis: `.${Classes.TABLE_TRUNCATED_TEXT}`

packages/table/test/editableCellTests.tsx
```diff
@@ -39,10 +39,10 @@ describe("<EditableCell>", () => {
         const VALUE_2 = "bar";
 
         const elem = mount(<EditableCell value={VALUE_1} />);
-        expect(elem.find(".pt-editable-content").text()).to.equal(VALUE_1);
+        expect(elem.find(".bp-table-truncated-text").text()).to.equal(VALUE_1);
```
cmslewis: same (etc.)

packages/table/src/cell/editableCell.tsx
```diff
@@ -77,31 +82,46 @@ export class EditableCell extends React.Component<IEditableCellProps, IEditableC
...
+            cellContents = <div className={textClasses}>{savedValue}</div>;
+        }
+
         return (
             <Cell {...spreadableProps} truncated={false} interactive={interactive}>
                 <Draggable
                     onActivate={this.handleCellActivate}
                     onDoubleClick={this.handleCellDoubleClick}
-                    preventDefault={!interactive}
+                    preventDefault={false}
```
cmslewis: Would appreciate a short comment as to why this is necessary.

gscshoyru: ...because for some reason, on draggable, the default for preventDefault is _true_ . Which seems really dumb to me. If you know why this is, let me know?

llorca: @cmslewis any more thoughts?

cmslewis: Nah.

cmslewis: @gscshoyru I meant a short comment in the code. :) No worries though.

cmslewis:
@gscshoyru would it be smarter to merge this into a feature branch instead of `master`, or are you confident this won't break/worsen things in the interim?

gscshoyru:
I have no good reason to believe this would break anything? But who knows. Possibly we should ask some folks internally that we know are using editable cells... if there are any. I'm reasonably but not 100% confident this won't break anything.

cmslewis:
@gscshoyru I'm gonna take another look and then probably merge this in a moment.

cmslewis:
So the miss rate when double-clicking to edit does appear to have increased significantly. @gscshoyru can you either investigate that before merging this PR, or can you merge this into a `feature/table-editing` feature branch?

gscshoyru:
I did do some looking into that, as far as I can tell, every single miss isn't something _directly_ to do with the new code, because "handleCellDoubleClick" doesn't fire. I have no idea why it doesn't, yet. I can try to look into this before merging, now, I was saving the full investigation until after -- but if we're worse (which we do seem to be) I'll try and do that first.

llorca:
I mean, it might be slightly worst now but it's still unusable. Maybe we merge that and then properly work on all the editable cell interactions?

gscshoyru:
@cmslewis and everyone else -- the double-clicking on cells to edit should work, now. The problem was that the fact that it switches between editable and div, now, the click handlers were getting lost on state flip. Now we reattach _after_ render, and not before, in case the div changes (which makes perfect sense and I have no idea why we didn't do in the first place). Let me know if this approach is wrong and should be changed in some fashion.

llorca:
Are we sure those handlers are attached to the entire area of the cell? I'm still getting < 50% success rate getting into an editable state:

cmslewis:
Double-clicking feels good to me now. 👍  Any tests we need to add to verify the new `<Draggable>` change?

cmslewis:
(p.s. @gscshoyru you have merge conflicts)

gscshoyru:
@llorca I'm not seeing that at all. How are you getting it to fail at all?

In any case: I'm reasonably certain that even if that was still messed up somehow, unlike the other thing, it wouldn't be a regression the way the other thing was.

llorca:
@gscshoyru probably not a regression, but it's still bad -- can fix later when we do a real pass on editable interactions. I just don't get how clicking on the padding of a cell selects it, but double clicking on it doesn't give editable mode...

gscshoyru:
@llorca -- oh, I see the behavior you're talking about now. You've gotta be _really_ close to the edge. 

That said this is going to be... very hard to fix, if possible at all. The reason why the behaviors are different is that the selection is handled by a `Draggable` on the _entire_ table body, which uses coordinates to figure out the relevant things. Selection there isn't done on a per-cell level, but over the whole table, because the cells don't care whether they're selected or not; it makes no difference in rendering them. In the editable cells case, the `Draggable` sits inside the cell itself, and inside whatever padding that exists. Modifying this would be extremely hard, and while we could check for the double-clicks at the same toplevel, and plausibly should... we'd have to migrate down the "is this cell being edited" downwards to the cell itself via props, and do some weird manipulations. Fixing what you're saying isn't going to be straightforward, but hopefully you should see why now at least, yes?

cmslewis:
@gscshoyru good to merge?

gscshoyru:
Yup!

PR #1596, opened 2017-09-21
[Table] Introduce TableBodyCells component to avoid unnecessary cell renders

cmslewis:
#### Checklist

- [x] ~Include tests~ (no need; just an under-the-hood refactor)

_Before_:
- `TableBody` used to invoke `props.renderCell` for every cell in view, every time `selectedRegions` changed, just to run a diff against the currently rendered cells. That's a lot of wasted work for an interaction that doesn't affect the cells themselves.

_After:_
- The `TableBodyCells` component now encapsulates all cell rendering and doesn't respond to `selectedRegions` changes. This means that we now do _zero_ work in the `TableBody` when that prop changes. Woo!
    - This also greatly simplifies our `shouldComponentUpdate` logic in `TableBody`.

- Look at the React Dev Tools in Chrome, and tick the "Highlight updates" checkbox.
- Click to move the selection around.
- Note that nothing in the table body is highlighted, because nothing in the table body repaints on the `selectedRegions` change anymore.


cmslewis:
```
FAILED TESTS:
    onCompleteRender
      ✖ triggers immediately on mount/update with RenderMode.BATCH for very small batches
```

Hmm, will look into this after lunch.

PR #1597, opened 2017-09-22
[Labs/Select] Support for controlled select query input

brieb:

### DIFF
--- a/packages/labs/src/components/select/select.tsx
+++ b/packages/labs/src/components/select/select.tsx
@@ -47,7 +47,7 @@ export interface ISelectProps<T> extends IListItemsProps<T> {
 
     /**
      * Current query value, for controlled usage.
-     * Providing this prop will put the component in controlled mode.
+     * Providing this prop will control the query input.
      */
     query?: string;
 
@@ -163,7 +163,7 @@ export class Select<T> extends React.Component<ISelectProps<T>, ISelectState<T>>
 
     public componentWillReceiveProps(nextProps: ISelectProps<T>) {
         const nextQuery = getQuery(nextProps);
-        if (nextQuery !== undefined && this.state.query !== nextQuery) {
+        if (nextQuery !== undefined && nextQuery !== this.state.query) {
             this.setState({ query: nextQuery });
         }
     }
@@ -178,15 +178,15 @@ export class Select<T> extends React.Component<ISelectProps<T>, ISelectState<T>>
         // not using defaultProps cuz they're hard to type with generics (can't use <T> on static members)
         const { filterable = true, disabled = false, inputProps = {}, popoverProps = {} } = this.props;
 
-        const { ref, value, ...htmlInputProps } = inputProps;
+        const { ref, ...htmlInputProps } = inputProps;
         const input = (
             <InputGroup
                 autoFocus={true}
                 leftIconName="search"
                 placeholder="Filter..."
                 rightElement={this.maybeRenderInputClearButton()}
-                value={listProps.query}
                 {...htmlInputProps}
+                value={listProps.query}
                 onChange={this.handleQueryChange}
             />
         );
@@ -310,31 +310,34 @@ export class Select<T> extends React.Component<ISelectProps<T>, ISelectState<T>>
     };
 
     private handleQueryChange = (event: React.FormEvent<HTMLInputElement>) => {
-        const { inputProps = {}, onQueryChange } = this.props;
-        const query = event.currentTarget.value;
+        const { inputProps = {} } = this.props;
+        Utils.safeInvoke(inputProps.onChange, event);
+        this.setQuery(event.currentTarget.value);
+    };
+
+    private resetQuery = () => {
+        const { items } = this.props;
+        this.setState({ activeItem: items[0] });
+        this.setQuery("");
+    };
+
+    private setQuery(query: string): void {
         // Only set the query state ourselves if not controlled
         if (getQuery(this.props) === undefined) {
             this.setState({ query });
         }
-        Utils.safeInvoke(inputProps.onChange, event);
-        Utils.safeInvoke(onQueryChange, query);
-    };
 
-    private resetQuery = () => {
-        const { items, onQueryChange } = this.props;
-        const query = "";
-        this.setState({ activeItem: items[0], query });
-        Utils.safeInvoke(onQueryChange, query);
-    };
+        Utils.safeInvoke(this.props.onQueryChange, query);
+    }
 }
 
-function getQuery(props: ISelectProps<any>, fallback?: string): string | undefined {
+function getQuery(props: ISelectProps<any>, defaultValue?: string): string | undefined {
     const { inputProps = {} } = props;
     if (props.query !== undefined) {
         return props.query;
     } else if (inputProps.value !== undefined) {
         return inputProps.value;
     } else {
-        return fallback;
+        return defaultValue;
     }
 }
--- a/packages/labs/src/components/select/select.tsx
+++ b/packages/labs/src/components/select/select.tsx
@@ -133,7 +133,7 @@ export class Select<T> extends React.Component<ISelectProps<T>, ISelectState<T>>
     constructor(props?: ISelectProps<T>, context?: any) {
         super(props, context);
 
-        this.state = { isOpen: false, query: getQuery(props) };
+        this.state = { isOpen: false, query: getQuery(props, "") };
     }
 
     public render() {
@@ -163,7 +163,7 @@ export class Select<T> extends React.Component<ISelectProps<T>, ISelectState<T>>
 
     public componentWillReceiveProps(nextProps: ISelectProps<T>) {
         const nextQuery = getQuery(nextProps);
-        if (this.state.query !== nextQuery) {
+        if (nextQuery !== undefined && this.state.query !== nextQuery) {
             this.setState({ query: nextQuery });
         }
     }
@@ -312,7 +312,10 @@ export class Select<T> extends React.Component<ISelectProps<T>, ISelectState<T>>
     private handleQueryChange = (event: React.FormEvent<HTMLInputElement>) => {
         const { inputProps = {}, onQueryChange } = this.props;
         const query = event.currentTarget.value;
-        this.setState({ query });
+        // Only set the query state ourselves if not controlled
+        if (getQuery(this.props) === undefined) {
+            this.setState({ query });
+        }
         Utils.safeInvoke(inputProps.onChange, event);
         Utils.safeInvoke(onQueryChange, query);
     };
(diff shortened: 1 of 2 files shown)


packages/labs/src/components/select/select.tsx
```diff
@@ -46,6 +46,18 @@ export interface ISelectProps<T> extends IListItemsProps<T> {
...
 
     /**
+     * Current query value, for controlled usage.
+     * Providing this prop will put the component in controlled mode.
+     */
+    query?: string;
+
+    /**
+     * Callback invoked when the query value changes,
+     * through user input or when the filter is reset.
```
cmslewis: We're woefully inconsistent at this, but can we align documentation to 80 chars? There's a VSCode plugin to help (`Align`).

packages/labs/src/components/select/select.tsx
```diff
@@ -149,6 +161,13 @@ export class Select<T> extends React.Component<ISelectProps<T>, ISelectState<T>>
         );
     }
 
+    public componentWillReceiveProps(nextProps: ISelectProps<T>) {
+        const nextQuery = getQuery(nextProps);
+        if (nextQuery !== undefined && this.state.query !== nextQuery) {
```
cmslewis: nit: with `if` statements, compare the new thing to the existing thing:

**`nextQuery !== this.state.query`** instead of ~`this.state.query !== nextQuery`~

packages/labs/src/components/select/select.tsx
```diff
@@ -291,10 +310,31 @@ export class Select<T> extends React.Component<ISelectProps<T>, ISelectState<T>>
...
+        }
         Utils.safeInvoke(inputProps.onChange, event);
+        Utils.safeInvoke(onQueryChange, query);
     };
 
-    private resetQuery = () => this.setState({ activeItem: this.props.items[0], query: "" });
+    private resetQuery = () => {
+        const { items, onQueryChange } = this.props;
+        const query = "";
+        this.setState({ activeItem: items[0], query });
```
cmslewis: Seems weird to let the query be mutated in controlled mode. Shouldn't we just emit the _desired_ new query via `onQueryChange`, and then let the caller change the controlled value? Otherwise we get into a weird hybrid place.

packages/labs/src/components/select/select.tsx
```diff
@@ -291,10 +310,31 @@ export class Select<T> extends React.Component<ISelectProps<T>, ISelectState<T>>
...
+        Utils.safeInvoke(onQueryChange, query);
+    };
+}
+
+function getQuery(props: ISelectProps<any>, fallback?: string): string | undefined {
```
cmslewis: I'd rename `fallback` => `defaultValue` for naming consistency with the rest of the codebase.

packages/labs/test/selectTests.tsx
```diff
@@ -54,7 +53,8 @@ describe("<Select>", () => {
     });
 
     it("itemRenderer is called for each filtered child", () => {
-        select({}, "1999");
+        const wrapper = select({ query: "" });
+        wrapper.setProps({ query: "1999" });
```
cmslewis: Modifying the query via a prop (i.e. controlling it) isn't necessarily going to test the same thing as modifying the query via `setState`. We should keep these uncontrolled tests uncontrolled and have explicit tests for controlled mode elsewhere in the file.

packages/labs/test/selectTests.tsx
```diff
@@ -98,39 +98,127 @@ describe("<Select>", () => {
...
         assert.isTrue(popoverWillOpen.calledOnce);
     });
 
+    it("popover can be controlled with popoverProps.isOpen", () => {
+        const wrapper = select({ popoverProps: { isOpen: false } });
+        wrapper.find("table").simulate("click");
+        assert.strictEqual(wrapper.find(Popover).prop("isOpen"), false);
+        wrapper.setProps({ popoverProps: { isOpen: true } });
+        assert.strictEqual(wrapper.find(Popover).prop("isOpen"), true);
+    });
```
cmslewis: There are enough tests in here now that the file would benefit from an extra layer of `describe` blocks:

```tsx
describe("<Select>", () => {
    describe("Basic behavior");
    /** When query is not provided via props. */
    describe("Uncontrolled usage");
    /** When query is provided via props. */
    describe("Controlled usage");
});

File: packages/labs/test/selectTests.tsx [hunk 7]
```diff
@@ -98,39 +98,127 @@ describe("<Select>", () => {
...
+
     it("returns focus to focusable target after popover closed");
 
-    function select(props: Partial<ISelectProps<Film>> = {}, query?: string) {
-        const wrapper = mount(
+    function select(props: Partial<ISelectProps<Film>> = {}) {
```
cmslewis: Per my comment above, would prefer to undo changes to this function to let uncontrolled mode remain testable.

packages/labs/src/components/select/select.tsx
```diff
@@ -46,6 +46,18 @@ export interface ISelectProps<T> extends IListItemsProps<T> {
     itemRenderer: (itemProps: ISelectItemRendererProps<T>) => JSX.Element;
 
     /**
+     * Current query value, for controlled usage.
+     * Providing this prop will put the component in controlled mode.
```
giladgray: not the entire component, just this one feature... "Providing this prop will control the query input."

packages/labs/src/components/select/select.tsx
giladgray: I'm fine with this. 80 chars would require two lines anyway and this is just so handsomely formatted 😍

packages/labs/src/components/select/select.tsx
```diff
@@ -159,7 +178,7 @@ export class Select<T> extends React.Component<ISelectProps<T>, ISelectState<T>>
         // not using defaultProps cuz they're hard to type with generics (can't use <T> on static members)
         const { filterable = true, disabled = false, inputProps = {}, popoverProps = {} } = this.props;
 
-        const { ref, ...htmlInputProps } = inputProps;
+        const { ref, value, ...htmlInputProps } = inputProps;
```
giladgray: move the `value` prop below the spread (swap lines 188 and 189) and you won't have to do this.

how to JSX spread:

```jsx
<Tag
    overridableProps...
    {...spread}
    notOverridableProps...
/>
```

packages/labs/src/components/select/select.tsx
giladgray: might be useful to introduce a helper function that can be used by both `resetQuery` and `handleQueryChange` to ensure the correct events are called.

packages/labs/test/selectTests.tsx
```diff
@@ -98,39 +98,127 @@ describe("<Select>", () => {
         assert.strictEqual(wrapper.state("query"), "");
     });
 
+    it("if query is non-empty, the input value will stay in sync with the query prop", () => {
```
giladgray: "if query [prop] is non-empty" ? can't tell what you're trying to achieve from title.

packages/labs/test/selectTests.tsx
```diff
@@ -98,39 +98,127 @@ describe("<Select>", () => {
...
+        const query = "nailed it";
+        const wrapper = select({ query });
+        const input = wrapper.find("input");
+
+        assert.equal(input.prop("value"), query);
+        assert.equal(wrapper.state("query"), query);
+
+        (input.getDOMNode() as HTMLInputElement).value = "some other value";
+        input.simulate("change");
```
giladgray: I'm not sure this works like you think it does. I thought you had to specify event data for anything to happen... 

`input.simulate("change", { currentTarget: { value: "..." } })` 

either way, I find it cleaner and less hacky to use the eventData approach.

packages/labs/test/selectTests.tsx
```diff
@@ -98,39 +98,127 @@ describe("<Select>", () => {
...
         const value = "nailed it";
         const onChange = sinon.spy();
 
-        const input = select({ inputProps: { value, onChange } }).find("input");
+        const wrapper = select({ inputProps: { value, onChange } });
+        assert.equal(wrapper.state("query"), value);
+
+        const input = wrapper.find("input");
         assert.equal(input.prop("value"), value);
+        assert.equal(wrapper.state("query"), value);
```
giladgray: not convinced that we need to validate that state matches input prop _everywhere_. just doing it in the relevant tests should be enough.

packages/labs/test/selectTests.tsx
giladgray: probably best saved for another PR though, that'll be a big unrelated diff.

packages/labs/src/components/select/select.tsx
brieb: Unfortunately, prettier doesn't have an opinion on this matter. :)

packages/labs/src/components/select/select.tsx
brieb: Great catch, thanks!

packages/labs/test/selectTests.tsx
brieb: I tried `input.simulate("change", { currentTarget: { value: "..." } })` and `input.simulate("change", { target: { value: "..." } })` and even `input.simulate("change", { currentTarget: { value: "..." }, target: { value: "..." } })` with no luck

packages/labs/test/selectTests.tsx
brieb: good point - should be testing one thing

brieb:
So, I tried looking into the off-by-one-px issue, and I narrowed it down to the possibility that the scrollbar might have something to do with it. Not sure what the right solution is here.

Before:
screen shot 2017-09-26 at 10 27 29 pm

After unconditionally adding [`document.body.classList.add(Classes.OVERLAY_OPEN);`](https://github.com/palantir/blueprint/blob/5626890d117b4c21164eb932980c4f76e847e6aa/packages/core/src/components/overlay/overlay.tsx#L295) which applies 
```
body.pt-overlay-open {
  overflow: hidden;
}
```
screen shot 2017-09-26 at 10 28 03 pm

Easier to see with a border around the overlay (as it's actually a shadow)

Before:
screen shot 2017-09-26 at 10 29 30 pm

After:
screen shot 2017-09-26 at 10 30 01 pm

Can we file a separate issue for this?
Is this considered blocking?

packages/labs/src/components/select/select.tsx
```diff
@@ -291,10 +310,31 @@ export class Select<T> extends React.Component<ISelectProps<T>, ISelectState<T>>
     };
 
     private handleQueryChange = (event: React.FormEvent<HTMLInputElement>) => {
-        const { inputProps = {} } = this.props;
-        this.setState({ query: event.currentTarget.value });
+        const { inputProps = {}, onQueryChange } = this.props;
+        const query = event.currentTarget.value;
```
cmslewis: @brieb do the tests work with `input.simulate("change", { target: { value: "..." } })` if you change this to `event.target.value`?

adidahiya:
closing stale PR

PR #1599, opened 2017-09-22
[Button] Add pt-placeholder modifier

brieb:
Motivation is to have a placeholder style for buttons that resembles ::placeholder style for inputs.
An example use case would be a `<Select>` target with a default value.

@llorca This PR adds the pt-placeholder class we had talked about.
Styles are pretty basic - just adds muted text color. Is there more we should do here?

Also, do I need to do anything with `$control-group-stack`?
https://github.com/palantir/blueprint/blob/625f343a48111025aa44b01002d5312e1e35eb5c/packages/core/src/components/forms/_common.scss#L36-L55

### DIFF
--- a/packages/core/examples/buttonsExample.tsx
+++ b/packages/core/examples/buttonsExample.tsx
@@ -19,6 +19,7 @@ export interface IButtonsExampleState {
     loading?: boolean;
     large?: boolean;
     minimal?: boolean;
+    placeholder?: boolean;
     wiggling?: boolean;
 }
 
@@ -29,6 +30,7 @@ export class ButtonsExample extends BaseExample<IButtonsExampleState> {
         large: false,
         loading: false,
         minimal: false,
+        placeholder: false,
         wiggling: false,
     };
 
@@ -37,6 +39,7 @@ export class ButtonsExample extends BaseExample<IButtonsExampleState> {
     private handleLargeChange = handleBooleanChange(large => this.setState({ large }));
     private handleLoadingChange = handleBooleanChange(loading => this.setState({ loading }));
     private handleMinimalChange = handleBooleanChange(minimal => this.setState({ minimal }));
+    private handlePlaceholderChange = handleBooleanChange(placeholder => this.setState({ placeholder }));
     private handleIntentChange = handleNumberChange((intent: Intent) => this.setState({ intent }));
 
     private wiggleTimeoutId: number;
@@ -49,6 +52,7 @@ export class ButtonsExample extends BaseExample<IButtonsExampleState> {
         const classes = classNames({
             [Classes.LARGE]: this.state.large,
             [Classes.MINIMAL]: this.state.minimal,
+            [Classes.PLACEHOLDER]: this.state.placeholder,
         });
 
         return (
@@ -114,6 +118,12 @@ export class ButtonsExample extends BaseExample<IButtonsExampleState> {
                     key="minimal"
                     label="Minimal"
                     onChange={this.handleMinimalChange}
+                />,
+                <Switch
+                    checked={this.state.placeholder}
+                    key="placeholder"
+                    label="Placeholder"
+                    onChange={this.handlePlaceholderChange}
                 />,
             ],
             [<IntentSelect intent={this.state.intent} key="intent" onChange={this.handleIntentChange} />],
--- a/packages/core/src/components/button/_button.scss
+++ b/packages/core/src/components/button/_button.scss
@@ -16,6 +16,7 @@ Markup:
 
 :disabled - Disabled state
 .pt-disabled - Disabled appearance
+.pt-placeholder - Placeholder appearance
 .pt-intent-primary - Primary intent
 .pt-intent-success - Success intent
 .pt-intent-warning - Warning intent
@@ -171,6 +172,7 @@ Styleguide pt-button
   <button type="button" class="pt-button pt-minimal pt-icon-add {{.modifier}}" {{:modifier}}>Button</button>
 
   .pt-disabled - Disabled appearance
+  .pt-placeholder - Placeholder appearance
   .pt-intent-primary - Primary intent
   .pt-intent-success - Success intent
   .pt-intent-warning - Warning intent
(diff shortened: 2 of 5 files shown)


packages/core/src/components/button/_button.scss
```diff
@@ -16,6 +16,7 @@ Markup:
 
 :disabled - Disabled state
 .pt-disabled - Disabled appearance
+.pt-placeholder - Placeholder appearance
```
cmslewis: @llorca Would it make sense to alphabetize these? They seem kind of arbitrary right now.

cmslewis:
@brieb the `$control-group-stack` shouldn't be relevant here unless we want to give placeholder buttons a different `z-index` relative to their peers when inside a control group. That doesn't seem necessary to me, but I defer to @llorca.

packages/core/src/components/button/_button.scss
llorca: I believe the order is such that it looks good on the example docs

packages/core/src/components/button/_common.scss
```diff
@@ -64,19 +64,23 @@ $button-gradient: linear-gradient(to bottom, rgba($white, 0.8), rgba($white, 0))
 $button-intent-gradient: linear-gradient(to bottom, rgba($white, 0.1), rgba($white, 0)) !default;
 $dark-button-gradient: linear-gradient(to bottom, rgba($white, 0.05), rgba($white, 0)) !default;
 
+$button-color-placeholder: $pt-text-color-muted !default;
```
llorca: I think we should use `$pt-text-color-disabled` here. The difference with `-muted` is so small that I might be unnoticeable in some cases

packages/core/src/components/button/_common.scss
```diff
@@ -64,19 +64,23 @@ $button-gradient: linear-gradient(to bottom, rgba($white, 0.8), rgba($white, 0))
...
 $button-color-disabled: $pt-text-color-disabled !default;
 $button-background-color: $light-gray5 !default;
 $button-background-color-hover: $light-gray4 !default;
 $button-background-color-active: $light-gray2 !default;
 $button-background-color-disabled: rgba($light-gray1, 0.5) !default;
 $button-background-color-active-disabled: rgba($light-gray1, 0.7) !default;
+$button-intent-color-placeholder: rgba($white, 0.7);
```
llorca: What about `$button-intent-color-placeholder: $button-intent-color-disabled;`? I find 0.7 a bit too strong

packages/core/examples/buttonsExample.tsx
```diff
@@ -19,6 +19,7 @@ export interface IButtonsExampleState {
     loading?: boolean;
     large?: boolean;
     minimal?: boolean;
+    placeholder?: boolean;
```
llorca: I've been thinking about the naming and I'm torn between `placeholder` and `muted`. We've been using `muted` already as a modifier for text, and it seems a bit more generic than `placeholder`. Would love to hear your thoughts! Also @cmslewis what do you think?

cmslewis: No strong opinion, but now that you mention it, `placeholder` elsewhere means a `string`:

image

So maybe `muted` would be clearer?

llorca: Very good point. What do you think about changing to `.pt-muted` @brieb? I think I buy it

giladgray: yeah def cannot use `placeholder` cuz that's an HTML prop that accepts a string.

we should reuse `.pt-text-muted`, if anything.

packages/core/src/components/button/_common.scss
giladgray: how about `$input-placeholder-color: $pt-text-color-disabled !default;`?

also, one of these names is wrong? should be `button-placeholder-color`. https://github.com/palantir/blueprint/wiki/Coding-guidelines#sass

packages/core/src/components/button/_common.scss
giladgray: also always end variables with `!default`

packages/core/src/components/button/_common.scss
```diff
@@ -292,6 +328,10 @@ $button-intents: (
     color: $pt-text-color;
   }
 
+  &.pt-placeholder {
+    color: $pt-text-color-muted;
```
giladgray: 😱 I thought we had a local var for this? and all the ones below.

packages/core/src/components/button/_common.scss
```diff
@@ -239,6 +263,10 @@ $button-intents: (
   background-image: none;
 }
 
+@mixin pt-dark-button-placeholder() {
```
giladgray: not convinced these mixins are helpful, since they just apply a single variable. please remove.

packages/core/src/components/button/_common.scss
llorca: Good catch with `button-placeholder-color`

llorca: `$button-placeholder-color: $pt-text-color-disabled !default;`

packages/core/src/components/button/_button.scss
brieb: Given that, does it make sense where I added it?

packages/core/examples/buttonsExample.tsx
brieb: Named it `placeholder` to more clearly associate it with `::placeholder`. 
Muted works too, though.

packages/core/src/components/button/_common.scss
brieb: sounds good -- pattern-matched too closely.

giladgray:
**YO CRAZY API PROPOSAL AHEAD**

```css
.pt-button[placeholder]::after {
  color: $pt-text-color-muted;
  content: attr(placeholder);
}
```

```tsx
<Button placeholder="Placeholder text" />
```
❓ ❔ ❓

llorca:
whoa that's cool, sounds even better than a modifier for the purpose

brieb:
Sounds good. Going to close this PR in favor of the alternative proposal.

PR #1614, opened 2017-09-25
Table Preview Updates

themadcreator:
Separate mutable table into its own file.
Add session local storage for saving settings.
Add reset button to revert to defaults.


llorca:
@themadcreator does this overlap with #1603? Also instead of a reset button, what do you think about https://github.com/palantir/blueprint/pull/1603#issuecomment-331560009

packages/table/preview/localStore.ts
```diff
@@ -0,0 +1,42 @@
+/**
+ * Copyright 2017 Palantir Technologies, Inc. All rights reserved.
+ * Licensed under the BSD-3 License as modified (the “License”); you may obtain a copy
+ * of the license at https://github.com/palantir/blueprint/blob/master/LICENSE
+ * and https://github.com/palantir/blueprint/blob/master/PATENTS
+ *
+ * Demonstrates sample usage of the table component.
+ */
+
+import { defaults } from "lodash";
```
cmslewis: I see `lodash` is only a dev dependency, but still, do we really need it for this one function? Would prefer not to pull it in if we've gone this long without it.

Alternatively, we could just cherry-pick the function we need, as mentioned in their docs:

```tsx
// Cherry-pick methods for smaller browserify/rollup/webpack bundles.
var at = require('lodash/at');
var curryN = require('lodash/fp/curryN');
```

packages/table/preview/localStore.ts
```diff
@@ -0,0 +1,42 @@
...
+
+export class LocalStore<T> {
+    private storage: Storage;
+
+    constructor(private key: string, session = false){
+        this.storage = session ? sessionStorage : localStorage;
+    }
+
+    public getWithDefaults(defs?: T): T | {} {
```
cmslewis: Would prefer to call this `defaults`. `defs` means "definitions" to me.

packages/table/preview/mutableTable.tsx
```diff
@@ -0,0 +1,969 @@
...
+function getRandomInteger(min: number, max: number) {
+    // min and max are inclusive
+    return Math.floor(min + Math.random() * (max - min + 1));
+}
+
+function contains(arr: any[], value: any) {
+    return arr.indexOf(value) >= 0;
+}
+
+const DEFAULT_STATE: IMutableTableState = {
```
cmslewis: Can we define this _below_ the `IMutableTableState` interface? Keeps the `DEFAULTS` definition closer to its usage in the constructor, and that order would just make more sense to me.

packages/table/preview/localStore.ts
themadcreator: name already taken by the lodash method, but i'll come up w somethign

packages/table/preview/localStore.ts
themadcreator: not important for preview, but i'll switch it

packages/table/preview/localStore.ts
cmslewis: Ah, boo.

packages/table/preview/localStore.ts
themadcreator: when I switched to `import * as defaults from "lodash/defaults"` the types couldn't be resolved.

PR #1621, opened 2017-09-26
[Table] Don't render LEFT and TOP_LEFT quadrants if they're not needed

cmslewis:
#### Checklist

- [x] Include tests (updated existing tests!)

- If `isRowHeaderShown=false` and `numFrozenColumns=0`, don't render the `LEFT` or `TOP_LEFT` quadrants at all. 

- **Edge case:**
    1. Set `isRowHeaderShown=false`.
    2. Scroll downward and rightward a bit.
    3. Set `isRowHeaderShown=true`.
    4. *Expected:*
        - The row header should appear pre-scrolled to the current scroll offset (as shown in the GIF below).
        - The menu element should appear sync'd to the proper row-header width.

### DIFF
--- a/packages/table/src/quadrants/_quadrants.scss
+++ b/packages/table/src/quadrants/_quadrants.scss
@@ -104,13 +104,6 @@ $table-quadrant-scroll-container-overflow: 20px;
     height: auto;
     overflow-x: hidden;
   }
-
-  // correct missing left selecton border when selection includes
-  // cells in the first column (#516). we do this by requiring at least
-  // 1px of width on the body-cell wrapper element.
-  .bp-table-body-virtual-client {
-    min-width: 1px;
-  }
 }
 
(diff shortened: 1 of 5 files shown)


llorca:
With 1000 rows, disabling row headers, scrolling all the way to the bottom right, enabling row headers, I got this:
image

Seems good otherwise!

cmslewis:
@gscshoyru because `isColumnHeaderShown` is not a thing yet (see: #1806 for that FR), thus the column header is always shown, thus there is always something in the `TOP` quadrant.

cmslewis:
@llorca I investigated. since the row-header cells batch render, the synchronous `syncUpdateViews` call executes before row `1000` is present on the page, so it has no hope of measuring it. :/

I think toggling row headers is sufficiently uncommon as to make this a P3 or less, particularly when scrolling a little will fix the width right away.

Issue #1444, opened 2017-08-14
[Table] Fix visual regressions and styles from frozen row/col work 


cmslewis:
### Bug report

- __Package version(s)__: v1.24.0

#### Problems:
- [x] **Missing left border on leftmost-column selection overlay when `numFrozenColumns={0}`**
    ![image](https://user-images.githubusercontent.com/443450/29291188-a22fbce2-80f7-11e7-85e3-bbf05499b311.png)
    _Variant_: Missing focus-cell left edge on `FULL_ROWS` selection
    ![image](https://user-images.githubusercontent.com/443450/29291201-b4c6a730-80f7-11e7-839b-b98af4f00112.png)
- [ ] **Stray selection border when selection ends at the border of a frozen quadrant.**
    ![2017-08-14 13 53 28](https://user-images.githubusercontent.com/443450/29291333-12916206-80f8-11e7-86a4-6f073b9715a1.gif)
    ![2017-08-14 13 56 58](https://user-images.githubusercontent.com/443450/29291450-7c3f55e6-80f8-11e7-9745-3ec12a8e04e3.gif)
- [x] **Weird overscroll effect when scrolling with force toward the rightmost edge of the main quadrant.**
    - _Hmm, can't repro on `feature/table-frozen-rows-cols` the way I could in the #1439 PR..._  
- [ ] **Add more prominent border at the edge of frozen panes.**
    - _Here's what it looks like now:_
    ![image](https://user-images.githubusercontent.com/443450/29291569-d95dd07c-80f8-11e7-86e3-24e387f0827f.png)
    - Fixed in #1698 
- [x] **Stray focused-cell border when focused cell is flush with a quadrant boundary.**
- [ ] **White dot at the top of the A1 cell**
![image](https://user-images.githubusercontent.com/199754/30620012-a66423f8-9d57-11e7-86bc-910a07823a84.png)
- [ ] Covered selection border along menu element right side in dark theme:
![image](https://user-images.githubusercontent.com/443450/31466207-44746802-ae8b-11e7-8dd2-ad9603427d0b.png)

- **Others? List them here!**

tgreenwatts:
![focus](https://user-images.githubusercontent.com/24398123/30294719-ea563cf6-96f2-11e7-9812-0ecfea18ebd1.gif)

PR #1622, opened 2017-09-26
[Table] Fix missing selection border on left edge of grid

cmslewis:

- 🐛  __FIXED__ `Table` now displays a left border on selections that are flush with the left side of the grid.

- You'll notice a visual regression with the left border when the row header is hidden. This is fixed in #1621, which hides the `LEFT` and `TOP_LEFT` quadrants when they're not needed. :)

image

### DIFF
--- a/packages/table/src/quadrants/_quadrants.scss
+++ b/packages/table/src/quadrants/_quadrants.scss
@@ -105,9 +105,9 @@ $table-quadrant-scroll-container-overflow: 20px;
     overflow-x: hidden;
   }
 
-  // correct missing left selecton border when selection includes
-  // cells in the first column (#516). we do this by requiring at least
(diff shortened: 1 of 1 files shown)


packages/table/src/quadrants/_quadrants.scss
```diff
@@ -104,6 +104,13 @@ $table-quadrant-scroll-container-overflow: 20px;
     height: auto;
     overflow-x: hidden;
   }
+
+  // correct missing left selecton border when selection includes
+  // cells in the first column (#516). we do this by requiring at least
```
cmslewis: Should say #1444.

cmslewis:
@llorca I agree. I have a potential fix in mind.

PR #1624, opened 2017-09-26
[Table] TruncatedFormat popover improvements and performance tuning for dev preview

themadcreator:
~Switches `<TrunctedFormat>` to measure text with a canvas element, which does not trigger a relayout.~

Also adding switches to the dev preview to demonstrate how the table performs with:
- Slow native layout
- An isolated layout boundary


packages/table/src/cell/formats/truncatedFormat.tsx
```diff
@@ -92,12 +95,24 @@ export class TruncatedFormat extends React.Component<ITruncatedFormatProps, ITru
...
     };
 
-    public state: ITruncatedFormatState = { isTruncated: false };
+    public state: ITruncatedFormatState = {
+        isPopoverOpen: false,
+        isTruncated: false,
+    };
 
     private contentDiv: HTMLDivElement;
+    private cachedFontString: string;
```
themadcreator: we cache the font string on first use

themadcreator: this will be out of date if the style changes, but async font loading will not be an issue since the fontstring will still apply once the font is loaded.

packages/table/src/cell/formats/truncatedFormat.tsx
```diff
@@ -145,16 +140,58 @@ export class TruncatedFormat extends React.Component<ITruncatedFormatProps, ITru
...
+                    isOpen={true}
+                    onClose={this.handlePopoverClose}
+                    useSmartArrowPositioning={true}
+                >
+                    <Icon iconName="more" />
+                </Popover>
+            );
+        } else {
+            return (
+                <span className={Classes.TABLE_TRUNCATED_POPOVER_TARGET} onClick={this.handlePopoverOpen}>
```
themadcreator: this reproduces `<Popover>`'s internal structure without calling any layout-triggering listeners

packages/table/preview/mutableTable.tsx
```diff
@@ -220,43 +227,19 @@ export class MutableTable extends React.Component<{}, IMutableTableState> {
...
-                    onVisibleCellsChange={this.onVisibleCellsChange}
-                    onRowHeightChanged={this.onRowHeightChanged}
-                    onRowsReordered={this.onRowsReordered}
-                    ref={this.refHandlers.table}
-                    renderBodyContextMenu={this.renderBodyContextMenu}
-                    renderMode={renderMode}
-                    renderRowHeader={this.renderRowHeader}
-                    selectionModes={this.getEnabledSelectionModes()}
-                    styledRegionGroups={this.getStyledRegionGroups()}
+                <SlowLayoutStack
```
gscshoyru: ..not that it really matters, and maybe I'm misunderstanding something -- but would it have been better to split the changes to truncated format and the changes to the preview into separate commits?

packages/table/src/cell/formats/truncatedFormat.tsx
```diff
@@ -178,34 +215,18 @@ export class TruncatedFormat extends React.Component<ITruncatedFormatProps, ITru
...
-        this.setState({ isTruncated: shouldTruncate });
+        if (this.cachedFontString == null) {
+            this.cachedFontString = Utils.getFontStringFromDom(this.contentDiv);
+        }
+
+        const contentWidth = Utils.measureText(this.contentDiv.textContent, this.cachedFontString).width;
+        const containerWidth = parseInt(this.props.parentCellWidth, 10);
+        const availableWidth = containerWidth - CONTAINER_PADDING;
+        const isTruncated = contentWidth > availableWidth;
+        this.setState({ isTruncated } as ITruncatedFormatState);
```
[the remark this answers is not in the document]
gscshoyru: // appears in a puff of smoke, surrounded by chanting and candles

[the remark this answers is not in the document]
gscshoyru: ...um. Where'd height go? Why are we only measuring the width of the text itself? This doesn't actually work for any cases where we've got work wrap turned on. Did you test with that?

packages/table/src/cell/formats/truncatedFormat.tsx
```diff
@@ -11,12 +11,12 @@ import * as PureRender from "pure-render-decorator";
...
 
-// amount in pixels that the content div width changes when truncated vs when
-// not truncated. Note: could be modified by styles
-// Note 2: this doesn't come from the width of the popover element, but the "right" style
-// on the div, which comes from styles
-const CONTENT_DIV_WIDTH_DELTA = 25;
+// Since we measure only the `textContent` of the cell to determine the
+// truncation state, we must account for the padding that is applied via CSS to
+// the cell.
+const CONTAINER_PADDING = 20;
```
gscshoyru: Why the change?

packages/table/src/cell/formats/truncatedFormat.tsx
```diff
@@ -145,16 +140,58 @@ export class TruncatedFormat extends React.Component<ITruncatedFormatProps, ITru
...
-    }
+    private renderPopover() {
+        const { children, preformatted } = this.props;
 
-    public componentDidUpdate() {
-        this.setTruncationState();
+        // <Popover> will always check the content's position on update
+        // regardless if it is open or not. This negatively affects perf due to
+        // layout thrashing. So instead we manage the popover state ourselves
+        // and mimic its popover target
```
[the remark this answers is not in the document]
themadcreator: really, this should just be done by popover. the dom measuring should not be done on update, but rather on interaction

packages/table/src/cell/formats/truncatedFormat.tsx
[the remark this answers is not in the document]
themadcreator: so, uh, the only thing you can measure with canvas is width because that's all that exists on [TextMetrics](https://developer.mozilla.org/en-US/docs/Web/API/TextMetrics).

the text measurement assumes no wrapping in the text and is only useful at detecting horizontal overflow.

packages/table/preview/mutableTable.tsx
themadcreator: yeah, and might still need to do that, but i'm using this slowlayoutstack to test my changes to truncatedformat.

packages/table/src/cell/formats/truncatedFormat.tsx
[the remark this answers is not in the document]
themadcreator: image

😢

cmslewis:
@gscshoyru @themadcreator 

**RE: not considering parent-cell height:**

It looks like the only side effect of that omission is that we'll occasionally show the `···` icon _more_ often that we need to (e.g. if a long block of text fits in a cell with wrapping enabled but wouldn't with wrapping disabled). I'm okay with that slight goofiness given the huge perf benefits of this PR.

packages/table/preview/mutableTable.tsx
```diff
@@ -167,17 +177,20 @@ const DEFAULT_STATE: IMutableTableState = {
...
     enableColumnResizing: true,
     enableColumnSelection: true,
     enableContextMenu: false,
     enableFullTableSelection: true,
+    enableLayoutBoundary: false,
     enableMultiSelection: true,
     enableRowReordering: false,
     enableRowResizing: false,
     enableRowSelection: true,
+    enableSlowLayout: false,
```
cmslewis: @llorca FWIW the state-key names in this example have been a great testing ground for the ideas in #1625. Notice that we have both `enable...` and `show...` happening here. We really just need the former I think.

(No action required; just FYI)

packages/table/src/cell/formats/truncatedFormat.tsx
cmslewis: ping @themadcreator

themadcreator: Previously we had to consider the current truncation state when doing the calculation for content width because the element containing the text would have a different size if the "..." button was shown.

Now, since we're measuring with canvas, the size of the text content element is irrelevant. Instead we just need to consider the amount of padding between the parent element and the text content element.

packages/table/preview/index.html
```diff
@@ -52,6 +76,7 @@
     .sidebar {
       background-color: #EBF1F5;
       flex-basis: 300px;
+      flex-shrink: 0;
```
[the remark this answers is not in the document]
themadcreator: i had some trouble getting the preview layout to work with the slowlayoutstack. i'd like to revisit these styles and solidify how we're going to recommend creating layout boundaries.

it might even be worth adding a `<LayoutBoundary>` react component that uses style attributes (to avoid any class overrides from breaking them)

themadcreator:
some test in core is failing:
```
  Toaster
    with autoFocus set to true
      ✖ focuses on newly created toast
```

packages/table/preview/mutableTable.tsx
```diff
@@ -111,11 +124,45 @@ function handleNumberChange(handler: (value: number) => void) {
     return handleStringChange(value => handler(+value));
 }
 
-function getRandomInteger(min: number, max: number) {
+function getRandomObject(propCount: number, depth = 0): object {
+    const childPropCount = propCount;
+    const obj: any = {};
+    while (propCount-- >= 0) {
```
cmslewis: nit: I prefer to avoid the `--` and `++` operators, because they're so sneaky and tend to make lines denser.

packages/table/src/common/utils.ts
```diff
@@ -184,18 +206,60 @@ export const Utils = {
...
-     * Measures the bounds of supplied element's textContent.
+     * Measures the bounds of supplied element's textContent assuming no
+     * text wrapping.
+     *
+     * To improve performance, provide the `fontProperties` parameter. Otherwise,
+     * we use the computed font from the supplied element.
+     *
+     * Returns a `TextMetrics` object.
+     */
+    measureElementTextContent(element: Element, fontProperties?: string | ICssFontProperties) {
```
cmslewis: Can we revert all these utility changes?

PR #1627, opened 2017-09-27
[Table] Move useInteractionBar prop to <Table>, deprecate the <ColumnHeaderCell> one

cmslewis:
#### Checklist

- [x] Include tests

- 🔄  __CHANGED__ `Table` now supports a top-level `useInteractionBar` prop.
    - `ColumnHeaderCell`'s `useInteractionBar` prop is now `@deprecated`.
    - This has two benefits:
        1. It makes more sense from a design perspective to make the interaction bar an all-or-nothing commitment. It shouldn't be possible to enable it for only some column-header cells.
        2. Specifying `useInteractionBar` on the `Table` fixes a bug where the column header wouldn't resize properly when `useInteractionBar` was toggled on and off.

- I added a `measurableElementRef` prop to `ColumnHeader`, which is meant to return the element of the most-immediate container of the header cells. Measuring this element via `clientHeight`—instead of the outermost header element that also contains the fixed-height `menu` corner—ensures that we won't measure the header incorrectly when `useInteractionBar` is toggled off.
    - Maybe I should rename it `headerRef` or `headerRowRef` to make its purposes slightly clearer?

#### Screenshot

_BEFORE:_
Note: this is also how things will continue to work if you _don't_ pass `useInteractionBar` to the `Table` directly.

AFTER:
Passing `useInteractionBar` to the `Table` directly:


packages/table/src/common/errors.ts
```diff
@@ -9,6 +9,7 @@ const ns = "[Blueprint Table]";
 const deprec = `${ns} DEPRECATION:`;
 
 export const COLUMN_HEADER_CELL_MENU_DEPRECATED = `${deprec} <ColumnHeaderCell> menu is deprecated. Use renderMenu instead.`;
+export const COLUMN_HEADER_CELL_USE_INTERACTION_BAR_DEPRECATED = `${deprec} <ColumnHeaderCell> useInteractionBar is deprecated. Pass the prop to the top-level <Table> instead. When you pass the prop to <Table>, that value will override the one provided directly to <ColumnHeaderCell>.`;
```
cmslewis: @llorca @tgreenwatts thoughts on language?

packages/table/src/table.tsx
```diff
@@ -917,22 +930,47 @@ export class Table extends AbstractComponent<ITableProps, ITableState> {
...
+        const tableUseInteractionBar = this.props.useInteractionBar;
+
+        if (renderColumnHeaderCell != null) {
+            const columnHeaderCell = renderColumnHeaderCell(columnIndex);
+            const columnHeaderCellLoading = columnHeaderCell.props.loading;
+
+            // print a deprecation warning here if needed. we can't do this in
+            // ColumnHeaderCell, because we wouldn't be able to tell if
+            // useInteractionBar was injected by Table or provided by the user.
+            if (columnHeaderCell.props.useInteractionBar && !CoreUtils.isNodeEnv("production")) {
```
cmslewis: @gscshoyru @themadcreator @llorca FYI I'm printing a deprecation warning from here in the `table.tsx` instead from of `ColumnHeaderCell.validateProps`. Do the comments above explain why clearly enough?

packages/table/src/common/errors.ts
llorca: Looks good!

packages/table/src/table.tsx
llorca: Yes, that makes sense to me

packages/table/test/columnHeaderCellTests.tsx
```diff
@@ -106,6 +107,52 @@ describe("<ColumnHeaderCell>", () => {
...
+
+            afterEach(() => {
+                consoleWarn.reset();
+            });
+
+            after(() => {
+                consoleWarn.restore();
+            });
+
+            it("prints a deprecation warning if useInteractionBar=true", () => {
```
adidahiya: I don't really like testing deprecations; these tests seem tautological and eat up build time. Would prefer to remove the first 3 tests here.

cmslewis: Removed

PR #1630, opened 2017-09-27
Prepare release table-1.26.1

llorca:

### 🐛  Bug fixes
- `@blueprintjs/table` now depends on `@blueprintjs/core` 1.25.0, which contains the new `Icon` component used throughout the table

### DIFF
--- a/packages/site-docs/package.json
+++ b/packages/site-docs/package.json
@@ -1,6 +1,6 @@
 {
   "name": "blueprintjs.com/docs",
-  "version": "1.29.0",
+  "version": "1.29.1",
   "description": "Blueprint Docs",
   "private": true,
   "dependencies": {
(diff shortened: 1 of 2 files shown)


llorca:
Flaky test...

packages/site-docs/package.json
```diff
@@ -1,14 +1,14 @@
 {
   "name": "blueprintjs.com/docs",
-  "version": "1.29.0",
```
cmslewis: Ah weird. I thought `site-docs` and `core` moved in lockstep, but there's definitely no reason to bump `core` for this patch release. Ah well, we'll sync them up again in `1.30.0`.

llorca: I bumped `site-docs` but not `core`. What should I do?
- [ ] Bump both `core` and `site-docs` to reflect the patch number
- [ ] Bump neither
- [ ] Leave as is

adidahiya: they don't need to be in sync. what you've done in this PR is good.

PR #1632, opened 2017-09-27
[Table] Resize quadrants properly after user resizes frozen rows/columns

### LINKED ISSUE #1561: [Table] Double-click to resize frozen column fails if main-quadrant column is out of view
### Bug report

- __Package version(s)__: v1.27.0

#### Steps to reproduce

1. Create a `Table` with `numFrozenColumns={1}` and `isColumnResizable={true}` and enough columns to require horizontal scrolling.
1. Scroll all the way to the right, such that Column A is complete out of view (and not present in the bleed area either).
1. Double-click the Column A resize handle.

#### Actual behavior

Column A collapses to a tiny width.

#### Expected behavior

Column A resizes to the proper width to fit all cell content.

cmslewis:
#### Checklist

- [x] Include tests (will write one or two more though)

- 🐛 __FIXED__ `Table` now properly resizes frozen columns on double-click when the table is scrolled.
    - Before, we were always measuring the width of the corresponding column from the `MAIN` quadrant. When sufficiently scrolled, that column might not even be rendered, which returned a width measurement of `0px`.
    - Now, we measure the frozen column from the `LEFT` quadrant and other columns from the `MAIN` quadrant (and same thing for rows, but with `TOP`/`MAIN`).
- This PR also explicitly synchronizes quadrant sizes after resizing columns OR rows. This is useful when resizing frozen columns and rows in particular, since changing the size of those elements changes the size of the encompassing `LEFT`/`TOP`/`TOP_LEFT` quadrants.

- Thoughts on the `didColumnOrRowSizesChangeInLatestUpdate: boolean` approach? Is that too goofy? Definitely saves us some diff'ing effort; otherwise, we'd have to deeply compare the entire `rowHeights` and `columnWidths` arrays on every update (e.g. on every scroll event), which would be SLOW.

#### Screenshot

BEFORE

AFTER


packages/table/src/locator.ts
```diff
@@ -101,28 +101,35 @@ export class Locator implements ILocator {
...
-            const cellValue = cells.item(i).querySelector(`.${Classes.TABLE_TRUNCATED_VALUE}`);
-            const cellTruncatedFormatText = cells.item(i).querySelector(`.${Classes.TABLE_TRUNCATED_FORMAT_TEXT}`);
-            const cellTruncatedText = cells.item(i).querySelector(`.${Classes.TABLE_TRUNCATED_TEXT}`);
+        // consider only body cells, hence the extra Classes.TABLE_CELL specificity
+        const columnCellSelector = this.getColumnCellSelector(columnIndex);
+        const columnBodyCells = this.tableElement.querySelectorAll(`${columnCellSelector}.${Classes.TABLE_CELL}`);
+
+        let maxHeight = 0;
+        for (let i = 0; i < columnBodyCells.length; i++) {
+            const cell = columnBodyCells.item(i);
```
themadcreator: 👍

packages/table/src/quadrants/tableQuadrantStack.tsx
```diff
@@ -248,6 +248,19 @@ export class TableQuadrantStack extends AbstractComponent<ITableQuadrantStackPro
...
     }
 
+    /**
+     * Synchronizes quadrant sizes and scroll offsets based on the current
+     * column, row, and header sizes. Useful for correcting quadrant sizes after
+     * explicitly resizing columns and rows, for instance.
+     *
+     * Invoking this method imperatively is cheaper than providing columnWidths
+     * or rowHeights array props to TableQuadrantStack and forcing it to run
+     * expensive array diffs upon every update.
```
themadcreator: are you certain it's less expensive to make dom measurements in `syncQuadrantViews` than to diff some array values? won't it force a layout? what if the layout is slow?

packages/table/src/table.tsx
```diff
@@ -410,6 +410,11 @@ export class Table extends AbstractComponent<ITableProps, ITableState> {
     private rowHeaderElement: HTMLElement;
     private scrollContainerElement: HTMLElement;
 
+    // when this variable is true, we'll need to synchronize quadrant views
+    // after the update. this variable lets us avoid expensively diff'ing
+    // columnWidths and rowHeights on each update.
+    private didColumnOrRowSizesChangeInLatestUpdate: boolean = false;
```
themadcreator: wont this member be typed as `boolean` automatically if you set it to false?

packages/table/src/table.tsx
```diff
@@ -660,7 +666,7 @@ export class Table extends AbstractComponent<ITableProps, ITableState> {
     public componentDidMount() {
         this.validateGrid();
 
-        this.locator = new Locator(this.mainQuadrantElement, this.scrollContainerElement, this.cellContainerElement);
+        this.locator = new Locator(this.rootTableElement, this.scrollContainerElement, this.cellContainerElement);
```
themadcreator: oh man, are you sure this is right? changing the element of the locator makes me nervous

packages/table/src/quadrants/tableQuadrantStack.tsx
cmslewis: We invoke this method only when `didColumnOrRowSizesChangeInLatestUpdate=true`, so we're doing reflows only when we have to.

On a related/unrelated note, I'm not positive I handled controlled-mode changes correctly: i.e., what happens when you pass in a new `rowHeights` or `columnWidths` array to the `<Table>` component? Need to make sure I update the quadrant sizes appropriately in that case.

cmslewis:
I'm gonna wait on this one. Need to make sure controlled mode works properly.

packages/table/src/locator.ts
```diff
@@ -101,28 +101,35 @@ export class Locator implements ILocator {
...
-        for (let i = 0; i < cells.length; i++) {
-            const cellValue = cells.item(i).querySelector(`.${Classes.TABLE_TRUNCATED_VALUE}`);
-            const cellTruncatedFormatText = cells.item(i).querySelector(`.${Classes.TABLE_TRUNCATED_FORMAT_TEXT}`);
-            const cellTruncatedText = cells.item(i).querySelector(`.${Classes.TABLE_TRUNCATED_TEXT}`);
+        // consider only body cells, hence the extra Classes.TABLE_CELL specificity
+        const columnCellSelector = this.getColumnCellSelector(columnIndex);
+        const columnBodyCells = this.tableElement.querySelectorAll(`${columnCellSelector}.${Classes.TABLE_CELL}`);
+
+        let maxHeight = 0;
+        for (let i = 0; i < columnBodyCells.length; i++) {
```
gscshoyru: I realize we were doing this with a for loop before -- but what exactly is the structure returned from the querySelectorAll ? Can we use some form of iterator or foreach instead?

packages/table/src/table.tsx
gscshoyru: Is there... anything better that you can name this? I can't think of anything, but, while I agree with the usage, the name is pretty awful.

packages/table/src/quadrants/tableQuadrantStack.tsx
cmslewis: Turns out we don't honor controlled `columnWidths` or `rowHeights` at all after the first render... Therefore not a blocker for this PR, but we should fix it.

packages/table/src/table.tsx
cmslewis: Changed to `didUpdateColumnOrRowSizes`.

packages/table/src/locator.ts
cmslewis: Seems like we don't have typings for any of this stuff: https://developer.mozilla.org/en-US/docs/Web/API/NodeList. So need to leave as is.

Issue #1633, opened 2017-09-27
Better support resizing column at end of table


tgreenwatts:
Today if your table takes up the full width it is very difficult to expand the last column. Let's make this easier.

tgreenwatts:
For reference: was previously filed https://github.com/palantir/blueprint/issues/470 but closed. I think it's worth investigating now. 

SA: https://pl.ntr/gl

PR #1637, opened 2017-09-28
[Table] Prevent overscroll, refactor scrolling logic in TableQuadrantStack

cmslewis:
#### Checklist

- [x] ~Include tests~ (it's tricky/prohibitive to add a test for the prevent-overscroll case, since it's such a momentary effect and is more of a rendering artifact than anything)

- #1444 `Table` no longer allows inertial scroll beyond the last row or column.
- #1573 `Table` now doesn't stick _as much_ when you try to scroll slowly leftward/upward from the last column/row.
    - Not really sure what the problem is here.
- `Table` now caches the scroll container's `clientWidth`/`clientHeight` upon the first `wheel` event in a gesture, reducing DOM reads during the gesture.  
    - The cached value is cleared in the debounced view-sync method after the gesture has ended.
    - This eliminates DOM reads during the wheel gesture.
    - ⚠️  This assumes the `clientWidth` and `clientHeight` won't change during the wheel gesture.

- **Does scrolling actually feel better, or at least not worse?** 
- Is the ⚠️  assumption above okay for now? I guess this assumption may break if we eventually expand the table size while dragging to resize the last column (#1633).
    - If this is a problem, I can whittle that stuff out of this PR.
- If someone can help me figure out why the scrolling is slightly "stuck" at the end of the table, I'd appreciate it.

#### Screenshot

Things to note:
- No "overscroll" when you scroll to the end of the table.
- There's still a little stickiness at the end. :/
- There's not as much stickiness as there used to be (i.e. it doesn't take as forceful of a scroll to get out of the sticky zone).

### DIFF
--- a/packages/table/src/quadrants/tableQuadrantStack.tsx
+++ b/packages/table/src/quadrants/tableQuadrantStack.tsx
@@ -516,9 +516,10 @@ export class TableQuadrantStack extends AbstractComponent<ITableQuadrantStackPro
         const nextScrollLeft = mainScrollContainer.scrollLeft;
         const nextScrollTop = mainScrollContainer.scrollTop;
 
-        // update the cache.
-        this.setScrollOffset("scrollLeft", nextScrollLeft, /* setOffsetExplicitly */ false);
-        this.setScrollOffset("scrollTop", nextScrollTop, /* setOffsetExplicitly */ false);
+        // with the "scroll" event, scroll offsets are updated prior to the
+        // event's firing, so no explicit update needed.
+        this.handleScrollOffsetChange("scrollLeft", nextScrollLeft);
+        this.handleScrollOffsetChange("scrollTop", nextScrollTop);
 
         // sync less important view stuff when scrolling/wheeling stops.
         this.syncQuadrantViewsDebounced();
@@ -543,8 +544,10 @@ export class TableQuadrantStack extends AbstractComponent<ITableQuadrantStackPro
         // of this onWheel callback, so we need to manually update the affected
         // quadrant's scroll position to make up for that. note that these DOM
         // writes are batched together after the reads above.
-        this.setScrollOffset("scrollLeft", nextScrollLeft);
-        this.setScrollOffset("scrollTop", nextScrollTop);
+        this.quadrantRefs[QuadrantType.MAIN].scrollContainer.scrollLeft = nextScrollLeft;
+        this.quadrantRefs[QuadrantType.MAIN].scrollContainer.scrollTop = nextScrollTop;
+        this.handleScrollOffsetChange("scrollLeft", nextScrollLeft);
+        this.handleScrollOffsetChange("scrollTop", nextScrollTop);
 
         // sync less important view stuff when scrolling/wheeling stops.
         this.syncQuadrantViewsDebounced();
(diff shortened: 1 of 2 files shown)


packages/table/src/quadrants/tableQuadrantStack.tsx
```diff
@@ -776,13 +789,40 @@ export class TableQuadrantStack extends AbstractComponent<ITableQuadrantStackPro
...
+        this.cache.setScrollOffset(scrollKey, offset);
+
+        const dependentQuadrantType = scrollKey === "scrollLeft" ? QuadrantType.TOP : QuadrantType.LEFT;
+        this.maybeSetQuadrantScrollOffset(dependentQuadrantType, scrollKey);
+    };
+
+    private updateScrollContainerClientSize(isHorizontal: boolean) {
+        // this function is called 'update' instead of 'set', because a 'set'
+        // function typically takes the new value as a parameter. we avoid that
+        // to keep the isHorizontal logic tree contained within this function.
```
themadcreator: nit: comments about the function itself should be above defn. since this is a private method, it's going to be added to docs anyway.

packages/table/src/quadrants/tableQuadrantStack.tsx
```diff
@@ -536,48 +533,59 @@ export class TableQuadrantStack extends AbstractComponent<ITableQuadrantStackPro
...
+            : this.cache.getScrollContainerClientHeight();
+
+        if (clientSize == null) {
+            // should trigger only on the first scroll of the wheel gesture
+            clientSize = this.updateScrollContainerClientSize(isHorizontal);
+        }
 
-            mainScrollContainer[scrollKey] = nextScrollOffset;
+        const containerSize = isHorizontal
+            ? this.cache.getScrollContainerClientWidth() - this.cache.getRowHeaderWidth()
```
themadcreator: this code is a little difficult to read because above you access `getScrollContainerClientWidth` which according the conditional on line 574, could be nully. then right after, you invoke them again without a null check. this implies that `updateScrollContainerClientSize` will be setting the right value, but in order to verify, i have to ready that method and look at it's conditionals.

packages/table/src/quadrants/tableQuadrantStack.tsx
```diff
@@ -608,24 +620,24 @@ export class TableQuadrantStack extends AbstractComponent<ITableQuadrantStackPro
 
     // Rows
 
-    private handleRowResizeGuideMain = (verticalGuides: number[]) => {
-        this.invokeRowResizeHandler(verticalGuides, QuadrantType.MAIN);
+    private handleRowResizeGuideMain = (horizontalGuides: number[]) => {
```
cmslewis: Unrelated change, but figured I'd fix it since it's just a typo.

packages/table/src/quadrants/tableQuadrantStack.tsx
```diff
@@ -712,6 +724,11 @@ export class TableQuadrantStack extends AbstractComponent<ITableQuadrantStackPro
         // prevents unnecessary reflows in the future.
         this.cache.setRowHeaderWidth(rowHeaderWidth);
         this.cache.setColumnHeaderHeight(columnHeaderHeight);
+        // ...however, we also clear the cached client size, so we can read it
+        // again when a new scroll begins. not safe to assume this won't change.
+        // TODO: maybe use the ResizeSensor?
```
cmslewis: Not in this PR, maybe not ever.

packages/table/src/quadrants/tableQuadrantStack.tsx
```diff
@@ -740,7 +757,7 @@ export class TableQuadrantStack extends AbstractComponent<ITableQuadrantStackPro
         }
     };
 
-    private maybeSetQuadrantOffset = (quadrantType: QuadrantType, side: "right" | "bottom", value: number) => {
+    private maybeSetQuadrantPositionOffset = (quadrantType: QuadrantType, side: "right" | "bottom", value: number) => {
```
cmslewis: Name change for clarity.

cmslewis:
@llorca I'd rather have this robust, explicit logic in place than risk the "overscroll" bug returning again. Would prefer to merge this after all.

PR #1641, opened 2017-09-29
[Table] Add more switches for text layout to dev preview

themadcreator:
Adds switches to enabled "fixed truncation" which will truncate a string based on how many characters it has.


themadcreator:
image

themadcreator:
I think it still works, it's just that there is always so much JSON that it's always truncated.

PR #1652, opened 2017-10-03
[Table] Fix TS errors in features.tsx

cmslewis:


packages/table/preview/features.tsx
```diff
@@ -264,13 +264,16 @@ class EditableTable extends React.Component<{}, IEditableTableState> {
     private setArrayState<T>(key: string, index: number, value: T) {
         const values = (this.state as any)[key].slice() as T[];
         values[index] = value;
-        this.setState({ [key]: values });
+        // `any` necessary to fix weird TS errors. related SO post:
+        // https://stackoverflow.com/questions/46361905/property-is-missing-in-type-x-string-string
+        this.setState({ [key]: values } as any);
```
adidahiya: can you cast to `IEditableTableState` instead?

I'd like to avoid linking to this SO post; it doesn't add much insight. instead explain the typing problem in words here

cmslewis: FWIW all of my tricks failed me, and casting as `IEditableTableState` yields:
```
[ts]
Type '{ [x: string]: { [x: string]: T; }; }' cannot be converted to type 'IEditableTableState'.
  Property 'intents' is missing in type '{ [x: string]: { [x: string]: T; }; }'.
[tslint] Type assertion on object literals is forbidden, use a type annotation instead. (no-object-literal-type-assertion)
```

PR #1656, opened 2017-10-03
[Table] Turn off text truncation by default and raise fixed truncation limit

tgreenwatts:
- Benefits: No more detecting truncation, which is expensive. Performance benefits
- Drawbacks: A mix of little … and big …s, but I think this will be uncommon and the big … can change meaning from “contents don’t fit cell” to “contents are really big”, also an icon change will help
- Drawbacks: Wrapped text no longer gets … of any sort (unless real big). I’m fine with this, it’s pretty common. Apps that wrap should also resizeByTallest
-How we do it: 
— Change the default from detectTruncation to fixedTruncation
— Choose sane default
— Write good release notes

I may have missed spots

### DIFF
--- a/packages/table/src/cell/formats/jsonFormat.tsx
+++ b/packages/table/src/cell/formats/jsonFormat.tsx
@@ -35,7 +35,7 @@ export class JSONFormat extends React.Component<IJSONFormatProps, {}> {
         omitQuotesOnStrings: true,
         showPopover: TruncatedPopoverMode.WHEN_TRUNCATED,
         stringify: (obj: any) => JSON.stringify(obj, null, 2),
-        truncateLength: 500,
+        truncateLength: 2000,
     };
 
     public render() {
--- a/packages/table/src/cell/formats/jsonFormat.tsx
+++ b/packages/table/src/cell/formats/jsonFormat.tsx
@@ -31,10 +31,11 @@ export interface IJSONFormatProps extends ITruncatedFormatProps {
 
 export class JSONFormat extends React.Component<IJSONFormatProps, {}> {
     public static defaultProps: IJSONFormatProps = {
-        detectTruncation: true,
+        detectTruncation: false,
         omitQuotesOnStrings: true,
         showPopover: TruncatedPopoverMode.WHEN_TRUNCATED,
         stringify: (obj: any) => JSON.stringify(obj, null, 2),
+        truncateLength: 500,
     };
 
     public render() {
--- a/packages/table/src/cell/formats/truncatedFormat.tsx
+++ b/packages/table/src/cell/formats/truncatedFormat.tsx
@@ -67,7 +67,7 @@ export interface ITruncatedFormatProps extends IProps {
      * Number of characters that are displayed before being truncated and appended with the
      * `truncationSuffix` prop. A value of 0 will disable truncation. This prop is ignored if
      * `detectTruncation` is `false`.
-     * @default 500
+     * @default 2000
      */
     truncateLength?: number;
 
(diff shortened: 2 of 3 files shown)


tgreenwatts:
I believe this is a UX change, but no different than a UX change like adding reorder handles to the columns without an interaction bar - it's a change, but one we believe is better for everyone. This should not disrupt any users. We will call it out in release notes clearly so if anyone wants to opt in to `detectTruncation` they are welcome to. 

Before:
before

After:
image

This column is too long for it's current width, but not so long that it can't reasonably be displayed, so we do not truncate.

tgreenwatts:
I played with this in the dev preview and think the default was a little too low, so up-ed it to 2000.

packages/table/src/cell/formats/jsonFormat.tsx
```diff
@@ -31,10 +31,11 @@ export interface IJSONFormatProps extends ITruncatedFormatProps {
 
 export class JSONFormat extends React.Component<IJSONFormatProps, {}> {
     public static defaultProps: IJSONFormatProps = {
-        detectTruncation: true,
+        detectTruncation: false,
```
cmslewis: Can delete this line and fall back to the default props defined in `TruncatedFormat`.

packages/table/src/cell/formats/jsonFormat.tsx
```diff
@@ -31,10 +31,11 @@ export interface IJSONFormatProps extends ITruncatedFormatProps {
...
         omitQuotesOnStrings: true,
         showPopover: TruncatedPopoverMode.WHEN_TRUNCATED,
         stringify: (obj: any) => JSON.stringify(obj, null, 2),
+        truncateLength: 2000,
```
cmslewis: Same

packages/table/src/cell/formats/truncatedFormat.tsx
```diff
@@ -66,8 +66,8 @@ export interface ITruncatedFormatProps extends IProps {
     /**
      * Number of characters that are displayed before being truncated and appended with the
      * `truncationSuffix` prop. A value of 0 will disable truncation. This prop is ignored if
-     * `detectTruncation` is `true`.
-     * @default 80
+     * `detectTruncation` is `false`.
```
cmslewis: Hmm, should this have changed? See https://github.com/palantir/blueprint/pull/1656/files#diff-820e3fef53ee313553edff37ddbf4737R116. By that logic, we'll never refer to the `truncateLength` if `detectTruncation` is _`true`_.

packages/table/test/formatsTests.tsx
```diff
@@ -25,21 +25,7 @@ describe("Formats", () => {
...
-                ghosts include greater poets than Keats, scientists greater
-                than Newton. We know this because the set of possible people
-                allowed by our DNA so massively outnumbers the set of actual
-                people. In the teeth of these stupefying odds it is you and I,
-                in our ordinariness, that are here. We privileged few, who won
-                the lottery of birth against all odds, how dare we whine at
-                our inevitable return to that prior state from which the vast
-                majority have never stirred?
-            `;
+            const str = createStringOfLength(TruncatedFormat.defaultProps.truncateLength + 1);
```
cmslewis: Beau-tee-ful!

packages/table/test/formatsTests.tsx
```diff
@@ -71,11 +57,11 @@ describe("Formats", () => {
                 majority have never stirred?
             `;
 
-            const style = { height: "200px", position: "relative" };
+            const style = { height: "300px", width: "300px", position: "relative" };
```
cmslewis: I'd love a comment here.

```
// fix the container's width and height to ensure this test passes
// regardless of the page's dimensions.
```

cmslewis:
@tgreenwatts - Wait wait, did you see my requested changes? Would like to see those implemented before merging. Reverting.

packages/table/src/cell/formats/jsonFormat.tsx
tgreenwatts: Wouldn't this be true for `showPopover` as well? it's default here is `showPopover: TruncatedPopoverMode.WHEN_TRUNCATED,` which is the default

PR #1663, opened 2017-10-05
[Table] Fix empty states

cmslewis:
#### Checklist

- [x] Include tests

#### Changes proposed in this pull request:

- 🐛  __FIXED__ `Table` empty states are now both sane and beautiful.
    - Column-header borders are no longer missing.
    - Column header no longer collapses in size when `numCols = 0`
    - Selections are cleared and selecting is disabled when `numRows = 0` and/or `numCols === 0`

#### Reviewers should focus on:

- This work primarily involves style fixes (CSS) and more aggressive quadrant-size syncing.
- Please test anything and everything in the dev preview that could change the layout of the table. Cases I caught:
    - Num. rows changes to/from `0`
    - Num. columns changes to/from `0`
    - Num. frozen columns changes to/from `0`
    - Num. frozen rows changes to/from `0`
    - Row headers show/hide
    - Ghost cells show/hide
    - Column header-cell content gets taller (try scrolling past column `Z` to test this!)
    - Row header-cell content gets wider (try scrolling past row `100` to test this!)
    - Column header cells go in/out of loading state
    - Row header cells go in/out of loading state (bug when coming out of loading state: due to batch rendering, the widest-number cells at the bottom may not finish rendering before the quadrants are sync'd; this is a minor bug that fixes itself on scroll)
- One thing you may notice: with `numCols = 0` and `numRows > 0` with row headers disabled, the table container has this weird line at the top. This is an issue with the default `inline-block` container _around_ the table, not the table container itself. When you enable "Isolate layout boundary," it goes away.
image

#### Screenshot
image

image

image

image

image

image

image

_Column selection disabled, but reordering and resizing still possible:_

_Row selection disabled, but resizing still possible (reordering isn't, because you have to select rows before you reorder them):_

_Resizes when loading state changes:_

_Syncs quadrant sizes after scroll when column-header height changes:_

_Selections disappear when you go to 0 rows or 0 columns:_

PR #1671, opened 2017-10-05
[Table] Add toggle for custom column headers to preview

gscshoyru:

Add toggle for custom headers, which was missing

### DIFF
--- a/packages/table/preview/mutableTable.tsx
+++ b/packages/table/preview/mutableTable.tsx
@@ -603,6 +603,7 @@ export class MutableTable extends React.Component<{}, IMutableTableState> {
                 {this.renderSwitch("Loading state", "showColumnHeadersLoading")}
                 {this.renderSwitch("Menus", "showColumnMenus")}
                 <h6>Interactions</h6>
+                {this.renderSwitch("Custom Headers", "enableColumnCustomHeaders")}
                 {this.renderSwitch("Editing", "enableColumnNameEditing")}
                 {this.renderSwitch("Reordering", "enableColumnReordering")}
                 {this.renderSwitch("Resizing", "enableColumnResizing")}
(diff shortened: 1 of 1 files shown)


packages/table/preview/mutableTable.tsx
```diff
@@ -603,6 +603,7 @@ export class MutableTable extends React.Component<{}, IMutableTableState> {
                 {this.renderSwitch("Loading state", "showColumnHeadersLoading")}
                 {this.renderSwitch("Menus", "showColumnMenus")}
                 <h6>Interactions</h6>
+                {this.renderSwitch("Custom Headers", "enableColumnCustomHeaders")}
```
cmslewis: I'd say this belongs in the `Display` section. 🤓

tgreenwatts: Well if we do that, we should disable "Editing" with a note about this.

cmslewis: @gscshoyru you can make the editing toggle dependent on this one:

```tsx
{this.renderSwitch("Editing", "enableColumnNameEditing", "enableColumnCustomHeaders", false)}
``` 

Then the editing switch will be disabled and show a tooltip that says:
`Requires enableColumnCustomHeaders=false` on hover.

cmslewis: Oh, and also, `Sentence cased` please.

themadcreator:
toggle-col-header
is this right?

gscshoyru:
...that would be a no. 

Not related to this fix, obviously is bug in table that's been missed up until now. Please file.

tgreenwatts:
Pretty sure that bug was fixed by Chris's empty stats PR @themadcreator

cmslewis:
Yep just fixed that in #1663 @themadcreator.

PR #1672, opened 2017-10-05
[Submenus] Switch popover from relative to absolute positioning

### LINKED ISSUE #1668: [Menu] [IE11] Sub-menu's do not position correctly on IE11
If I hover over the Style sub-menu in the Blueprint docs on IE11, when there isn't enough room for it to display on it's default side:

image

Then in Chrome it'll reposition the menu to the other side:

image

But in IE11 it'll reposition it on top of the menu:

image

llorca:
#### Checklist

- [ ] Include tests

Change the `position` of submenu popovers to be `absolute` instead of `relative`

Any particular reason why it was `relative` in the first place?

### DIFF
--- a/packages/core/src/components/menu/_submenu.scss
+++ b/packages/core/src/components/menu/_submenu.scss
@@ -35,7 +35,7 @@
 
   // minor adjustments to popover position so top submenu item lines up with submenu trigger item
   .pt-popover {
-    position: relative;
+    position: absolute;
     top: -$half-grid-size;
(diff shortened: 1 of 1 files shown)


tgreenwatts:
I'll check it out when the preview comes up on IE!

tgreenwatts:
Looks great
image

cmslewis:
The line `git blame`s to @giladgray. Gilad, remember why we used `position: relative`?

Issue #875, opened 2017-03-21
Remove bourbon dependency


adidahiya:
follow-up of https://github.com/palantir/blueprint/pull/719#issuecomment-281855926, https://github.com/palantir/blueprint/issues/123

PR #1673, opened 2017-10-05
[Table] Make editable cells fill cells vertically

### LINKED ISSUE #1235: [Table] EditableCell doesn't grow to match the row like EditableName does with defaultRowHeight set
### Bug report

- __Package version(s)__: core: 1.20.0, table: 1.17.0 
- __Browser and OS versions__: Chrome 58.0.3029.110, OS X 10.12.5

#### Steps to reproduce

```jsx
<Table numRows={1} defaultRowHeight={30}>
    <Column
        name="foo"
        renderCell={() => <EditableCell/>}
        renderColumnHeader={() => <ColumnHeaderCell renderName={() => <EditableName/>} />}
    />
</Table>
```

#### Actual behavior

Header has the right size:

image

But the cell doesn't:

image

#### Expected behavior

`EditableCell`s grow to fit the size of their containing row.

This is extra sad because then the click target for the editable area is small even if the row is very large.

gscshoyru:

Make some style modifications to the surrounding div of the editable text, with a new class, to make it always horizontally fill the cell. Double-clicking anywhere in the cell will open the input now, and it takes up the whole space. 

Is position: absolute really the best way of doing this? I'm not happy with it, but I couldn't come up with anything better.

### DIFF
--- a/packages/table/src/cell/_cell.scss
+++ b/packages/table/src/cell/_cell.scss
@@ -54,7 +54,8 @@
 .bp-table-editable-text {
   position: absolute;
   top: 0;
-  right: 10px;
+  right: 0;
   bottom: 0;
-  left: 10px;
+  left: 0;
+  padding: $cell-padding;
 }
--- a/packages/table/src/cell/_cell.scss
+++ b/packages/table/src/cell/_cell.scss
@@ -50,3 +50,11 @@
 .bp-table-editable-name input {
   height: $cell-height; // fixed height for IE11
 }
+
+.bp-table-editable-text {
+  position: absolute;
+  top: 0;
+  right: 10px;
+  bottom: 0;
+  left: 10px;
+}
--- a/packages/table/src/headers/_headers.scss
+++ b/packages/table/src/headers/_headers.scss
@@ -314,11 +314,8 @@ $dark-selectable-header-menu-selected-hover-background:
 
       &.bp-table-editable-text {
         &::before {
-          top: -$cell-padding-vertical + $cell-border-width - 1px;
           right: $cell-border-width;
-          bottom: -$cell-padding-vertical + 1px;
           left: $cell-border-width - 1px;
-          cursor: $select-text-cursor;
         }
       }
     }
--- a/packages/table/src/headers/_headers.scss
+++ b/packages/table/src/headers/_headers.scss
@@ -303,12 +303,24 @@ $dark-selectable-header-menu-selected-hover-background:
       box-shadow: none;
     }
 
-    &.pt-editable-editing::before {
-      top: -$cell-padding-vertical + $cell-border-width - 1px;
-      right: -$cell-padding-horizontal + $cell-border-width;
-      bottom: -$cell-padding-vertical + 1px;
-      left: -$cell-padding-horizontal + $cell-border-width - 1px;
-      cursor: $select-text-cursor;
+    &.pt-editable-editing {
+      &::before {
+        top: -$cell-padding-vertical + $cell-border-width - 1px;
+        right: -$cell-padding-horizontal + $cell-border-width;
+        bottom: -$cell-padding-vertical + 1px;
+        left: -$cell-padding-horizontal + $cell-border-width - 1px;
+        cursor: $select-text-cursor;
+      }
+
+      &.bp-table-editable-text {
+        &::before {
+          top: -$cell-padding-vertical + $cell-border-width - 1px;
+          right: $cell-border-width;
+          bottom: -$cell-padding-vertical + 1px;
+          left: $cell-border-width - 1px;
+          cursor: $select-text-cursor;
+        }
+      }
     }
   }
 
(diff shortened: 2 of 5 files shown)


packages/table/src/headers/editableName.tsx
```diff
@@ -23,37 +23,55 @@ export interface IEditableNameProps extends IIntentProps, IProps {
...
+        this.invokeCallback(this.props.onChange, value);
+    };
+
+    private handleConfirm = (value: string) => {
+        this.invokeCallback(this.props.onConfirm, value);
+    };
+
+    private invokeCallback(callback: (value: string, rowIndex?: number, columnIndex?: number) => void, value: string) {
+        // pass through the row and column indices if they were provided as props by the consumer
+        // const { columnIndex } = this.props;
```
cmslewis: delete commented code?

packages/table/src/headers/editableName.tsx
```diff
@@ -23,37 +23,55 @@ export interface IEditableNameProps extends IIntentProps, IProps {
...
+
+    private handleChange = (value: string) => {
+        this.invokeCallback(this.props.onChange, value);
+    };
+
+    private handleConfirm = (value: string) => {
+        this.invokeCallback(this.props.onConfirm, value);
+    };
+
+    private invokeCallback(callback: (value: string, rowIndex?: number, columnIndex?: number) => void, value: string) {
```
cmslewis: These callback params look inconsistent with the ones in the props interface above.

In the props interface: `value, columnIndex`
Here: `value, rowIndex, columnIndex`

packages/table/src/cell/_cell.scss
```diff
@@ -50,3 +50,11 @@
 .bp-table-editable-name input {
   height: $cell-height; // fixed height for IE11
 }
+
+.bp-table-editable-text {
+  position: absolute;
+  top: 0;
+  right: 10px;
+  bottom: 0;
+  left: 10px;
```
cmslewis: Are these offsets not accomplished with padding within the editable text component itself?

Also, probably want to use `$pt-grid-size`. // CC @llorca

gscshoyru:
Gah, my bad, that file shouldn't be in the PR in the first place, was something else I was working on. Whoops.

packages/table/src/cell/_cell.scss
llorca: +1 use `$pt-grid-size`

gscshoyru: So the issue here is that if I use padding it messes up the ::before absolutely positioned element that shows up to make the input the right size, and messing with those values messes with the editable headers. 

I'll do it the other way, though.

packages/table/src/headers/_headers.scss
```diff
@@ -303,12 +303,24 @@ $dark-selectable-header-menu-selected-hover-background:
...
+      &::before {
+        top: -$cell-padding-vertical + $cell-border-width - 1px;
+        right: -$cell-padding-horizontal + $cell-border-width;
+        bottom: -$cell-padding-vertical + 1px;
+        left: -$cell-padding-horizontal + $cell-border-width - 1px;
+        cursor: $select-text-cursor;
+      }
+
+      &.bp-table-editable-text {
+        &::before {
```
giladgray: I'd merge this selector onto line above, but whatever 👍  `&.bp-table-editable-text::before`

packages/table/src/headers/_headers.scss
```diff
@@ -303,12 +303,24 @@ $dark-selectable-header-menu-selected-hover-background:
...
+          top: -$cell-padding-vertical + $cell-border-width - 1px;
+          right: $cell-border-width;
+          bottom: -$cell-padding-vertical + 1px;
```
giladgray: seems like `top` and `bottom` and `cursor` don't change from selector above. maybe they can be removed from this block, since it necessarily matches the selector above as well?

packages/table/src/cell/_cell.scss
```diff
@@ -50,3 +50,12 @@
 .bp-table-editable-name input {
   height: $cell-height; // fixed height for IE11
 }
+
+.bp-table-editable-text {
+  position: absolute;
+  top: 0;
+  right: 0;
+  bottom: 0;
+  left: 0;
```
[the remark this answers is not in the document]
llorca: `@include position(absolute, 0)`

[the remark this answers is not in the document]
giladgray: bourbon!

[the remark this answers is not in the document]
gscshoyru: So far as I can tell, this thing doesn't exist, or at least SASS was unhappy with it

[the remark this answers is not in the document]
giladgray: it's a bourbon mixin. search codebase for other usages. but don't block this PR on it cuz #875  http://bourbon.io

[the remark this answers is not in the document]
cmslewis: ack

cmslewis:
:shipit:

PR #1674, opened 2017-10-06
[Table] Stop preventDefault'ing on wheel

### LINKED ISSUE #1611: Table should not disable scrolling if there is no actual vertical scrolling for the table.
### Bug report

master version.

#### Steps to reproduce

Go to example page. See attached gif.

#### Actual behavior

Scroll event is disabled on the root div of the table.

#### Expected behavior

Original behaviour, if there was no vertical overflow scrolling on the table, then the scroll event still allowed for scrolling the page.

ezgif-1-10c2e8dd87

cmslewis:

- 🐛 __FIXED__ `Table` no longer swallows wheel events that would otherwise scroll the surrounding page.

- Turns out that explicitly setting the `scrollLeft`/`scrollTop` of the target quadrant is sufficient to keep it in sync with its dependent quadrant in each animation frame.
- This means we don't need to `preventDefault` on the wheel event anymore (we were doing that before to prevent the target quadrant from scrolling faster than dependent quadrants).

### DIFF
--- a/packages/table/src/quadrants/tableQuadrantStack.tsx
+++ b/packages/table/src/quadrants/tableQuadrantStack.tsx
@@ -239,15 +239,11 @@ export class TableQuadrantStack extends AbstractComponent<ITableQuadrantStackPro
     public constructor(props: ITableQuadrantStackProps, context?: any) {
         super(props, context);
 
-        // a few points here:
-        // - we throttle onScroll/onWheel callbacks to making scrolling look more fluid.
-        // - we declare throttled functions on the component instance, since they're stateful.
-        // - "wheel"-ing triggers super-fluid onScroll behavior by default, but relying on that
-        //   causes sync'd quadrants to lag behind. thus, we preventDefault for onWheel and instead
-        //   manually update all relevant quadrants using event.delta{X,Y} later, in the callback.
-        //   this keeps every sync'd quadrant visually aligned in each animation frame.
+        // callbacks trigger too frequently unless we throttle scroll and wheel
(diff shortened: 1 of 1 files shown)


packages/table/src/quadrants/tableQuadrantStack.tsx
```diff
@@ -239,15 +239,11 @@ export class TableQuadrantStack extends AbstractComponent<ITableQuadrantStackPro
...
-        // - we declare throttled functions on the component instance, since they're stateful.
-        // - "wheel"-ing triggers super-fluid onScroll behavior by default, but relying on that
-        //   causes sync'd quadrants to lag behind. thus, we preventDefault for onWheel and instead
-        //   manually update all relevant quadrants using event.delta{X,Y} later, in the callback.
-        //   this keeps every sync'd quadrant visually aligned in each animation frame.
+        // callbacks trigger too frequently unless we throttle scroll and wheel
+        // events. declare these functions on the component instance since
+        // they're stateful.
         this.throttledHandleMainQuadrantScroll = CoreUtils.throttleReactEventCallback(this.handleMainQuadrantScroll);
-        this.throttledHandleWheel = CoreUtils.throttleReactEventCallback(this.handleWheel, { preventDefault: true });
```
themadcreator: But now:
scroll-pull

cmslewis: I'm good with this; it's essentially expected browser behavior, and the benefits of not having your scroll swallowed are worth it (see the "Features" page to feel that benefit in action).

PR #1678, opened 2017-10-06
[Table] Stop rendering superfluous frozen cells

cmslewis:
#### Checklist

- [x] Include tests

- 🐛 __FIXED__ `Table` no longer renders superfluous cells when frozen rows/columns are active.
    - Before, we were rendering cells in the `MAIN` quadrant that were covered by frozen rows/columns, leading to lots of wasted effort painting duplicate cells that would never be visible.
        - Now, we render only the `MAIN`-quadrant cells that lie beyond the frozen regions.
    - Before, we were always painting 1 extra frozen row or column, because we forgot to convert the end index from 1-indexing to 0-index. This extra row/column was never visible, so it was wasted effort.
        - Now, we render only frozen rows/columns that are actually visible.
- Add `Resize rows by tallest cell` button in the dev preview (has an effect only when `Wrap text` is enabled, of course):
image

- This was so easy. I'm a dunce for letting it take this long to fix.
- To test: please try setting the number of rows and columns to various values in `[0, 1, 2, 5]` especially, then Right Click + `Inspect Element` on each quadrant's cells to verify that only the expected indices are painted in each quadrant.


cmslewis:
Test flake. Opened PR #1681 to disable the `Toaster` test in question. Rebuilding.

packages/table/src/table.tsx
```diff
@@ -1144,16 +1168,19 @@ export class Table extends AbstractComponent<ITableProps, ITableState> {
...
+        // start beyond the frozen area if rendering unrelated quadrants, so we
+        // don't render duplicate cells underneath the frozen ones.
+        const columnIndexStart = showFrozenColumnsOnly ? 0 : columnIndices.columnIndexStart + numFrozenColumns;
+        const rowIndexStart = showFrozenRowsOnly ? 0 : rowIndices.rowIndexStart + numFrozenRows;
+
+        // if rendering frozen rows/columns, subtract one to convert to
+        // 0-indexing. if the 1-indexed value is 0, this sets the end index
+        // to -1, which avoids rendering absent frozen rows/columns at all.
+        const columnIndexEnd = showFrozenColumnsOnly ? numFrozenColumns - 1 : columnIndices.columnIndexEnd;
+        const rowIndexEnd = showFrozenRowsOnly ? numFrozenRows - 1 : rowIndices.rowIndexEnd;
```
cmslewis: These few lines are the meat of the change in this PR. Everything else is dressing.

packages/table/src/table.tsx
```diff
@@ -644,8 +662,8 @@ export class Table extends AbstractComponent<ITableProps, ITableState> {
                     isVerticalScrollDisabled={this.shouldDisableVerticalScroll()}
                     loadingOptions={loadingOptions}
                     numColumns={React.Children.count(children)}
-                    numFrozenColumns={this.getNumFrozenColumnsClamped()}
-                    numFrozenRows={this.getNumFrozenRowsClamped()}
+                    numFrozenColumns={this.state.numFrozenColumnsClamped}
+                    numFrozenRows={this.state.numFrozenRowsClamped}
```
themadcreator: nit: destructure these

packages/table/src/table.tsx
```diff
@@ -136,12 +136,16 @@ export interface ITableProps extends IProps, IRowHeights, IColumnWidths {
     loadingOptions?: TableLoadingOption[];
 
     /**
-     * The number of columns to freeze to the left side of the table, counting from the leftmost column.
+     * The number of columns to freeze to the left side of the table, counting
+     * from the leftmost column.
+     * @default 0
```
themadcreator: is there any reason why we still need to maintain these values since we appear to be using `numFrozenColumnsClamped` everywhere? If there is a place where we still use the raw `numFrozenColumns` can you add something to this comment explaining why we have to keep this value around?

cmslewis: Yeah you're right! Will delete the `@default`s.

themadcreator: false alarm

cmslewis: Sync'd up offline. We're on the same page now. Adding the `@default`s back in.

PR #1679, opened 2017-10-06
[Table] Always disable batch-rendering in headers

cmslewis:

- 🔄 __CHANGED__ `Table` now always disables batch rendering for header cells.
    - __Why?__ Because we kept seeing race conditions where we tried to sync quadrant row-header sizes before the final header cells finished rendering, leading to mismeasurements. The cost here is likely minimal, since header cells are less numerous than body cells and generally won't have tons of content in them (e.g. they won't have huge JSON blobs or `detectTruncation` logic).

### DIFF
--- a/packages/table/src/headers/header.tsx
+++ b/packages/table/src/headers/header.tsx
@@ -10,7 +10,6 @@ import * as classNames from "classnames";
 import * as React from "react";
 
 import { Grid } from "../common";
-// import { Batcher } from "../common/batcher";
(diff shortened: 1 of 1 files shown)


llorca:
Feels pretty good for column headers. Is it now also disabled for row headers? They don't particularly feel that much faster

cmslewis:
@llorca yep, it's disabled in the generic `Header` component, which both row and column headers use internally.

Issue #1680, opened 2017-10-06
Fix flaky tests


cmslewis:
These tests often flaked and caused builds to fail. They're `it.skip`'d for now, but we should find a way to make them more reliable and reenable them:
- [ ] `Toaster > with autoFocus set to true > focuses on newly created toast`
- [x] `<Overlay> > Focus management > brings focus to overlay if autoFocus=true`

PR #1681, opened 2017-10-06
[Core] Skip flaky test, add comments mentioning FLUP issue

cmslewis:

- __FIXED__ `Toaster` flaky test has been disabled until we can fix it.
    - Follow up issue: https://github.com/palantir/blueprint/issues/1680


cmslewis:
Eh not really. GH issues are the real solution, which is what I've done here.

cmslewis:
Honestly, IMO `TODO`s should be reserved for work that should be done before opening the PR. Don't always do that though.

PR #1684, opened 2017-10-06
[Core/Overlay] Fix overlay unmount issue that prevented scrolling

hmarr:
#### Checklist

- [x] Include tests

When overlays are unmounted, we currently only re-enable scrolling if there are no other overlays in the stack. However, if the none of the remaining overlays would prevent scrolling, we should also re-enable scrolling on unmount.

This has been causing the following issue:
1. A modal dialog is opened (which uses an overlay that prevents scrolling)
2. A user action triggers a toast (which also uses an overlay, but shouldn't prevent scrolling)
3. The dialog closes, but the toast remains visible. At this stage, scrolling should be re-enabled, but isn't.

The new tests, and the logic for determining whether an overlay in the stack would prevent scrolling.


palantirtech:
Thanks for your interest in [palantir/blueprint](https://github.com/palantir/blueprint), @hmarr! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

PR #1647, opened 2017-10-02
[DateTime] Add dayPickerProps to datetime components

tobilen:
#### Adresses
* https://github.com/palantir/blueprint/issues/1646
* https://github.com/palantir/blueprint/issues/1574
* https://github.com/palantir/blueprint/issues/1548
* https://github.com/palantir/blueprint/issues/1414
* https://github.com/palantir/blueprint/issues/1184

#### Changes proposed in this pull request:

~adds `firstDayOfWeek` prop to datetime components. This prop gets passed to react-day-picker to determine which day should be the first day of the week (i.e. sunday for US, monday for EU)~

adds `dayPickerProps` prop to datetime components. This prop can be used to pass props directly to react-day-picker.
Props that overlap into blueprint (i.e. event handlers, `disabledDays`, `locale`) are reconciliated on a case-by-case basis.

---------------------------------------------------------------------------
Incorporate dayPickerProps into event handlers and native props receonzile app logic, write lots of tests for:
* [x] DatePicker
* [x] DateRangePicker
* [x] DateInput
* [x] DateRangeInput

PR #1693, opened 2017-10-09
Revert "[DateTime] Add dayPickerProps to datetime components"

cmslewis:
Reverts palantir/blueprint#1647

Will re-merge after 1.31.0 is cut in a moment and include this in 1.32.0, just so we can polish some final things.


cmslewis:
Huh, weird that there's a `Revert "Prepare Release 1.31.0 (#1687)"` commit here also. But regardless, the only changes reverted were from #1647, as desired.

PR #1696, opened 2017-10-09
Revert "Revert "[DateTime] Add dayPickerProps to datetime components""

cmslewis:
Reverts palantir/blueprint#1693

Undoing the revert to get the #1647 changes back into `master` now that we've cut the 1.31.0 release.


cmslewis:
Merging because this code was already reviewed and has clear followups (from #1647).

PR #1704, opened 2017-10-11
[Table] Prototype: resizeRowsByApproximateHeight

cmslewis:

- 🎉 __Prototype:__ An scheme for resizing rows to their approximate actual/"desired" size assuming fixed column widths, via **`resizeRowsByApproximateHeight`**!
    - Given fixed column widths, sets each row height to the maximum approximate height of the cells in that row, based on average character width and line height (currently hardcoded).
    - Verdict: the scheme works rather well!
    - Does _zero_ DOM access...at the moment.
    - If we want to invest more effort, we can use canvas's `measureText` function to:
        - determine the average alpha character width using the provided font. 
        - determine the line height using the provided font (maybe using https://stackoverflow.com/questions/1134586/how-can-you-find-the-height-of-text-on-an-html-canvas)

To test:
1. Set `Word wrap: on`.
1. Set `Cell content: "Long text"`
1. Click the `Resize rows by approx height` button.

**What do you think?**

### DIFF
--- a/packages/table/src/table.tsx
+++ b/packages/table/src/table.tsx
@@ -543,6 +543,45 @@ export class Table extends AbstractComponent<ITableProps, ITableState> {
         this.quadrantStackInstance.scrollToPosition(correctedScrollLeft, correctedScrollTop);
     }
 
+    public resizeRowsByApproximateHeight(getCellText: (rowIndex: number, columnIndex: number) => string) {
+        const APPROX_CHAR_WIDTH = 8;
+        const APPROX_LINE_HEIGHT = 18;
+        const NUM_EXTRA_LINES = 1; // extra lines
+        const CELL_HORIZONTAL_PADDING = 10;
+
+        const { numRows } = this.props;
+        const { columnWidths } = this.state;
+        const numColumns = columnWidths.length;
+
+        const rowHeights: number[] = [];
+
+        for (let rowIndex = 0; rowIndex < numRows; rowIndex++) {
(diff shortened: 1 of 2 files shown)


packages/table/src/table.tsx
```diff
@@ -543,6 +543,45 @@ export class Table extends AbstractComponent<ITableProps, ITableState> {
         this.quadrantStackInstance.scrollToPosition(correctedScrollLeft, correctedScrollTop);
     }
 
+    public resizeRowsByApproximateHeight(getCellText: (rowIndex: number, columnIndex: number) => string) {
+        const APPROX_CHAR_WIDTH = 8;
+        const APPROX_LINE_HEIGHT = 18;
+        const NUM_EXTRA_LINES = 1; // extra lines
```
themadcreator: if i were using this API i would want access to the above variables as optional parameters

cmslewis: Reasonable. 👍

cmslewis:
Cool. I'll close this prototype PR and add some more params/font-checking then.

adidahiya:
Do we really need font checking right now? At Palantir we only use the table with one font. I would be fine with leaving that as a future enhancement.

llorca:
You sure we only use on font family, size, weight, letter spacing, etc.? I wouldn't be so certain. We don't have to do checking just yet, but we should at least parametrize the current static values so consumers have an escape hatch. Consumers can customize every cell individually if they want to, that's the kind of granularity we're dealing with.

Issue #1706, opened 2017-10-12
[Table] Shift-click to expand selection is buggy when not focused in last-selected region 


cmslewis:
### Bug report

- __Package version(s)__: 
- __Browser and OS versions__: 

#### Steps to reproduce

With `allowMultipleSelection=true`:

1. `Cmd/Ctrl + Click` to select a few disjoint regions.
1. `Tab`/`Shift + Tab`/`Enter`/`Shift + Enter` to move from the last-selected region to another one.
1. `Shift + Click` to expand the focused region.

#### Actual behavior

The last-selected region is cleared and replaced with a region that has been expanded from the focused cell.

![2017-10-12 00 31 42](https://user-images.githubusercontent.com/443450/31484357-db263006-aee4-11e7-9db2-7df785396cdc.gif)

#### Expected behavior

Only the focused region is affected:

![2017-10-12 00 34 42](https://user-images.githubusercontent.com/443450/31484446-322435ce-aee5-11e7-90f2-eb164ea0c95f.gif)

PR #1707, opened 2017-10-12
[Table] Shift + arrow keys to resize selection

cmslewis:
#### Checklist

- [x] Include tests

- __NEW__ `Table` <kbd>Shift + ⬆️ / ⬇️ / ⬅️ / ➡️ </kbd> hotkeys to resize selection (using focused cell as anchor if present).
    - See function comments in the diff for details (in particular: [this comment](https://github.com/palantir/blueprint/pull/1707/files#diff-3c6700ce4c8768ee12a26d3b6b187392R14)).

- Really would like to decompose `<Table>`, but not in this PR. 
- I discovered a bug with somewhat-related behavior: #1706 

#### Screenshot

When `allowMultipleSelection=true`:

### DIFF
--- a/packages/table/src/common/internal/directionUtils.ts
+++ b/packages/table/src/common/internal/directionUtils.ts
@@ -1,0 +1,24 @@
+/**
+ * Copyright 2017 Palantir Technologies, Inc. All rights reserved.
+ * Licensed under the BSD-3 License as modified (the “License”); you may obtain a copy
+ * of the license at https://github.com/palantir/blueprint/blob/master/LICENSE
+ * and https://github.com/palantir/blueprint/blob/master/PATENTS
+ */
+
+import { Direction } from "../direction";
+import { IMovementDelta } from "../movementDelta";
+
+export function directionToDelta(direction: Direction): IMovementDelta {
+    switch (direction) {
+        case Direction.UP:
+            return { rows: -1, cols: 0 };
+        case Direction.DOWN:
+            return { rows: +1, cols: 0 };
+        case Direction.LEFT:
+            return { rows: 0, cols: -1 };
+        case Direction.RIGHT:
+            return { rows: 0, cols: +1 };
+        default:
+            return undefined;
+    }
+}
--- a/packages/table/src/common/internal/selectionUtils.ts
+++ b/packages/table/src/common/internal/selectionUtils.ts
@@ -11,20 +11,17 @@ import { Direction } from "../direction";
 import { IMovementDelta } from "../movementDelta";
 import * as FocusedCellUtils from "./focusedCellUtils";
 
-export function resizeLastSelectedRegion(
-    selectedRegions: IRegion[],
-    direction: Direction,
-    focusedCell?: IFocusedCellCoordinates,
-) {
-    const lastSelectedRegion = getLastSelectedRegion(selectedRegions);
-    const cardinality = Regions.getRegionCardinality(lastSelectedRegion);
-    const delta = directionToMovementDelta(direction);
-
-    if (cardinality === RegionCardinality.FULL_TABLE) {
-        return selectedRegions;
+export function resizeSelectedRegion(region: IRegion, direction: Direction, focusedCell?: IFocusedCellCoordinates) {
+    if (Regions.getRegionCardinality(region) === RegionCardinality.FULL_TABLE) {
+        // return the same instance to maintain referential integrity and avoid
+        // unnecessary updates.
+        return region;
     }
 
-    const nextRegion: IRegion = Regions.copy(lastSelectedRegion);
+    const nextRegion: IRegion = Regions.copy(region);
+
+    let affectedRowIndex: number = 0;
+    let affectedColumnIndex: number = 0;
 
     if (focusedCell != null) {
         const isAtTop = FocusedCellUtils.isFocusedCellAtRegionTop(nextRegion, focusedCell);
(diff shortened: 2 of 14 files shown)


packages/table/src/common/internal/selectionUtils.ts
```diff
@@ -0,0 +1,120 @@
...
+ * Other notes:
+ * - A CELLS region can be resized vertically or horizontally.
+ * - A FULL_ROWS region can be resized only vertically.
+ * - A FULL_COLUMNS region can be resized only horizontally.
+ * - A FULL_TABLE region cannot be resized.
+ *
+ * This function does not clamp the indices of the returned region; that is the
+ * responsibility of the caller.
+ */
+export function resizeSelectedRegion(region: IRegion, direction: Direction, focusedCell?: IFocusedCellCoordinates) {
```
cmslewis: How are these terms: `resize` (used on shift + arrow keys) vs `expand` (used on shift + click). Note that they behave differently when the focused cell is floating within the selection (due to <kbd>Tab</kbd>ing or <kbd>Enter</kbd>ing)! Would love to figure out some elegant way to marry all this logic and do away with the need for two names in the first place.

Shift + arrow keys:


Shift + click:

packages/table/src/common/internal/selectionUtils.ts
```diff
@@ -0,0 +1,120 @@
...
+    // sanitizing the result.
+    return nextRegion;
+}
+
+/**
+ * Expands the last-selected region to the new region, and replaces the
+ * last-selected region with the expanded region. If a focused cell is provided,
+ * the focused cell will serve as an anchor for the expansion.
+ */
+export function expandLastSelectedRegion(
```
cmslewis: In fixing #1706, I'd like to rename this to `expandSelectedRegion` and let the caller determine which region to operate on (based on the `focusedSelectionIndex` etc).

packages/table/src/common/movementDelta.ts
```diff
@@ -0,0 +1,14 @@
+/**
+ * Copyright 2017 Palantir Technologies, Inc. All rights reserved.
+ * Licensed under the BSD-3 License as modified (the “License”); you may obtain a copy
+ * of the license at https://github.com/palantir/blueprint/blob/master/LICENSE
+ * and https://github.com/palantir/blueprint/blob/master/PATENTS
+ */
+
+export interface IMovementDelta {
```
cmslewis: How's the name / how's the construct here?

packages/table/src/common/internal/directionUtils.ts
```diff
@@ -0,0 +1,24 @@
...
+ * Copyright 2017 Palantir Technologies, Inc. All rights reserved.
+ * Licensed under the BSD-3 License as modified (the “License”); you may obtain a copy
+ * of the license at https://github.com/palantir/blueprint/blob/master/LICENSE
+ * and https://github.com/palantir/blueprint/blob/master/PATENTS
+ */
+
+import { Direction } from "../direction";
+import { IMovementDelta } from "../movementDelta";
+
+export function directionToDelta(direction: Direction): IMovementDelta {
```
cmslewis: Would also like to use this in `handleFocusMove`, but not in this PR:
https://github.com/palantir/blueprint/blob/be0f870933c5e8b71d87ac248ee680973056efaa/packages/table/src/table.tsx#L1704-L1719

packages/table/src/common/internal/directionUtils.ts
```diff
@@ -0,0 +1,24 @@
...
+    switch (direction) {
+        case Direction.UP:
+            return { rows: -1, cols: 0 };
+        case Direction.DOWN:
+            return { rows: +1, cols: 0 };
```
cmslewis: `+` for clarity / pretty vertical alignment in code

packages/table/src/interactions/selectable.tsx
```diff
@@ -181,7 +182,7 @@ export class DragSelectable extends React.Component<IDragSelectableProps, {}> {
         }
 
         const nextSelectedRegions = this.didExpandSelectionOnActivate
-            ? this.expandSelectedRegions(selectedRegions, region, focusedCell)
+            ? SelectionUtils.expandLastSelectedRegion(selectedRegions, region, focusedCell)
```
cmslewis: Didn't _really_ need to move this code for this PR after all, but I like it in its new home so that you can see the parallelism of naming / possible parallelism of implementation / etc.

packages/table/src/regions.ts
```diff
@@ -169,6 +169,16 @@ export class Regions {
     }
 
     /**
+     * Returns a deep copy of the provided region.
+     */
+    public static copy(region: IRegion): IRegion {
```
cmslewis: Shallower copies from spreads: `{ ...region }` were causing `shouldComponentUpdate` checks to pass erroneously early in this work. Introduced this function as a hammer; may not need this though. LMK what you think.

packages/table/src/regions.ts
```diff
@@ -211,16 +221,37 @@ export class Regions {
 
     /**
      * Replaces the region at the end of a cloned copy of the supplied region
-     * array.
+     * array, or at the specific index if one is provided.
      */
-    public static update(regions: IRegion[], region: IRegion) {
+    public static update(regions: IRegion[], region: IRegion, index?: number) {
```
cmslewis: `index` param necessary now that we need to modify regions at particular indices in the `selectedRegions` array.

packages/table/src/regions.ts
```diff
@@ -211,16 +221,37 @@ export class Regions {
...
+        }
         return copy;
     }
 
     /**
+     * Clamps the region's start and end indices between 0 and the provided
+     * maximum values.
+     */
+    public static clampRegion(region: IRegion, maxRows: number, maxColumns: number) {
+        const nextRegion = Regions.copy(region);
```
cmslewis: Also makes this line nice and concise.

packages/table/src/table.tsx
```diff
@@ -781,6 +784,188 @@ export class Table extends AbstractComponent<ITableProps, ITableState> {
         }
     }
 
+    // Hotkeys
+    // =======
+
+    private maybeRenderCopyHotkey() {
```
cmslewis: Moved all these existing hotkey renders into one area of the file.

packages/table/src/table.tsx
```diff
@@ -781,6 +784,188 @@ export class Table extends AbstractComponent<ITableProps, ITableState> {
...
+                    combo="mod+c"
+                    onKeyDown={this.handleCopy}
+                />
+            );
+        } else {
+            return undefined;
+        }
+    }
+
+    private maybeRenderSelectionResizeHotkeys() {
```
cmslewis: This is actually new code.

packages/table/src/table.tsx
```diff
@@ -781,6 +784,188 @@ export class Table extends AbstractComponent<ITableProps, ITableState> {
...
+        const { allowMultipleSelection, selectionModes } = this.props;
+        const isSomeSelectionModeEnabled = selectionModes.length > 0;
+
+        if (allowMultipleSelection && isSomeSelectionModeEnabled) {
```
cmslewis: Any other cases we'd want this feature disabled?

packages/table/src/table.tsx
```diff
@@ -781,6 +784,188 @@ export class Table extends AbstractComponent<ITableProps, ITableState> {
...
+                    combo="mod+a"
+                    onKeyDown={this.handleSelectAllHotkey}
+                />
+            );
+        } else {
+            return undefined;
+        }
+    }
+
+    // Selection resize
```
cmslewis: More new code from here down.

packages/table/src/table.tsx
```diff
@@ -781,6 +784,188 @@ export class Table extends AbstractComponent<ITableProps, ITableState> {
...
+                    combo="shift+right"
+                    onKeyDown={this.handleSelectionResizeRight}
+                />,
+            ];
+        } else {
+            return undefined;
+        }
+    }
+
+    private maybeRenderFocusHotkeys() {
```
cmslewis: Old code again.

packages/table/src/table.tsx
```diff
@@ -781,6 +784,188 @@ export class Table extends AbstractComponent<ITableProps, ITableState> {
...
+        e.preventDefault();
+        e.stopPropagation();
+
+        const { focusedCell, selectedRegions } = this.state;
+
+        if (selectedRegions.length === 0) {
+            return;
+        }
+
+        const index = FocusedCellUtils.getFocusedOrLastSelectedIndex(selectedRegions, focusedCell);
```
cmslewis: Here's where we determine which index to actually affect if we want the feature to behave properly (unlike #1706).

packages/table/src/common/internal/selectionUtils.ts
cmslewis: Another naming thought for this function: we're effectively creating a minimum bounding box around the focused cell (or if it doesn't exist: the last-selected region) and the provided `newRegion`. Maybe we can name the two functions:

- `getResizedRegion`
- `getBoundingRegion` or `getMinimalBoundingRegion`

?

cmslewis: It's also somewhat ambiguous why these functions live in `SelectionUtils` vs. somewhere else. Maybe everything belongs in `Regions.ts`?

cmslewis:
@llorca @giladgray @themadcreator @gscshoyru - preview is up. Ready for CR.

packages/table/src/table.tsx
gscshoyru: Why does allowMultipleSelection need to be necessary? You should be able to resize selection even in single-select modes.

gscshoyru: ...oh, I see, allowMultipleSelection doesn't let you have more than one cell. I thought it was more than one region. Ignore this comment.

cmslewis: When `allowMultipleSelection=false`, that actually means only one cell/column/row (or the full table) can be selected at a time:



The prop name is unclear; that's why I want to rename it. See #1625.

cmslewis:
Hmm, my reply to @gscshoyru isn't showing up for me here on the main thread. Here's the GIF of what _should_ happen when `allowMultipleSelection=false`:



Note that only a single cell can be selected at a time, which means we should disable shift + arrow keys.

gscshoyru:
Yeah, I commented on the thing as well. Don't know why it won't show up in the main thread. I tested it out afterwards and saw how it actually behaved. 

Speaking of renaming it -- _should_ we add a prop that makes it behave the way I expected this one to?

cmslewis:
Yep, I think so. Would be nice to support one expandable selection, but not necessarily disjoint selections.

PR #1714, opened 2017-10-12
[Core/Button] Remove outdated note about minimal buttons in button groups

rootulp:
For consistency with other notes like [here](http://blueprintjs.com/docs/#core/components/callout.css-api)


llorca:
I think we can actually remove this note -- button groups do support minimal buttons: http://blueprintjs.com/docs/#core/components/button-group.css-api

rootulp:
Removed note

PR #1717, opened 2017-10-13
don't calculate the position of the tab-indicator-wrapper if it won't…

mcintyret:
… be shown

Don't calculate the new position of the selected tab indicator if not required. (i.e. when not animating). This method involves calculating elements widths and therefore forces a layout which is non-trivial for performance (see  https://developers.google.com/web/fundamentals/performance/rendering/avoid-large-complex-layouts-and-layout-thrashing#avoid-forced-synchronous-layouts).

Also include a mini-optimisation where the indicator div is never created if it won't be used.


palantirtech:
Thanks for your interest in [palantir/blueprint](https://github.com/palantir/blueprint), @mcintyret! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

PR #1729, opened 2017-10-17
[Table] Edit focused cells on f2

gscshoyru:
Editable cells can be edited when focused by pressing `f2` -- so one can edit editable tables merely with `f2`, `enter`, and the arrow keys -- no mouse required!

This is done by making cells able to be browser-focused, and having them focus themselves whenever, well, they become focused. This also lets the focus stay on the table when you're done editing a cell. 

There's... a bit of a horrible hack to allow the hotkeys to... pass through to the div beneath. I'm not sure if we should keep this or if there's a better way, or what? We could make some sort of slightly better version of cells having these endpoints, and make it less of a hack, maybe? Or fix hotkeys to find the DOM element underneath using `ReactDOM.findDOMNode`?

This doesn't work as well with the cell render batching if you're far down and to the right enough -- you can press f2 before the rendering is caught up and be editing the wrong cell for a short period of time before the rendering catches up and snaps it away. I'm... not sure what to do about this. 

#### Screenshot

oct-17-2017 17-31-51

### DIFF
--- a/packages/table/src/cell/_cell.scss
+++ b/packages/table/src/cell/_cell.scss
@@ -21,6 +21,11 @@
     @include cell-content-align-vertical();
     color: transparent;
   }
+
+  &:focus {
+    // disable focus outline on cells; we already have the focus border
+    outline: none;
+  }
 }
 
 .bp-table-cell-interactive {
--- a/packages/table/src/cell/cell.tsx
+++ b/packages/table/src/cell/cell.tsx
@@ -71,12 +71,12 @@ export interface ICellProps extends IIntentProps, IProps {
     tabIndex?: number;
 
     /**
-     * Callback to be called when the cell is focused and a key is pressed down.
+     * Callback invoked when the cell is focused and a key is pressed down.
      */
     onKeyDown?: React.KeyboardEventHandler<HTMLElement>;
 
     /**
-     * Callback to be called when the cell is focused and a key is released.
+     * Callback invoked when the cell is focused and a key is released.
      */
     onKeyUp?: React.KeyboardEventHandler<HTMLElement>;
 
--- a/packages/table/src/cell/cell.tsx
+++ b/packages/table/src/cell/cell.tsx
@@ -64,6 +64,26 @@ export interface ICellProps extends IIntentProps, IProps {
      * @default false
      */
     wrapText?: boolean;
+
+    /**
+     * Allows for setting a tab index on the cell, so the cell can be browser-focusable.
+     */
+    tabIndex?: number;
+
+    /**
+     * Callback to be called when cell is focused and key is pressed down.
+     */
+    onKeyDown?: React.KeyboardEventHandler<HTMLElement>;
+
+    /**
+     * Callback to be called when cell is focused and key is released.
+     */
+    onKeyUp?: React.KeyboardEventHandler<HTMLElement>;
+
+    /**
+     * A ref handle to capture the outer div of this cell. Used internally.
+     */
+    cellRef?: (ref: HTMLElement) => void;
 }
 
 export type ICellRenderer = (rowIndex: number, columnIndex: number) => React.ReactElement<ICellProps>;
@@ -85,7 +105,20 @@ export class Cell extends React.Component<ICellProps, {}> {
     }
 
     public render() {
-        const { style, intent, interactive, loading, tooltip, truncated, className, wrapText } = this.props;
+        const {
+            cellRef,
+            tabIndex,
+            onKeyDown,
+            onKeyUp,
+            style,
+            intent,
+            interactive,
+            loading,
+            tooltip,
+            truncated,
+            className,
+            wrapText,
+        } = this.props;
 
         const classes = classNames(
             Classes.TABLE_CELL,
@@ -132,7 +165,15 @@ export class Cell extends React.Component<ICellProps, {}> {
         const content = <div className={textClasses}>{modifiedChildren}</div>;
 
         return (
-            <div className={classes} style={style} title={tooltip}>
+            <div
+                className={classes}
+                style={style}
+                title={tooltip}
+                tabIndex={tabIndex}
+                onKeyDown={onKeyDown}
+                onKeyUp={onKeyUp}
+                ref={cellRef}
+            >
                 <LoadableContent loading={loading} variableLength={true}>
                     {content}
                 </LoadableContent>
--- a/packages/table/src/cell/cell.tsx
+++ b/packages/table/src/cell/cell.tsx
@@ -71,12 +71,12 @@ export interface ICellProps extends IIntentProps, IProps {
     tabIndex?: number;
 
     /**
-     * Callback to be called when cell is focused and key is pressed down.
+     * Callback to be called when the cell is focused and a key is pressed down.
      */
     onKeyDown?: React.KeyboardEventHandler<HTMLElement>;
 
     /**
-     * Callback to be called when cell is focused and key is released.
+     * Callback to be called when the cell is focused and a key is released.
      */
     onKeyUp?: React.KeyboardEventHandler<HTMLElement>;
 
@@ -165,15 +165,7 @@ export class Cell extends React.Component<ICellProps, {}> {
         const content = <div className={textClasses}>{modifiedChildren}</div>;
 
         return (
-            <div
-                className={classes}
-                style={style}
-                title={tooltip}
-                tabIndex={tabIndex}
-                onKeyDown={onKeyDown}
-                onKeyUp={onKeyUp}
-                ref={cellRef}
-            >
+            <div className={classes} title={tooltip} ref={cellRef} {...{ style, tabIndex, onKeyDown, onKeyUp }}>
                 <LoadableContent loading={loading} variableLength={true}>
                     {content}
                 </LoadableContent>
--- a/packages/table/src/cell/editableCell.tsx
+++ b/packages/table/src/cell/editableCell.tsx
@@ -8,13 +8,18 @@
 import * as classNames from "classnames";
 import * as React from "react";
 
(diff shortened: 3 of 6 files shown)


packages/table/src/cell/cell.tsx
```diff
@@ -64,6 +64,26 @@ export interface ICellProps extends IIntentProps, IProps {
...
      */
     wrapText?: boolean;
+
+    /**
+     * Allows for setting a tab index on the cell, so the cell can be browser-focusable.
+     */
+    tabIndex?: number;
+
+    /**
+     * Callback to be called when cell is focused and key is pressed down.
```
cmslewis: Articles or no articles? That is the question. Whether 'tis nobler in (the?) mind to suffer...

packages/table/src/cell/cell.tsx
```diff
@@ -64,6 +64,26 @@ export interface ICellProps extends IIntentProps, IProps {
...
+     */
+    onKeyDown?: React.KeyboardEventHandler<HTMLElement>;
+
+    /**
+     * Callback to be called when cell is focused and key is released.
```
cmslewis: No articles

packages/table/src/cell/cell.tsx
```diff
@@ -64,6 +64,26 @@ export interface ICellProps extends IIntentProps, IProps {
...
+     */
+    onKeyUp?: React.KeyboardEventHandler<HTMLElement>;
+
+    /**
+     * A ref handle to capture the outer div of this cell. Used internally.
```
cmslewis: Articles

packages/table/src/cell/editableCell.tsx
```diff
@@ -8,14 +8,19 @@
...
-import { EditableText, Utils as CoreUtils } from "@blueprintjs/core";
+import { EditableText, Hotkey, Hotkeys, HotkeysTarget, Utils as CoreUtils } from "@blueprintjs/core";
 
 import * as Classes from "../common/classes";
 import { Draggable } from "../interactions/draggable";
 import { Cell, ICellProps } from "./cell";
 
 export interface IEditableCellProps extends ICellProps {
     /**
+     * Whether or not the given cell is the current active/focused cell.
```
cmslewis: Do we need the `or not`?

https://english.stackexchange.com/questions/3382/whether-or-not-vs-whether

packages/table/src/cell/editableCell.tsx
```diff
@@ -128,6 +144,31 @@ export class EditableCell extends React.Component<IEditableCellProps, IEditableC
         );
     }
 
+    public renderHotkeys() {
+        return (
+            <Hotkeys>
+                <Hotkey
+                    key="edit-cell"
+                    label="Edit the currely focused cell"
```
cmslewis: typo `currely` => `currently`

packages/table/src/cell/editableCell.tsx
```diff
@@ -128,6 +144,31 @@ export class EditableCell extends React.Component<IEditableCellProps, IEditableC
...
+    }
+
+    private checkShouldFocus() {
+        if (this.props.isFocused && !this.state.isEditing) {
+            // don't focus if we're editing -- we'll lose the fact that we're editing
+            this.cellRef.focus();
+        }
+    }
+
+    private handleCellRef = (cellRef: HTMLElement) => {
```
cmslewis: Our convention for ref handlers is:

```tsx
private cellRef: HTMLElement;

// ...

private refHandlers = {
    cell: (ref: HTMLElement) => this.cellRef = ref;
};

File: packages/table/src/tableBodyCells.tsx [hunk 7]
```diff
@@ -24,6 +25,11 @@ export interface ITableBodyCellsProps extends IRowIndices, IColumnIndices, IProp
     cellRenderer: ICellRenderer;
 
     /**
+     * The location of the focused cell, for setting the "focused" property on cells
```
cmslewis: 1. Better phrasing: "The _coordinates_ of the currently focused cell" or "The currently focused cell"? No need to explain why it's needed IMO. Or, since this is just a codebase-internal component, we can explicitly say "..., for setting the 'isFocused' prop on cells."
1. Need trailing period.

packages/table/src/tableBodyCells.tsx
```diff
@@ -187,7 +193,8 @@ export class TableBodyCells extends React.Component<ITableBodyCellsProps, {}> {
         const cellLoading = baseCell.props.loading != null ? baseCell.props.loading : loading;
 
         const style = { ...baseCell.props.style, ...Rect.style(rect) };
-        return React.cloneElement(baseCell, { className, key, loading: cellLoading, style });
+        const isFocused = focusedCell != null && focusedCell.row === rowIndex && focusedCell.col === columnIndex;
```
cmslewis: Can we add a `CellUtils.cellsEqual` utility function? (in `common/internal/cellUtils` + a few quick unit tests). This check seems like something that could pop up again.

packages/table/src/cell/_cell.scss
```diff
@@ -21,6 +21,11 @@
     @include cell-content-align-vertical();
     color: transparent;
   }
+
+  &:focus {
+    // disable focus outline on cells; we already have the focus border
+    outline: none;
```
themadcreator: does this rule affect accessibility? I'm not sure how accessibility features in browsers deal with focus style overrides. maybe it doesn't matter?

packages/table/src/cell/cell.tsx
themadcreator: You should probably add to this comment whether or not these callbacks are called while the cell is being edited.

packages/table/src/cell/cell.tsx
```diff
@@ -132,7 +165,15 @@ export class Cell extends React.Component<ICellProps, {}> {
         const content = <div className={textClasses}>{modifiedChildren}</div>;
 
         return (
```
themadcreator: I've started using destructuring assignment in bp-chart for cases like this. e.g.:
```
{...{ style, tooltip, tabIndex, onKeyDown, onKeyUp }}
```

packages/table/src/tableBodyCells.tsx
themadcreator: Don't we have something like this for regions?

packages/table/src/cell/_cell.scss
llorca: This only affects visuals, not interactivity

packages/table/src/cell/cell.tsx
gscshoyru: They are... but since you're in a text field they're ignored. These are used by the hotkeys -- i.e. this is the hack, passing the props down to the actual dom element.

packages/table/src/tableBodyCells.tsx
cmslewis: We do.

giladgray:
found a bug! can press f2 repeatedly to put a bunch of cells in editing state:
screen shot 2017-10-18 at 3 04 46 pm

packages/table/src/tableBodyCells.tsx
gscshoyru: I'm... not sure how cellsEqual or anything of the like would make sense here. I'm comparing a focused coordinates with two numbers passed into the function. If I made a cellsEqual, what would I be passing it as arguments, exactly?

gscshoyru:
@themadcreator I think we can do the different and additonal hotkeys stuff in a later commit, yes? Unless you all think it'd be better if it happened now, in this commit. 

Also the hackery is in Cell.tsx -- hotkeys passes down a tabIndex and onKeyDown/Up props -- which if the child is not a DOM element (such as in this case) they are ignored. So I added them as props to the Cell and pass them along manually. 

@giladgray how exactly are you surfacing that?

packages/table/src/tableBodyCells.tsx
cmslewis: Ah, didn't look closely at the comparison values. You're right.

packages/table/src/cell/cell.tsx
```diff
@@ -64,6 +64,26 @@ export interface ICellProps extends IIntentProps, IProps {
...
      */
     wrapText?: boolean;
+
+    /**
+     * Allows for setting a tab index on the cell, so the cell can be browser-focusable.
+     */
+    tabIndex?: number;
+
+    /**
+     * Callback to be called when the cell is focused and a key is pressed down.
```
cmslewis: nit: `Callback invoked` is more concise

packages/table/src/cell/cell.tsx
```diff
@@ -64,6 +64,26 @@ export interface ICellProps extends IIntentProps, IProps {
...
+     */
+    onKeyDown?: React.KeyboardEventHandler<HTMLElement>;
+
+    /**
+     * Callback to be called when the cell is focused and a key is released.
```
cmslewis: Same

packages/table/src/cell/editableCell.tsx
```diff
@@ -53,12 +58,20 @@ export interface IEditableCellState {
...
     public static defaultProps = {
         truncated: true,
         wrapText: false,
     };
 
+    private cellRef: HTMLElement;
+    private refHandlers = {
+        cell: (ref: HTMLElement) => {
+            this.cellRef = ref;
+        },
```
cmslewis: nit: Change this to a one-liner:

```tsx
cell: (ref: HTMLElement) => this.cellRef = ref,
```

cmslewis:
@gscshoyru feel free to address the nits if you want, then merge.

packages/table/src/cell/editableCell.tsx
gscshoyru: Prettier does this. Or at least, adds parens if I don't use curlies, which is honestly silly.

PR #1746, opened 2017-10-24
[Dialog] Fix `user-select` value

llorca:
Change `user-select` from `initial` to `text` for `.pt-dialog`. Chrome 62 regression.

Can someone check that it works as expected with Chrome <= 61?


cmslewis:
Works in Chrome 61 just fine.

PR #1748, opened 2017-10-24
Remove hover in sample text

rootulp:
I don't see any hover interaction here


rootulp:
@llorca where should I be hovering?

rootulp:
Ahh sorry I was moving too quickly to see the hover activated tooltip

PR #1759, opened 2017-10-27
[Core/Button] Fix erroneous padding when text=""

cmslewis:
#### Checklist

- [x] Include tests

- 🐛  __FIXED__ `Button` (and `AnchorButton`) with `iconName` no longer include erroneous right padding when `text=""`.
    - I'm coercing `""` children to `undefined` so that this rule is applied:
    ```scss
    // ensure icon button with no text is a perfect square
    &[class*="pt-icon-"]:empty {
        padding: 0;
    
        &::before {
          margin-right: 0;
        }
    }
    ```

#### Screenshot

___BEFORE:___

`text=""`

image

`iconName="edit"`

image

`rightIconName="edit"`

image

___AFTER:___

`text=""`
image

`iconName="edit"`
image

`rightIconName="edit"` _(unchanged)_
image

### DIFF
--- a/packages/core/src/components/button/abstractButton.tsx
+++ b/packages/core/src/components/button/abstractButton.tsx
@@ -117,8 +117,11 @@ export abstract class AbstractButton<T> extends React.Component<React.HTMLProps<
         const { loading, rightIconName, text } = this.props;
 
         const children = React.Children.map(this.props.children, (child, index) => {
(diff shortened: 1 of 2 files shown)


cmslewis:
Use case: a configurable button-text constant that revealed this bug when we set it to `""`.

PR #1760, opened 2017-10-27
[Table/EditableCell] Type in a cell to edit it now

gscshoyru:

Typing any character now replaces the cell content with that character and puts it in edit mode, essentially letting you arrow around and type in the correct cell.

### DIFF
--- a/packages/table/src/cell/cell.tsx
+++ b/packages/table/src/cell/cell.tsx
@@ -79,6 +79,11 @@ export interface ICellProps extends IIntentProps, IProps {
     onKeyUp?: React.KeyboardEventHandler<HTMLElement>;
 
     /**
+     * Callback invoked when a character-key is pressed
+     */
+    onKeyPress?: React.KeyboardEventHandler<HTMLElement>;
+
+    /**
      * A ref handle to capture the outer div of this cell. Used internally.
      */
     cellRef?: (ref: HTMLElement) => void;
@@ -108,6 +113,7 @@ export class Cell extends React.Component<ICellProps, {}> {
             tabIndex,
             onKeyDown,
             onKeyUp,
+            onKeyPress,
             style,
             intent,
             interactive,
@@ -163,7 +169,12 @@ export class Cell extends React.Component<ICellProps, {}> {
         const content = <div className={textClasses}>{modifiedChildren}</div>;
 
         return (
-            <div className={classes} title={tooltip} ref={cellRef} {...{ style, tabIndex, onKeyDown, onKeyUp }}>
+            <div
+                className={classes}
+                title={tooltip}
+                ref={cellRef}
+                {...{ style, tabIndex, onKeyDown, onKeyUp, onKeyPress }}
+            >
                 <LoadableContent loading={loading} variableLength={true}>
                     {content}
                 </LoadableContent>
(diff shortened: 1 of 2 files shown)


packages/table/src/cell/editableCell.tsx
```diff
@@ -168,6 +174,14 @@ export class EditableCell extends React.Component<IEditableCellProps, IEditableC
         }
     }
 
+    private handleKeyPress = () => {
+        if (this.state.isEditing) {
+            return;
+        }
+        // setting dirty value to empty string because apparently the text field will pick up the key and write it in there
```
cmslewis: This is 123 characters long. Curious why `prettier` didn't catch this. Can we line-wrap comments to 80 for readability?

Also, can you elaborate a little more on this case?

llorca:
Working well. A couple of details that I spotted:
- behaves weirdly when the focus cell isn't enabled: you can type in and edit the cell, confirm, but then cannot type in again in the same cell.
- doesn't work quite well when typing and going to the next cell super fast.

Also, for some reason, I'm getting much much better success rates when entering edit mode with double click... 🎉

packages/table/src/cell/editableCell.tsx
gscshoyru: Not sure why it wasn't caught. Basically, apparently, you don't need to insert the character yourself. When switched over to an empty text input (or non-empty one, technically), the bubbling up to the top of the browser will apparently get it to write a character. This is... incredibly weird, but, clearly what is happening, and I'm much happier to let chrome type the character than hope that we got the right one out of the keyboard event.

gscshoyru:
@llorca doing this without focus cell enabled is... unsupported. Unless you think it should be? It is certainly somewhat strange (I would agree) that you can type in it at all without double click with focus cell off, and I can try to figure out how to tweak it to turn this off (not... entierly sure how) -- but regardless I don't think it should be a supported use-case. 

Can you clarify what you mean by "doesn't work quite well when typing and going to the next cell super fast."?

llorca:
@gscshoyru yeah I agree, shouldn't be supported, but editable cells without focus cells aren't explicitly disabled right now, which is why it's problematic.

What I mean by "doesn't work quite well when typing and going to the next cell super fast." is: typing in stuff in a cell, then quickly press "enter" then quickly type in again. You'll see that the focus cell unsyncs from the cell being edited because of the fast interactions. If that doesn't make sense, I'll make a GIF tomorrow showcasing the issue. In any case, it'd be neat to fix but I suppose it's lower priority

gscshoyru:
Hmmm... so there's an issue here with the fact that how I'm controlling/have to control focus doesn't work very well with batched rendering. @cmslewis is there a good way to override batching behavior so that it always renders a specific given cell (the focused one) _first_ before anything else? That way focus behavior will behave appropriately, and not have to catch up to itself.

cmslewis:
That's not currently possible, no.

gscshoyru:
Hm, if that's not possible then... I'm not sure how to make this work properly/quickly, with batching. Possibly we need to get cells to keep from constantly re-rendering when they don't need to -- maybe look into the logic some more. 

Also, @llorca I've disabled the ability to type-to-edit cells if focus is disabled, which is how it should be :)

packages/table/src/cell/cell.tsx
```diff
@@ -79,6 +79,11 @@ export interface ICellProps extends IIntentProps, IProps {
     onKeyUp?: React.KeyboardEventHandler<HTMLElement>;
 
     /**
+     * Callback invoked when a character-key is pressed
```
cmslewis: Need trailing period for consistency.

Issue #877, opened 2017-03-22
[DateRangePicker] [DateRangeInput] Allow min/max bounds to be removed


michael-yx-wu:
Is it possible to remove min / max dates on date-related inputs? There are times when _any_ date is valid and that's just currently not possible.

- __Package version(s)__: Datetime 1.9.1

cmslewis:
Reasonable. As a workaround, can you provide your own distant `minDate` and `maxDate` prop values to fit your needs?

michael-yx-wu:
Yup this is not blocking for me!

adidahiya:
This is pretty low priority, we'd accept a PR if anyone really needs this feature

tnrich:
+1 for this feature :)

michael-yx-wu:
I played around with this and it doesn't seem like there's a good way to update `<select>` options while the select is open. Would we accept a pre-commit that replaces the `<select>` with `BPLabs#select`?

michael-yx-wu:
Would an acceptable solution be to set a large max / min range i.e. +/- 100 years of today and increase range when users select dates?

tnrich:
Just chiming in here that if you naively extend the Date picker to have a large/small min/max the whole picker gets slowed down quite a bit:

https://codesandbox.io/s/rrrkq5jj0q

Seems like a more intuitive solution could be thought up for the no-bounds case.

tnrich:
For example: http://www.daterangepicker.com/#examples

adidahiya:
Closing for housekeeping, feel free to comment if this is still important to you

adidahiya:
This feature has been requested again, it would be nice to have the component handle unbounded dates in a performant & convenient way... without the noticeable lag that a workaround like this produces (this gif is compressed, so frames are dropped... the lag is more noticeable in the real thing):

![2020-02-10 17 27 13](https://user-images.githubusercontent.com/723999/74196151-a3e12480-4c2a-11ea-989a-96876c0f109d.gif)

PR #1762, opened 2017-10-29
[Core/Slider] Slider now supports vertical orientation 

### LINKED ISSUE #1645: [Slider] Add support for vertical slider orientation
Hi Blueprint Team,

It would be awesome if you could come out with a vertical slider along with the horizontal one. Use cases involve volume level, high-to-low index, etc...

Thanks

cmslewis:
#### Checklist

- [x] Include tests

- 🎉 __NEW__ `Slider` now supports vertical orientation!
    - Reworked existing SCSS mixins to contain only orientation-specific code (before, they were decomposed despite being used in only one place.)
    - Reworked SCSS variables but kept aliases to the old ones (`$handle-height`, `$track-height`).
    - Min/default height is `150px`, the same as the current `min-width` for sliders.

- We don't export mixins, right? So it's okay that I changed them?
- This work has uncovered a subtle bug with pixel => value logic: https://github.com/palantir/blueprint/issues/1769.

#### Screenshot

____SLIDER____

____RANGE SLIDER____

### DIFF
--- a/packages/core/examples/rangeSliderExample.tsx
+++ b/packages/core/examples/rangeSliderExample.tsx
@@ -40,7 +40,7 @@ export class RangeSliderExample extends BaseExample<IRangeSliderExampleState> {
 
     protected renderOptions() {
         return [
-            [<Switch checked={this.state.vertical} label={"Vertical"} key="vertical" onChange={this.toggleVertical} />],
+            [<Switch checked={this.state.vertical} label="Vertical" key="vertical" onChange={this.toggleVertical} />],
         ];
     }
 
--- a/packages/core/examples/rangeSliderExample.tsx
+++ b/packages/core/examples/rangeSliderExample.tsx
@@ -6,17 +6,21 @@
 
 import * as React from "react";
 
-import { NumberRange, RangeSlider } from "@blueprintjs/core";
-import { BaseExample } from "@blueprintjs/docs";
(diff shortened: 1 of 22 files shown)


cmslewis:
FYI this is delayed due to some tricky logic that behaves incorrectly only in the bottom-up coordinate system used in the vertical orientation.

cmslewis:
Removing reviewers for now.

cmslewis:
@tobilen FYI your `DateRangeInput` [tests](https://github.com/palantir/blueprint/pull/1647/files#diff-323ba28e2783ea39a76dda6c3eb4dc10R175
) from #1647 are failing as of this month, because the default `maxDate` is set to `Dec 31` of the current year, and the lack of a `defaultValue` means there's no "next" button to click.

image

Related to #877.

cmslewis:
@llorca done

image

packages/core/examples/rangeSliderExample.tsx
```diff
@@ -28,10 +32,17 @@ export class RangeSliderExample extends BaseExample<IRangeSliderExampleState> {
...
                     value={this.state.range}
+                    vertical={this.state.vertical}
                 />
             </div>
         );
     }
 
+    protected renderOptions() {
+        return [
+            [<Switch checked={this.state.vertical} label={"Vertical"} key="vertical" onChange={this.toggleVertical} />],
```
giladgray: `label="Vertical"`

PR #1775, opened 2017-11-02
[DateTime] Fix datetime tests by applying a defaultValue

cmslewis:
#### Checklist

- [x] Include tests

From the linked comment:

> **@[redacted]** FYI your `DateRangeInput` [tests](https://github.com/palantir/blueprint/pull/1647/files#diff-323ba28e2783ea39a76dda6c3eb4dc10R175
) from #1647 are failing as of this month, because the default `maxDate` is set to `Dec 31` of the current year, and the lack of a `defaultValue` means there's no "next" button to click.

image

Related to #877.

I've encountered this kind of thing frequently when working with date/time tests. There might be more tests that will break toward the end of the year; we'll fix those as they arise.


cmslewis:
FYI This is blocking https://github.com/palantir/blueprint/pull/1762.

PR #1793, opened 2017-11-06
[Table] Better editable names

### LINKED ISSUE #1262: [Table] Resolve differences between <EditableCell> and <EditableName>

gscshoyru:

Clone the header name so it gets the index as props when created, so it can use it to pass it in the callbacks. Also improve the header so that it actually is controlled mode, more or less.

### DIFF
--- a/packages/table/src/headers/columnHeaderCell.tsx
+++ b/packages/table/src/headers/columnHeaderCell.tsx
@@ -154,7 +154,9 @@ export class ColumnHeaderCell extends AbstractComponent<IColumnHeaderCellProps, 
 
         const nameComponent = (
             <LoadableContent loading={loading} variableLength={true}>
-                {renderName == null ? defaultName : renderName(name, index)}
+                {renderName == null
+                    ? defaultName
+                    : React.cloneElement(renderName(name, index) as JSX.Element, { index })}
             </LoadableContent>
         );
 
--- a/packages/table/src/headers/editableName.tsx
+++ b/packages/table/src/headers/editableName.tsx
@@ -4,7 +4,7 @@
  * Licensed under the terms of the LICENSE file distributed with this project.
  */
 
-import { EditableText, IIntentProps, IProps } from "@blueprintjs/core";
+import { EditableText, IIntentProps, IProps, Utils as CoreUtils } from "@blueprintjs/core";
 import * as classNames from "classnames";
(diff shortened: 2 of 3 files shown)


packages/table/src/headers/columnHeaderCell.tsx
```diff
@@ -154,7 +154,9 @@ export class ColumnHeaderCell extends AbstractComponent<IColumnHeaderCellProps,
 
         const nameComponent = (
             <LoadableContent loading={loading} variableLength={true}>
-                {renderName == null ? defaultName : renderName(name, index)}
+                {renderName == null
```
cmslewis: Isn't multiline JSX forbidden by our TSLint config? Regardless, I'd prefer to kick this out into a variable above to keep the JSX lean.

packages/table/src/headers/editableName.tsx
```diff
@@ -35,24 +35,80 @@ export interface IEditableNameProps extends IIntentProps, IProps {
...
+        this.setState({ dirtyName: value });
+        this.invokeCallback(this.props.onChange, value);
+    };
+
+    private handleConfirm = (value: string) => {
+        this.setState({ isEditing: false, savedName: value, dirtyName: undefined });
+        this.invokeCallback(this.props.onConfirm, value);
+    };
+
+    private invokeCallback(callback: (value: string, columnIndex?: number) => void, value: string) {
```
cmslewis: How about `maybeInvokeCallback`, since these aren't required?

packages/table/src/headers/editableName.tsx
```diff
@@ -35,24 +35,80 @@ export interface IEditableNameProps extends IIntentProps, IProps {
...
             />
         );
     }
+
+    private handleEdit = () => {
+        this.setState({ isEditing: true, dirtyName: this.state.savedName });
+    };
+
+    private handleCancel = (value: string) => {
+        // don't strictly need to clear the dirtyName, but it's better hygiene
```
cmslewis: This comment looks familiar. 😛

PR #1796, opened 2017-11-07
[EditableText] Add extra code to make sure we move scroll position to the end

### LINKED ISSUE #1761: [EditableText] Opening an EditableText on Chrome should scroll to the end of the input
When you double click on an editable cell, the cursor is at the end of the contents, but if the contents are long, the end isn't actually visible. The contents should shift so the cursor is visible. 
double

gscshoyru:

Set scrollLeft as well as selection position to guarentee that the scroll is correct.

I tried, repeatedly, to figure out why setSelectionRange doesn't seem to work properly, for us, in chrome. I... had a lot of trouble playing with it to get it to do it programmatically in _any_ browser. Even using existing examples. 

This method of doing it works in chrome, and other browsers, and will make sure we're scrolled to the end when we need to be.

### DIFF
--- a/packages/core/src/components/editable-text/editableText.tsx
+++ b/packages/core/src/components/editable-text/editableText.tsx
@@ -128,6 +128,9 @@ export class EditableText extends AbstractComponent<IEditableTextProps, IEditabl
                 input.focus();
                 const { length } = input.value;
                 input.setSelectionRange(this.props.selectAllOnFocus ? 0 : length, length);
+                if (!this.props.selectAllOnFocus) {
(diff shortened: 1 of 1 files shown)


cmslewis:
Can you post a GIF or add a test?

gscshoyru:
GIF:

nov-07-2017 16-13-19

I'm not sure how one would go about testing this, exactly -- other than checking that `scrollLeft` equals `scrollWidth` which would be pretty much just testing that javascript works.

tgreenwatts:
Looks like it works for me! Chrome & FF

PR #1825, opened 2017-11-20
[DateInput/DateRangeInput] Commit typed value on 'Enter' key press

### LINKED ISSUE #1798: [DateTime] Fields don't "commit" the inputed date on 'Enter'
This is a feature request to support the use of "enter" when selecting dates from the Blueprint Date Picker.

cmslewis:
#### Checklist

- [x] Include tests

- `DateInput` and `DateRangeInput` now save the typed value(s) on <kbd>Enter</kbd>.
    - They also move/transfer the focus state appropriately.

- The tests don't corroborate my understanding of what should be happening. Maybe I just need to sleep on it. I've commented or omitted some checks in the meantime.
- Idea for followup: Revert to the old value on <kbd>Esc</kbd>? Thoughts?
- Does controlled mode look good?

#### Screenshot

_`DateInput`_

_`DateRangeInput`_

### DIFF
--- a/packages/datetime/src/dateInput.tsx
+++ b/packages/datetime/src/dateInput.tsx
@@ -16,6 +16,7 @@ import {
     InputGroup,
     IPopoverProps,
     IProps,
+    Keys,
     Popover,
     Position,
     Utils,
@@ -248,6 +249,7 @@ export class DateInput extends AbstractComponent<IDateInputProps, IDateInputStat
                     onChange={this.handleInputChange}
                     onClick={this.handleInputClick}
                     onFocus={this.handleInputFocus}
+                    onKeyDown={this.handleInputKeyDown}
                     value={dateString}
                 />
             </Popover>
@@ -390,7 +392,7 @@ export class DateInput extends AbstractComponent<IDateInputProps, IDateInputStat
     };
 
     private handleInputBlur = (e: React.FocusEvent<HTMLInputElement>) => {
-        const valueString = this.state.valueString;
+        const { valueString } = this.state;
         const value = this.createMoment(valueString);
         if (
             valueString.length > 0 &&
@@ -418,6 +420,13 @@ export class DateInput extends AbstractComponent<IDateInputProps, IDateInputStat
             }
         }
         this.safeInvokeInputProp("onBlur", e);
+    };
+
+    private handleInputKeyDown = (e: React.KeyboardEvent<HTMLInputElement>) => {
+        if (e.which === Keys.ENTER) {
+            this.handleDateChange(new Date(this.state.valueString), true);
+            this.inputRef.blur();
+        }
     };
 
     private setInputRef = (el: HTMLElement) => {
--- a/packages/datetime/src/dateInput.tsx
+++ b/packages/datetime/src/dateInput.tsx
@@ -437,6 +437,11 @@ export class DateInput extends AbstractComponent<IDateInputProps, IDateInputStat
             const nextValue = this.createMoment(this.state.valueString);
             const nextDate = fromMomentToDate(nextValue);
             this.handleDateChange(nextDate, true, true);
+        } else if (e.which === Keys.TAB && e.shiftKey) {
+            // close the popover if focus will move to the previous element on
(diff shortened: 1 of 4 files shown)


llorca:
does this PR change any interaction?

+1 on reverting to previous value on `Esc`, if current value hasn't been submitted

cmslewis:
@llorca no, it should be strictly additive.

packages/datetime/src/dateInput.tsx
```diff
@@ -423,6 +425,15 @@ export class DateInput extends AbstractComponent<IDateInputProps, IDateInputStat
         this.safeInvokeInputProp("onBlur", e);
     };
 
+    private handleInputKeyDown = (e: React.KeyboardEvent<HTMLInputElement>) => {
+        if (e.which === Keys.ENTER) {
+            const nextValue = this.createMoment(this.state.valueString);
+            const nextDate = fromMomentToDate(nextValue);
+            this.handleDateChange(nextDate, true);
+            this.inputRef.blur();
+        }
```
cmslewis: Whoops, need to do `this.safeInvokeInputProp("onKeyDown", e);`.

packages/datetime/src/dateRangeInput.tsx
```diff
@@ -542,9 +542,12 @@ export class DateRangeInput extends AbstractComponent<IDateRangeInputProps, IDat
     // - if focused in start field, Tab moves focus to end field
     // - if focused in end field, Shift+Tab moves focus to start field
     private handleInputKeyDown = (e: React.KeyboardEvent<HTMLInputElement>) => {
```
cmslewis: Here too?

packages/datetime/src/dateInput.tsx
```diff
@@ -423,6 +425,15 @@ export class DateInput extends AbstractComponent<IDateInputProps, IDateInputStat
         this.safeInvokeInputProp("onBlur", e);
     };
 
+    private handleInputKeyDown = (e: React.KeyboardEvent<HTMLInputElement>) => {
+        if (e.which === Keys.ENTER) {
+            const nextValue = this.createMoment(this.state.valueString);
+            const nextDate = fromMomentToDate(nextValue);
+            this.handleDateChange(nextDate, true);
+            this.inputRef.blur();
```
giladgray: 🤔 this `.blur()` feels intuitively wrong, as it kills the focus flow: I can't hit `enter` and then `tab` to move to the next form input.

can we do everything else but simply not blur? same goes for DRI.

cmslewis: I buy that.

cmslewis:
@giladgray is this good to merge?

packages/datetime/src/dateInput.tsx
giladgray: @cmslewis is this comment still relevant?

giladgray:
@cmslewis pressing `enter` in `DateInput` does not expand to full date. works great for `DRI` though.

adidahiya:
> pressing enter in DateInput does not expand to full date

seems fine to address in a follow up. current behavior seems fine to me. @erinmarchand @LindsayWard @dalessi can you verify in the "documentation" link above?

giladgray:
wait @adidahiya how is this fine to merge? it literally doesn't work properly in one of the title use cases.

adidahiya:
It seems to address the original intent of the filed issue though. Once you type something in DateInput, you get it committed when you blur out of the input (the popover gets closed, which suggests to the user that something worked). Seems reasonable in most cases, and we can make further adjustments if we get more feedback. But I'm also cool with waiting to hear from the people tagged in my last comment ^

llorca:
Check out this GIF:
What's good: in both DI and DRI, the date picker activates with the date typed in the field
What's not good: DRI updates the input value on `enter`, DI doesn't. Instead, DI updates the input value on blur.

Not a fan of that inconsistency, as @giladgray pointed out.

cmslewis:
@llorca @adidahiya @giladgray Sorry for the delay. Should be good to go now.

adidahiya:
nice, lgtm 👍

PR #1827, opened 2017-11-20
[Table] Prevent DOM reflow by optimizing popover dark-ancestor checks

zerovox:
Prevent DOM reflow on popover mount by only checking for dark ancestors when popover is open.

Seeing some really bad perf in tables where every cell / header has a popover (right click menus, column menus), and each componentDidMount is causing a reflow when the popover does the element.closest("pt-dark") check. We don't need to check this unless the popover is actually open (or about to become open) as the class is only applied to the non-inline portal anyway, which is only visible when the popover is open.

#### Checklist

- [ ] Include tests
- [ ] Update documentation

### DIFF
--- a/packages/core/src/components/popover/popover.tsx
+++ b/packages/core/src/components/popover/popover.tsx
@@ -672,7 +672,7 @@ export class Popover extends AbstractComponent<IPopoverProps, IPopoverState> {
     }
 
     private setOpenStateDirectly(isOpen: boolean | ((previousIsOpen: boolean) => boolean)) {
-        if (typeof isOpen === "function") {
+        if (Utils.isFunction(isOpen)) {
             this.setState(previousState => {
                 const nextIsOpen = isOpen(previousState.isOpen);
                 if (nextIsOpen) {
--- a/packages/core/src/components/popover/popover.tsx
+++ b/packages/core/src/components/popover/popover.tsx
@@ -360,11 +360,11 @@ export class Popover extends AbstractComponent<IPopoverProps, IPopoverState> {
 
         if (nextProps.isOpen == null && nextProps.isDisabled && !this.props.isDisabled) {
             // ok to use setOpenState here because isDisabled and isOpen are mutex.
-            this.setOpenState(false);
+            this.setOpenStateByPropCallback(false);
         } else if (nextProps.isOpen !== this.props.isOpen) {
             // propagate isOpen prop directly to state, circumventing onInteraction callback
             // (which would be invoked if this went through setOpenState)
-            this.setState({ isOpen: nextProps.isOpen });
+            this.setOpenStateDirectly(nextProps.isOpen);
         }
     }
 
@@ -432,8 +432,10 @@ export class Popover extends AbstractComponent<IPopoverProps, IPopoverState> {
             });
         }
         if (!this.props.inline) {
-            this.hasDarkParent = isDarkTheme(this.targetElement);
             this.updateTether();
+        }
+        if (!this.props.inline && this.state.isOpen) {
+            this.checkDOMForDarkTheme();
         }
     }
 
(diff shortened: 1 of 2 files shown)


packages/core/src/components/popover/popover.tsx
```diff
@@ -669,6 +671,23 @@ export class Popover extends AbstractComponent<IPopoverProps, IPopoverState> {
         }
     }
 
+    private setOpenStateDirectly(isOpen: boolean | ((previousIsOpen: boolean) => boolean)) {
+        if (typeof isOpen === "function") {
```
cmslewis: nit: Use `Utils.isFunction`

packages/core/src/components/popover/popover.tsx
```diff
@@ -669,6 +671,23 @@ export class Popover extends AbstractComponent<IPopoverProps, IPopoverState> {
...
+            this.setState(previousState => {
```
cmslewis: Do we need this callback approach? We don't really use it elsewhere in the codebase these days (for better or for worse). More concisely:

```tsx
private setOpenStateDirectly(isOpen: boolean | ((previousIsOpen: boolean) => boolean)) {
    const nextIsOpen = Utils.isFunction(isOpen) ? isOpen(this.state.isOpen) : isOpen;
    if (nextIsOpen) {
        this.checkDOMForDarkTheme();
    }
    this.setState({ isOpen: nextIsOpen });
});
```

EDIT: I wrote this before my comment on L590 above. If we can in fact do away with the function param support, that'd be preferable.

packages/core/src/components/popover/popover.tsx
```diff
@@ -555,39 +557,39 @@ export class Popover extends AbstractComponent<IPopoverProps, IPopoverState> {
...
+            this.setOpenStateByPropCallback(false, e);
         }
     };
 
     private handleTargetClick = (e: React.MouseEvent<HTMLElement>) => {
         // ensure click did not originate from within inline popover before closing
         if (!this.props.isDisabled && !this.isElementInPopover(e.target as HTMLElement)) {
             if (this.props.isOpen == null) {
-                this.setState(prevState => ({ isOpen: !prevState.isOpen }));
+                this.setOpenStateDirectly(previousIsOpen => !previousIsOpen);
```
cmslewis: Can we do:

```tsx
this.setOpenStateDirectly(!this.state.previousIsOpen);
```

and then get rid of the helper's function-type argument altogether?

packages/core/src/components/popover/popover.tsx
zerovox: Presumable the reason the original code has the `this.setState(state => ...` format is to avoid reading from `this.state` which can be potentially stale if there are multiple `setStates` prior to a `willUpdate` being called. (https://stackoverflow.com/questions/42038590/when-to-use-react-setstate-callback)

packages/core/src/components/popover/popover.tsx
zerovox: fixed

packages/core/src/components/popover/popover.tsx
cmslewis: That post is arguing for a callback as the _second_ parameter to `setState`. Above, we're providing a callback as the first parameter. Since we're only referring to the old `isOpen` value in the following lines, it's not clear to me that we need a callback at all.

zerovox: Sorry, I was mostly referencing the opening point "Yes there is, since setState works in an asynchronous way. That means after calling setState the this.state variable is not immediately changed". If some code path has recently called setState and changed the value of isOpen, before calling into this function we could potentially miss the update by doing a stale read. Happy to change it if that seems off.

cmslewis: My understanding is that `state` changes may be batched until the next render cycle, but they won't be applied piecemeal before then (that would seem to violate React's fundamental lifecycle guarantees). I don't see any `setState` calls interleaved here, so I think we're fine not to use the callback.

@giladgray what say you?

tgreenwatts:
Do we have any numbers on how much this helps perf?

zerovox:
This change doesn't really affect the perf of the popover component itself,
it makes it more performant when mounting a bunch of them at once. I can
pull perf numbers of this change when it makes it to a release and I can do
the side-by-side in contour.

packages/core/src/components/popover/popover.tsx
giladgray: @cmslewis you're correct that `state` changes may be batched, which is precisely why the callback is necessary: by the time your `setState` is executed, the actual `state` value may have changed. the callback syntax allows you to respond to the exact latest value, **which is actually critical in this component** to ensure `isOpen` consistency at all times.

i believe i used the callback intentionally here to handle `isOpen` flickering, like if you click the button a whole bunch of times before one transition finishes.

cmslewis: Sweet. 👍

Issue #1813, opened 2017-11-15
Blueprint types are overly strict


rherms:
Some blueprint types that require strings actually work with JSX.Element as well.
This mostly include the text for certain components such as "cancelButtonText" in Alert or "text" in Button. For example, I can pass in a <span> to these and cast it to "any" and it will render correctly. It would be nice to not have to cast these to "any".

llorca:
`Alert` is explicitly opinionated and restrictive. If you need more flexibility, use `Dialog`.

We've never had issues with `text` prop in `Button` being a string as we allow other button aspects to be customized through the other props.

Gonna close this out, though feel free to reopen if you have a real use case to share regarding button text.

rherms:
@llorca The components all work fine when passing in elements such as <span>s into them as props. However, the types for typescript do not allow this. Why is that? Instead of making these props strings, shouldn't they be something like `string | JSX.Element`?

cmslewis:
@rherms types are a means of restricting "bad" usage that doesn't conform to the design system. 

e.g. in a `MenuItem`, `text` must currently be a `string`, because allowing `span`s (`JSX.Element`s) would give people a way to bold, italicize, and underline text in menu items, which we didn't want to support for lack of a clear use case when we designed the component.

As @llorca mentioned, `Alert`s and `Button`s are similarly restrictive - the typings confine you to good practice based on our understanding of the use case when we designed the component.

If we see clear use cases that argue for broader type support, we can discuss changing the types. For your use case, what if you just unpack the text from your `span`s and add any relevant `className`s to the parent components? Would that be sufficient for you?

rherms:
Ah, that makes sense. I think we should just go back and look again at how we do internationalization and make it easier to get a string instead of a span back for the text.

PR #1840, opened 2017-11-22
[Core] Add JS APIs for ControlGroup, FileUpload, Label, TextArea

cmslewis:
#### Checklist

- [x] Include tests
- [x] Update documentation

- Add JS APIs for `ControlGroup`, `FileUpload`, `Label`, `TextArea`.
- Add example for `ControlGroup`.
- Add tests for `FileUpload`.

- There are some subtleties to the `FileUpload` component, so I added tests.
- I really think we should have at least a simple test for every component. Exact quotes from my experience trying to use the `FileUpload` component in a test:
    - "Damn, I accidentally specified a prop as required instead of optional"
    - "Damn, I accidentally forgot to export the component from `index.ts`"

I'd so much rather rely on a test for those things than on fallible human cognition in a code review.

### DIFF
--- a/packages/core/src/components/forms/controlGroup.tsx
+++ b/packages/core/src/components/forms/controlGroup.tsx
@@ -1,0 +1,49 @@
+/**
+ * Copyright 2017 Palantir Technologies, Inc. All rights reserved.
+ *
+ * Licensed under the terms of the LICENSE file distributed with this project.
+ */
+
+import * as classNames from "classnames";
+import * as PureRender from "pure-render-decorator";
+import * as React from "react";
+import * as Classes from "../../common/classes";
+import { IProps } from "../../common/props";
+
+export interface IControlGroupProps extends React.HTMLProps<HTMLDivElement>, IProps {
+    /**
+     * Whether the control group should take up the full width of its container.
+     */
+    fill?: boolean;
+
+    /**
+     * Whether the button group should appear with vertical styling.
+     */
+    vertical: boolean;
+}
+
+// this component is simple enough that tests would be purely tautological.
+/* istanbul ignore next */
+@PureRender
+export class ControlGroup extends React.Component<IControlGroupProps, {}> {
+    public static displayName = "Blueprint.ControlGroup";
+
+    public render() {
+        const { children, className, fill, vertical, ...htmlProps } = this.props;
+
+        const rootClasses = classNames(
+            Classes.CONTROL_GROUP,
+            {
+                [Classes.FILL]: fill,
+                [Classes.VERTICAL]: vertical,
+            },
+            className,
+        );
+
+        return (
+            <div {...htmlProps} className={rootClasses}>
+                {children}
+            </div>
+        );
+    }
+}
--- a/packages/core/src/components/forms/controlGroup.tsx
+++ b/packages/core/src/components/forms/controlGroup.tsx
@@ -47,3 +47,5 @@ export class ControlGroup extends React.Component<IControlGroupProps, {}> {
         );
     }
 }
+
+export const ControlGroupFactory = React.createFactory(ControlGroup);
--- a/packages/core/src/components/forms/controlGroup.tsx
+++ b/packages/core/src/components/forms/controlGroup.tsx
@@ -19,7 +19,7 @@ export interface IControlGroupProps extends React.HTMLProps<HTMLDivElement>, IPr
     /**
      * Whether the button group should appear with vertical styling.
      */
-    vertical: boolean;
+    vertical?: boolean;
 }
 
 // this component is simple enough that tests would be purely tautological.
--- a/packages/core/src/components/forms/controlGroup.tsx
+++ b/packages/core/src/components/forms/controlGroup.tsx
@@ -10,7 +10,7 @@ import * as React from "react";
 import * as Classes from "../../common/classes";
 import { IProps } from "../../common/props";
 
-export interface IControlGroupProps extends React.HTMLProps<HTMLDivElement>, IProps {
+export interface IControlGroupProps extends React.AllHTMLAttributes<HTMLDivElement>, IProps {
     /**
      * Whether the control group should take up the full width of its container.
      */
--- a/packages/core/src/components/forms/file-upload.md
+++ b/packages/core/src/components/forms/file-upload.md
@@ -9,4 +9,18 @@ Wrap that all in a `label` with class `pt-file-upload`.
     you must implement it separately in JS.
 </div>
 
+@## CSS API
+
 @css pt-file-upload
+
+@## JavaScript API
+
+The `FileUpload` component is available in the @blueprintjs/core package. Make sure to review the general usage docs for JS components.
+
+This components is a simple wrapper around the corresponding CSS API. It supports the full range of HTML props.
+
+```tsx
+<FileUpload disabled={true} text="Choose file..." inputProps={{ onChange: /* ... */ }} />
+```
+
+@interface IFileUploadProps
--- a/packages/core/src/components/forms/file-upload.md
+++ b/packages/core/src/components/forms/file-upload.md
@@ -20,7 +20,7 @@ The `FileUpload` component is available in the __@blueprintjs/core__ package. Ma
 This component is a simple wrapper around the corresponding CSS API. It supports the full range of HTML props.
 
 ```tsx
-<FileUpload disabled={true} text="Choose file..." inputProps={{ onChange: /* ... */ }} />
+<FileUpload disabled={true} text="Choose file..." onInputChange={...} />
 ```
 
 @interface IFileUploadProps
--- a/packages/core/src/components/forms/file-upload.md
+++ b/packages/core/src/components/forms/file-upload.md
@@ -15,7 +15,7 @@ Wrap that all in a `label` with class `pt-file-upload`.
 
 @## JavaScript API
 
-The `FileUpload` component is available in the @blueprintjs/core package. Make sure to review the general usage docs for JS components.
+The `FileUpload` component is available in the __@blueprintjs/core__ package. Make sure to review the [general usage docs for JS components](#blueprint.usage).
 
 This component is a simple wrapper around the corresponding CSS API. It supports the full range of HTML props.
 
--- a/packages/core/src/components/forms/file-upload.md
+++ b/packages/core/src/components/forms/file-upload.md
@@ -17,7 +17,7 @@ Wrap that all in a `label` with class `pt-file-upload`.
 
 The `FileUpload` component is available in the @blueprintjs/core package. Make sure to review the general usage docs for JS components.
 
-This components is a simple wrapper around the corresponding CSS API. It supports the full range of HTML props.
+This component is a simple wrapper around the corresponding CSS API. It supports the full range of HTML props.
 
 ```tsx
 <FileUpload disabled={true} text="Choose file..." inputProps={{ onChange: /* ... */ }} />
--- a/packages/core/src/components/forms/fileUpload.tsx
+++ b/packages/core/src/components/forms/fileUpload.tsx
@@ -11,7 +11,7 @@ import { Utils } from "../../common";
 import * as Classes from "../../common/classes";
 import { IProps } from "../../common/props";
 
-export interface IFileUploadProps extends React.HTMLProps<HTMLLabelElement>, IProps {
+export interface IFileUploadProps extends React.AllHTMLAttributes<HTMLLabelElement>, IProps {
     /**
      * Whether the file upload is non-interactive.
      * Setting this to `true` will automatically disable the child input too.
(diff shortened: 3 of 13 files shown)


packages/core/src/components/forms/label.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+export interface ILabelProps extends React.HTMLProps<HTMLDivElement>, IProps {
+    /**
+     * Whether the label is non-interactive.
+     * Be sure to explicitly disable any child controls as well.
+     */
+    disabled?: boolean;
+
+    /** The helper text to show next to the label. */
+    helperText?: JSX.Element | string;
```
cmslewis: Supports JSX.Element too, in case you want to **bold** or _italicize_ things.

packages/core/src/components/forms/file-upload.md
```diff
@@ -9,4 +9,18 @@ Wrap that all in a `label` with class `pt-file-upload`.
     you must implement it separately in JS.
 </div>
 
+@## CSS API
+
 @css pt-file-upload
+
+@## JavaScript API
+
+The `FileUpload` component is available in the __@blueprintjs/core__ package. Make sure to review the [general usage docs for JS components](#blueprint.usage).
```
giladgray: I might hold off on this component until 2.0 as we want to rename to `FileInput` #1453. would be weird to ship and deprecate immediately.

packages/core/src/components/forms/label.tsx
giladgray: use `React.ReactNode` to support anything renderable.

packages/core/src/components/forms/label.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+    /** The text to show in the label. */
+    text: JSX.Element | string;
```
giladgray: `React.ReactNode`

packages/core/src/components/forms/label.tsx
```diff
@@ -0,0 +1,52 @@
...
+            {
+                [Classes.DISABLED]: disabled,
+            },
+            className,
+        );
+
+        return (
+            <div {...htmlProps} className={rootClasses}>
+                {text}
+                <span className={classNames(Classes.TEXT_MUTED)}>{helperText}</span>
```
giladgray: I thought `helperText` came after `children`?
also no need for `classNames` with only one class

packages/core/test/forms/fileUploadTests.tsx
```diff
@@ -0,0 +1,79 @@
...
+        assert.strictEqual(span.text(), "Choose file...");
+
+        // custom text
+        wrapper.setProps({ text: "Upload file..." });
+        assert.strictEqual(span.text(), "Upload file...");
+    });
+
+    it("onChange works", () => {
+        const onChange = sinon.spy();
+        const wrapper = shallow(<FileUpload inputProps={{ onChange }} />);
```
giladgray: `onChange` should also be supported at the root level, rather than requiring `inputProps` object, since it'll be set almost every time. can you please add that?

packages/docs-app/src/examples/core-examples/controlGroupExample.tsx
```diff
@@ -0,0 +1,48 @@
+/*
+ * Copyright 2016 Palantir Technologies, Inc. All rights reserved.
```
giladgray: 2017. please audit other years too.

packages/core/src/components/forms/label.tsx
giladgray: oh wait I see, from the examples. hmm alright.

packages/core/src/components/forms/label.tsx
cmslewis: `React.ReactNode` also support arrays of elements though (and `ReactPortal` and other weird stuff). I'd rather explicitly spell out only the stuff we care about.

packages/core/src/components/forms/file-upload.md
cmslewis: Yeah thought about that. I could also just call it `FileInput` now. That okay?

packages/core/src/components/forms/label.tsx
cmslewis: Yeah for `pt-label`, helper text is inline with the label.

image

packages/core/src/components/forms/label.tsx
giladgray: but technically those values are supported! this is the crux of #1813

packages/core/src/components/forms/file-upload.md
giladgray: it's inconsistent and 2.0 is right around the corner... not sure what to do.

packages/core/src/components/forms/file-upload.md
```diff
@@ -9,4 +9,18 @@ Wrap that all in a `label` with class `pt-file-upload`.
...
+
+This component is a simple wrapper around the corresponding CSS API. It supports the full range of HTML props.
+
+```tsx
+<FileUpload disabled={true} text="Choose file..." inputProps={{ onChange: /* ... */ }} />
```
giladgray: `onInputChange={...}`

packages/core/src/components/forms/fileUpload.tsx
```diff
@@ -0,0 +1,94 @@
...
+
+    /**
+     * Callback invoked on `input` `change` events.
+     *
+     * This callback is offered as a convenience; it is equivalent to passing
+     * `onChange` to `inputProps`.
+     *
+     * __Note:__ If you pass `onChange` as a top-level prop, it will be passed
+     * to the wrapping `label` rather than the `input`, which may not be what
+     * you expect.
```
giladgray: 👌

packages/core/src/components/forms/controlGroup.tsx
```diff
@@ -0,0 +1,49 @@
...
+ * Licensed under the terms of the LICENSE file distributed with this project.
+ */
+
+import * as classNames from "classnames";
+import * as PureRender from "pure-render-decorator";
+import * as React from "react";
+import * as Classes from "../../common/classes";
+import { IProps } from "../../common/props";
+
+export interface IControlGroupProps extends React.HTMLProps<HTMLDivElement>, IProps {
```
adidahiya: as discussed offline, use `HTMLAttributes`?

packages/core/src/components/forms/file-upload.md
adidahiya: Prefer to just ship this as `FileUpload` right now.

packages/core/src/components/forms/fileUpload.tsx
adidahiya: What does `onChange` on `<label>` even mean? Is it even valid? I really think we should consider just piping it through to the `<input>`; there doesn't seem to be a valid use case for `<label onChange={...}>` in practice.

cmslewis: The component's `props` should be spreadable to the outermost node of the component. That node is a `<label>` here, so the props interface looks like this:

```
interface IFileUploadProps extends React.HTMLProps<HTMLLabelElement>, IProps {
```

Since `onChange` uses the custom type `HTMLLabelElement`, it can't be spread to the `input` (due to the type conflict). Maybe I can cast it, but I'd rather not pick and choose which HTML top-level props are spread where.

packages/core/src/components/forms/file-upload.md
cmslewis: K

packages/core/src/components/forms/controlGroup.tsx
cmslewis: @giladgray @adidahiya - `HTMLAttributes`, or `AllHTMLAttributes`?

```tsx
interface HTMLAttributes<T> extends DOMAttributes<T> {
    // React-specific Attributes
    defaultChecked?: boolean;
    defaultValue?: string | string[];
    suppressContentEditableWarning?: boolean;

    // Standard HTML Attributes
    accessKey?: string;
    className?: string;
    contentEditable?: boolean;
    contextMenu?: string;
    dir?: string;
    draggable?: boolean;
    hidden?: boolean;
    id?: string;
    lang?: string;
    slot?: string;
    spellCheck?: boolean;
    style?: CSSProperties;
    tabIndex?: number;
    title?: string;

    // Unknown
    inputMode?: string;
    is?: string;
    radioGroup?: string; // <command>, <menuitem>

    // WAI-ARIA
    role?: string;

    // RDFa Attributes
    about?: string;
    datatype?: string;
    inlist?: any;
    prefix?: string;
    property?: string;
    resource?: string;
    typeof?: string;
    vocab?: string;

    // Non-standard Attributes
    autoCapitalize?: string;
    autoCorrect?: string;
    autoSave?: string;
    color?: string;
    itemProp?: string;
    itemScope?: boolean;
    itemType?: string;
    itemID?: string;
    itemRef?: string;
    results?: number;
    security?: string;
    unselectable?: boolean;
}
```

```tsx
interface AllHTMLAttributes<T> extends HTMLAttributes<T> {
    // Standard HTML Attributes
    accept?: string;
    acceptCharset?: string;
    action?: string;
    allowFullScreen?: boolean;
    allowTransparency?: boolean;
    alt?: string;
    as?: string;
    async?: boolean;
    autoComplete?: string;
    autoFocus?: boolean;
    autoPlay?: boolean;
    capture?: boolean;
    cellPadding?: number | string;
    cellSpacing?: number | string;
    charSet?: string;
    challenge?: string;
    checked?: boolean;
    cite?: string;
    classID?: string;
    cols?: number;
    colSpan?: number;
    content?: string;
    controls?: boolean;
    coords?: string;
    crossOrigin?: string;
    data?: string;
    dateTime?: string;
    default?: boolean;
    defer?: boolean;
    disabled?: boolean;
    download?: any;
    encType?: string;
    form?: string;
    formAction?: string;
    formEncType?: string;
    formMethod?: string;
    formNoValidate?: boolean;
    formTarget?: string;
    frameBorder?: number | string;
    headers?: string;
    height?: number | string;
    high?: number;
    href?: string;
    hrefLang?: string;
    htmlFor?: string;
    httpEquiv?: string;
    integrity?: string;
    keyParams?: string;
    keyType?: string;
    kind?: string;
    label?: string;
    list?: string;
    loop?: boolean;
    low?: number;
    manifest?: string;
    marginHeight?: number;
    marginWidth?: number;
    max?: number | string;
    maxLength?: number;
    media?: string;
    mediaGroup?: string;
    method?: string;
    min?: number | string;
    minLength?: number;
    multiple?: boolean;
    muted?: boolean;
    name?: string;
    nonce?: string;
    noValidate?: boolean;
    open?: boolean;
    optimum?: number;
    pattern?: string;
    placeholder?: string;
    playsInline?: boolean;
    poster?: string;
    preload?: string;
    readOnly?: boolean;
    rel?: string;
    required?: boolean;
    reversed?: boolean;
    rows?: number;
    rowSpan?: number;
    sandbox?: string;
    scope?: string;
    scoped?: boolean;
    scrolling?: string;
    seamless?: boolean;
    selected?: boolean;
    shape?: string;
    size?: number;
    sizes?: string;
    span?: number;
    src?: string;
    srcDoc?: string;
    srcLang?: string;
    srcSet?: string;
    start?: number;
    step?: number | string;
    summary?: string;
    target?: string;
    type?: string;
    useMap?: string;
    value?: string | string[] | number;
    width?: number | string;
    wmode?: string;
    wrap?: string;
}
```

cmslewis: Also, if we don't use `HTMLProps`, will you not be able to pass a `ref`?

adidahiya: HTMLAttributes, or AllHTMLAttributes -- which one does HTMLProps use?

cmslewis: `AllHTMLAttributes`

adidahiya: let's go with AllHTMLAttributes then, to be safer

packages/core/src/components/forms/fileUpload.tsx
adidahiya: shrug... I still don't like the API tradeoffs we're making here but we can revisit in 2.0

giladgray: oof typings biting butts again. would be nice and logical for `onChange` to apply to the input, but makes sense to spread all props to `label` element. we have precedent for `onInputChange` in other input-enveloping components.

packages/core/src/components/forms/label.md
```diff
@@ -30,3 +32,21 @@ must add the `:disabled` attribute directly to any nested elements to disable th
...
+
+This component is a simple wrapper around the corresponding CSS API. It supports the full range of HTML props.
+
+```tsx
+<Label
+    helperText="Helper text with details..."
+    text="Label A"
+    required={true}
+>
+    <input id="text-input" placeholder="Placeholder text" />
```
giladgray: `class="pt-input"`

cmslewis: _`className`_ 😛

PR #1853, opened 2017-11-29
[Table] Added option to truncated format for truncated by approx

gscshoyru:

Create a faster, less-accurate "truncate by approx" that is synchronous rather than async and doesn't require measuring the DOM, for extra speeds. 

It's definitely somewhat faster in batch-on-update mode, and _much_ faster (visibly takes much less time) on "none" render mode. Pick "json" for the cell contents to make this more obvious. 

This isn't... quite done yet -- there are a couple open questions. Namely:
- currently the defaults are hardcoded into the class, the same way table is -- we should probably be pulling this out somewhere common, yes? Where should they go?
- additionally -- the defaults for "buffer lines" are slightly different -- having a buffer line if you're not wrapping the text means the cell needs to be twice as long, which is silly. Should we be passing the "wrap text" down from the parent the same we do for other things?

### DIFF
--- a/packages/karma-build-scripts/createKarmaConfig.js
+++ b/packages/karma-build-scripts/createKarmaConfig.js
@@ -6,7 +6,7 @@ const path = require("path");
 const webpack = require("webpack");
 const webpackConfig = require("./webpack.karma.config");
 
-const COVERAGE_PERCENT = 80;
+const COVERAGE_PERCENT = 75;
 const COVERAGE_PERCENT_HIGH = 90;
 const KARMA_SERVER_PORT = 9876;
 
--- a/packages/karma-build-scripts/createKarmaConfig.js
+++ b/packages/karma-build-scripts/createKarmaConfig.js
@@ -6,7 +6,7 @@ const path = require("path");
 const webpack = require("webpack");
 const webpackConfig = require("./webpack.karma.config");
 
-const COVERAGE_PERCENT = 75;
+const COVERAGE_PERCENT = 80;
 const COVERAGE_PERCENT_HIGH = 90;
 const KARMA_SERVER_PORT = 9876;
 
--- a/packages/table/src/cell/formats/truncatedFormat.tsx
+++ b/packages/table/src/cell/formats/truncatedFormat.tsx
@@ -13,7 +13,6 @@ import * as Classes from "../../common/classes";
 import { Utils } from "../../common/utils";
 
 import { Locator } from "../../locator";
-import { IResizeRowsByApproximateHeightOptions } from "../../table";
 
 // amount in pixels that the content div width changes when truncated vs when
 // not truncated. Note: could be modified by styles
@@ -42,12 +41,12 @@ export interface ITruncatedFormatProps extends IProps {
     /**
      * Height of the parent cell. Used by shouldComponentUpdate only
      */
-    parentCellHeight?: string;
+    parentCellHeight?: number;
 
     /**
      * Width of the parent cell. Used by shouldComponentUpdate only
      */
-    parentCellWidth?: string;
+    parentCellWidth?: number;
 
     /**
      * Sets the popover content style to `white-space: pre` if `true` or
@@ -80,6 +79,37 @@ export interface ITruncatedFormatProps extends IProps {
      * @default "..."
      */
     truncationSuffix?: string;
+
+    /**
+     * Approximate character width, used to determine whether to display the popover in approx truncation mode.
+     * The default value should work for normal table styles,
+     * but should be changed as necessary if the fonts or styles are changed significantly.
+     * @default 8
+     */
+    approximateCharWidth?: number;
+
+    /**
+     * Approximate character width, used to determine whether to display the popover in approx truncation mode.
+     * The default value should work for normal table styles, but should be changed if the fonts or styles are changed significantly.
+     * @default 18
+     */
+    approximateLineHeight?: number;
+
+    /**
+     * Total horizonal cell padding (both sides), used to determine whether to display the popover in approx truncation mode.
+     * The default value should work for normal table styles,
+     * but should be changed as necessary if the fonts or styles are changed significantly.
+     * @default 20
+     */
+    cellHorizontalPadding?: number;
+
+    /**
+     * Number of buffer lines desired, used to determine whether to display the popover in approx truncation mode.
+     * The default value should work for normal table styles,
+     * but should be changed as necessary if the fonts or styles are changed significantly.
+     * @default 0
+     */
+    numBufferLines?: number;
 }
 
 export interface ITruncatedFormatState {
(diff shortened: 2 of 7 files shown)


packages/karma-build-scripts/createKarmaConfig.js
```diff
@@ -6,7 +6,7 @@ const path = require("path");
 const webpack = require("webpack");
 const webpackConfig = require("./webpack.karma.config");
 
-const COVERAGE_PERCENT = 80;
+const COVERAGE_PERCENT = 75;
```
cmslewis: :(

packages/table/src/cell/formats/truncatedFormat.tsx
```diff
@@ -92,6 +97,18 @@ export class TruncatedFormat extends React.Component<ITruncatedFormatProps, ITru
...
 
+    // these default values for `resizeRowsByApproximateHeight` have been
+    // fine-tuned to work well with default Table font styles.
+    private static resizeRowsByApproximateHeightDefaults: Record<
+        keyof IResizeRowsByApproximateHeightOptions,
+        number
+    > = {
+        getApproximateCharWidth: 8,
+        getApproximateLineHeight: 18,
+        getCellHorizontalPadding: 2 * Locator.CELL_HORIZONTAL_PADDING,
```
cmslewis: Kinda weird to import Locator just to grab the constants within it. Can we pull these constants out into a separate `tableConstants.ts` file?

packages/table/src/cell/formats/truncatedFormat.tsx
```diff
@@ -202,6 +219,32 @@ export class TruncatedFormat extends React.Component<ITruncatedFormatProps, ITru
...
                     : truncateLength > 0 && content.length > truncateLength;
+            case TruncatedPopoverMode.WHEN_TRUNCATED_APPROX:
+                if (!detectTruncation) {
+                    return truncateLength > 0 && content.length > truncateLength;
+                }
+                const {
+                    getApproximateCharWidth: approxCharWidth,
+                    getApproximateLineHeight: approxLineHeight,
+                    getCellHorizontalPadding: horizontalPadding,
+                    getNumBufferLines: numBufferLines,
```
cmslewis: Do you plan to support functions for these properties too? The names are misleading otherwise. (Granted, if there's no public API, it doesn't matter, but it probably deserves a comment then.)

packages/table/src/cell/formats/truncatedFormat.tsx
```diff
@@ -202,6 +219,32 @@ export class TruncatedFormat extends React.Component<ITruncatedFormatProps, ITru
...
+                } = TruncatedFormat.resizeRowsByApproximateHeightDefaults;
+
+                const cellWidth = parseInt(this.props.parentCellWidth, 10);
```
cmslewis: `parentCellWidth` is an optional prop. Can you add a null check before invoking `parseInt`?

packages/table/src/cell/formats/truncatedFormat.tsx
```diff
@@ -202,6 +219,32 @@ export class TruncatedFormat extends React.Component<ITruncatedFormatProps, ITru
...
+                    content,
+                    cellWidth,
+                    approxCharWidth,
+                    approxLineHeight,
+                    horizontalPadding,
+                    numBufferLines,
+                );
+
+                let shouldTruncate = false;
+                if (approxCellHeight > parseInt(this.props.parentCellHeight, 10)) {
```
cmslewis: Same.

Should probably just return early on L223 if either of these doesn't exist.

packages/table/src/cell/formats/truncatedFormat.tsx
cmslewis: Also, why is this prop a string in the first place?

packages/table/src/cell/formats/truncatedFormat.tsx
```diff
@@ -202,6 +219,32 @@ export class TruncatedFormat extends React.Component<ITruncatedFormatProps, ITru
...
+                    shouldTruncate = true;
+                }
+                return shouldTruncate;
```
cmslewis: I like the variable for clarity, but you can use a ternary:

```tsx
const shouldTruncate = approxCellHeight > parentCellHeightAsNumber;
return shouldTruncate;
```

packages/table/src/cell/formats/truncatedFormat.tsx
```diff
@@ -218,7 +261,6 @@ export class TruncatedFormat extends React.Component<ITruncatedFormatProps, ITru
         }
 
         const { isTruncated } = this.state;
-
```
cmslewis: Is this deletion needed? I tend to like whitespace.

packages/table/src/cell/formats/truncatedFormat.tsx
gscshoyru: ...I mean, that in and of itself seems somewhat strange to me. Should we put other constants in there too? Which?

packages/table/src/cell/formats/truncatedFormat.tsx
gscshoyru: This is what I get for copy/pasting :P 

We should probably support a public API for this, but given that this is on the cell level they probably don't need to be functions. I'll change the names and make them public, I think.

packages/table/src/cell/formats/truncatedFormat.tsx
gscshoyru: It's only optional because the end-user isn't expected to use it -- the thing gets cloned and the value injected whenever it's used in the right spot inside the table. In fact any values end-users put there will get overwritten. I can still put the null checks, though. 

It's a string because it's not "10" it's "10px", because it comes from styles. And we didn't care about filtering it down above, because it was about triggering changes.

gscshoyru: ...do we want to do the parsing at the cell level instead of here? I think that's probably slightly better, will-do.

packages/table/src/cell/formats/truncatedFormat.tsx
```diff
@@ -80,6 +79,37 @@ export interface ITruncatedFormatProps extends IProps {
...
+    /**
+     * Approximate character width, used to determine whether to display the popover in approx truncation mode.
+     * The default value should work for normal table styles,
+     * but should be changed as necessary if the fonts or styles are changed significantly.
+     * @default 8
+     */
+    approximateCharWidth?: number;
+
+    /**
+     * Approximate character width, used to determine whether to display the popover in approx truncation mode.
```
adidahiya: "Approximate line height..."

packages/table/src/cell/formats/truncatedFormat.tsx
```diff
@@ -80,6 +79,37 @@ export interface ITruncatedFormatProps extends IProps {
      * @default "..."
      */
     truncationSuffix?: string;
+
+    /**
+     * Approximate character width, used to determine whether to display the popover in approx truncation mode.
```
adidahiya: should note that this is `(in pixels)`

packages/table/src/cell/formats/truncatedFormat.tsx
adidahiya: also `(in pixels)` here

packages/table/src/cell/formats/truncatedFormat.tsx
```diff
@@ -80,6 +79,37 @@ export interface ITruncatedFormatProps extends IProps {
...
+    /**
+     * Total horizonal cell padding (both sides), used to determine whether to display the popover in approx truncation mode.
+     * The default value should work for normal table styles,
+     * but should be changed as necessary if the fonts or styles are changed significantly.
+     * @default 20
+     */
+    cellHorizontalPadding?: number;
+
+    /**
+     * Number of buffer lines desired, used to determine whether to display the popover in approx truncation mode.
```
adidahiya: Hmm it's unclear to me what a buffer line is from reading these docs.

Also I wonder if these props should all be grouped in an object like `styleLayout: { approximateCharWidth, approximateLineHeight, ...}` because you're likely to set them all at once or not at all.

adidahiya:
Overall the perf improvements look nice here.

An unrelated issue I'm seeing:

when resizing columns, I spend 200ms (2 x 100ms) doing a forced layout in `measureDesiredRowHeaderWidth`. this seems superfluous since row headers shouldn't change at all when resizing columns:

image

packages/table/src/cell/formats/truncatedFormat.tsx
gscshoyru: Could certainly do this! The main issue with doing so is I'd have to merge the default props myself, since defaultProps are only shallowly merged.

adidahiya: honestly we can get away with _not_ merging those default props for now. make the `styleLayout` object optional but all its fields required -- if this ends up feeling like a clunky API to actual users, we can reconsider, but I see no reason to do the extra work right now.

PR #1858, opened 2017-12-01
[KeyCombo] Render span element instead of div

isbjoernfiona:
Per offline discussion with @giladgray, return span instead of div such that the keycombo can be used inline, e.g. to prompt users to use keyboard combinations to do things.


palantirtech:
Thanks for your interest in [palantir/blueprint](https://github.com/palantir/blueprint), @isbjoernfiona! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

PR #1867, opened 2017-12-05
Emit ES modules in dev scripts, CJS in CI scripts

adidahiya:
This improves the dev experience based on something I saw in @ethanbond's workflow.

We now compile ES modules in the default dev workflows and only produce CommonJS modules when ~building the distribution (for backwards compatibility)~ running tests or building the distribution.

Here's how I think about the split:

- _compilation_: tasks that allow you to test your development work, run unit tests, and produce code that runs the docs app or table-dev-app
- _distribution_: tasks that produce code that is only used by blueprint's NPM consumers. this is things like the CJS modules and unpkg bundles

---

Now we compile CJS (via `compile-libs` at the root) for the isomorphic tests because mocha can't handle ES modules natively. I spent a few minutes trying to get this to work with `@std/esm` but when I hit a few issues I decided to leave that for another day (in theory, this shouldn't be too hard, just going to punt it to later). At the same time, this does offer some benefits -- we are able to dogfood _both_ the ES modules and the CommonJS modules in the CI build.


packages/core/package.json
```diff
@@ -4,7 +4,7 @@
   "description": "Core styles & components",
   "main": "dist/index.js",
   "module": "dist/esm/index.js",
-  "typings": "dist/index.d.ts",
+  "typings": "dist/esm/index.d.ts",
```
adidahiya: this is tricky... we might need to produce typings at __both__ locations, e.g. next to the CJS files and next to the ESM files

packages/labs/webpack.config.js
```diff
@@ -8,15 +8,15 @@ const path = require("path");
 
 module.exports = Object.assign({}, baseConfig, {
     entry: {
-        "labs.bundle": [
+        labs: [
```
zerovox: does this change the name of any emitted .css files?

adidahiya: good shout -- it does not because extract text works based on the name of the `.scss` file, which we name the same as the output css file by convention. this is also verified by `assert-package-layout` and I checked it just now locally.

PR #1871, opened 2017-12-06
Migrate to CircleCI 2.0

adidahiya:
Changes in this PR:

- Remove `circle.yml` in favor of `.circleci/config.yml` to configure [CircleCI 2.0](https://circleci.com/docs/2.0/)
- Optimize CI build pipeline to reduce build time from 10min -> 4min
- Move `generate-docs-data` and the resulting JSON it produces into a new package called `@blueprintjs/docs-data` so that we can have a clean separation of `typescript` versions (`documentalist > ts-quick-docs` still requires TS 2.1, while all other compilation requires TS 2.4)


.circleci/config.yml
```diff
@@ -0,0 +1,168 @@
...
+  test-datetime:
+    docker:
+      - image: circleci/node:latest
+    steps:
+      - checkout
+      - attach_workspace:
+          at: '.'
+      - restore_cache:
+          key: v1-dependency-cache-{{ checksum "yarn.lock" }}
+      - run: ./node_modules/.bin/lerna run test --scope '@blueprintjs/datetime'
```
giladgray: `yarn lerna run test ...args`?

`yarn <binary>` is equivalent to `./node_modules/.bin/<binary>`

adidahiya: yeah makes sense

giladgray:
build failures look like a typescript version mismatch? 
it's complaining about default generics syntax `Foo<T = {}>`.

adidahiya:
getting closer...
image

PR #1891, opened 2017-12-12
Graduate Labs Components - Part 2: Replace Popover/Tooltip with Popover2/Tooltip2

cmslewis:
- Delete `Tooltip` and `Popover` from `core/`
- Rename `Popover2` and `Tooltip2` to `Popover` and `Tooltip`, respectively
- Delete `tether` dependency
- Fix build failures.
- Migrate deprecated internal usages to new props (`isDisabled`, `useModal`, `hasBackdrop`)

Note: `Context Menu` is super screwy now; its calculated `transform` offset is totally wrong. ***Can I get some help with the popper modifiers for context menus?*** Not sure I changed them properly.

#### Next steps (each in a separate PR, probably):
1. ~Migrate deprecated internal usages to new props (e.g. `Popover2.isDisabled`).~ (done in this PR after all)
1. Write tests for `Tooltip2`, `Popover`, `TagInput`, and remove them from karma `coverageExcludes`; bring back old documentation that is still relevant.


packages/core/src/common/errors.ts
```diff
@@ -35,6 +35,7 @@ export const NUMERIC_INPUT_STEP_SIZE_NON_POSITIVE =
     ns + ` <NumericInput> requires stepSize to be strictly greater than zero.`;
 export const NUMERIC_INPUT_STEP_SIZE_NULL = ns + ` <NumericInput> requires stepSize to be defined.`;
 
+// TODO (clewis): Delete old Popover errors that aren't used anymore.
```
giladgray: can we get this resolved before merging?

packages/core/src/components/popover/popover.md
```diff
@@ -1,281 +1,73 @@
-@# Popovers
```
giladgray: oh there's some very useful stuff in here that we don't want to delete.

packages/core/src/components/popover/popover.tsx
```diff
@@ -107,15 +82,28 @@ export interface IPopoverProps extends IOverlayableProps, IProps {
      * Prevents the popover from appearing when `true`.
      * @default false
      */
+    disabled?: boolean;
+
+    /**
+     * Prevents the popover from appearing when `true`.
+     * @deprecated use `disabled`
+     */
     isDisabled?: boolean;
```
giladgray: lets just jump straight to the non-deprecated props

packages/core/src/components/popover/popover.md
cmslewis: Yep, adding it back in Part 4.

packages/core/src/common/errors.ts
cmslewis: Yep

packages/core/karma.conf.js
```diff
@@ -17,9 +17,9 @@ module.exports = function (config) {
             "src/components/popover/*",
             "src/components/tabs/*",
             // TODO (clewis): write tests for these component as part of the 2.0 effort:
-            "src/components/popover2/*",
+            "src/components/popover/*",
```
adidahiya: this is now a duplicate of line 17

packages/core/src/components/popover/popover.md
```diff
@@ -1,281 +1,73 @@
...
+  - `isDisabled` &rarr; `disabled` for consistency with HTML elements
+- ...and except for the handful of Tether-specific props, which are now Popper.js-specific:
+  - `position: Position` &rarr; [`placement: PopperJS.Placement`](#labs/popover2.placement)
+  - `tetherOptions: ITetherOptions` &rarr; [`modifiers: PopperJS.Modifiers`](#labs/popover2.modifiers)
+- ...and one special addition:
+  - `minimal: boolean` applies minimal styles, which includes removing the arrow and minimizing the transition
 
 @reactExample PopoverExample
 
-@## JavaScript API
```
giladgray: I think we likely want to keep this entire section, and all its subsections, as they document key features of `Popover` that are still relevant.

packages/karma-build-scripts/createKarmaConfig.js
```diff
@@ -70,7 +70,7 @@ module.exports = function createKarmaConfig({ dirname, coverageExcludes, coverag
             [path.join(dirname, "test/index.ts")]: "webpack",
         },
         reporters: ["mocha", "coverage"],
-        singleRun: true,
+        singleRun: false,
```
adidahiya: questionable edit... this is probably why the tests are hanging

cmslewis: DERP.

cmslewis:
image

image

packages/core/src/common/errors.ts
cmslewis: @giladgray Turns out `Popover2` doesn't use _any_ of these. Seems like it should still handle these cases, so I'd be happy to leave these and file a follow-up issue. Thoughts?

packages/core/src/components/context-menu/contextMenu.tsx
```diff
@@ -27,8 +27,8 @@ interface IContextMenuState {
     onClose?: () => void;
 }
 
-const TETHER_OPTIONS = {
-    constraints: [{ attachment: "together", pin: true, to: "window" }],
+const POPPER_MODIFIERS: PopperModifiers = {
+    preventOverflow: { boundariesElement: "window" },
```
cmslewis: Are these settings more or less equivalent to the old `constraints`?

packages/core/src/components/context-menu/contextMenu.tsx
```diff
@@ -41,23 +41,33 @@ class ContextMenu extends AbstractPureComponent<{}, IContextMenuState> {
...
-        const popoverClassName = classNames(Classes.MINIMAL, { [Classes.DARK]: this.state.isDarkTheme });
+        const popoverClassName = classNames({ [Classes.DARK]: this.state.isDarkTheme });
+
+        // the target isn't relevant in this case. the context menu simply
+        // appears overlayed at the desired offset on the screen.
+        const emptyTarget = <div />;
+
+        // make the positioned div a wrapper instead of the target. Popper.js
+        // inserts the popover into the DOM as a *sibling* of the target, so
+        // this is the best way to ensure the popover gets positioned properly.
```
cmslewis: Think this comment really needs to be here after this PR merges?

packages/core/src/components/index.ts
```diff
@@ -48,11 +48,8 @@ export * from "./non-ideal-state/nonIdealState";
 export * from "./overlay/overlay";
 export * from "./text/text";
 export * from "./popover/popover";
-export * from "./popover2/popover2";
-export * from "./popover/svgPopover";
 export * from "./portal/portal";
 export * from "./progress/progressBar";
-export * from "./tooltip/svgTooltip";
```
cmslewis: @adidahiya @giladgray React 16 was causing `SVGTooltip` and `SVGPopover` tests to fail, because they were trying to render `<g>` tags outside of an `<svg>`. Since these components are such thin wrappers around the core `Tooltip` and `Popover`, respectively, I opted just to delete them and consider adding documentation about working with these components in SVG land.

Does that sound okay?

packages/core/src/components/menu/menuItem.tsx
```diff
@@ -133,32 +137,36 @@ export class MenuItem extends AbstractPureComponent<IMenuItemProps, IMenuItemSta
...
                     enforceFocus={false}
                     hoverCloseDelay={0}
                     inline={true}
                     interactionKind={PopoverInteractionKind.HOVER}
                     position={this.state.alignLeft ? Position.LEFT_TOP : Position.RIGHT_TOP}
-                    useSmartArrowPositioning={false}
                     {...popoverProps}
-                    content={submenuElement}
+                    content={submenuRef}
+                    minimal={true}
```
cmslewis: Use the `minimal` prop instead of applying the `pt-minimal` class alone. This ensures the menu won't leave room for an arrow that doesn't even exist.

packages/core/src/components/menu/menuItem.tsx
```diff
@@ -174,47 +182,56 @@ export class MenuItem extends AbstractPureComponent<IMenuItemProps, IMenuItemSta
...
+        }
+    };
 
-    private measureSubmenu = (el: Menu) => {
-        if (el != null) {
-            const submenuRect = ReactDOM.findDOMNode(el).getBoundingClientRect();
-            const parentWidth = this.liElement.parentElement.getBoundingClientRect().width;
-            const adjustmentWidth = submenuRect.width + parentWidth;
+    private maybeAlignSubmenuLeft() {
+        if (this.popoverRef == null) {
```
cmslewis: Small style improvement: put this check on top and de-intent everything below.

packages/core/src/components/tooltip/tooltip.md
```diff
@@ -1,77 +1,7 @@
-@# Tooltips
+@# Tooltip
```
cmslewis: Will audit these docs and re-include relevant info in a follow-up PR.

packages/core/src/components/tooltip/tooltip.tsx
```diff
@@ -1,17 +1,16 @@
 /*
- * Copyright 2015 Palantir Technologies, Inc. All rights reserved.
+ * Copyright 2017 Palantir Technologies, Inc. All rights reserved.
```
cmslewis: 2015 because it's the same file in the diff? 2017 because that's when we wrote `Tooltip2`?

packages/core/src/components/index.ts
giladgray: code search reveals a very small number (~5) of `SVGPopover` usages, and it's possible some of those are stale. I hate to say it, but I think we'll want to at least preserve support for this use case, which may require some work to be able to customize elements.

packages/core/src/components/menu/menuItem.tsx
```diff
@@ -174,47 +182,56 @@ export class MenuItem extends AbstractPureComponent<IMenuItemProps, IMenuItemSta
         }
     }
 
-    private liRefHandler = (r: HTMLElement) => (this.liElement = r);
+    private handlePopoverDidOpen = () => {
+        if (this.props.useSmartPositioning) {
+            // Popper.js renders the popover in the DOM before relocating its
+            // position on the next tick. We need to rAF to wait for that to happen.
+            requestAnimationFrame(() => this.maybeAlignSubmenuLeft());
```
giladgray: 🤔 Popper.js _should_ be powerful enough to handle all the positioning logic: we should let _it_ decide when to flip to the left, because it likely knows better than us. Happy for this to be a follow-up, though.

packages/core/src/components/popover/popover.md
giladgray: wait what's part 3? is it possible to just not delete it now?

packages/core/src/components/tooltip/tooltip.tsx
giladgray: yeah lol I'm not sure! probably 2015 cuz it's the same file.

packages/core/src/components/tooltip/tooltip.tsx
```diff
@@ -21,17 +20,16 @@ export interface ITooltipProps extends IProps, IIntentProps {
     content: JSX.Element | string;
 
     /**
-     * Constraints for the underlying Tether instance.
-     * See http://github.hubspot.com/tether/#constraints
-     * @deprecated since v1.12.0; use `tetherOptions` instead.
+     * Initial opened state when uncontrolled.
+     * @default false
```
giladgray: remove this, it defaults to `undefined`

packages/core/test/menu/menuTests.tsx
```diff
@@ -219,9 +223,8 @@ describe("MenuItem", () => {
                             <MenuItem iconName="underline" text="Underline" />
                         </MenuItem>
                     </MenuItem>
-                </MenuItem>,
-                childContainer,
-            ) as MenuItem;
+                </React.Fragment>,
```
giladgray: ✨ 😱 ✨

packages/core/test/menu/menuTests.tsx
```diff
@@ -293,6 +290,15 @@ describe("MenuItem", () => {
         function assertClassNameCount(className: string, count: number) {
             assert.strictEqual(childContainer.queryAll(`.${className}`).length, count, `${count}x .${className}`);
         }
+
+        function mountMenuItem(props?: IMenuItemProps, childItems?: React.ReactNode) {
+            return ReactDOM.render(
```
giladgray: awkward we're not using `enzyme` here

packages/core/src/components/tooltip/tooltip.tsx
cmslewis: Doesn't it default explicitly to `false` below?

https://github.com/palantir/blueprint/pull/1891/files/08312ffb58f969cfa74dc3738982393d142cdc07#diff-0f4e5f29db46149ff7a16aa9869d0ff9R130

packages/core/test/menu/menuTests.tsx
cmslewis: @giladgray Looks like we need real DOM to properly evaluate all the real-estate-dependent dynamic-layout behavior. Also looks like you were the original test author, so feel free to tell me if that's not the case!

packages/core/test/popover/popoverTests.tsx
```diff
@@ -33,13 +46,13 @@ describe("<Popover>", () => {
         testsContainerElement.remove();
     });
 
-    describe("validation:", () => {
+    describe.skip("validation:", () => {
```
adidahiya: use `{ expectPropValidationError } from "@blueprintjs/test-commons";`
fine to do in a follow-up PR, just make a note in this PR description

packages/core/test/popover/popoverTests.tsx
```diff
@@ -538,51 +551,132 @@ describe("<Popover>", () => {
...
+            );
         });
-        assert.notEqual(0, root.state().targetWidth, "targetWidth should not equal 0");
-        assert.notEqual(0, root.state().targetHeight, "targetHeight should not equal 0");
     });
 
-    it("componentDOMChange does not update targetHeight/targetWidth state when useSmartArrowPositioning=false", () => {
-        const root = renderPopover({
-            useSmartArrowPositioning: false,
+    describe("deprecated prop shims", () => {
```
adidahiya: should remove this `describe` block in a follow-up PR

cmslewis:
Unrelated test flake failed the build. I'll run it one more time and then just merge into my feature branch if it flakes again (I have a proposed fix for the flaky test btw; will open a new PR if it's foolproof).

cmslewis:
Flaked again. Opened this PR to (hopefully) fix the flaky `Overlay` test: https://github.com/palantir/blueprint/pull/1943. Merging this for now.

PR #1892, opened 2017-12-12
Add docs for truncated-by-approx

gscshoyru:
Missed the docs for it in the props options, adding it now


tgreenwatts:
Thanks Geoff!

gscshoyru:
I do normally have a check for whitespace -- but I made this change in github directly, and their editor doesn't have a nice linter. Sadly.

PR #1902, opened 2017-12-13
Switch to Flexbox layout for various components

### LINKED ISSUE #1425: Refactor buttons, input groups and others into a 3-column flex layout
Applicable to:
- [x] `.pt-menu-item` #2075 
- [x] `.pt-button` #2078 
- [ ] `.pt-tag-input`
- ~`.pt-input-group`~ (cannot flex as `input` must surround left/right for focus state)

marvinsum:
#### Checklist

- [x] Update `Button`
- [x] Update `MenuItem`
- [ ] Update `TagInput`
- [ ] Update `Toast`
- [ ] Test in Chrome, Firefox and Edge

Switch to using Flexbox layout for `.pt-button`, `.pt-input-group`, `.pt-tag-input` and `.pt-menu-item`. This PR removes a lot of the older layout/positioning hacks which were required prior to Flexbox.

#### Breaking changes

- `MenuItem` requires `<span class="pt-menu-item-text">...</span>` wrapped around the text when using the CSS API

#### Screenshot

No visual changes.


packages/core/src/components/forms/_input-group.scss
```diff
@@ -10,20 +10,20 @@ Input groups
...
 </div>
 <div class="pt-input-group {{.modifier}}">
-  <span class="pt-icon pt-icon-search"></span>
+  <span class="pt-icon-standard pt-icon-search"></span>
   <input type="text" class="pt-input" {{:modifier}} placeholder="Search" />
   <button class="pt-button pt-minimal pt-intent-primary pt-icon-arrow-right" {{:modifier}}></button>
 </div>
 
-:disabled - Disabled input. Must be added separately to the <code>&#60;input&#62;</code> and <code>&#60;button&#62;</code>. Also add <code>.pt-disabled</code> to <code>.pt-input-group</code> for icon coloring (not shown below).
+.pt-disabled - Disabled input. Must be added separately to the <code>&#60;input&#62;</code> and <code>&#60;button&#62;</code>. Also add <code>.pt-disabled</code> to <code>.pt-input-group</code> for icon coloring (not shown below).
```
llorca: that doesn't seem right

packages/core/src/components/button/_button.scss
```diff
@@ -191,7 +191,7 @@ Styleguide pt-button
     #{$icon-classes} {
       margin-right: $button-icon-spacing-large;
 
-      &.pt-align-right {
+      &:last-child {
```
giladgray: skeptical of this change... using the `.pt-align-right` to change margins is much more reliable (though slightly more work for users) than affecting the last-child without thinking.

packages/core/src/common/_mixins.scss
```diff
@@ -103,6 +103,39 @@ $pt-dark-intent-text-colors: (
   @include force-hardware-acceleration();
 }
 
+// A collection of mixins that utilize the CSS Flexbox model to standardize
+// the layout and composition of Blueprint components.
+
+// pt-flex-parent styles
+@mixin pt-flex-parent() {
+  display: inline-flex;
```
giladgray: probably not enough to put `inline-flex` everywhere... block elements are quite useful.

packages/core/src/components/forms/_input-group.scss
```diff
@@ -37,38 +41,48 @@ $input-button-height: $pt-button-height - 6px !default;
...
+  &.pt-fill {
+    @include pt-flex-parent-fill();
+  }
 
-  .pt-input {
-    // explicit position prevents shadow clipping https://github.com/palantir/blueprint/pull/490#issuecomment-273342170
-    position: relative;
-    width: 100%;
+  // apply the :focus state on .pt-input-group, not .pt-input
+  &:focus,
```
giladgray: 🙅‍♂️ unfortunately this is probably a non-starter. can't shift the `:focus` state to the parent element because
a) `.pt-input-group` not actually focusable (no tab index, etc), and
b) it can't be made to appear focused **in css only** when its child `input` is focused because CSS can only go down the tree, not up.

you'll note that InputGroups now have no focus state whatsoever when typing in them.
https://6591-71939872-gh.circle-artifacts.com/0/home/circleci/project/packages/docs-app/dist/index.html#core/components/forms/input-group
screen shot 2017-12-14 at 1 24 50 pm

packages/core/src/components/forms/_input-group.scss
```diff
@@ -78,20 +92,21 @@ $input-button-height-large: $pt-button-height !default;
...
-    @include pt-icon($pt-icon-size-standard);
-
-    // bump icon up so it sits above input
-    z-index: 1;
+  > .pt-icon-standard {
     margin: 0 ($pt-input-height - $pt-icon-size-standard) / 2;
-    line-height: $pt-input-height;
+  }
+
+  .pt-icon-standard {
```
giladgray: we used `.pt-icon` before so that we could adapt the icon size if `large`. seems we never actually use large icons, so this is probably fine.

packages/core/src/components/forms/_input-group.scss
```diff
@@ -144,53 +160,53 @@ $input-button-height-large: $pt-button-height !default;
...
-  &.pt-round {
-    .pt-button,
-    .pt-input,
-    .pt-tag {
-      border-radius: $pt-input-height;
-    }
+  &.pt-round,
+  &.pt-round .pt-button,
+  &.pt-round .pt-tag, {
+    border-radius: $pt-input-height;
```
giladgray: why make this change? the previous syntax was shorter and more obvious.

packages/core/src/components/forms/inputGroup.tsx
```diff
@@ -106,17 +99,15 @@ export class InputGroup extends React.Component<HTMLInputProps & IInputGroupProp
...
-            this.setState({ rightElementWidth: 0 });
-        }
-    }
+    private handleOnBlur = (e: React.FocusEvent<HTMLInputElement>) => {
+        this.props.onBlur != null ? this.props.onBlur(e) : undefined;
+        this.setState({ isActive: !this.state.isActive });
+    };
+
+    private handleOnFocus = (e: React.FocusEvent<HTMLInputElement>) => {
+        this.props.onFocus != null ? this.props.onFocus(e) : undefined;
```
giladgray: we have a hammer for this: `Utils.safeInvoke(this.props.onFocus, e)`

packages/core/src/components/forms/inputGroup.tsx
```diff
@@ -106,17 +99,15 @@ export class InputGroup extends React.Component<HTMLInputProps & IInputGroupProp
...
+        this.setState({ isActive: !this.state.isActive });
```
giladgray: should probably be explicit: set to `true` here and `false` in blur, since we know the desired end state and it's possible that multiple quick state transitions could break this.

packages/core/src/components/forms/inputGroup.tsx
```diff
@@ -66,34 +66,27 @@ export class InputGroup extends React.Component<HTMLInputProps & IInputGroupProp
             Classes.intentClass(intent),
             {
                 [Classes.DISABLED]: this.props.disabled,
+                [Classes.ACTIVE]: this.state.isActive,
```
giladgray: _oh I see_. still, this breaks the CSS API for this component pretty wholly.

packages/core/src/common/_mixins.scss
```diff
@@ -103,6 +103,39 @@ $pt-dark-intent-text-colors: (
...
+  align-items: center;
+  justify-content: center;
+}
+
+@mixin pt-flex-parent-fill() {
```
giladgray: this mixin really has nothing to do with flex and it's an unnecessary level of abstraction.
let's just remove it and write the one line of CSS where necessary.

packages/core/src/common/_mixins.scss
```diff
@@ -103,6 +103,39 @@ $pt-dark-intent-text-colors: (
...
+  width: 100%;
+}
+
+@mixin pt-flex-parent-direction-horizontal {
+  flex-direction: row;
+}
+
+@mixin pt-flex-parent-direction-vertical {
+  flex-direction: column;
+}
```
giladgray: these mixin names are uncomfortably long.
how about a simple argument to `pt-flex-parent($direction)`?

packages/core/src/common/_mixins.scss
```diff
@@ -103,6 +103,39 @@ $pt-dark-intent-text-colors: (
...
+
+// pt-flex-child styles
```
giladgray: remove unnecessary comment (and line 109): not helpful as divider, and immediately obvious from actual code.

llorca:
Nice stuff @marvinsum! Some bugs I noticed:
- [ ] Gotta fix the rounded input groups:
image
- [ ] Lost the focus state on input groups
image

packages/core/src/common/_mixins.scss
```diff
@@ -103,6 +103,39 @@ $pt-dark-intent-text-colors: (
...
+@mixin pt-flex-child {
+  flex: 0 0 auto;
+}
+
+@mixin pt-flex-child-fill {
+  flex: 1 1 auto;
+  min-width: 0;
+  min-height: 0;
```
giladgray: why clear the min sizes? what if I want min sizes?

packages/core/src/common/_mixins.scss
```diff
@@ -103,6 +103,39 @@ $pt-dark-intent-text-colors: (
   @include force-hardware-acceleration();
 }
 
+// A collection of mixins that utilize the CSS Flexbox model to standardize
+// the layout and composition of Blueprint components.
+
+// pt-flex-parent styles
+@mixin pt-flex-parent() {
```
giladgray: new syntax proposal:
```scss
@mixin pt-flex-parent($direction: row, $inline: false) {
    @if $inline == true { 
        display: inline-flex; 
    } @else { 
        display: flex; 
    }
    flex-direction: $direction;
    align-items: center;
    justify-content: center;
}

packages/core/src/common/_mixins.scss
marvinsum: Should this be added individually by flex-children themselves? If you look at the docs, the `.pt-input` (somehow, browser default?) has `min-width: 146px` by default. If you shrink the `.pt-input-group` significantly, you'd need to overwrite its `min-width`.

https://6488-71939872-gh.circle-artifacts.com/0/home/circleci/project/packages/docs-app/dist/index.html#core/components/forms/input-group

packages/core/src/components/forms/_input-group.scss
marvinsum: @giladgray mmm, I'm not sure haha. Will revert.

packages/core/src/components/forms/inputGroup.tsx
marvinsum: What's a hammer? So it should be `this.props.onFocus != null ? Utils.safeInvoke(this.props.onFocus, e) : undefined;`?

giladgray: no, just safeInvoke. search codebase for other usages, you'll see what it does.

packages/core/src/components/forms/inputGroup.tsx
marvinsum: yep

packages/core/src/components/forms/inputGroup.tsx
llorca: basically, `safeInvoke` does this logic for you. line should be `Utils.safeInvoke(this.props.onFocus, e);`

adidahiya:
Please merge master to get the new required status check

packages/core/src/components/forms/inputGroup.tsx
marvinsum: @giladgray stashing this and punting for now

packages/core/src/components/forms/_input-group.scss
marvinsum: @giladgray stashing this and punting for now

packages/core/src/components/forms/_input-group.scss
marvinsum: @giladgray stashing this and punting for now

packages/core/src/components/forms/_input-group.scss
marvinsum: @giladgray stashing this and punting for now

packages/core/src/components/forms/inputGroup.tsx
marvinsum: @giladgray stashing this and punting for now

packages/core/src/components/forms/inputGroup.tsx
marvinsum: @llorca @giladgray stashing this and punting for now

packages/core/src/components/button/_button.scss
marvinsum: @giladgray ok, so we should bias towards being more verbose in the html, which would lead to inadvertently selecting the wrong element?

packages/core/src/common/_mixins.scss
marvinsum: @giladgray seems reasonable

packages/core/src/common/_mixins.scss
marvinsum: @giladgray seems reasonable

packages/core/src/common/_mixins.scss
marvinsum: @giladgray seems reasonable

packages/core/src/common/_mixins.scss
marvinsum: yep

packages/core/src/common/_mixins.scss
marvinsum: @giladgray fair, only `Button` uses `inline-flex`.

giladgray:
closing in favor of smaller PRs, starting with #2075

PR #1872, opened 2017-12-07
Add active state to Popover2 when popover is open and target is a button

gjavitt:
#### Checklist

- [x] [Enable CircleCI for your fork](https://circleci.com/add-projects)
- [ ] Include tests
- [ ] Update documentation

#### Changes proposed in this pull request:
Adds active state to Popover2 when popover is open and target is a button. (Does not make button active if popover is triggered by hover interaction instead of click.)

#### Screenshot
image

PR #1909, opened 2017-12-13
Prepare release labs@0.14.1

cmslewis:
## `@blueprintjs/labs 0.14.1`

> Patch release because these are just small bugfixes / ERs

### 💎 New features

#1843 `TagInput` `onKeyDown` and `onKeyUp` callbacks now receive the active tag index as a second parameter 
#1872 `Popover2` adds `pt-active` class the target while the popover is open (🎩  @gjavitt)

### 🐛 Bug fixes
#1879 `TimezonePicker` Fix popoverProps typing to use `IPopover2Props`


adidahiya:
not necessary, please use `release/1.x` branch

PR #1910, opened 2017-12-13
Prepare release datetime@1.25.1

cmslewis:
## `@blueprintjs/datetime 1.25.1`

> Patch release because these are just small bugfixes / ERs

### 💎 New features
#1825 `DateInput` and `DateRangeInput` now commit the typed value when 'Enter' is pressed


adidahiya:
not necessary, please use `release/1.x` branch

PR #1924, opened 2017-12-14
Add new @blueprintjs/timezone package

hellochar:
Followed https://github.com/palantir/blueprint/wiki/Creating-a-new-package


package.json
```diff
@@ -21,8 +21,9 @@
     "dev:landing": "lerna run dev --parallel --scope '@blueprintjs/{core,landing-app}'",
     "dev:select": "lerna run dev --parallel --scope '@blueprintjs/{core,select,docs-app}'",
     "dev:table": "lerna run dev --parallel --scope '@blueprintjs/table-dev-app'",
+    "dev:timezone": "lerna run dev --parallel --scope '@blueprintjs/{core,select,timezone,docs-app}'",
```
gscshoyru: Why select?

packages/docs-app/src/tags/reactExamples.ts
```diff
@@ -36,6 +36,6 @@ function addPackageExamples(
 
 addPackageExamples("core", CoreExamples as any);
 addPackageExamples("datetime", DateExamples as any);
-addPackageExamples("labs", LabsExamples as any);
```
gscshoyru: No labs examples at all, anymore?

packages/labs/src/common/index.ts
```diff
@@ -4,6 +4,5 @@
  * Licensed under the terms of the LICENSE file distributed with this project.
  */
 
-import * as classes from "../common/classes";
-
-export const Classes = classes;
+// Empty module since there's nothing to export.
+export {};
```
gscshoyru: ...I guess so, then.

package.json
adidahiya: I would just skip this command / it's not very useful (I prefer running the scripts in separate terminals) / we can add it back if anyone actually wants it

packages/labs/src/common/index.ts
adidahiya: I thought we were planning to re-export `@blueprintjs/select` components? That can be another PR though.

packages/timezone/package.json
```diff
@@ -0,0 +1,71 @@
...
+    "@blueprintjs/core": "^1.34.0",
+    "@blueprintjs/select": "^1.0.0",
+    "classnames": "^2.2",
+    "fuzzaldrin-plus": "^0.5.0",
+    "moment": "^2.14.1",
+    "moment-timezone": "^0.5.13",
+    "tslib": "^1.5.0"
+  },
+  "devDependencies": {
+    "@blueprintjs/node-build-scripts": "^0.2.0",
```
adidahiya: this needs to be `0.3.0` otherwise your build will fail on master

packages/timezone/package.json
```diff
@@ -0,0 +1,71 @@
...
+    "@blueprintjs/test-commons": "^0.1.0",
```
adidahiya: same here, `^0.3.0` (match semver range of whatever the package currently is in the monorepo)

packages/labs/package.json
```diff
@@ -32,8 +32,6 @@
     "@blueprintjs/select": "^1.0.0",
     "classnames": "^2.2",
     "fuzzaldrin-plus": "^0.5.0",
```
adidahiya: you can now remove this as well as its types below

packages/docs-app/src/tags/reactExamples.ts
hellochar: Yep no more examples at all, because there's no components, or subitems in labs. It's just an empty shell for now.

packages/labs/src/common/index.ts
hellochar: future PR, I still havea a few reservations

adidahiya:
Please merge master to fix conflicts and pick up the new required status check

adidahiya:
pretty sure you need to bump the dependency cache to `v4`

hellochar:
Looks like v4 didn't work because a previous build had already hit the cache. v5 ftw!

PR #1939, opened 2017-12-18
Circle CI build improvements

cmslewis:
Inspired by build issues in #1891:
- Switch from `yarn lerna run test --scope '@blueprintjs/____` to `cd packages/____ && yarn test`.
    - The former was hanging without printing anything.
- To make the above work, add `packages/____/node_modules` to `save_cache.paths`.
- Bust the cache (changed v5 => v6).
- Add missing `mocha` dev dependency to `packages/timezone`.
    - Because of this: `"test:iso": "mocha test/isotest.js",`

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -21,6 +21,7 @@ jobs:
             - packages/core/node_modules
             - packages/datetime/node_modules
             - packages/docs/node_modules
+            - packages/icons/node_modules
             - packages/labs/node_modules
             - packages/select/node_modules
             - packages/table/node_modules
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -7,17 +7,24 @@ jobs:
       - checkout
       - restore_cache:
           keys:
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -7,17 +7,25 @@ jobs:
...
             - ~/.cache/yarn
             - node_modules
+            - packages/core/node_modules
+            - packages/datetime/node_modules
+            - packages/docs/node_modules
+            - packages/icons/node_modules
+            - packages/labs/node_modules
+            - packages/select/node_modules
+            - packages/table/node_modules
+            - packages/timezone/node_modules
```
adidahiya: :( I don't want to list out these package names again like this. let's move back to the old approach of `persist_to_workspace` with `packages/*/node_modules/.bin`

cmslewis: If the `singleRun` edit fixes the other branch, I'm happy to just close this whole PR.

cmslewis: It worked.

PR #1943, opened 2017-12-19
Fix flaky Overlay tests (hopefully)

cmslewis:
Use nested `setTimeout`s instead of an arbitrary short delay for one `setTimeout`. Goal is to delay the code execution until the end of the next frame, not just to the end of the current frame (which may in some instances cause code to invoke _before_ focus has changed / `requestAnimationFrame` has executed).


cmslewis:
...Initial build failed due to yet another (unrelated) flaky test. The irony.

cmslewis:
@adidahiya we can try that in a follow-up, yes.

PR #1945, opened 2017-12-19
Slim down docs-app bundle by loading fewer locales (#1930)

### LINKED ISSUE #1930: [blueprintjs/datetime] allow for optional moment locales
react-day-picker is trying to get `.firstDayOfWeek()` but its trying to get its through `moment.localeData`
https://github.com/gpbl/react-day-picker/blob/ac85a53929e47ba025b523b05d3f6e1c7f112da7/src/addons/MomentLocaleUtils.js#L3

which is commonly excluded in webpack via:

```
const plugins = [
    new webpack.IgnorePlugin(/^\.\/locale$/, /moment$/),
];
```

this will break for some i18n users

quick solution is to add back locales, but we remove it since it increases build size. it would be nice to be able to use the DatePicker without include locales

clekili:
Fixes #1930

Webpack only pull moment EN locale by default and datetime package will only pull moment EN/DE/FR locales by default

Locale defaults


adidahiya:
This only affects the docs-app bundle size, right? What's the overall goal here?

PR #1948, opened 2017-12-20
Merge release/1.x into master

adidahiya:
- Regenerate 1.x docs


adidahiya:
CI is reporting the wrong thing because that check ran on `release/1.x`, which has a different CI setup than `master`. I checked the built artifact manually and it fixes the problem I wanted -- `@blueprintjs/select` no longer shows up in the Releases dropdown.

PR #1960, opened 2017-12-21
Fix TagInput test coverage

cmslewis:
Turned out the be easier than I expected. :)


adidahiya:
I thought we skipped these flaky tests by this point...

cmslewis:
I might have forked before that.

PR #1968, opened 2017-12-23
Graduate Labs Components: Fix/Improve Popover Examples

cmslewis:
- Fix docs-app scroll-jumping issue on the `PopoverInlineExample` by setting `enforceFocus={false}`
- Fix issue where `PopoverInlineExample` scroll containers weren't being centered on mount.
- Fix issue where `PopoverInlineExample` non-inline popover wasn't tracking the target.
- Set `enforceFocus={false}` on the `CLICK_TARGET_ONLY` example, so user can still click on things in the rest of the docs app. 
- Remove mentions of old props in Popover error-message constants.

### DIFF
--- a/packages/docs-app/src/examples/core-examples/popoverExample.tsx
+++ b/packages/docs-app/src/examples/core-examples/popoverExample.tsx
@@ -271,9 +271,13 @@ export class PopoverExample extends BaseExample<IPopoverExampleState> {
 
     private centerScroll = (div: HTMLDivElement) => {
         if (div != null) {
-            const container = div.parentElement;
-            container.scrollTop = div.clientHeight / 4;
-            container.scrollLeft = div.clientWidth / 4;
+            // if we don't requestAnimationFrame, this function apparently executes
+            // before styles are applied to the page, so the centering is way off.
+            requestAnimationFrame(() => {
+                const container = div.parentElement;
+                container.scrollTop = div.clientHeight / 4;
+                container.scrollLeft = div.clientWidth / 4;
+            });
         }
     };
 }
--- a/packages/docs-app/src/examples/core-examples/popoverInlineExample.tsx
+++ b/packages/docs-app/src/examples/core-examples/popoverInlineExample.tsx
@@ -9,7 +9,15 @@ import * as React from "react";
 import { Button, IPopoverProps, Popover, Position } from "@blueprintjs/core";
 import { BaseExample } from "@blueprintjs/docs";
 
-export class PopoverInlineExample extends BaseExample<{}> {
+export interface IPopoverInlineExampleState {
+    hasMounted?: boolean;
+}
+
+export class PopoverInlineExample extends BaseExample<IPopoverInlineExampleState> {
+    public state: IPopoverInlineExampleState = {
+        hasMounted: false,
+    };
+
     protected className = "docs-popover-inline-example";
 
     private scrollContainer1Ref: HTMLDivElement;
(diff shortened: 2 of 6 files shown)


llorca:
Might be unrelated, though are you aware of this `inline={true}` problem in the main example?
image

cmslewis:
@llorca yeah, that example is just kind of wonky all over. Would prefer to replace it with a bunch of simpler, orthogonal examples but didn't prioritize that.

llorca:
Ha, I love the detailed sections explaining each feature of the component, but also quite dig the single playground example where you can turn everything on/off, similarly to the table test page. I think it'd be useful to keep it, don't you? Is it wonky because of the example code or because of the component code?

packages/docs-app/src/examples/core-examples/popoverInlineExample.tsx
```diff
@@ -20,38 +28,53 @@ export class PopoverInlineExample extends BaseExample<{}> {
     };
 
     public componentDidMount() {
-        this.recenter();
+        // if we don't requestAnimationFrame, this function apparently executes
+        // before styles are applied to the page, so the centering is way off.
+        requestAnimationFrame(this.recenter);
+        // likewise, Popper.js doesn't handle controlled, non-inline popovers
+        // properly if rendered with isOpen={true} on initial mount: popovers
+        // won't follow the target when scrolled. to fix, defer opening.
```
giladgray: can this be done inside `Popover` itself? or maybe inside `Portal`, which now tracks a `hasMounted` state field? 

seems like useful behavior that will be desirable across the board.

packages/docs-app/src/examples/core-examples/popoverInlineExample.tsx
```diff
@@ -86,4 +109,19 @@ export class PopoverInlineExample extends BaseExample<{}> {
             scrollContainer.scrollLeft = contentWidth / 4;
         }
     };
+
+    private syncScroll1 = () => {
+        // use rAF to throttle scroll-sync calculations; otherwise, scrolling is noticeably choppy.
+        return requestAnimationFrame(() => this.syncScroll(this.scrollContainer1Ref, this.scrollContainer2Ref));
+    };
+
+    private syncScroll2 = () => {
```
giladgray: `Left,Right` > `1,2` ?

packages/docs-app/src/examples/core-examples/popoverInlineExample.tsx
cmslewis: I'm hesitant to make controlled Popover and/or Portal rendering asynchronous...

Can I file an issue to track this? Would prefer to see how things fare in the wild as is—then return if it proves painful for more folks than just me.

packages/docs-app/src/examples/core-examples/popoverExample.tsx
```diff
@@ -271,9 +271,13 @@ export class PopoverExample extends BaseExample<IPopoverExampleState> {
 
     private centerScroll = (div: HTMLDivElement) => {
         if (div != null) {
-            const container = div.parentElement;
-            container.scrollTop = div.clientHeight / 4;
-            container.scrollLeft = div.clientWidth / 4;
+            // if we don't requestAnimationFrame, this function apparently executes
+            // before styles are applied to the page, so the centering is way off.
```
adidahiya: this sounds highly suspect... how are styles not on the page yet?

cmslewis: Don't know, but IIRC, when I stopped at a breakpoint in here, nothing was styled on the page at all. @giladgray any idea why this might have happened?

packages/docs-app/src/examples/core-examples/popoverInlineExample.tsx
cmslewis: Done

packages/docs-app/src/examples/core-examples/popoverInlineExample.tsx
cmslewis: Either way, I'll merge this PR now and follow up somehow after.

adidahiya:
wait, why merge this? I don't expect to be reviewing the PR that merges this feature branch into master, so the code review should happen here. I am worried about the increasing usage of `requestAnimationFrame` all over the place.

PR #2007, opened 2018-01-16
Graduate labs components to core

adidahiya:
Feature branch: DO NOT SQUASH


packages/core/src/common/errors.ts
```diff
@@ -40,9 +39,11 @@ export const NUMERIC_INPUT_STEP_SIZE_NON_POSITIVE =
     ns + ` <NumericInput> requires stepSize to be strictly greater than zero.`;
 export const NUMERIC_INPUT_STEP_SIZE_NULL = ns + ` <NumericInput> requires stepSize to be defined.`;
 
+// TODO (clewis): Migrate old Popover validation errors to the component formerly known as Popover2.
```
cmslewis: Stray TODO. It's done though; can delete.

packages/datetime/src/dateInput.tsx
```diff
@@ -362,6 +361,10 @@ export class DateInput extends AbstractPureComponent<IDateInputProps, IDateInput
     };
 
     private handleInputClick = (e: React.SyntheticEvent<HTMLInputElement>) => {
+        // stop propagation to the Popover's internal handleTargetClick handler;
+        // otherwise, the popover will flicker closed as soon as it opens.
+        e.stopPropagation();
```
adidahiya: strange, how come this wasn't necessary before?

packages/core/test/menu/menuTests.tsx
```diff
@@ -151,39 +151,42 @@ describe("MenuItem", () => {
...
                 assert.isNotNull(childContainer.query(`.${Classes.OVERLAY_OPEN}`));
                 assert.isNull(childContainer.query(`.${Classes.ALIGN_LEFT}`));
                 done();
             });
         });
 
-        it("children will continue displaying in the same direction if possible", done => {
-            menuItem = ReactDOM.render(
-                <MenuItem iconName="style" text="Style">
+        // TODO (clewis): Commented this out during the Popover2 => Popover refactor. Get this working again.
```
cmslewis: Never got this test working, even though the functionality works. Should I try again now?

packages/core/test/menu/menuTests.tsx
```diff
@@ -203,9 +205,11 @@ describe("MenuItem", () => {
             });
         });
 
-        it("children will flip direction after no more room in the existing direction", done => {
-            menuItem = ReactDOM.render(
-                <MenuItem iconName="style" text="Style">
+        // TODO (clewis): Commented this out during the Popover2 => Popover refactor. Get this working again.
```
cmslewis: Same

yarn.lock
```diff
@@ -114,7 +114,7 @@
   version "0.2.4"
   resolved "https://registry.yarnpkg.com/@types/tapable/-/tapable-0.2.4.tgz#8181a228da46185439300e600c5ae3b3b3982585"
 
-"@types/tether@^1.1.27", "@types/tether@^1.4.3":
+"@types/tether@^1.4.3":
```
adidahiya: hmm, we should probably be able to drop this entry

packages/core/test/menu/menuTests.tsx
adidahiya: yes, but it shouldn't block this PR. please file a follow-up ticket and add it to the 2.0.0 milestone

packages/core/test/menu/menuTests.tsx
adidahiya: same

yarn.lock
adidahiya: done

packages/core/test/menu/menuTests.tsx
cmslewis: https://github.com/palantir/blueprint/issues/2008

packages/datetime/src/dateInput.tsx
cmslewis: Not sure, but `DateRangeInput` already had something like this FWIW:
https://github.com/palantir/blueprint/blob/develop/packages/datetime/src/dateRangeInput.tsx#L620

adidahiya:
Table test failure looks like a flake. I'll refactor the harness to use Enzyme harnesses in a separate PR.

I noticed a Popover regression -- the placement dropdown doesn't seem to do anything here

image

cmslewis:
@adidahiya I'll take a look.

adidahiya:
latest commits lgtm

adidahiya:
looks great!! :shipit:

cmslewis:
@adidahiya FWIW it's possible to get into a state in the example when `isOpen={true}` and the popover content is too big to let you toggle set the `isOpen` switch back to false:

image

adidahiya:
you can probably work around that by getting a bigger monitor. or zooming out :P

PR #2027, opened 2018-01-23
[release/1.x] [table] measureTextContentWithExclusions

adidahiya:
### Motivation

We'd like to add some text to our column headers (column descriptions) which should be shown in the header cell (as an affordance to users that they can click and view more details about the column) but not taken into account for double-click-to-resize interactions on the column.

### DIFF
--- a/packages/table/src/common/utils.ts
+++ b/packages/table/src/common/utils.ts
@@ -189,11 +189,11 @@ export const Utils = {
      * We use the computed font from the supplied element and a non-DOM canvas
      * context to measure the text.
      */
-    measureElementTextContent(element: HTMLElement): TextMetrics {
+    measureElementTextContent(element: Element): TextMetrics {
         const context = document.createElement("canvas").getContext("2d");
-        const style = getComputedStyleWithExclusions(element);
+        const style = getComputedStyle(element);
         context.font = CSS_FONT_PROPERTIES.map(prop => style.getPropertyValue(prop)).join(" ");
-        return context.measureText(element.textContent);
+        return measureTextContentWithExclusions(context, element);
     },
 
     /**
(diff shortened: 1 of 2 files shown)


packages/table/src/common/utils.ts
```diff
@@ -408,3 +407,27 @@ export const Utils = {
...
+ * Wrapper around document.getComputedStyle which applies some extra logic to exclude certain elements
+ * from the computation.
+ */
+function getComputedStyleWithExclusions(element: HTMLElement) {
+    const elementsToExclude = element.querySelectorAll(`.${CLASSNAME_EXCLUDED_FROM_TEXT_MEASUREMENT}`);
+
+    // hide all elements to exclude
+    for (let i = 0; i < elementsToExclude.length; i++) {
+        const el = elementsToExclude.item(i) as HTMLElement;
+        el.setAttribute("style", "display: none");
```
adidahiya: we should save off the old `style` attribute value and set it back to what it was previously.

adidahiya:
h/o, this won't work, give me a min

packages/table/src/common/utils.ts
```diff
@@ -408,3 +407,27 @@ export const Utils = {
...
+
+/**
+ * Wrapper around document.getComputedStyle which applies some extra logic to exclude certain elements
+ * from the computation.
+ */
+function getComputedStyleWithExclusions(element: HTMLElement) {
+    const elementsToExclude = element.querySelectorAll(`.${CLASSNAME_EXCLUDED_FROM_TEXT_MEASUREMENT}`);
+
+    // hide all elements to exclude
+    for (let i = 0; i < elementsToExclude.length; i++) {
```
cmslewis: nit: use `forEach`?

adidahiya:
updated to do the "ignore me" part in the right spot

packages/table/src/common/utils.ts
```diff
@@ -184,17 +186,14 @@ export const Utils = {
...
      * We use the computed font from the supplied element and a non-DOM canvas
      * context to measure the text.
-     *
-     * Returns a `TextMetrics` object.
      */
-    measureElementTextContent(element: Element) {
+    measureElementTextContent(element: Element): TextMetrics {
         const context = document.createElement("canvas").getContext("2d");
-        const style = getComputedStyle(element, null);
+        const style = getComputedStyle(element);
```
adidahiya: should probably add back the second param

packages/table/src/locator.ts
```diff
@@ -101,7 +101,9 @@ export class Locator implements ILocator {
 
     public getWidestVisibleCellInColumn(columnIndex: number): number {
         const columnCellSelector = this.getColumnCellSelector(columnIndex);
-        const columnHeaderAndBodyCells = this.tableElement.querySelectorAll(columnCellSelector);
+        const columnHeaderAndBodyCells = this.tableElement.querySelectorAll(columnCellSelector) as NodeListOf<
+            HTMLElement
+        >;
```
adidahiya: this change is no longer necessary

adidahiya:
@giladgray or @themadcreator can I get a quick +1? @tgreenwatts has already tested this and it works

packages/table/src/common/utils.ts
```diff
@@ -42,7 +44,7 @@ export const Utils = {
      * element's original className and any other classes passed in with variadic
      * arguments matching the `classNames` api.
      */
-    assignClasses<P extends IProps>(elem: React.ReactElement<P>, ...extendedClasses: ClassValue[]) {
+    assignClasses<P extends IProps>(elem: React.ReactElement<P>, ...extendedClasses: ClassValue[]): React.ReactNode {
```
rcchen: Why all the explicit return signatures?

adidahiya: I find it good hygiene for public-ish APIs. Not entirely relevant to the change at hand.

PR #2138, opened 2018-02-15
[Dialog] Change dialog header to h4

llorca:
Per https://github.com/palantir/blueprint/pull/2134#issue-169207095


giladgray:
@llorca you ok with this from a design perspective? it's not too big now?
image

adidahiya:
@giladgray looks fine to me

packages/core/src/components/dialog/_dialog.scss
```diff
@@ -107,7 +107,7 @@ $dialog-padding: $pt-grid-size * 2 !default;
     color: $pt-icon-color;
   }
 
-  h5 {
+  h4 {
```
adidahiya: it's not good practice to use a tag selector -- this becomes a CSS breaking change.

I think we should refactor this CSS to use a `.pt-dialog-header-title` class in _both_ in 1.x and 2.x to avoid this subtle break.

adidahiya: I opened https://github.com/palantir/blueprint/pull/2141

packages/core/src/components/dialog/dialog.tsx
```diff
@@ -115,7 +115,7 @@ export class Dialog extends AbstractPureComponent<IDialogProps, {}> {
         return (
             <div className={Classes.DIALOG_HEADER}>
                 <Icon icon={icon} iconSize={Icon.SIZE_LARGE} />
-                <h5>{title}</h5>
+                <h4>{title}</h4>
```
adidahiya: you need to add the selector here too.

giladgray: @llorca this

packages/core/src/components/dialog/dialog.tsx
```diff
@@ -115,7 +115,7 @@ export class Dialog extends AbstractPureComponent<IDialogProps, {}> {
         return (
             <div className={Classes.DIALOG_HEADER}>
                 <Icon icon={icon} iconSize={Icon.SIZE_LARGE} />
-                <h5>{title}</h5>
+                <h4 className={Classes.DIALOG_HEADER_TITLE}>{title}</h4>
```
giladgray: gotta merge 1.x into develop (and then into this branch) for `Classes` constant to exist

pkwi:
screen shot 2018-02-16 at 11 07 51 am

Getting this ^

giladgray:
yep @pkwi due to my comment above. gotta merge 1.x into develop to fix this PR.

adidahiya:
doing the merge here https://github.com/palantir/blueprint/pull/2154

PR #2166, opened 2018-02-21
[release/1.x] port @blueprintjs/docs-theme package

adidahiya:
- remove docs package
- port docs-theme package from develop
- upgrade to documentalist 1.0.0
- revert some typography changes to avoid #2167


packages/docs-theme/package.json
```diff
@@ -1,12 +1,12 @@
 {
-  "name": "@blueprintjs/docs",
-  "version": "1.3.1",
+  "name": "@blueprintjs/docs-theme",
+  "version": "1.0.0",
```
rcchen: will this also go into the `latest` channel?

adidahiya: yes, but I expect most sensible users to be using ranges dictated by numbers, not channel names

PR #2274, opened 2018-03-19
Tweak popover docs colors

llorca:
- better color for positioning buttons
- same style for inset elements


giladgray:
scrollbars look awkward on new examples frames:
image

llorca:
ack, ended up simplifying styles with no border/shadows, and solid background colors

PR #2288, opened 2018-03-21
[Docs] Tick message in monospace

llorca:
image


packages/docs-app/src/styles/_colors.scss
```diff
@@ -40,7 +40,7 @@ $swatch-copied-message: "#{$pt-icon-tick} Copied to clipboard";
   // override more specific selector above
   //stylelint-disable declaration-no-important
   &.docs-clipboard-copied .docs-clipboard-message::after {
-    font-family: $pt-font-family;
+    font-family: $pt-font-family-monospace, "Icons16", sans-serif;
```
giladgray: should not need `sans-serif` but whatever

llorca:
for posterity the sans-serif is because of this lint rule:
image

I could have disabled it but went for the sans-serif instead 🤷‍♂️

PR #2310, opened 2018-03-25
add react and react-dom as peer dependency for v2

yaodingyd:
If by any chances someone dumb like me skips this page "http://blueprintjs.com/docs/v2/#blueprint/whats-new" and still uses react 16.0.0, they have a nice way to know to upgrade to 16.2.0 when doing "npm install"

add peerDependency in package.json

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,6 +60,10 @@
     "stylelint-scss": "^2.5.0",
     "typescript": "~2.7.2"
   },
+  "peerDependencies": {
+    "react": "^16.2.0",
+    "react-dom": "^16.2.0"
+  },
   "engines": {
     "node": ">=6.1"
   },


palantirtech:
Thanks for your interest in [palantir/blueprint](https://github.com/palantir/blueprint), @yaodingyd! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

giladgray:
@yaodingyd this repo is a monorepo, meaning it contains many separate NPM [packages](https://github.com/palantir/blueprint/blob/develop/packages). as a result, the root package.json (which you've edited here) is marked `private` and is never published to NPM; it contains build dependencies for the entire repo (mostly typings).

the _correct_ place for this change (core package.json) already declares these `peerDependencies`:
https://github.com/palantir/blueprint/blob/2f08ded9bff03da77f42a8989a79c7c8d2ca095e/packages/core/package.json#L47-L51

any version of react 16 should be sufficient. did you have problems with 16.0 that were resolved by using 16.2?

yaodingyd:
Thanks! Did not notice the repo structure, my bad.

I think @blueprintjs/core’s Abstract Button use react  fragment which is
only in react 16.2.0. Thus React 16 has trouble rendering Button.

PR #2428, opened 2018-04-27
[Tag] Improve outline behavior

llorca:
- Add `tabIndex` on interactive tags only, for proper keyboard navigation
- Clean up focus styles


giladgray:
@llorca `TagInput` backspace & arrow key interactions look broken -- no focus state

packages/core/src/components/tag-input/_tag-input.scss
```diff
@@ -17,34 +21,47 @@ $ti-icon-padding-large: ($pt-input-height-large - $pt-icon-size-large) / 2;
...
     }
   }
 
   .#{$ns}-tag {
+    position: relative;
     // NOTE: in order to wrap long words, you must set explicit width on TagInput,
     // or use .#{$ns}-fill CSS class modifier.
     overflow-wrap: break-word;
+
+    &::before {
```
giladgray: not a fan of this special case, and even less a fan of not simply using `outline` / the default focus styles (which it used to before this PR)

packages/core/src/components/tag/tag.tsx
```diff
@@ -88,7 +88,7 @@ export class Tag extends React.PureComponent<ITagProps, {}> {
         ) : null;
 
         return (
-            <span {...htmlProps} className={tagClasses}>
+            <span {...htmlProps} className={tagClasses} tabIndex={interactive ? 0 : -1}>
```
giladgray: i think we want `undefined` in the false case (emphasis mine):

> A negative value (usually `tabindex="-1"`) means that the element **should be focusable, but should not be reachable** via sequential keyboard navigation. Mostly useful to create accessible widgets with JavaScript.
https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/tabindex

packages/core/src/components/tag-input/_tag-input.scss
```diff
@@ -17,34 +21,39 @@ $ti-icon-padding-large: ($pt-input-height-large - $pt-icon-size-large) / 2;
...
+    margin-right: $tag-input-icon-padding;
 
     > * {
-      margin-bottom: $ti-padding;
+      margin-bottom: $tag-input-padding;
     }
   }
 
   .#{$ns}-tag {
+    position: relative;
```
giladgray: don't think this chance is necessary. leftover from previous attempt?

packages/core/src/components/tag-input/_tag-input.scss
```diff
@@ -17,34 +21,39 @@ $ti-icon-padding-large: ($pt-input-height-large - $pt-icon-size-large) / 2;
...
     // NOTE: in order to wrap long words, you must set explicit width on TagInput,
     // or use .#{$ns}-fill CSS class modifier.
     overflow-wrap: break-word;
+
+    &.#{$ns}-active {
+      @include focus-outline(0);
```
giladgray: i kinda think we want to adjust outline offset on _all_ tags, not just in tag inputs:

image
(note how the outline touches the neighboring tags)

image

so this change should move back to core tag styles.

giladgray:
@llorca please merge latest master for the new build steps

packages/core/src/components/tag-input/_tag-input.scss
llorca: I don't think that's the right move, but did it anyway, Not sure what to do with outline design in general

packages/core/src/components/tag-input/_tag-input.scss
```diff
@@ -5,10 +5,14 @@
...
 
-$ti-padding: ($pt-input-height - $tag-height) / 2;
+$tag-input-padding: ($pt-input-height - $tag-height) / 2;
 
-$ti-icon-padding: ($pt-input-height - $pt-icon-size-standard) / 2;
-$ti-icon-padding-large: ($pt-input-height-large - $pt-icon-size-large) / 2;
+$tag-input-icon-padding: ($pt-input-height - $pt-icon-size-standard) / 2;
+$tag-input-icon-padding-large: ($pt-input-height-large - $pt-icon-size-large) / 2;
+
+$tag-input-outline-color: $pt-outline-color;
```
giladgray: remove this variable, use original name directly

packages/core/src/components/tag-input/_tag-input.scss
```diff
@@ -5,10 +5,14 @@
...
+$tag-input-outline-offset: 1px;
+$tag-input-outline-size: 2px;
```
giladgray: these two vars are unused; please remove

packages/core/src/components/tag/_common.scss
```diff
@@ -120,6 +121,7 @@ $tag-padding-large: ($tag-height-large - $pt-icon-size-large) / 2 !default;
       background-color: rgba($background-color, $opacity + 0.1);
     }
 
+    &.pt-active,
```
giladgray: `#{$ns}`

packages/core/src/components/tag/_common.scss
```diff
@@ -106,6 +106,7 @@ $tag-padding-large: ($tag-height-large - $pt-icon-size-large) / 2 !default;
       background-color: rgba($background-color, $opacity - 0.15);
     }
 
+    &.pt-active,
```
giladgray: `pt` &rArr; `#{$ns}`

packages/core/src/components/tag-input/_tag-input.scss
```diff
@@ -5,10 +5,14 @@
 @import "../forms/common";
 @import "../tag/common";
 
-$ti-padding: ($pt-input-height - $tag-height) / 2;
+$tag-input-padding: ($pt-input-height - $tag-height) / 2;
 
-$ti-icon-padding: ($pt-input-height - $pt-icon-size-standard) / 2;
-$ti-icon-padding-large: ($pt-input-height-large - $pt-icon-size-large) / 2;
+$tag-input-icon-padding: ($pt-input-height - $pt-icon-size-standard) / 2;
+$tag-input-icon-padding-large: ($pt-input-height-large - $pt-icon-size-large) / 2;
```
giladgray: can you add `!default` to all these (surviving) vars please?

packages/core/src/components/tag/_common.scss
```diff
@@ -27,6 +27,10 @@ $tag-padding-large: ($tag-height-large - $pt-icon-size-large) / 2 !default;
   color: $pt-dark-text-color;
   font-size: $pt-font-size-small;
 
+  &.#{$ns}-active {
```
giladgray: ah dang actually i think this should be `&:focus` here, just to adjust the outline-offset.

and bring back this selector in `.pt-tag-input .pt-tag` like you had earlier

😄

llorca: pushed another commit, hopefully I got it right this time!

PR #2474, opened 2018-05-07
[Callout] Flex callouts

### LINKED ISSUE #2472: Button inside of Callout will display icon in absolute position
### Bug report

- __Package version(s)__: 2.3.1

#### Steps to reproduce
```jsx
<Callout intent={Intent.WARNING} title={"Some title"}>
    <Button icon={IconNames.ADD} intent={Intent.WARNING}
                            text={"Click Me"} />
</Callout>
```

#### Actual behavior

Icon for Button will display in the upper left corner of the Callout along with the Callout icon

I attached Chrome styling output for reference

screen shot 2018-05-07 at 4 41 06 pm

screen shot 2018-05-07 at 4 37 53 pm

#### Expected behavior

Icon shows inside of button

#### Possible solution

Possible CSS attribute for the callout icon to differentiate it from the Button icon

llorca:
#### fixes #2472

- `Callout` now renders wrapper `<div>` around title and children. (CSS usage is unchanged)
- 🔥 remove `Classes.CALLOUT_ICON`.


packages/core/src/components/callout/_callout.scss
```diff
@@ -45,7 +45,7 @@ Styleguide callout
   &.#{$ns}-callout-icon {
     padding-left: ($pt-grid-size * 2) + $pt-icon-size-large;
 
-    .#{$ns}-icon {
+    > .#{$ns}-icon {
```
giladgray: might actually want to add `:first-child` here to avoid targeting potential icons in the callout text.

llorca:
@giladgray pushed a commit for flex callouts: achieved with one extra `<div>` wrapper, removed unnecessary classes and CSS

giladgray:
image

oops.

packages/core/src/components/callout/_callout.scss
```diff
@@ -42,15 +45,10 @@ Styleguide callout
...
-    .#{$ns}-icon {
-      position: absolute;
-      top: $pt-grid-size;
-      left: $pt-grid-size;
-      color: $pt-icon-color;
-    }
+  > .#{$ns}-icon {
+    flex: 0 0 $pt-icon-size-large;
+    margin-right: $pt-grid-size;
+    margin-left: -$pt-grid-size * 0.2;
```
giladgray: why this negative margin? does not seem necessary.

packages/core/src/components/callout/_callout.scss
```diff
@@ -29,10 +27,12 @@ Styleguide callout
   border-radius: $pt-border-radius;
   background-color: rgba($gray3, 0.15);
   width: 100%;
-  padding: $pt-grid-size ($pt-grid-size * 1.2) ($pt-grid-size * 0.9);
+  padding: $pt-grid-size;
```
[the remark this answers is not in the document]
giladgray: i noticed these but declared them changes, not regressions.

packages/core/src/components/callout/_callout.scss
```diff
@@ -8,10 +8,8 @@ Callout
 
 Markup:
 <div class="#{$ns}-callout {{.modifier}}">
-  <div>
-    <h4 class="#{$ns}-heading">Callout heading</h4>
-    Lorem ipsum dolor sit amet, consectetur adipisicing elit. Ex, delectus!
-  </div>
+  <h4 class="#{$ns}-heading">Callout heading</h4>
```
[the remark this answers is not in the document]
giladgray: ah yep CSS callouts without icons. do not like

giladgray:
hmm my official take here is that we should not change this CSS. the current approach works fine: either a CSS `.pt-icon-*` class or a`.pt-callout-icon` to make space for the icon. that extra class is added automatically by the component so no sweat on the user.

llorca:
discussed offline and added a CALLOUT_TEXT class for that wrapper thing. also tweaked callouts in our markdown docs. everything is now pixel perfect, whether it's the CSS API or the React component, with or without icon/heading etc 👍

packages/core/src/components/callout/_callout.scss
```diff
@@ -23,14 +25,17 @@ Styleguide callout
...
+  display: flex;
   border-radius: $pt-border-radius;
   background-color: rgba($gray3, 0.15);
   width: 100%;
   padding: $pt-grid-size ($pt-grid-size * 1.2) ($pt-grid-size * 0.9);
 
   // CSS API support
   &[class*="#{$ns}-icon-"] {
+    // no flex layout as icon is absolute positioned within padding-left.
+    display: block;
```
giladgray: this change is not necessary anymore, and it allows callouts with icons _and_ headings to avoid the `.pt-callout-text` in HTML. for instance, you missed a callout on the buttons page.

llorca: that's right, very cool that we don't need it in markdown. can't wait for MDX tho!!

llorca:
reverted all the .md changes as they're not needed

llorca:
whoops my bad, so we do need that wrapper in those markdown files, right? what did you mean when you said:
> it allows callouts with icons and headings to avoid the .pt-callout-text in HTML

llorca:
@giladgray ok, Markdown callouts should now look sane!

packages/core/src/components/callout/_callout.scss
```diff
@@ -23,14 +25,16 @@ Styleguide callout
...
-  position: relative;
+  display: flex;
   border-radius: $pt-border-radius;
   background-color: rgba($gray3, 0.15);
   width: 100%;
   padding: $pt-grid-size ($pt-grid-size * 1.2) ($pt-grid-size * 0.9);
 
   // CSS API support
   &[class*="#{$ns}-icon-"] {
+    display: block;
```
giladgray: look, this line amounts to a special case for our own docs usage of the CSS API, which is a major code smell. the CSS API functions quite differently when you add an icon, which is weird.

all this to say i'm still not sold on the value of this change, and i'm quite uncomfortable with this special case to avoid changing all our docs.

granted, if we used MDX or could somehow generate those callouts using the component then this would not be a problem.

llorca:
ok fine. you let me know how we fix things like #2472 without this CALLOUT_TEXT wrapper. and I'm still very much against a modifier like CALLOUT_ICON, it clearly is out of place when you look at the rest of the codebase

llorca:
reopening cause I'm still not convinced CALLOUT_TEXT is bad, and I'd like to address #2472. I don't really care what ends up in .md files

giladgray:
let's go back to the `> :first-child` approach from the first commit, but be sure to fix intents as well.

llorca:
gonna go for the :first-child solution from the beginning

PR #2488, opened 2018-05-11
Add 3.0 upgrade script; improve 2.0 script

JKillian:

**Changes to 2.0 script**:
* rename `IconClasses` when used in a NIS `visual`
* only warn about remaining uses of `IconClasses`, trying to replace it with invalid TS just ends badly

**New 3.0 script**:
* Same functions and help docs as 2.0. (Sidenote: I think we should've just done these in node.js instead of bash for more cross-platform compat, but too late now)
* Handles NIS `visual` -> `icon`
* Renames classes to `Classes.HEADING`

### DIFF
--- a/packages/core/scripts/upgrade-blueprint-3.0.0-rename
+++ b/packages/core/scripts/upgrade-blueprint-3.0.0-rename
@@ -1,0 +1,171 @@
+#!/bin/bash
+
+function usage() {
+    cat << EOF
+
+Blueprint 3.0.0 Upgrade Script
+
+Usage
+
+    $0 [--path=path] [--include=glob] [--exclude=glob] [--prefix=regexp]
+
+Description
+
+    This script finds instances of various renamed methods and properties and
+    replaces them in place. This includes renames for React props, css classes,
+    and enum constants.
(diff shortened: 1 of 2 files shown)


packages/core/scripts/upgrade-blueprint-3.0.0-rename
```diff
@@ -0,0 +1,171 @@
...
+function renameString() {
+    # Add simple word boundaries
+    rename $1 $2 "\\b$1\\b"
+}
+
+echo "
+
+Renames for @blueprintjs/core"
+renameProp visual icon
+renameString "Classes\.CALLOUT_TITLE" "Classes.HEADING"
```
JKillian: I didn't actually test these class renames, maybe I should before we merge?

llorca: we can always do follow ups and stuff, but let's make sure it works before merging yes

JKillian: tested manually, seems to work correctly 👍

PR #2491, opened 2018-05-14
[Spinner] Slightly thinner stroke

llorca:
Current is too large compared to large, and too small compared to small

Before:
image

After:
image


llorca:
I'm more attached to the thickness of the ring rather than the code rhythm. 8 doesn't even make sense when you think about it, the stroke is actually about 4px thick

PR #2536, opened 2018-05-22
[Core] MultiSlider component

jscheiny:
#### Checklist

- [x] Include tests
- [ ] Update documentation

---

This adds a very general `MultiRangeSlider` component that ultimately should be able to subsume the `Slider` and `RangeSlider` components.

Notes about the component API:
- Instead of taking a value/values the component takes a list of handles specified by props. Each handle has:
 - `value` - where it gets placed on the slider
 - `type` - specifies how the slider handle should be displayed
   - `"full"` - handle appears like it does in a `Slider` (default)
   - `"lower"` - handle appears like the first slider in a `RangeSlider`
   - `"upper"` - handle appears like the second slider in a `RangeSlider`
 - `trackIntent{Above,Below}` specifies the color of the track fill above and below (numerically) of the given handle.

Handles are always displayed in ascending numeric order and the values that are returned by `onChange` and `onRelease` are always sorted.

In addition, `MultiRangeSlider` takes a `defaultTrackIntent` prop which is used in the resolution of how to render the color of the track between two given handles. The way that this is resolved is as follows (and is one of my concerns about the API).

- If the left handle has an above intent, use that
- If the right handle has a below intent, use that
- If the component has a default intent, use that
- Otherwise don't color that section

**Important:** Note that there is a difference between the `NONE` intent and the absence of an intent entirely (`undefined`). The former will render a gray track between the two handles and the latter will fall-through to the next color in the hierarchy above.

I previously allowed the specification of colors instead of intents but that didn't seem very blueprint-y.

---

I have explicitly **not** written tests/docs for this yet, as I would first like to iterate on the API as I'm not convinced what I've written is particularly nice (worked for our use case but don't know if it generalizes well). Once we've iterated and nailed down an API I will gladly write tests and docs for this component.

I've written some questions in the code, but one additional question I have is should this component live in `labs` first to incubate before moving into `core`?

---

#### Not addressed in this PR (requires follow-up):

To keep this PR simple(r) and easy(er) to review, I have not made many changes to the existing slider components but in 1-2 follow up PRs I would like to:

1. Refactor `Slider` and `RangeSlider` to be simple wrappers around `MultiRangeSlider`
2. Remove `CoreSlider` and refactor into `MultiRangeSlider`

The second one is questionable, especially if there are plans to add a slider that does not fit into the multi range slider model.

---

#### Screenshots

Two tailed slider:

screen shot 2018-04-19 at 10 43 24 am

Disabled:

screen shot 2018-04-19 at 11 09 44 am

One tailed:

screen shot 2018-04-19 at 11 10 39 am

Weird:

screen shot 2018-04-19 at 11 12 27 am

Vertical:

screen shot 2018-04-19 at 11 14 13 am

In action:

mulit-range-slider


jscheiny:
Here's the lock/push interaction:

lock-push

The inner handles use the push interaction the outer handles use the lock interaction.

packages/core/src/components/slider/multiSlider.tsx
```diff
@@ -0,0 +1,245 @@
...
+        } else {
+            newValues[newIndex] = newValues[index];
+        }
+
+        return newValues;
+    }
+
+    private handleChange = (newValues: number[]) => {
+        const oldValues = this.getSortedHandles().map(handle => handle.value);
+        if (!Utils.arraysEqual(newValues, oldValues) && Utils.isFunction(this.props.onChange)) {
```
llorca: `Utils.isFunction` -> `safeInvoke`?

jscheiny:
Who knows, maybe someone will find this useful? Here's a range slider with upper and lower quartiles highlighted.

track-quartile-stops

jscheiny:
Update: The styles fixes I made to make track stops work broke the styles for `<Slider>` (but not the `<MultiSlider>` implementation for `<Slider>`). So instead of trying to get both to cooperate I went ahead and refactored `<Slider>`. To do that, I also broke the existing implementation of `<RangeSlider>`, so I refactored that one as well. I was thinking about doing this refactor in another PR but I think making the styles work for both implementations would've been more annoying.

packages/core/src/components/slider/rangeSlider.tsx
```diff
@@ -34,103 +36,22 @@ export interface IRangeSliderProps extends ICoreSliderProps {
...
-            // if the values are equal, return the handle that is *able* to move
-            // in the necessary direction.
-            return value < firstHandleValue ? firstHandle : secondHandle;
-        }
+    public render() {
+        const { value, className, ...props } = this.props;
+        return (
+            <MultiSlider {...props} className={classNames(className, Classes.RANGE_SLIDER)}>
+                <SliderHandle value={value[RangeIndex.START]} type="start" trackIntentAfter={Intent.PRIMARY} />
+                <SliderHandle value={value[RangeIndex.END]} type="end" trackIntentBefore={Intent.PRIMARY} />
```
giladgray: no need to set intent twice

packages/core/src/components/slider/rangeSlider.tsx
```diff
@@ -34,103 +36,22 @@ export interface IRangeSliderProps extends ICoreSliderProps {
...
-            return secondHandle;
-        } else {
-            // if the values are equal, return the handle that is *able* to move
-            // in the necessary direction.
-            return value < firstHandleValue ? firstHandle : secondHandle;
-        }
+    public render() {
+        const { value, className, ...props } = this.props;
+        return (
+            <MultiSlider {...props} className={classNames(className, Classes.RANGE_SLIDER)}>
```
giladgray: style: className prop should come last

packages/core/src/components/slider/slider.tsx
```diff
@@ -31,72 +34,32 @@ export interface ISliderProps extends ICoreSliderProps {
...
-        if (this.handle != null) {
-            this.handle.beginHandleMovement(event);
-        }
-    }
-
-    protected handleTrackTouch(event: React.TouchEvent<HTMLElement>) {
-        if (this.handle != null) {
-            this.handle.beginHandleTouchMovement(event);
-        }
+    private getHandler(callback: (value: number) => void) {
```
giladgray: not worth the render penalty here (new instances each invoke). just make two `handleChange` and `handleRelease` members.

packages/core/src/components/slider/sliderTrackStop.tsx
```diff
@@ -0,0 +1,18 @@
...
+ * Licensed under the terms of the LICENSE file distributed with this project.
+ */
+
+import * as React from "react";
+import { Intent } from "../../common";
+
+export interface ISliderTrackStopProps {
+    value: number;
+    trackIntentAfter?: Intent;
+    trackIntentBefore?: Intent;
```
giladgray: thoughts on simply `intentAfter` and `intentBefore`? i feel like "before/after" suggests that it affects the track.

packages/core/src/components/slider/sliderTrackStop.tsx
```diff
@@ -0,0 +1,18 @@
...
+}
+
+export class SliderTrackStop extends React.PureComponent<ISliderTrackStopProps> {
```
giladgray: i feel like we could easily fold this into `SliderHandle`. 
thoughts on `SliderHandleType.HIDDEN` or `SliderInteractionKind.NONE`?

packages/core/src/components/slider/multiSlider.tsx
```diff
@@ -0,0 +1,272 @@
...
+        return (newValue: number) => {
+            Utils.safeInvoke(callback, this.getNewHandleValues(newValue, index));
+        };
+    };
+
+    private getNewHandleValues(newValue: number, oldIndex: number) {
+        const oldValues = this.sortedHandleProps.map(handle => handle.value);
+        const start = oldValues.slice(0, oldIndex);
+        const end = oldValues.slice(oldIndex + 1);
+        const newValues = [...start, newValue, ...end];
```
giladgray: wait why not use `array.splice()`?

but actually, does insertion index really matter if you're just gonna sort again?

packages/core/src/components/slider/multiSlider.tsx
```diff
@@ -0,0 +1,272 @@
...
+    }
+
+    private findFirstLockedHandleIndex(startIndex: number, endIndex: number): number | undefined {
+        const inc = startIndex < endIndex ? 1 : -1;
+        for (let index = startIndex + inc; index !== endIndex + inc; index += inc) {
+            if (this.sortedHandleProps[index].interactionKind !== SliderHandleInteractionKind.PUSH) {
+                return index;
+            }
+        }
+        return undefined;
```
giladgray: `return -1` as is the convention for `indexOf`

packages/core/src/components/slider/multiSlider.tsx
```diff
@@ -0,0 +1,272 @@
...
+        const endOffset = formatPercentage(1 - endOffsetRatio);
+
+        const style: React.CSSProperties = this.props.vertical
+            ? { bottom: startOffset, top: endOffset, left: 0 }
+            : { left: startOffset, right: endOffset, top: 0 };
+
+        const classes = classNames(Classes.SLIDER_PROGRESS, intentClass(intent), {
+            [Classes.START]: start.type === SliderHandleType.START,
+            [Classes.END]: end.type === SliderHandleType.END,
+            [`${Classes.SLIDER_PROGRESS}-empty`]: intent === Intent.NONE,
```
giladgray: this class is not essential. intent can override background color

jscheiny: This was previously required, need to rethink if its still needed with the removal of other slider implementations.

packages/core/src/components/slider/sliderTrackStop.tsx
jscheiny: I like it!

packages/core/src/components/slider/sliderTrackStop.tsx
jscheiny: Will explore the possibility, I think it could create certain code weirdnesses but I'm not positive.

packages/core/src/components/slider/multiSlider.tsx
jscheiny: Not really, we just need to replace the value at `oldIndex` with `newValue`, refactored it in a nice way I think.

packages/core/src/components/slider/sliderTrackStop.tsx
giladgray: cool can leave that for next time

giladgray: like, after this round of review. still in 3.0

packages/core/src/components/slider/multiSlider.tsx
```diff
@@ -0,0 +1,446 @@
...
+     * @default 0
+     */
+    min?: number;
+
+    /**
+     * Whether a solid bar should be rendered on the track between current and initial values,
+     * or between handles for `RangeSlider`.
+     * @default true
+     */
+    showTrackFill?: boolean;
```
giladgray: @jscheiny is this prop respected by your impl?

jscheiny: Yes, here: https://github.com/palantir/blueprint/pull/2536/files#diff-b1bce65425b54a1075cd383becfa1663R258

giladgray:
example is broken. will fix shortly

packages/core/src/components/slider/handleProps.tsx
```diff
@@ -0,0 +1,60 @@
...
+    NONE: "none" as "none",
+};
+export type HandleInteractionKind = typeof HandleInteractionKind[keyof typeof HandleInteractionKind];
+
+export interface IHandleProps extends IProps {
+    /** Numeric value of this handle. */
+    value: number;
+
+    /** Whether this handle is non-interactive. */
+    disabled?: boolean;
```
jscheiny: I don't think this prop is ever read.

Issue #1745, opened 2017-10-20
Standalone icons don't fully support dark theme with intents


adidahiya:
- intents only work on standard icons, not large (at least in docs)
- intent class selectors don't actually work in practice; color overrides to be made overly explicit to override this rule: https://github.com/palantir/blueprint/blob/39f142f33009910fefe3f8bee5ec52aa09a82f83/packages/core/src/_typography.scss#L182
  - this rule does not get applied in the docs site (sounds like a build system bug)

PR #2541, opened 2018-05-22
[Core] support dark intent colors on icons

llorca:
#### Fixes https://github.com/palantir/blueprint/issues/1745

also added a intent select in the icon example, so it's easy to test out


andio57:
Dark,darka,darkla,darkala,darkja=1compress_environment_elementeri+android

PR #2642, opened 2018-07-05
[Core] Add panel stack component

kadhirvelm:
#### Bugs
There's a strange bug that if you smash the `Open new panel` button rapidly, it spits out the panels incorrectly in the `onOpen` callback handler.
screen shot 2018-07-05 at 10 32 57 am
The panel stack isn't really meant to be used like that, but either way I don't think the component should be merged with a bug like that. Wanted to get the PR up in the meantime and get some feedback while I continue to figure this out.

#### Checklist

- [X] Include tests
- [X] Update documentation

Adds the panel stack component to blueprint core:
screen shot 2018-07-05 at 10 35 32 am

The core component was already reviewed, so best to focus on the example, documentation, and that overall the component was properly integrated into blueprint.

Also I've been having issues getting the transition to work correctly, so any thoughts on how to fix that nicely would be very appreciated.

#### GIF

panelstack


palantirtech:
Thanks for your interest in [palantir/blueprint](https://github.com/palantir/blueprint), @kadhirvelm! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

packages/core/src/components/panel-stack/panel-stack.md
```diff
@@ -0,0 +1,28 @@
...
+    <h4 class="@ns-heading">A note about closing a panel</h4>
+    Try and avoid calling `closePanel()` as much as possible since there is a back button in the header.
+</div>
+
+@reactExample PanelStackExample
+
+@## Props
+
+The `Panel Stack` component is available in the __@blueprintjs/core__ package.
+Make sure to review the [general usage docs for JS components](#blueprint.usage).
```
giladgray: remove this paragraph, we don't do this anymore.

packages/core/src/components/panel-stack/panel-stack.scss
```diff
@@ -0,0 +1,99 @@
+// Copyright 2018 Palantir Technologies, Inc. All rights reserved.
+
+@import "../../common/variables";
+@import "~@blueprintjs/core/src/common/react-transition";
+
+.#{$ns}-panel-stack {
+  position: relative;
+  overflow: hidden;
+
+  .push-#{$ns}-panel-stack-transition-enter {
```
giladgray: $ns must be the first thing in the class. `#{$ns}-panel-stack-transition-push-*`

packages/core/src/components/panel-stack/panel-stack.scss
```diff
@@ -0,0 +1,99 @@
...
+    transform: translateX(100%);
+  }
+
+  .push-#{$ns}-panel-stack-transition-enter-active {
```
giladgray: TODO: use the mixin

packages/core/src/components/panel-stack/panelStack.tsx
```diff
@@ -0,0 +1,142 @@
...
+}
+
+export interface IPanelOptions {
+    /** The title to be displayed above this panel. It is also used as the back button for any panel opened by this panel. */
+    title: string;
+}
+
+export interface IPanelStackProps extends IProps {
+    /** The panel to show when initially rendering the component */
+    initialPanel: IPanel;
```
giladgray: must this be required?

packages/core/src/common/classes.ts
```diff
@@ -167,6 +167,15 @@ export const OVERLAY_INLINE = `${OVERLAY}-inline`;
...
 export const OVERLAY_SCROLL_CONTAINER = `${OVERLAY}-scroll-container`;
 
+export const PANELSTACK = `${NS}-panel-stack`;
+export const PANELSTACK_CHEVRON = `${PANELSTACK}-chevron`;
+export const PANELSTACK_HEADER = `${PANELSTACK}-header`;
+export const PANELSTACK_HEADER_BACK = `${PANELSTACK}-header-back`;
+export const PANELSTACK_HEADER_SPACER = `${PANELSTACK}-header-spacer`;
+export const PANELSTACK_HEADER_TITLE = `${PANELSTACK}-header-title`;
+export const PANELSTACK_TRANSITION = `${PANELSTACK}-transition`;
+export const PANELSTACK_VIEW = `${PANELSTACK}-view`;
```
giladgray: my gut says that roughly half of these classes should not be necessary through clever use of flex, but i'll have to dig into the code to figure out which ones.

packages/core/src/components/panel-stack/panel-stack.scss
```diff
@@ -0,0 +1,99 @@
...
+  box-shadow: 0 1px $pt-divider-black;
+  height: 30px;
+
+  .#{$ns}-panel-stack-header-back {
+    display: flex;
+    align-items: center;
+    cursor: pointer;
+    padding: 0 5px;
+    text-transform: uppercase;
+    letter-spacing: 0.5px;
```
giladgray: whoa, prefer not messing with text-transform or letter-spacing, they're too design-specific

packages/core/src/components/panel-stack/panel-stack.scss
```diff
@@ -0,0 +1,99 @@
...
+  flex-shrink: 0;
+  z-index: $pt-z-index-content + 2;
+  box-shadow: 0 1px $pt-divider-black;
+  height: 30px;
+
+  .#{$ns}-panel-stack-header-back {
+    display: flex;
+    align-items: center;
+    cursor: pointer;
+    padding: 0 5px;
```
giladgray: compute from `$pt-grid-size` variable where ever possible

packages/core/src/components/panel-stack/panel-stack.scss
```diff
@@ -0,0 +1,99 @@
...
+    text-transform: uppercase;
+    letter-spacing: 0.5px;
+    color: $pt-text-color-muted;
```
giladgray: 🌔  needs dark theme (also hover state)

packages/core/src/components/panel-stack/panel-stack.scss
```diff
@@ -0,0 +1,99 @@
...
+  .#{$ns}-panel-stack-header-spacer {
+    display: flex;
+    flex: 1;
+    align-items: stretch;
+  }
+
+  .#{$ns}-panel-stack-header-title {
+    margin: 0 10px;
+    // aligns title with back button
+    line-height: 28px;
```
giladgray: um whoa this should be computed. no magic numbers

packages/core/src/components/panel-stack/panel-stack.scss
```diff
@@ -0,0 +1,99 @@
...
+  }
+
+  .#{$ns}-panel-stack-header-spacer {
+    display: flex;
+    flex: 1;
+    align-items: stretch;
+  }
+
+  .#{$ns}-panel-stack-header-title {
+    margin: 0 10px;
```
giladgray: `$pt-grid-size`

packages/core/src/components/panel-stack/panel-stack.scss
```diff
@@ -0,0 +1,99 @@
...
+    // aligns title with back button
+    line-height: 28px;
+    color: $pt-text-color;
```
giladgray: this is the default, why must it be reset? if so, needs dark theme

packages/core/src/components/panel-stack/panel-stack.scss
```diff
@@ -0,0 +1,99 @@
...
+    color: $pt-text-color-muted;
+    font-size: $pt-font-size-small;
+    font-weight: 600;
+
+    &:hover {
+      color: $pt-text-color;
+    }
+
+    .#{$ns}-panel-stack-chevron {
+      font-size: $pt-font-size;
```
giladgray: font-size has no effect on svg icons. please remove.

packages/core/src/components/panel-stack/panel-stack.scss
```diff
@@ -0,0 +1,99 @@
...
+
+.#{$ns}-panel-stack-view {
+  display: flex;
+  flex-direction: column;
+  position: absolute;
+  top: 0;
+  right: 0;
+  bottom: 0;
+  left: 0;
+  background-color: $white;
```
giladgray: needs dark theme

packages/core/src/components/panel-stack/panel-stack.scss
```diff
@@ -0,0 +1,99 @@
...
+}
+
+.#{$ns}-panel-stack-view {
+  display: flex;
+  flex-direction: column;
+  position: absolute;
+  top: 0;
+  right: 0;
+  bottom: 0;
+  left: 0;
```
giladgray: `@include position-all(absolute, 0)`

packages/core/src/components/panel-stack/panelHeader.tsx
```diff
@@ -0,0 +1,44 @@
...
+            </div>
+        );
+    }
+
+    private maybeRenderBack() {
+        if (this.props.back === undefined) {
+            return null;
+        }
+        return (
+            <div className={Classes.PANELSTACK_HEADER_BACK} onClick={this.props.back.onClick}>
```
giladgray: 🎨 i feel like we should just use a small minimal button here instead of all this custom style

packages/core/src/components/panel-stack/panelStack.tsx
invliD: @giladgray I think it must. The only way to open a panel (after the initial panel becomes the stack) is from a panel. If there are no more panels, the stack will stay empty, forever. I don't think there's a valid use case for that.

giladgray: oh yeah that makes a lot of sense. probably worth calling that out in some usage docs.

kadhirvelm:
screen shot 2018-07-10 at 11 45 00 am

kadhirvelm:
So I tried to add the `react-transition-phase` mixin, but I'm definitely not using it right. Could use some help.

But on the plus side, changing the duration to `pt-transition-duration` fixed the bug mentioned above!

packages/core/src/components/panel-stack/panelHeader.tsx
kadhirvelm: Tried to remove whatever I could, now it's just flex + cursor + color

packages/core/src/components/panel-stack/panelStack.tsx
```diff
@@ -0,0 +1,142 @@
...
+    /** Whether the stack is currently animating the push or pop of a panel. */
+    direction: "push" | "pop";
+    /** The current stack of panels. The first panel in the stack will be displayed. */
+    stack: IPanel[];
+}
+
+export class PanelStack extends React.PureComponent<IPanelStackProps, IPanelStackState> {
+    public constructor(props: IPanelStackProps, context?: any) {
+        super(props, context);
+        this.state = {
```
giladgray: pull this out to member, remove constructor. you can access `this.props` in member definitions.

packages/core/src/components/panel-stack/panelStack.tsx
```diff
@@ -0,0 +1,142 @@
...
+        const { stack } = this.state;
+        const [activePanel, previousPanel] = stack;
+        return (
+            <CSSTransition
+                classNames={`${Classes.PANELSTACK_TRANSITION}-${this.state.direction}`}
+                key={stack.length}
+                timeout={{
+                    enter: 400,
+                    exit: 400,
+                }}
```
giladgray: `timeout={400}`?

packages/core/src/components/panel-stack/panelStack.tsx
```diff
@@ -0,0 +1,142 @@
...
+        );
+    }
+
+    private renderCurrentPanel() {
+        const { stack } = this.state;
+        const [activePanel, previousPanel] = stack;
+        return (
+            <CSSTransition
+                classNames={`${Classes.PANELSTACK_TRANSITION}-${this.state.direction}`}
+                key={stack.length}
```
giladgray: this does not feel safe to use as a key. the size of the stack is not a unique identifier of the panel. not really sure what to do tho, needs investigation.

packages/core/src/components/panel-stack/panelStack.tsx
```diff
@@ -0,0 +1,142 @@
...
+                timeout={{
+                    enter: 400,
+                    exit: 400,
+                }}
+            >
+                <PanelView
+                    key={stack.length}
```
giladgray: no need for key here, it's not in an array.

packages/core/src/components/panel-stack/panelStack.tsx
```diff
@@ -0,0 +1,142 @@
...
+    private closePanel = (stackSize: number) => {
+        this.setState(state => {
+            const allowed =
+                // Make sure this function only works for the panel it was issued to,
+                // otherwise rapidly clicking the back link will close multiple panels.
+                state.stack.length === stackSize &&
+                // Prevent the last panel from being closed.
+                state.stack.length > 1;
+            if (allowed && this.props.onClose !== undefined) {
+                this.props.onClose(state.stack[0]);
```
giladgray: `Utils.safeInvoke` instead of undefined check

packages/core/src/components/panel-stack/panelHeader.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+import * as Classes from "../../common/classes";
+
+export interface IBackButton {
+    title: string;
+    onClick(): void;
+}
+
+export interface IPanelHeaderProps {
+    back?: IBackButton;
```
giladgray: prefer to avoid objects as props as it often breaks the pure behavior by creating new objects every render ([you do this below](#diff-e8ab461dbd0d083bb148b0cb65f9dac2R21)). spell them out as `backTitle` and `onBackClick`.

packages/core/src/components/panel-stack/panelHeader.tsx
```diff
@@ -0,0 +1,44 @@
...
+    title: string;
```
giladgray: why not use `children` for this? `<PanelHeader>content</PanelHeader>`

packages/core/src/components/panel-stack/panelStack.tsx
invliD: not sure about transition group 2.x, but in 1.x the key was required for their bookkeeping, not react's. Maybe it needs to be on the transition now?

packages/core/src/components/panel-stack/panelStack.tsx
```diff
@@ -0,0 +1,142 @@
...
+    }
+
+    private closePanel = (stackSize: number) => {
+        this.setState(state => {
+            const allowed =
+                // Make sure this function only works for the panel it was issued to,
+                // otherwise rapidly clicking the back link will close multiple panels.
+                state.stack.length === stackSize &&
+                // Prevent the last panel from being closed.
+                state.stack.length > 1;
```
invliD: this can be refactored to just `if (condition) { return null }`
I initially wrote it this way because I had problems with `setState` typings, but returning `null` should work and allows you to remove the `allowed` variable entirely.

packages/core/src/components/panel-stack/panelHeader.tsx
```diff
@@ -0,0 +1,53 @@
...
+import { Text } from "../text/text";
+
+import * as Classes from "../../common/classes";
+
+export interface IPanelHeaderProps {
+    /**
+     * The name of the previous panel header in the stack. Will not render unless an onBackClick
+     * handler is defined.
+     */
+    backTitle?: string;
```
kadhirvelm: Oh hum, should this ever be undefined? [title](https://github.com/palantir/blueprint/pull/2642/files#diff-8a6652a4df13b0fb84b0d08566c49beaR37)

Should we allow a panel to be defined without a title?

packages/core/src/components/panel-stack/panelStack.tsx
```diff
@@ -0,0 +1,132 @@
...
+        const [activePanel, previousPanel] = stack;
+        return (
+            <CSSTransition
+                classNames={`${Classes.PANELSTACK_TRANSITION}-${this.state.direction}`}
+                key={activePanel.title + "_" + stack.length}
+                timeout={100}
+            >
+                <PanelView
+                    key={stack.length}
+                    panelClassName={this.props.className}
```
giladgray: @kadhirvelm why is this prop passed down here? it gets rendered on two elements, which seems undesirable.

PR #2704, opened 2018-07-18
Fix path to docs app

JKillian:
Fix the broken localhost link to the docs app in the docs app readme


JKillian:
Seems like a test failure flake, no way this change should've broke things...

PR #2760, opened 2018-08-04
match event handler prop name with code and docs

dhenson02:

Corrected example of ResizeSensor API to use `onResize` vs. `onChange` so it matches the surrounding instructions and the imported class def.

The example code's usage of `onResize` compared with expectations in `resizeSensor.tsx`.


palantirtech:
Thanks for your interest in [palantir/blueprint](https://github.com/palantir/blueprint), @dhenson02! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

PR #2761, opened 2018-08-04
Fix tabs querySelector on jsdom

mizchi:
blueprint's `tabs` rendering fail on jsdom. It may  happen on browser.

`document.querySelector` returns `null` if document.body has no matched selector.

We have to consider this.

```
> document.querySelector('missing-selector')
null
```

#### Screenshot

My test case screenshot


palantirtech:
Thanks for your interest in [palantir/blueprint](https://github.com/palantir/blueprint), @mizchi! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

PR #2802, opened 2018-08-14
Tabs overflow

qcharlieshi:
#### Checklist
Just need to fix the tests that are broken in addition to making any more changes to this which I forsee will happen.

- [ ] Fix Tests
- [ ] Update documentation

Got PR up to quality, fixed linting and typescript errors. Made some of the suggested changes. Others will require some discussion

-What overflowListProps should we enable?
-If we can't grab the overflowList state, how should we find how which elements have been determined to be part of the overflow.
-Check typescript types, I added two props to tabs which may not be desirable on your end.
-Not sure if we can make it so they can determine how many tabs to show, but we could adjust tolerance of the overflowList, adjusting this tolerance number would in turn make more or less tabs disappear. (line 198 of overflowList has this, (this.spacer.getBoundingClientRect().width < 1), the one could be adjusted to allow more or less tabs but y'all might not want that value to be adjustable by the end user.

#### Screenshot

http://recordit.co/ikXWJINHMN


packages/core/src/components/tabs/tabs.tsx
```diff
@@ -98,13 +109,19 @@ export class Tabs extends AbstractPureComponent<ITabsProps, ITabsState> {
     public static displayName = `${DISPLAYNAME_PREFIX}.Tabs`;
 
     private tablistElement: HTMLDivElement;
+    private overflowElement: React.Component<ITabsProps, IOverflowListState<any>>;
```
giladgray: use the actual component type. `private overflowElement: OverflowList`

packages/core/src/components/tabs/tabs.tsx
```diff
@@ -98,13 +109,19 @@ export class Tabs extends AbstractPureComponent<ITabsProps, ITabsState> {
...
     private refHandlers = {
+        overflowList: (tabElement: React.Component<ITabsProps, IOverflowListState<any>>) =>
+            (this.overflowElement = tabElement),
         tablist: (tabElement: HTMLDivElement) => (this.tablistElement = tabElement),
     };
 
     constructor(props?: ITabsProps) {
         super(props);
         const selectedTabId = this.getInitialSelectedTabId();
+        this.overflowElement = null;
```
giladgray: not necessary. remove

packages/core/src/components/tabs/tabs.tsx
```diff
@@ -98,13 +109,19 @@ export class Tabs extends AbstractPureComponent<ITabsProps, ITabsState> {
...
+
+        this.overflowRenderer = this.overflowRenderer.bind(this);
```
giladgray: no! use fat arrows in the definition.

`private overflowRenderer = (args) => // bound`

packages/core/src/components/tabs/tabs.tsx
```diff
@@ -141,15 +158,70 @@ export class Tabs extends AbstractPureComponent<ITabsProps, ITabsState> {
                     role="tablist"
                 >
                     {tabIndicator}
-                    {tabTitles}
+                    {this.props.overflow ? (
+                        <OverflowList
+                            // @ts-ignore
```
giladgray: do not ignore compiler errors, fix them!

packages/core/src/components/tabs/tabs.tsx
```diff
@@ -141,15 +158,70 @@ export class Tabs extends AbstractPureComponent<ITabsProps, ITabsState> {
...
+                            items={tabTitles}
+                            collapseFrom={"start"}
+                            observeParents={true}
```
giladgray: prefer prop defaults. do not change this; let the user override with `overflowListProps` _if needed_.

packages/core/src/components/tabs/tabs.tsx
```diff
@@ -141,15 +158,70 @@ export class Tabs extends AbstractPureComponent<ITabsProps, ITabsState> {
...
+                    } else {
+                        return item;
+                    }
+                })}
+            </Menu>
+        );
+
+        return (
+            <Popover content={overflowMenu}>
+                <span style={{ lineHeight: "30px", cursor: "pointer" }}>Overflow</span>
```
giladgray: gonna need to customize this.

packages/core/src/components/tabs/tabs.tsx
```diff
@@ -160,20 +232,22 @@ export class Tabs extends AbstractPureComponent<ITabsProps, ITabsState> {
...
-        } else if (prevState.selectedTabId != null) {
-            // comparing React nodes is difficult to do with simple logic, so
-            // shallowly compare just their props as a workaround.
-            const didChildrenChange = !Utils.arraysEqual(
-                this.getTabChildrenProps(prevProps),
-                this.getTabChildrenProps(),
-                Utils.shallowCompareKeys,
-            );
-            if (didChildrenChange) {
+        window.requestAnimationFrame(() => {
```
giladgray: this should go at the lowest level, namely around `this.moveSelectionIndicator();`, not around the entire if blocks. especially since state could actually change between the function body and the animation frame.

giladgray: actually, it should probably go _inside_ `this.moveSelectionIndicator();` since we seem to do it every time now.

packages/core/src/components/tabs/tabs.tsx
```diff
@@ -265,15 +339,40 @@ export class Tabs extends AbstractPureComponent<ITabsProps, ITabsState> {
 
         const tabIdSelector = `${TAB_SELECTOR}[data-tab-id="${this.state.selectedTabId}"]`;
         const selectedTabElement = this.tablistElement.querySelector(tabIdSelector) as HTMLElement;
+        // Check if there are any overflow elementsR
+        const isOverflowing = this.overflowElement && this.overflowElement.state.overflow.length;
```
giladgray: 🙅‍♂️ we must be able to do this without reaching _into_ another component's internals. if it's truly unavoidable then we should provide a hook via a new prop. 

treat the insides of other components as black boxes, much like you treat your friends 😄

packages/core/src/components/tabs/tabs.tsx
```diff
@@ -265,15 +339,40 @@ export class Tabs extends AbstractPureComponent<ITabsProps, ITabsState> {
...
+        };
+
+        // Set indicator below tab or below overflowList element
+        if (selectedTabElement !== null) {
+            const { clientHeight, clientWidth, offsetLeft, offsetTop } = selectedTabElement;
+            indicatorWrapperStyle = indicatorWrapperCreator(clientHeight, clientWidth, offsetLeft, offsetTop);
+        } else if (isOverflowing && selectedTabElement === null) {
+            // Find the overflow dropdown span
+            // @ts-ignore
+            const overflowHTMLElement = this.overflowElement.element;
```
giladgray: avoid reaching inside others.

packages/core/src/components/tabs/tabs.tsx
giladgray: (see #2505 for context)

murtagha:
qqet8jvcxutgxt0rg4rkf379cnx5hc799uzaamena3

packages/core/src/components/tabs/tabs.tsx
qcharlieshi: It might be unavoidable, I think state has been updated but the actually element has not rerendered to reflect that.

packages/core/src/components/tabs/tabs.tsx
qcharlieshi: it's passed in as props, I figured there should be a default one which could be overwritten but we can just move it into the examples instead, they are both required props.

packages/core/src/components/tabs/tabs.tsx
qcharlieshi: I suppose we can grab it using querySelector on the tablist element, either tab or a public getter method? I'll make the changes you suggested here and see if it works with the querySelector

qcharlieshi:
@giladgray I made all the changes and used the queryselector to grab the dom element instead of inspecting the component state, I still need to fix the tests and update some of the documentation which ill do if you think the functionality and implementation are fine. I also added those two new props to the tab interface, not too sure if that's okay but I didn't see a comment there.

giladgray:
@qcharlieshi before tests and docs, please fix the compile errors. seems you've got some messed up imports.

qcharlieshi:
@giladgray my b, we should be good now

giladgray:
@qcharlieshi well now you can do tests and docs :D

packages/core/src/components/tabs/tabs.tsx
```diff
@@ -141,15 +157,72 @@ export class Tabs extends AbstractPureComponent<ITabsProps, ITabsState> {
...
                 >
                     {tabIndicator}
-                    {tabTitles}
+                    {this.props.overflow ? (
+                        <OverflowList
+                            items={tabTitles}
+                            visibleItemRenderer={this.visibleItemsRenderer}
+                            ref={this.refHandlers.overflowList}
+                            {...this.props.overflowListProps}
+                            overflowRenderer={this.overflowRenderer}
```
giladgray: move this line above the `...` spread so the user can override the default renderer. then you won't have to do crazy logic in there.

packages/core/src/components/tabs/tabs.tsx
```diff
@@ -141,15 +157,72 @@ export class Tabs extends AbstractPureComponent<ITabsProps, ITabsState> {
...
+                ) : (
+                    <Popover content={overflowMenu}>
+                        <span style={{ lineHeight: "30px", cursor: "pointer" }}>Overflow</span>
+                    </Popover>
+                )}
+            </div>
+        );
+    };
+
+    public visibleItemsRenderer = (item: React.ReactChild) => {
```
giladgray: `visibleItemRenderer` (no `s`, one item at a time)

packages/core/src/components/tabs/tabs.tsx
```diff
@@ -141,15 +157,72 @@ export class Tabs extends AbstractPureComponent<ITabsProps, ITabsState> {
...
+        return item;
+    };
+
     public componentDidMount() {
-        this.moveSelectionIndicator();
+        // requires overflowList to render in order for it to figure out what elements are overflowing
+        window.requestAnimationFrame(() => {
+            this.moveSelectionIndicator();
+        });
```
giladgray: please move the `rAF` call _inside_ `this.moveSelectionIndicator()` because you do it every time now.

packages/core/src/components/tabs/tabs.tsx
```diff
@@ -265,15 +342,52 @@ export class Tabs extends AbstractPureComponent<ITabsProps, ITabsState> {
...
+        const overflowElement = this.tablistElement.querySelector(OVERFLOW_SELECTOR) as HTMLElement;
+        const overflowChildren = overflowElement && overflowElement.children;
+
+        // Check if there are any overflow elements
+        let isOverflowing = false;
+        if (overflowChildren) {
+            for (const element in overflowChildren) {
+                if (
+                    overflowChildren.hasOwnProperty(element) &&
+                    overflowChildren[element].className === OVERFLOW_CONTENT_CLASS
```
giladgray: this is super brittle and (i believe) would break immediately if a user writes their own `overflowRenderer`. we need a better solution.

packages/docs-app/src/examples/core-examples/tabsExample.tsx
```diff
@@ -66,24 +94,32 @@ export class TabsExample extends React.PureComponent<IExampleProps, ITabsExample
...
-                    vertical={this.state.vertical}
-                >
-                    <Tab id="rx" title="React" panel={<ReactPanel />} />
-                    <Tab id="ng" title="Angular" panel={<AngularPanel />} />
-                    <Tab id="mb" title="Ember" panel={<EmberPanel />} />
-                    <Tab id="bb" disabled={true} title="Backbone" panel={<BackbonePanel />} />
-                    <Tabs.Expander />
-                    <InputGroup className={Classes.FILL} type="text" placeholder="Search..." />
-                </Tabs>
+                <div style={{ width: `${this.state.overflowWidth}%`, marginRight: "auto" }}>
```
giladgray: this makes for some awful panel visuals. i would make only the top example (the `Navbar`) overflow because it has no panels and leave this one untouched.
image

giladgray:
#### [preview](https://31-98603600-gh.circle-artifacts.com/0/home/circleci/project/packages/docs-app/dist/index.html#core/components/tabs)

packages/core/src/components/tabs/tabs.tsx
qcharlieshi: Agreed, that's why I have a default overflowRenderer for the tabs that takes in the custom renderer a user wants to use from props. Everything renders out under the overflow content class which is just a class identifier that has no css. This happens regardless if they used the default or decided to write their own. This allows that loop to always find out if the overflowList has considered something to be overflowing and dropped it off to that renderer. I think this was the best approach based on what I saw in the codebase though you would know better. The other way I was thinking of grabbing that information is a public getter of some kind in the overflowList. Let me know if you can think of something else that's more appropriate for how y'all have architected things.

qcharlieshi:
@giladgray appreciate the comments, I think there's one main thing holding this up other than some cleanup, docs, tests etc... and it has to do with how we find out the state of the overflowList. Obviously  grabbing the state from the ref breaks encapsulation, so i went about wrapping all the overflow elements in a class. I think there's no real way of knowing how the user would customize the overflowRenderer so you need to wrap the whole thing up so you can identify what's overflowing and what's not. I left some more details in the comments. let me know, I'll go ahead and clean up the other things you identified.

packages/docs-app/src/examples/core-examples/tabsExample.tsx
qcharlieshi: It does technically need some sort of panel so the didChildrenChange in the componeneDidUpdate can figure out the panel width changed. So if you don't have a panel of any sort the animation of the indicator would not work when you resize the component. Though we can add something to make it work without the panels, would just need to cache the previous widths somewhere to compare

giladgray:
@qcharlieshi could we add a prop or two on `OverflowList` to solve this problem? perhaps an `onOverflow` callback?

qcharlieshi:
@giladgray I tried the onOverflow callback, there's some trickiness involved in making sure it doesn't loop infinitely since the two states will impact each other (overflow then calculate movementIndicator which triggers a rerender in the child of tabs aka the overflowList. Let me know what you think, I still like utilizing the css selectors better, I don't believe any passing in a overflowRenderer will cause any issues, it's simply wrapped in a div with no css properties.

qcharlieshi:
@giladgray Just checking in to see if there's anything I can do to help push this along, no rush though

packages/core/src/components/overflow-list/overflowList.tsx
```diff
@@ -161,6 +167,14 @@ export class OverflowList<T> extends React.PureComponent<IOverflowListProps<T>,
         entries.forEach(entry => this.previousWidths.set(entry.target, entry.contentRect.width));
     };
 
+    private callOnOverflow = (prevState: IOverflowListState<any>) => {
+        if (prevState.overflow.length === this.state.overflow.length) {
+            window.requestAnimationFrame(() => {
```
giladgray: why delay the callback here?

packages/core/src/components/tabs/tabTitle.tsx
```diff
@@ -13,13 +13,13 @@ import { ITabProps, TabId } from "./tab";
...
-    onClick: (id: TabId, event: React.MouseEvent<HTMLElement>) => void;
+    onClick?: (id: TabId, event: React.MouseEvent<HTMLElement>) => void;
 
     /** ID of the parent `Tabs` to which this tab belongs. Used to generate ID for ARIA attributes. */
-    parentId: TabId;
+    parentId?: TabId;
 
     /** Whether the tab is currently selected. */
-    selected: boolean;
+    selected?: boolean;
```
giladgray: why make these optional? would prefer to not change this component, as it's internal-only.

giladgray:
closing as stale. i'll build this out at some point.

fushar:
Hey @qcharlieshi @giladgray any updates on this? Would love to use overflow tabs on my personal project :)

PR #2900, opened 2018-09-03
[TagInput] Use $input-padding-horizontal when empty for consistency with <InputGroup>

### LINKED ISSUE #2872: [TagInput] If no values and no left icon, increase left padding of placeholder text
#### Environment

- __Package version(s)__: 3.4.0

#### Steps to reproduce

1. Create a `TagInput` with no left icon and no values in it yet.

#### Actual behavior

Left padding is too small: 
image

#### Expected behavior

Padding matches `InputGroup` left padding exactly.

cmslewis:
#### Checklist

- [x] ~[Enable CircleCI for your fork](https://circleci.com/add-projects)~ __N/A__
- [x] ~Include tests~ __N/A__
- [x] ~Update documentation~ __N/A__

- `<TagInput>` now uses `$input-padding-horizontal` when empty.
    - Same as `<InputGroup>`.

Minor as this fix may be, there were a few ways to do it, none of which felt great:
- Inline style on the component
    - Felt dirty given our current scheme driven by CSS classes.
- Add an `-empty` CSS class to `.bp3-tag-input`, change its padding-left to 10px.
    - __(This is what I did.)__
    - Modifies the element that sets the existing padding.
    - Minor con: it introduces a new (non-exported) CSS class into the vocabulary.
- Add an `-empty` CSS class to `.bp3-tag-input-values` to scope this clunky special class strictly _within_ the outer `<TagInput>` element.
    - Feels a little cleaner philosophically.
    - But then we have padding split between `bp3-tag-input` and `bp3-tag-input-values`.
- Use some CSS `:empty` wizardry on `.bp3-tag-input-values`.
    - `bp3-input-ghost` is always present as a child, so that selector won't be satisfied unless I add another child for the tags, which could break line-wrapping behavior for tags/ghost input.
    - Would rather not have conditional logic in CSS.

#### Screenshot

My icon styles are a little screwy for some reason after I pull from latest `develop`, but note that the padding is all that changed in this PR:

_BEFORE_

`<TagInput>` 5px padding !== vs `<InputGroup>` 10px padding:

image

image

_AFTER_

`<TagInput>` 10px padding == `<InputGroup>` 10px padding (and the padding reverts to 5px when there are tags present, as before):

image

### DIFF
--- a/packages/core/src/components/tag-input/tagInput.tsx
+++ b/packages/core/src/components/tag-input/tagInput.tsx
@@ -214,6 +214,10 @@ export class TagInput extends AbstractPureComponent<ITagInputProps, ITagInputSta
     public render() {
         const { className, disabled, fill, inputProps, large, leftIcon, placeholder, values } = this.props;
 
+        // use placeholder prop only if it's defined and values list is empty or contains only falsy values
+        const isSomeValueDefined = values.some(val => !!val);
+        const resolvedPlaceholder = placeholder == null || isSomeValueDefined ? inputProps.placeholder : placeholder;
+
         const classes = classNames(
             Classes.INPUT,
             Classes.TAG_INPUT,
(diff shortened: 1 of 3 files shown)


packages/core/src/components/tag-input/tagInput.tsx
```diff
@@ -222,15 +226,15 @@ export class TagInput extends AbstractPureComponent<ITagInputProps, ITagInputSta
                 [Classes.DISABLED]: disabled,
                 [Classes.FILL]: fill,
                 [Classes.LARGE]: large,
+                // add CSS class when empty so the placeholder padding can match that of an empty <InputGroup>
+                // (also, no need for this to be part of the public Classes API)
+                // see: https://github.com/palantir/blueprint/issues/2872
+                [`${Classes.TAG_INPUT}-empty`]: !isSomeValueDefined,
```
giladgray: what about if it has a `leftIcon`?

this can also be accomplished using `:first-child` twice - on input-values and input.

giladgray: image
 
too much left padding

cmslewis:
:borat:

PR #2935, opened 2018-09-13
[DateTime][WIP] Don't apply local timezone to passed dates

### LINKED ISSUE #2934: Datepicker uses local timezone
#### Environment

- __Package version(s)__: 3.2.0
- __Browser and OS versions__: MacOS 10.13.6 , Chrome 69.0.3497.9

#### Why is this a problem
When doing snapshot testing with the Datepicker component, the snapshots will be different depending on the local timezone of the machine running the tests. This is especially problematic if a CI server is set to a different timezone than the person creating or updating the snapshots.

#### Steps to reproduce

1. Pass date object to Datepicker (`defaultValue`, `maxDate`, or `minDate`)
2. Serialize the date and store it in a file.
3. Do the same thing with a different local timezone.

A good way to do this is to check out the branch i opened ( https://github.com/palantir/blueprint/pull/2935 ) and run `UPDATE=1 yarn test`. This will update the DatePicker snapshot with your local timezone applied to the dates.

I originally recorded the snapshots with UTC+2 and modified them with `TZ=UTC UPDATE=1 yarn test`, which led to this commit: https://github.com/palantir/blueprint/pull/2935/commits/130784301b615a00a364e9038c577bd4d442b85e

#### Actual behavior

The files will have different timestamps.

#### Expected behavior

Timezone will be part of the passed date. Local timezone will not be applied.

#### Possible solution

The Datepicker should keep from applying timezone offset on the dates it receives.

tobilen:
#### Fixes #2934 
related to #1721

#### Checklist

- [X] Include tests
- [X] Update documentation

First change is the introduction of snapshot testing in blueprint. We need this test pattern to visualize the bug i want to solve here.

After this, the idea is to make sure local timezone is not a factor in component output. I don't know what shape this will take as of yet, but will update the description once i do.

1. Snapshot testing pattern (i.e. folder location, config in the karma-build-script package)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -40,6 +40,7 @@ A typical contributor workflow looks like this:
       [Editor integration](https://github.com/palantir/blueprint/wiki/Editor-integration)). Run
       `yarn lint` to be 100% safe.
     - TypeScript lint errors can often be automatically fixed by TSLint. Run lint fixes with `yarn lint-fix`.
+    - Blueprint makes use of [snapshot testing](https://jestjs.io/blog/2016/07/27/jest-14.html#why-snapshot-testing). This means that changes in component output need to be verified by updating the stored snapshots. You can accomplish this by running `UPDATE=1 yarn test` and then committing the written changes. For more information regarding snapshot testing in blueprint, refer to https://github.com/localvoid/karma-snapshot.
 1. Submit a Pull Request on GitHub and fill out the template.
 1. Team members will review your code and merge it after approvals.
     - You may be asked to make modifications to code style or to fix bugs you may have not noticed.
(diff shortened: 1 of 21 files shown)


giladgray:
I am not a big fan of snapshot testing for React components because it's far too fiddly. I much prefer testing interactions over exact markup generation.

I think this feature could be easily and reliably tested using our existing infrastructure with a simple chai assertion, rather than adding snapshots all the way down.

also, this error message is beyond unhelpful:
image
jest's snapshot errors at least provide a helpful diff.

giladgray:
All this to say, let's focus on fixing the timezone and not changing the test infrastructure please.

giladgray:
closing this PR as it does not address the issue itself.

PR #2982, opened 2018-09-27
[Icon] Allow setting size on container span

invliD:
While #2884 fixed many issues with the icons by wrapping them in an html tag, it removed the option to override an icon's size from CSS easily. It was still possible using:
```css
.my-cool-icon {
  svg {
    width: 48px;
    height: 48px;
  }
}
```
This, however, requires using the icon component's implementation details.

Before that change—and after this one—this is easily done as expected:
```css
.my-cool-icon {
  width: 48px;
  height: 48px;
}
```

I'm not sure what kind of tests or examples are useful here. You can test this easily by going to the icon component examples and setting a width on the `span` around the icon svg in the inspector—the icon will change size. Before this PR, nothing would happen.

### DIFF
--- a/packages/core/src/components/icon/_icon.scss
+++ b/packages/core/src/components/icon/_icon.scss
@@ -22,6 +22,10 @@
   > svg {
     // prevent extra vertical whitespace
     display: block;
+    // Allow width and height to be set on the span in css. This is ignored if the
+    // span does not have an explicit size set.
+    width: inherit;
+    height: inherit;
 
     // inherit text color unless explicit fill is set
     &:not([fill]) {
--- a/packages/docs-app/src/examples/core-examples/iconExample.tsx
+++ b/packages/docs-app/src/examples/core-examples/iconExample.tsx
@@ -47,7 +47,7 @@ export class IconExample extends React.PureComponent<IExampleProps, IIconExample
 
         return (
             <Example options={options} {...this.props}>
-                <Icon icon={icon} iconSize={iconSize} intent={intent} />
+                <Icon icon={icon} iconSize={iconSize} intent={intent} key={iconSize} />
             </Example>
         );
     }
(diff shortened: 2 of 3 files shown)


giladgray:
icons do seem to render as expected by default, but there a number of weird bits i've noticed.

----
it's possible to fake out the pixel-grid-detection by setting a large size in CSS without updating the corresponding prop. this will actually happen quite often as the default 16px size uses the small grid so setting any larger size will result in the wrong image being rendered. note the `viewBox` values in the images below:
image

image

----
icons resize quite differently in the example--from the left edge instead of centered.
**deployed**
icon-current
**this PR**
icon-inherit

----
`iconSize={X}` feels much friendlier to me than setting width and height in CSS.

invliD:
(1) this was previously also possible in BP1 when you would choose one of the two sizes manually and then override the font-size in css
(2) ouch, that seems to be a Chrome bug. When you inspect the `span` you can see that it properly changes height, but not width. Just unchecking and re-checking `width: inherit` on the `svg` fixes it. I think this bug is only really relevant in the docs though, since you don't usually change size after the fact.
(3) I disagree strongly, and this is why I'm proposing this PR. 100% of our design lives in CSS. With this prop being the only good way of changing the icon's size, we now have to move this one thing of design to JS, keeping everything else in CSS. I've specifically noticed this converting some parts of my app from BP1 to BP3 and I now have to manage the font-size of some text in CSS and the size of the icon next to it in JS. That's not maintainable and will cause visual bugs in the future.

Sometimes using the prop is not really an option, e.g. when overriding styles for `InputGroup`. We have an input that's significantly smaller than usual, so we override its styles to work for us. Previously we could just override the nested `.pt-icon`'s styles, now we have to do what I described above in the first example, overriding styles on the nested `svg` element. Unless we add optional `iconProps` to all components wrapping them to allow changing their size, this, again, isn't really maintainable.

giladgray:
can you fix the example please?

packages/core/src/components/icon/_icon.scss
```diff
@@ -22,6 +22,10 @@
   > svg {
     // prevent extra vertical whitespace
     display: block;
+    // Allow width and height to be set on the span in css. This is ignored if the
+    // span does not have an explicit size set.
```
giladgray: how is it ignored? there's no logic that ignores it, we're relying on browser implementations it seems.

giladgray:
closing as stale and not ready to ship.

PR #3035, opened 2018-10-15
include integrity hashes in yarn lock

rgbkrk:
Running more recent `yarn` (v1.10.1) will generate these hashes. I love having a clean branch to work from so I'm making this as a standalone PR. This will reduce the noise for anyone on recent `yarn`.


rgbkrk:
Thanks @giladgray!

baqiyahwahab:
Good

PR #3096, opened 2018-10-29
[Select] Flag to optionally scroll to active item for select components

jscheiny:

Adds a flag `scrollToActiveItem` which defaults to `true` for select list components.

#### Screenshot

Prevents this undesirable scrolling behavior when updating active item based on mouse hover:

select-infinite-scroll

Note this is just by hovering, no scroll wheel was used in the creation of this gif.

### DIFF
--- a/packages/select/src/common/listItemsProps.ts
+++ b/packages/select/src/common/listItemsProps.ts
@@ -113,7 +113,9 @@ export interface IListItemsProps<T> extends IProps {
     resetOnSelect?: boolean;
 
     /**
-     * Whether the active item should be scrolled into view when the active item is changed.
+     * When the active item is controlled, whether the active item should be scrolled into view when
+     * an external change to the active item is made. External changes do not include selecting an item
+     * in the list, searching for an item, or changing the selection via up and down arrow keys.
      * @default true
      */
     scrollToActiveItem?: boolean;
--- a/packages/select/src/common/listItemsProps.ts
+++ b/packages/select/src/common/listItemsProps.ts
@@ -113,6 +113,12 @@ export interface IListItemsProps<T> extends IProps {
     resetOnSelect?: boolean;
 
     /**
+     * Whether the active item should be scrolled into view when the active item is changed.
+     * @default true
+     */
+    scrollToActiveItem?: boolean;
+
+    /**
      * Query string passed to `itemListPredicate` or `itemPredicate` to filter items.
      * This value is controlled: its state must be managed externally by attaching an `onChange`
      * handler to the relevant element in your `renderer` implementation.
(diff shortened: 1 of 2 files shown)


packages/select/src/components/query-list/queryList.tsx
```diff
@@ -152,6 +152,10 @@ export class QueryList<T> extends React.Component<IQueryListProps<T>, IQueryList
     }
 
     public scrollActiveItemIntoView() {
+        if (this.props.scrollToActiveItem === false) {
```
giladgray: turning this off _entirely_ is a no-go as it fully breaks keyboard interactions -- you can easily arrow right out of the viewport. instead, can we move this check to conditionally invoke this method in certain circumstances?

jscheiny: Yeah, that's a good point, I'll investigate if there's a way to do this.

jscheiny:
I haven't been able to figure out a way to get this prop to only respect external changes to active item. However, if we change the scrolling functionality to only scroll the element *just* into view, then that should work. Any objections to that change of functionality?

toombsy:
I haven’t got a clue what’s going on with any of it so I can not comment 

Neil Get

giladgray:
@jscheiny sure you can take a stab at adjusting how it scrolls. 

but also how about playing with when `this.shouldCheckActiveItemInViewport` is enabled? say, this line: https://github.com/palantir/blueprint/blob/develop/packages/select/src/components/query-list/queryList.tsx#L128

toombsy:
Neil Get

giladgray:
@toombsy you can unsubscribe yourself using the "Notifications" button on the right.

jscheiny:
@giladgray I investigated that and it looks like it will be a lot more complicated than is worth while, since we'd have to wire through a lot of information about the source of the change of the active item. I can try again but in the meantime I've opened up https://github.com/palantir/blueprint/pull/3125 with the change to the scrolling behavior.

toombsy:
https://www.paulkelly.com.au/



Toombsy

toombsy:
Neil Get

jscheiny:
For reference, here's the new behavior:

When `scrollToActiveItem={false}`, we will still scroll to the active item when an expected internal change to the active item is made. This includes:

- Clicking on an item
- Using keyboard inputs to move to a new item
- When updating a query and `resetOnQuery={true}`

However, if an external change to the active item is made which is unexpected (such as in the example above where hovering on an item makes it the active item), then we will not scroll to that item.

packages/select/src/common/listItemsProps.ts
```diff
@@ -112,6 +112,12 @@ export interface IListItemsProps<T> extends IProps {
      */
     resetOnSelect?: boolean;
 
+    /**
+     * Whether the active item should be scrolled into view when the active item is changed.
```
giladgray: explain the particular behavior of this prop -- it's not _always_ when the active item is changed.
it also only really applies when `props.activeItem` is controlled, right?

PR #3125, opened 2018-11-02
[Select] Update query list scroll behavior

jscheiny:

When scrolling the active item into view for query lists, only scroll that item exactly into view and no more, fixes the same issue that https://github.com/palantir/blueprint/pull/3096 attempts to fix but in a different and more consistent manner. I've tested this will menu items of multiple heights and it seems to work fine.

Should this belong under a flag prop?


jscheiny:
Found a decent solution in the other PR, closing this out for the moment.

PR #3137, opened 2018-11-08
Don't clear TagInput inputValue state if controlled

### LINKED ISSUE #3134: Multi select mistakenly clears selection when no tag is added
#### Environment

- Blueprint 3.8.0
- MacOS. Chrome 69

#### Steps to reproduce

1. Open MultiSelect docs
2. Enter gibberish so that nothing matches

image

3. Press enter

#### Actual behavior

Gibberish is no longer shown but the gibberish filter still applies.

image

#### Expected behavior

Nothing happens when pressing enter, and the gibberish is still shown.

image

#### Possible solution

A few, though I'm not sure which is most in line with the rest of the Blueprint API:

1. Don't automatically clear the `inputValue` in `TagInput.addTags` if `inputValue` is controlled
2. Add a prop to `TagInput` that controls whether the value is auto cleared
3. In `MultiSelect`, pass an `onAdd` function to the `TagInput` component that always returns false, because we don't want it to auto clear the input.

ericjeney:

This change modifies `TagInput` to not clear the value displayed in the `input` field when adding tags if the consumer passes in their own `inputValue` prop. Previously, adding tags would clear this field, which resulted in confusing behavior in `MultiSelect`, as described in the issue. This seems philosophically correct: If the consumer specifies an explicit value for `inputValue`, it should never be overridden in `TagInput`'s state.

### DIFF
--- a/packages/core/src/components/tag-input/tagInput.tsx
+++ b/packages/core/src/components/tag-input/tagInput.tsx
@@ -184,13 +184,14 @@ export class TagInput extends AbstractPureComponent<ITagInputProps, ITagInputSta
         addOnBlur: false,
         addOnPaste: true,
         inputProps: {},
+        inputValue: "",
         separator: /[,\n\r]/,
         tagProps: {},
     };
 
     public state: ITagInputState = {
         activeIndex: NONE,
-        inputValue: this.props.inputValue || "",
+        inputValue: this.props.inputValue,
         isInputFocused: false,
     };
 
@@ -206,7 +207,7 @@ export class TagInput extends AbstractPureComponent<ITagInputProps, ITagInputSta
         super.componentWillReceiveProps(nextProps);
 
         if (nextProps.inputValue !== this.props.inputValue) {
-            this.setState({ inputValue: nextProps.inputValue || "" });
+            this.setState({ inputValue: nextProps.inputValue });
         }
     }
 
@@ -262,8 +263,8 @@ export class TagInput extends AbstractPureComponent<ITagInputProps, ITagInputSta
     private addTags = (value: string) => {
         const { inputValue, onAdd, onChange, values } = this.props;
         const newValues = this.getValues(value);
-        let shouldClearInput = inputValue === undefined;
-        shouldClearInput = Utils.safeInvoke(onAdd, newValues) !== false && shouldClearInput;
+        let shouldClearInput =
+            Utils.safeInvoke(onAdd, newValues) !== false && inputValue === TagInput.defaultProps.inputValue;
         // avoid a potentially expensive computation if this prop is omitted
         if (Utils.isFunction(onChange)) {
             shouldClearInput = onChange([...values, ...newValues]) !== false && shouldClearInput;
--- a/packages/core/src/components/tag-input/tagInput.tsx
+++ b/packages/core/src/components/tag-input/tagInput.tsx
@@ -184,14 +184,13 @@ export class TagInput extends AbstractPureComponent<ITagInputProps, ITagInputSta
         addOnBlur: false,
         addOnPaste: true,
         inputProps: {},
-        inputValue: "",
         separator: /[,\n\r]/,
         tagProps: {},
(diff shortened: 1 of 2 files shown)


packages/core/src/components/tag-input/tagInput.tsx
```diff
@@ -184,14 +184,13 @@ export class TagInput extends AbstractPureComponent<ITagInputProps, ITagInputSta
         addOnBlur: false,
         addOnPaste: true,
         inputProps: {},
-        inputValue: "",
```
giladgray: i prefer one fully type-safe `defaultProps` over two not-strictly-typesafe boolean coercions `this.props.inputValue || ""`

packages/core/src/components/tag-input/tagInput.tsx
```diff
@@ -261,15 +260,16 @@ export class TagInput extends AbstractPureComponent<ITagInputProps, ITagInputSta
     }
 
     private addTags = (value: string) => {
-        const { onAdd, onChange, values } = this.props;
+        const { inputValue, onAdd, onChange, values } = this.props;
         const newValues = this.getValues(value);
-        let shouldClearInput = Utils.safeInvoke(onAdd, newValues);
+        let shouldClearInput = inputValue === undefined;
+        shouldClearInput = Utils.safeInvoke(onAdd, newValues) !== false && shouldClearInput;
```
giladgray: collapse these two to one line?
`let shouldClearInput = Utils.safeInvoke(onAdd, newValues) !== false && inputValue === undefined;`

packages/core/src/components/tag-input/tagInput.tsx
ericjeney: If I switch this back, I won't be able to explicitly check in `addTags` if `inputValue` is controlled, and will instead have to check if `inputValue === ""`. I think that works in this case, but is less explicit. Would you still prefer me to set the `inputValue` default here?

giladgray: oh hmm that's a thing. in other components, i've checked `=== X.defaultProps.Y` to great success.

packages/core/src/components/tag-input/tagInput.tsx
```diff
@@ -262,8 +263,8 @@ export class TagInput extends AbstractPureComponent<ITagInputProps, ITagInputSta
     private addTags = (value: string) => {
         const { inputValue, onAdd, onChange, values } = this.props;
         const newValues = this.getValues(value);
-        let shouldClearInput = inputValue === undefined;
-        shouldClearInput = Utils.safeInvoke(onAdd, newValues) !== false && shouldClearInput;
+        let shouldClearInput =
+            Utils.safeInvoke(onAdd, newValues) !== false && inputValue === TagInput.defaultProps.inputValue;
```
giladgray: ok on second thought, this doesn't work for strings:
```js
{} === {}
> false

[] === []
> false

"" === ""
> true // uh oh - controlled empty value looks like uncontrolled
```

i think we do have to remove the `defaultProps` entry and compare against undefined like you had before

packages/core/test/tag-input/tagInputTests.tsx
```diff
@@ -356,6 +362,13 @@ describe("<TagInput>", () => {
             pressEnterInInput(wrapper, NEW_VALUE);
             assert.strictEqual(wrapper.state().inputValue, "");
         });
+
+        it("does not clear the input if the input is controlled", () => {
+            const onChange = sinon.stub();
+            const wrapper = shallow(<TagInput onChange={onChange} values={VALUES} inputValue={NEW_VALUE} />);
```
giladgray: i'll bet if you did `inputValue=""` (controlled empty value) this test would fail.

ericjeney: I don't think it would fail in this particular case, but I in general agree that we should be more explicit in checking if this prop is controlled.

PR #3219, opened 2018-12-06
Removed es6-shim from @blueprintjs/table for 1.x version

### LINKED ISSUE #522: Remove es6-shim as a dependency for @blueprintjs/table

mktange:
#### Fixes #522
The fixes for this issue were not backported to the 1.x version. This PR aims to do this.

### DIFF
--- a/packages/docs-app/src/blueprint.md
+++ b/packages/docs-app/src/blueprint.md
@@ -42,6 +42,15 @@ dependencies:
   npm install --save react react-dom react-addons-css-transition-group
   ```
 
+1. Note that since the minimum supported version of React is [v16](https://reactjs.org/blog/2017/09/26/react-v16.0.html),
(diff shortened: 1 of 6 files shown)


mktange:
Closing this after talking with @giladgray, who pointed out that this is essentially a breaking change if consumers do not already have a shim in place for es6, and rely on the one that comes packaged with @blueprintjs/table.

As a workaround (if you have other shims in place), you can exclude the package entirely in your webpack config with:

```js
module.exports = {
  //...
  externals: {
    "es6-shim": "window"
  }
};
```

PR #3259, opened 2018-12-27
Suggest using `ContextMenuTarget` as a HOC

ClaudiuCeia:

Instead of falling back to the imperative API, suggest using `ContextMenuTarget`, where decorators are not an option. 
I couldn't find this suggestion in the docs. If the PR gets merged I can send one for the [hotkey decorator](https://blueprintjs.com/docs/#core/components/hotkeys.decorator) too.

I tested this locally already and I couldn't find any issues with this approach.

### DIFF
--- a/packages/core/src/components/context-menu/context-menu.md
+++ b/packages/core/src/components/context-menu/context-menu.md
@@ -57,8 +57,36 @@ class RightClickMe extends React.Component<{}, {}> {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


palantirtech:
Thanks for your interest in [palantir/blueprint](https://github.com/palantir/blueprint), @ClaudiuCeia! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

PR #3356, opened 2019-02-13
Fix shortcuts.tsx react import typings

philcchen:

- Fix shortcuts.tsx react import typings


philcchen:
Lint failure seems unrelated?

PR #3370, opened 2019-02-19
Upgrade dev dependencies and types

adidahiya:
I wanted to bump the `@types/dom4` dependency to 2.0.1 and ended up upgrading a handful of things in the root package.json. The major version bumps are all libraries which have dropped support for Node v4, which we didn't support anyway. Some react types changed, they seem to be a little stricter now.


packages/core/package.json
```diff
@@ -61,7 +61,7 @@
     "@blueprintjs/test-commons": "^0.8.1",
     "enzyme": "^3.3.0",
     "karma": "^3.1.4",
-    "mocha": "^4.1.0",
+    "mocha": "^5.2.0",
```
cmslewis: For the record, looks like the only break in Mocha 5.x was to drop support for IE9 and IE10, which seems fine. Here are the [release notes](https://github.com/mochajs/mocha/blob/master/CHANGELOG.md#500--2018-01-17).

packages/core/src/components/context-menu/contextMenu.tsx
```diff
@@ -34,8 +34,10 @@ const POPPER_MODIFIERS: PopperModifiers = {
 };
 const TRANSITION_DURATION = 100;
 
+type IContextMenuProps = IOverlayLifecycleProps;
```
cmslewis: 👍 This is definitely clearer.

packages/core/test/controls/numericInputTests.tsx
```diff
@@ -24,12 +30,20 @@ import {
     Position,
 } from "../../src/index";
 
+/**
+ * @see https://github.com/DefinitelyTyped/DefinitelyTyped/issues/26979#issuecomment-465304376
```
cmslewis: Ack.

packages/core/test/editable-text/editableTextTests.tsx
```diff
@@ -75,7 +75,7 @@ describe("<EditableText>", () => {
             const OLD_VALUE = "alphabet";
             const NEW_VALUE = "hello";
 
-            const component = mount(
+            const component = mount<EditableText>(
```
cmslewis: Oof, this is annoying. Hope you get a response to your question in DefinitelyTyped.

packages/docs-app/src/components/resources.tsx
```diff
@@ -18,11 +18,15 @@ const RESOURCES: IResourceProps[] = [
 export const Resources: React.SFC = () => (
     <>
         <div className="blueprint-resources">
-            {RESOURCES.map(resource => <ResourceCard key={resource.fileName} {...resource} />)}
+            {RESOURCES.map(resource => (
+                <ResourceCard key={resource.fileName} {...resource} />
+            ))}
```
cmslewis: This file was failing lint in my PR also. I went ahead and fixed it too.

packages/tslint-config/src/rules/blueprintClassesConstantsRule.ts
```diff
@@ -5,7 +5,7 @@
  */
 
 import * as Lint from "tslint";
-import * as utils from "tsutils";
+import { isExpressionStatement, isJsxAttribute, isPropertyAssignment, isStringLiteral } from "tsutils/typeguard/2.8";
```
cmslewis: Did `tsutils` change how it's exporting things? Possible to avoid the submodule import?

packages/core/test/editable-text/editableTextTests.tsx
adidahiya: I think it's unlikely to be resolved and we'll have to keep doing this

packages/docs-app/src/components/resources.tsx
adidahiya: yeah it was fixed on master with #3374

packages/tslint-config/src/rules/blueprintClassesConstantsRule.ts
adidahiya: no, since tsutils references types from different tsc versions, and this is the best way to deal with it

PR #3375, opened 2019-02-21
[Core] File input text styling for active selections

pkwi:
Matched text color with other input type components

@adidahiya or should we convert it into a placeholder?


packages/core/src/components/forms/_file-input.scss
```diff
@@ -128,6 +131,16 @@ $file-input-button-padding-large: ($pt-input-height-large - $pt-button-height) /
       @include pt-dark-button-active();
     }
   }
+
+  .#{ns}-file-input-has-selection {
```
adidahiya: missing `$` here

adidahiya:
@pkwi what do you think about my latest changes?

image

pkwi:
@adidahiya perfect! thanks for finishing this one for me.

PR #3382, opened 2019-02-27
Drawer default prop warning fix

### LINKED ISSUE #3380: <Drawer /> shows warning with default props
#### Environment

- __Package version(s)__: **"@blueprintjs/core": "3.13.0"**
- __Browser and OS versions__: Google Chrome Version 72.0.3626.109 (Official Build) (64-bit), Ubuntu 16.04.5 LTS

#### Steps to reproduce

1. Render `Drawer` component with default props (only fill required ones)
2. You can even try to modify other props (without adding `title`), even setting `isCloseButtonShown` to `false`, same result

#### Actual behavior

Throws out warning on each re-render (I assume):

`[Blueprint] <Dialog> isCloseButtonShown prop is ignored if title is omitted.`

#### Expected behavior

No warnings..

#### Possible solution

Only valid solution I found was to add empty `title` prop (`title=""`), but that causes empty header to draw within (which I don't want)

Ackos95:
Updated 'isCloseButtonShown' prop at 'Drawer' not to be defaulted regularly (through defaultProps) because of prop validation would throw error for default props (title == null && isCloseButtonShown == true).

#### Checklist

- [x] Includes tests - no need, behavior not chanaged
- [x] Update documentation - no need, documentation remains the same

Simply updated the way of defaulting `isCloseButtonShown` prop, to avoid unnecessary warnings.

Nothing specific, small PR

### DIFF
--- a/packages/core/src/components/drawer/drawer.tsx
+++ b/packages/core/src/components/drawer/drawer.tsx
@@ -79,7 +79,6 @@ export class Drawer extends AbstractPureComponent<IDrawerProps, {}> {
     public static displayName = `${DISPLAYNAME_PREFIX}.Drawer`;
     public static defaultProps: IDrawerProps = {
         canOutsideClickClose: true,
-        isCloseButtonShown: true,
         isOpen: false,
         style: {},
         vertical: false,
@@ -115,7 +114,9 @@ export class Drawer extends AbstractPureComponent<IDrawerProps, {}> {
     }
 
     private maybeRenderCloseButton() {
-        if (this.props.isCloseButtonShown) {
+        // `isCloseButtonShown` can't be defaulted through default props because of props validation
+        // so this check actually defaults it to true if it is not defined (== null)
+        if (this.props.isCloseButtonShown || this.props.isCloseButtonShown == null) {
             return (
                 <Button
                     aria-label="Close"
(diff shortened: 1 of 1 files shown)


palantirtech:
Thanks for your interest in [palantir/blueprint](https://github.com/palantir/blueprint), @Ackos95! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

packages/core/src/components/drawer/drawer.tsx
```diff
@@ -115,7 +114,9 @@ export class Drawer extends AbstractPureComponent<IDrawerProps, {}> {
     }
 
     private maybeRenderCloseButton() {
-        if (this.props.isCloseButtonShown) {
+        // `isCloseButtonShown` can't be defaulted through default props because of props validation
+        // so this check actually defaults it to true if it is not defined (== null)
+        if (this.props.isCloseButtonShown || this.props.isCloseButtonShown == null) {
```
adidahiya: you can simplify a bit with `if (this.props.isCloseButtonShown !== false)`

make sure to update the code comment too

Ackos95:
@palantirtech it's signed

packages/core/src/components/drawer/drawer.tsx
Ackos95: That was my first choice, but I saw that `== null` was already used in prop validation, so I wanted to be more consistent. No problem, I'll update it

PR #3386, opened 2019-02-28
[Core] Add position prop to Drawer component, deprecate vertical prop

### LINKED ISSUE #3360: Feature request: open drawers from the left
I would like to know if I can set the direction of drawer on the page. I would like to make drawer appear on the left of web page. I set left:0 for drawer component. It appears on the left. However, it still transform from right to left. This is not result I would like. How can I solve this problem?

Ackos95:
Added `reverse` prop to `Drawer` component to enable rendering on reverse side (if vertical then on the top instead of the bottom, otherwise on the left instead of the right side).

#### Checklist

- [x] Includes tests
- [x] Update documentation

Added new prop (`reversed`) to `Drawer` component to enable it's reversed rendering (so that it can be rendered on all sides of the screen, not only on right and top). Changes are mostly related to CSS.

#### Screenshot

**reversed**: `true` && **vertical**: `false`

image

**reversed**: `true` && **vertical**: `true`

image

Docs updated

image


Ackos95:
@adidahiya any news on this?

packages/core/src/components/drawer/drawer.tsx
```diff
@@ -37,6 +37,14 @@ export interface IDrawerProps extends IOverlayableProps, IBackdropProps, IProps
      */
     isOpen: boolean;
 
+    /**
+     * Whether the drawer should appear on the reversed side.
+     * Depending on the `vertical` prop, it will show on the left side (instead of the right),
+     * or on the top (instead of the bottom)
```
giladgray: trailing period please

packages/core/src/components/drawer/drawer.tsx
```diff
@@ -89,8 +98,9 @@ export class Drawer extends AbstractPureComponent<IDrawerProps, {}> {
     public static readonly SIZE_LARGE = "90%";
 
     public render() {
-        const { size, style, vertical } = this.props;
-        const classes = classNames(Classes.DRAWER, { [Classes.VERTICAL]: vertical }, this.props.className);
+        const { size, style, vertical, reversed } = this.props;
+        const additionalClasses = { [Classes.VERTICAL]: vertical, [Classes.REVERSED]: reversed };
```
giladgray: just define this object in the `classNames` call - we don't use this pattern anywhere else. prettier will fix the formatting.

giladgray:
one more thing: `reverse` or ` reversed`? i prefer the former; it's `minimal` not `minimized`

Ackos95:
@adidahiya Alright, I'll do that. Any particular way to deprecate `vertical`, or simply just to remove it?

adidahiya:
Do not remove it, that would be a break. Deprecate it with the `@deprecate` JSDoc annotation: https://github.com/palantir/blueprint/blob/ec81076a4b491989123cc0ed6c27f1c6768e8e03/packages/core/src/components/html-table/htmlTable.tsx#L33

adidahiya:
And you'd still have to support the old prop. Just make a note in the docs that the new prop overrides the deprecated one if both are present.

Ackos95:
@adidahiya Can you check again, I updated it do use `position?: Position` prop for location, deprecated `vertical` prop, and removed `reverse` prop. I left previous logic with classes `{ns}-reverse` and `{ns}-vertical`, just to avoid creating new classes, when it already works with given ones (and it is intuitive IMHO).

adidahiya:
I didn't look at the new code yet but I don't think you should introduce a `{ns}-reverse` class which has no relation to a prop of that name. Please keep the existing `{ns}-vertical` class, since I would like to avoid that subtle CSS API break. Add new `{ns}-position-{top,right,bottom-left}` classes.

Ackos95:
@adidahiya Just a question, why would I need `{ns}-vertical`? To insert it when position is not defined, even though I could resolve it with `{ns}-position-bottom`?

adidahiya:
@Ackos95 it's part our (unfortunately large) CSS API surface area. I intend to elaborate more on this in #3395, but for now: yes, "to insert it when position is not defined". please try to make it work so that the deprecated `vertical` prop applies `{ns}-vertical` and the relevant styling. There will be multiple CSS classes which end up doing the same thing.

Ackos95:
@adidahiya, it's done as you asked. I removed `vertical` from example, but left it in tests and documentation, `{ns}-reverse` class is removed, and position classes are added and used.

adidahiya:
documentation preview for latest commit: https://24956-71939872-gh.circle-artifacts.com/0/home/circleci/project/packages/docs-app/dist/index.html#core/components/drawer

packages/core/src/common/position.ts
```diff
@@ -43,3 +43,27 @@ export function isPositionVertical(position: Position) {
         position === Position.RIGHT_BOTTOM
     );
 }
+
+export function getPositionIgnoreAngles(position: Position) {
```
adidahiya: nice 👍

Ackos95:
@adidahiya thanks man, can you tell me in which release will this be included?

adidahiya:
I plan to cut a minor release with this today. You can always click through the commit to see which tags include it, for future reference:

image

PR #3398, opened 2019-03-07
Add option to grow text area with input

jroitgrund:
#### Checklist

- [x] Includes tests
- [x] Update documentation

Add a default-false prop to `TextArea`: `growVertically`. When set to `true`, the text area will expand vertically to accomodate all the user's input without scrolling.

Does this break any existing behavior?

#### Screenshot

isengard

### DIFF
--- a/packages/core/src/components/forms/textArea.tsx
+++ b/packages/core/src/components/forms/textArea.tsx
@@ -26,18 +26,29 @@ export interface ITextAreaProps extends IIntentProps, IProps, React.TextareaHTML
     small?: boolean;
 
     /**
+     * Whether the text area should automatically grow vertically to accomodate content.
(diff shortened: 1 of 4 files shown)


packages/core/src/components/forms/textArea.tsx
```diff
@@ -50,6 +61,29 @@ export class TextArea extends React.PureComponent<ITextAreaProps, {}> {
             className,
         );
 
-        return <textarea {...htmlProps} className={rootClasses} ref={inputRef} />;
+        const styleProps =
```
adidahiya: actually I just realized we should be merging this style object into the one in `htmlProps` if the user provided that. I'll fix it in a follow up

PR #3405, opened 2019-03-10
[Overlay] always listen for mousedown on document if overlay is inlin…

### LINKED ISSUE #3404: [Overlay] Inline overlays with backdrops do not close on outside click
#### Environment

- __Package version(s)__: 3.14.1
- __Browser and OS versions__: OSX/FF65

#### Steps to reproduce

1. On Overlay docs page, open an overlay with `usePortal={false}`, `hasBackdrop={true}` and `canOutsideCloseClick={true}`
2. Click anywhere on the page outside of the overlay's container (overlay does not close)
3. Click anywhere inside the overlay's backdrop (overlay will close)

#### Actual behavior

An outside click somewhere on the document while an overlay is open does not close it.

#### Expected behavior

When an overlay is open with `canOutsideClick={true}`, any click outside of the overlay content should close it.

#### Possible solution

Provided that this is not the desired behavior and by design, inline (non-portal) overlays should always bind a mousedown listener to document, regardless of whether or not they have a backdrop.

jaamison:
…e, even if it has a backdrop

### DIFF
--- a/packages/core/src/components/overlay/overlay.tsx
+++ b/packages/core/src/components/overlay/overlay.tsx
@@ -378,7 +378,7 @@ export class Overlay extends React.PureComponent<IOverlayProps, IOverlayState> {
             document.addEventListener("focus", this.handleDocumentFocus, /* useCapture */ true);
         }
 
-        if (this.props.canOutsideClickClose && !this.props.hasBackdrop) {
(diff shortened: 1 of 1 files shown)


jaamison:
Closing per discussion - existing behavior is by design

PR #3406, opened 2019-03-10
[Docs] Overlay Example: add button to toggle scrolling behavior

jaamison:
Add button to the Overlay content in the docs example that makes the Overlay container need to scroll

### DIFF
--- a/packages/docs-app/src/examples/core-examples/overlayExample.tsx
+++ b/packages/docs-app/src/examples/core-examples/overlayExample.tsx
@@ -11,6 +11,7 @@ import { Example, handleBooleanChange, IExampleProps } from "@blueprintjs/docs-t
 import { IBlueprintExampleData } from "../../tags/reactExamples";
 
 const OVERLAY_EXAMPLE_CLASS = "docs-overlay-example-transition";
+const OVERLAY_TALL_CLASS = "docs-overlay-example-tall";
 
 export interface IOverlayExampleState {
     autoFocus: boolean;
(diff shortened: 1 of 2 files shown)


packages/docs-app/src/examples/core-examples/overlayExample.tsx
```diff
@@ -46,7 +49,13 @@ export class OverlayExample extends React.PureComponent<IExampleProps<IBlueprint
     private handleOutsideClickChange = handleBooleanChange(val => this.setState({ canOutsideClickClose: val }));
 
     public render() {
-        const classes = classNames(Classes.CARD, Classes.ELEVATION_4, OVERLAY_EXAMPLE_CLASS, this.props.data.themeName);
+        const classes = classNames(
+            Classes.CARD,
+            Classes.ELEVATION_4,
+            OVERLAY_EXAMPLE_CLASS,
+            this.props.data.themeName,
+            this.state.scroll ? OVERLAY_TALL_CLASS : undefined
```
adidahiya: use the classNames syntax of `{ [OVERLAY_TALL_CLASS]: this.state.scroll }`

packages/docs-app/src/examples/core-examples/overlayExample.tsx
```diff
@@ -20,6 +21,7 @@ export interface IOverlayExampleState {
     hasBackdrop: boolean;
     isOpen: boolean;
     usePortal: boolean;
+    scroll: boolean;
```
adidahiya: a better name would be `useTallContent`

PR #3418, opened 2019-03-13
Update recommended tslint extension for vscode

lukyth:

[eg2.tslint](https://marketplace.visualstudio.com/items?itemName=eg2.tslint) has been deprecated in favor of [vscode-typescript-tslint-plugin](https://marketplace.visualstudio.com/items?itemName=ms-vscode.vscode-typescript-tslint-plugin), see https://marketplace.visualstudio.com/items?itemName=eg2.tslint for more info.

Whether this new extension should be recommended or not.

### DIFF
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -7,7 +7,7 @@
     // for consistent editor settings:
     "EditorConfig.EditorConfig",
     // for typescript source files:
-    "eg2.tslint",
+    "ms-vscode.vscode-typescript-tslint-plugin",
     // for sass source files:
     "shinnn.stylelint",
   ]


adidahiya:
The new plugin requires TS 3.2, which this repo doesn't use

lukyth:
I see. Sorry for the inconvenience.

Just out of curiosity, what's stopping this repo from using TypeScript@3.2+ ?

adidahiya:
@lukyth not much is blocking that upgrade, just need the time to do it. also we'd need to be careful about breaking changes in the generated typings so that we don't break consumers using older versions of TS

lukyth:
I see. Thank you for the explanation! 😄 I'm closing this for now then.

PR #3428, opened 2019-03-18
[MultiSelect] Enable pasting of multiple values (same as TagInput)

cmslewis:
### Checklist

- [x] Includes tests
- [x] Update documentation

Allow pasting of multiple values in `MultiSelect`.

#### Given:

The clipboard contains separated values (where the separator is configurable via `tagInputProps.separator`):

image

#### You can now:

* Select multiple _existing_ items on paste:
    
* _(If the "Create item" option is enabled)_
    Select a combination of _existing_ and _newly created_ items on paste:
    

* _(If the "Create item" option is enabled)_
    Select all _newly created_ items on paste:
    

* _(UX nicety)_
   If a pasted value is already selected, set that item as the active item on paste:
    _In general, the code moves the active item to the final existing item amongst the pasted values._
    

### Reviewers should focus on:

I originally tried to implement this outside of `MultiSelect` in a wrapping component (by listening to `tagInputProps.onAdd`, which is invoked when values are added via paste), but that didn't work.

First, note that adding values via pasting does _not_ invoke our normal on-select listeners, but it does still invoke `tagInputProps.onAdd`. Thus, we have to explicitly invoke `onItemSelect` or similar from within `tagInputProps.onAdd` in order for pasted values to be emitted.

Unfortunately, when you emit `onItemSelect` from `tagInputProps.onAdd`, you end up double-invoking it in the default selection case, which leads to a bug. In particular, pressing <kbd>Enter</kbd> or clicking to select an item will emit `onItemSelect` from `tagInputProps.onAdd`, then by the time the `MultiSelect` change handler is invoked, the pasted value will already be selected, and it would  be toggled to _deselected_ immediately.

In this PR, I now have `TagInput`'s `onAdd` emit a second parameter of type `TagInputAddMethod`, which tells callers _how_ the emitted values were added (`default`, on `blur`, or on `paste`).¹  Only if the values were added via `paste` do I take special action to select matching items or create new ones if possible; otherwise, I don't do anything, and the `MultiSelect` component behaves just as before. The second parameter here can be safely ignored, which avoids API breaks.

---

¹ This consolidates logic, because parent components won't need to do any key listening to figure out if a particular keydown will trigger a paste.

❌ In that world, you might have to do something 🤡clowny like this:
```tsx
class MyWrapper extends React.PureComponent {
    private wasLastKeyDownAPasteOperation = false; 

    // render a MultiSelect, and pass tagInputProps: { onAdd: this.handleTagInputAdd } 
    private handleKeyDown(e) {
        this.wasLastKeyDownAPasteOperation = e.which === Keys.LETTER_V && e.metaKey;
    }
    private handleTagInputAdd(values: string) {
        if (this.wasLastKeyDownAPasteOperation) {
            // Then we know the values were added on paste, which would have
            // bypassed our normal key and click selection listeners. We'll need
            // to emit the values manually. 
        }
    }
}
```

### Screenshot

See above.

### DIFF
--- a/packages/core/src/components/tag-input/tagInput.tsx
+++ b/packages/core/src/components/tag-input/tagInput.tsx
@@ -15,6 +15,26 @@ import * as Utils from "../../common/utils";
 import { Icon, IconName } from "../icon/icon";
 import { ITagProps, Tag } from "../tag/tag";
 
+/**
+ * An enumeration of methods in which a `TagInput` value might have been added.
+ */
+export enum TagInputAddMethod {
+    /** Indicates that a value was added in the default fasion, via manual selection. */
+    DEFAULT = "default",
+
+    /**
+     * Indicates that a value was added when the `TagInput` lost focus. This is
+     * only possible when `addOnBlur=true`.
+     */
+    BLUR = "blur",
+
+    /**
+     * Indicates that a value was added via paste. This is only possible when
+     * `addOnPaste=true`.
+     */
+    PASTE = "paste",
+}
+
 export interface ITagInputProps extends IIntentProps, IProps {
     /**
      * If true, `onAdd` will be invoked when the input loses focus.
@@ -74,7 +94,7 @@ export interface ITagInputProps extends IIntentProps, IProps {
      * returns `false`. This is useful if the provided `value` is somehow invalid and should
      * not be added as a tag.
      */
-    onAdd?: (values: string[]) => boolean | void;
+    onAdd?: (values: string[], method: TagInputAddMethod) => boolean | void;
 
     /**
      * Callback invoked when new tags are added or removed. Receives the updated list of `values`:
@@ -260,10 +280,10 @@ export class TagInput extends AbstractPureComponent<ITagInputProps, ITagInputSta
         );
     }
 
-    private addTags = (value: string) => {
+    private addTags = (value: string, method: TagInputAddMethod = TagInputAddMethod.DEFAULT) => {
         const { inputValue, onAdd, onChange, values } = this.props;
         const newValues = this.getValues(value);
-        let shouldClearInput = Utils.safeInvoke(onAdd, newValues) !== false && inputValue === undefined;
+        let shouldClearInput = Utils.safeInvoke(onAdd, newValues, method) !== false && inputValue === undefined;
         // avoid a potentially expensive computation if this prop is omitted
         if (Utils.isFunction(onChange)) {
             shouldClearInput = onChange([...values, ...newValues]) !== false && shouldClearInput;
@@ -342,7 +362,7 @@ export class TagInput extends AbstractPureComponent<ITagInputProps, ITagInputSta
             // defer this check using rAF so activeElement will have updated.
             if (!currentTarget.contains(document.activeElement)) {
                 if (this.props.addOnBlur && this.state.inputValue !== undefined && this.state.inputValue.length > 0) {
-                    this.addTags(this.state.inputValue);
+                    this.addTags(this.state.inputValue, TagInputAddMethod.BLUR);
                 }
                 this.setState({ activeIndex: NONE, isInputFocused: false });
             }
@@ -367,7 +387,7 @@ export class TagInput extends AbstractPureComponent<ITagInputProps, ITagInputSta
         let activeIndexToEmit = activeIndex;
 
         if (event.which === Keys.ENTER && value.length > 0) {
-            this.addTags(value);
+            this.addTags(value, TagInputAddMethod.DEFAULT);
         } else if (selectionEnd === 0 && this.props.values.length > 0) {
             // cursor at beginning of input allows interaction with tags.
             // use selectionEnd to verify cursor position and no text selection.
@@ -405,7 +425,7 @@ export class TagInput extends AbstractPureComponent<ITagInputProps, ITagInputSta
         }
 
         event.preventDefault();
-        this.addTags(value);
+        this.addTags(value, TagInputAddMethod.PASTE);
     };
 
     private handleRemoveTag = (event: React.MouseEvent<HTMLSpanElement>) => {
--- a/packages/core/src/components/tag-input/tagInput.tsx
+++ b/packages/core/src/components/tag-input/tagInput.tsx
@@ -16,24 +16,15 @@ import { Icon, IconName } from "../icon/icon";
 import { ITagProps, Tag } from "../tag/tag";
 
 /**
- * An enumeration of methods in which a `TagInput` value might have been added.
+ * A type reflecting the manner in which a `TagInput` value was added.
+ * - `"default"` - indicates that a value was added in the default fasion, via
+ *   manual selection.
+ * - `"blur"` - indicates that a value was added when the `TagInput` lost focus.
+ *   This is only possible when `addOnBlur=true`.
+ * - `"paste"` - indicates that a value was added via paste. This is only
+ *   possible when `addOnPaste=true`.
  */
-export enum TagInputAddMethod {
-    /** Indicates that a value was added in the default fasion, via manual selection. */
-    DEFAULT = "default",
-
-    /**
-     * Indicates that a value was added when the `TagInput` lost focus. This is
-     * only possible when `addOnBlur=true`.
-     */
-    BLUR = "blur",
-
-    /**
-     * Indicates that a value was added via paste. This is only possible when
-     * `addOnPaste=true`.
-     */
-    PASTE = "paste",
-}
+export type TagInputAddMethod = "default" | "blur" | "paste";
 
 export interface ITagInputProps extends IIntentProps, IProps {
     /**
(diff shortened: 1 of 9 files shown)


packages/core/src/components/tag-input/tagInput.tsx
```diff
@@ -15,6 +15,26 @@ import * as Utils from "../../common/utils";
 import { Icon, IconName } from "../icon/icon";
 import { ITagProps, Tag } from "../tag/tag";
 
+/**
+ * An enumeration of methods in which a `TagInput` value might have been added.
+ */
+export enum TagInputAddMethod {
```
giladgray: I would prefer the `const/type` approach so users can use string literals instead of importing this long-named type in order to `switch` on it.

packages/select/src/common/listItemsProps.ts
```diff
@@ -34,6 +34,13 @@ export interface IListItemsProps<T> extends IProps {
     /** Array of items in the list. */
     items: T[];
 
+    /**
+     * Whether the provided item fully matches the provided value. This is used
+     * to match pasted values to existing items, so that items can be emitted
+     * via `onItemsPaste`.
+     */
+    itemEqualsQuery?: (item: T, query: string) => boolean;
```
giladgray: isn't this just `itemPredicate`? https://github.com/palantir/blueprint/blob/develop/packages/select/src/common/predicate.ts#L23

i don't see the need for a new prop here, except that it slightly changes the semantics of `itemPredicate` in that you may want to define `itemListPredicate` for fast querying _and_ `itemPredicate` for pasting, instead of one-or-the-other.

cmslewis: @giladgray They're unfortunately different:
- `itemPredicate` is for filtering: an item matches if the query matches any substring of the item title (e.g.)
- `itemEqualsQuery` is for exact matching: an item matches iff the query === the full item title (e.g.) 

Possible to add an `exactMatch` flag or something in `ItemPredicate<T>`?

packages/core/src/components/tag-input/tagInput.tsx
cmslewis: @giladgray Product teams have requested string-literal enums in the past, but happy to do a simpler type. Is this what you mean?

```tsx
type TagInputAddMethod = "default" | "blur" | "paste";
```

Or perhaps just inline it?

```tsx
onAdd?: (values: string[], method: "default" | "blur" | "paste") => boolean | void;
```

adidahiya: I like the first one,

```
export type TagInputAddMethod = "default" | "blur" | "paste";
```

cmslewis: Done

packages/core/src/components/tag-input/tagInput.tsx
```diff
@@ -15,6 +15,17 @@ import * as Utils from "../../common/utils";
 import { Icon, IconName } from "../icon/icon";
 import { ITagProps, Tag } from "../tag/tag";
 
+/**
+ * A type reflecting the manner in which a `TagInput` value was added.
+ * - `"default"` - indicates that a value was added in the default fasion, via
```
adidahiya: nit: "fashion"

actually I would just simplify, remove the metaphor, and say "... was added by manual selection"

packages/core/src/components/tag-input/tagInput.tsx
```diff
@@ -15,6 +15,17 @@ import * as Utils from "../../common/utils";
 import { Icon, IconName } from "../icon/icon";
 import { ITagProps, Tag } from "../tag/tag";
 
+/**
+ * A type reflecting the manner in which a `TagInput` value was added.
```
adidahiya: nit: instead of "reflecting the manner ...", simpler language would be "The method in which a TagInput value was added."

packages/select/src/components/query-list/queryList.tsx
```diff
@@ -54,6 +54,24 @@ export interface IQueryListRendererProps<T>  // Omit `createNewItem`, because it
...
+     * `onItemsPaste` with those found items. Each pasted value that does not
+     * exactly match an item will be ignored.
+     *
+     * If creating items is enabled (by providing both `createNewItemFromQuery`
+     * and `createNewItemRenderer`), then pasted values that do not exactly
+     * match an existing item will emit a new item as created via
+     * `createNewItemFromQuery`.
+     *
+     * If `itemEqualsQuery` returns multiple matching items for a particular
+     * `value`, then only the first matching item will be emitted.
```
adidahiya: do you mean "for a particular query"? if not, what is `value` here?

packages/select/src/components/query-list/queryList.tsx
```diff
@@ -350,6 +369,40 @@ export class QueryList<T> extends React.Component<IQueryListProps<T>, IQueryList
...
+            return agg;
+        }, []);
+
+        // UX nicety: update the active item if we matched with at least one
+        // existing item.
+        if (nextActiveItem !== undefined) {
+            this.setActiveItem(nextActiveItem);
+        }
+
+        // No need to update the active item.
```
adidahiya: what's this comment about? you _may_ have updated the active item at this point, no?

packages/select/src/components/query-list/queryList.tsx
```diff
@@ -350,6 +369,40 @@ export class QueryList<T> extends React.Component<IQueryListProps<T>, IQueryList
...
 
+    private handlePaste = (values: string[]) => {
+        const { createNewItemFromQuery, onItemsPaste } = this.props;
+
+        let nextActiveItem: T | undefined;
+
+        // Find an exising item that exactly matches each pasted value, or
+        // create a new item if possible. Ignore unmatched values if creating
+        // items is disabled.
+        const pastedItemsToEmit = values.reduce<T[]>((agg, value) => {
```
adidahiya: nit: I would slightly prefer this as a regular for loop, since that would be easier to read than a `reduce` which produces a list

packages/select/src/components/query-list/queryList.tsx
```diff
@@ -443,6 +498,19 @@ function pxToNumber(value: string | null) {
     return value == null ? 0 : parseInt(value.slice(0, -2), 10);
 }
 
+function getMatchingItem<T>(query: string, { items, itemEqualsQuery }: IQueryListProps<T>): T | undefined {
+    if (Utils.isFunction(itemEqualsQuery)) {
+        // .find() doesn't exist in ES5. Alternative: use a for loop instead of
```
adidahiya: AFAIK, `Array.prototype.find` should be supported in blueprint's environment.. can you elaborate here why that doesn't work?

adidahiya:
An `exactMatch` flag for `itemPredicate` does sound like it would work a little better, have you tried that? I think you should

packages/select/src/components/query-list/queryList.tsx
cmslewis: Old cruft. Removed. 👍

packages/select/src/components/query-list/queryList.tsx
cmslewis: @adidahiya Yep, the compile step was failing on Circle. [Here is the failing build](https://circleci.com/gh/palantir/blueprint/24768?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link) from a few commits back.

cmslewis:
@adidahiya Thanks for the CR; I addressed your comments. The most recent commit includes a crack at using `itemPredicate` with an `exactMatch` flag. Two minor downsides are that the description for `itemPredicate` is now fairly complex, and attempting to handling pasting but not filtering will cause `QueryList` to invoke `itemPredicate` while typing a query when it otherwise wouldn't have. But that's probably a really uncommon case.

packages/select/src/components/query-list/queryList.tsx
adidahiya: that's unfortunate, I wish we had added it to the environment requirements. oh well.

packages/select/src/common/predicate.ts
```diff
@@ -14,10 +14,20 @@
 export type ItemListPredicate<T> = (query: string, items: T[]) => T[];
 
 /**
- * Customize querying of individual items. Return `true` to keep the item, `false` to hide.
- * This method will be invoked once for each item, so it should be performant. For more complex
- * queries, use `itemListPredicate` to operate once on the entire array.
+ * Customize querying of individual items.
  *
- * If defined with `itemListPredicate`, this prop will be ignored.
+ * __Filtering a list of items.__ This function is invoked to filter the list of
```
adidahiya: it's weird to have the exact same documentation block in two places. it's especially weird to reference the prop names here, when we're not in a react component class module... can you simplify this to a simple one-sentence description and tell people to "see its usage sites in `IListItemsProps` ..." ?

cmslewis:
@adidahiya Done! Now concatenating all unrecognized values into the input query on paste (only if the "Create item" option isn't enabled):

PR #3471, opened 2019-04-09
[datetime] Fix DateInput under Internet Explorer 11

gjvoosten:
#### Checklist

- [ ] Includes tests
- [ ] Update documentation

Add support for IE11 by using this check: https://stackoverflow.com/a/49325196
Tested under IE11, Firefox (60 ESR) and Chrome (73).

Note: testing under IE11 was slightly hindered by the fact that `yarn dev` (or `yarn dev:datetime`) gives an error: 
Screenshot - 090419 - 15:06:34
so I had to test it using our own application. And while testing the fix under Chrome, it turned out that `this.popoverContentEl` was sometimes `null` (hence the comment and the extra check).
Also, `yarn.lock` in `develop` appears out of date.

### DIFF
--- a/packages/datetime/src/dateInput.tsx
+++ b/packages/datetime/src/dateInput.tsx
@@ -379,8 +379,16 @@ export class DateInput extends AbstractPureComponent<IDateInputProps, IDateInput
 
     // focus DOM event listener (not React event)
     private handlePopoverBlur = (e: FocusEvent) => {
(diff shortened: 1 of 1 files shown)


gjvoosten:
'Interesting' failure in the [lint build](https://circleci.com/gh/palantir/blueprint/25126), but not much I can do about that I'm afraid…

PR #3484, opened 2019-04-17
Fix for NumericInput throws type error when value=null #3479

### LINKED ISSUE #3479: NumericInput throws type error when value=null
#### Environment

- __Package version(s)__: @blueprintjs/core 3.15.1
- __Browser and OS versions__: Chrome 73.0.3683.103, Windows 7

__Minimal repro__: https://codesandbox.io/s/j7r4p4kpy3

#### Steps to reproduce

1. Set the value of a NumericInput to null

#### Actual behavior

Component throws "TypeError Cannot read property 'toString' of null"

#### Expected behavior

Display an empty NumericInput box.

I was using blueprintjs/core 3.7.0, and null values would result in an empty numeric input box just as I expected. When I updated to blueprintjs/core 3.15.1, I started seeing the type error.

The [Controlled Mode](https://blueprintjs.com/docs/#core/components/numeric-input.controlled-mode) example initializes the NumericInput value to null. So I expected it to work. But now I see that the documentation for the `value `prop clearly states it's type is `number|string` which excludes null values. So this is arguably a usage error on my part.

#### Possible solution

Maybe update the example in the documentation to initialize the value to "" instead of null. The example also has some syntax errors (missing => in the handleValueChange function definition).

ibk6:
#### Checklist

- Updated documentation

Made initial state of NumericInput value as NumericInput.VALUE_EMPTY instead of null

NumericInput documentation update

### DIFF
--- a/packages/core/src/components/forms/numeric-input.md
+++ b/packages/core/src/components/forms/numeric-input.md
@@ -106,7 +106,7 @@ import * as SomeLibrary from "some-library";
 
 export class NumericInputExample extends React.Component<{}, { value?: number |
 string }> {
-    public state = { value: null };
+    public state = { value: NumericInput.VALUE_EMPTY };
 
     public render() {
         return (


palantirtech:
Thanks for your interest in [palantir/blueprint](https://github.com/palantir/blueprint), @ibk6! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

PR #3560, opened 2019-05-21
Add FileInput buttonText prop

thepocp:
#### Checklist

- [ ] Includes tests
- [x] Update documentation

Unfortunately, I do not know how to test this functionality. You can only check the value of the data attribute, but not the text itself.

Added a prop for FileInput, which changes the text on the button. This is necessary, for example, for the i18n.


packages/core/src/components/forms/fileInput.tsx
```diff
@@ -108,11 +128,25 @@ export class FileInput extends React.PureComponent<IFileInputProps, {}> {
...
 
+    private createRef = (input: HTMLSpanElement | null) => {
+        if (input !== null) {
+            this.input = input;
+        }
+    };
+
+    private changeButtonText = (text: string) => {
+        if (this.input) {
+            this.input.setAttribute("button-text", text);
```
adidahiya: we should namespace the attribute name, just to be safe
```suggestion
            this.input.setAttribute(`${NS}-button-text`, text);
```

... this would also require exporting `NS` from `packages/core/src/common/classes.ts` and importing that constant here

packages/core/src/components/forms/_file-input.scss
```diff
@@ -143,6 +143,11 @@ $file-input-button-padding-large: ($pt-input-height-large - $pt-button-height) /
   }
 }
 
+#blueprint-documentation [data-example-id="file-input"] .#{$ns}-file-upload-input::after {
+  content: "Browse";
```
adidahiya: why do you need this? shouldn't it be set as the default in the JS component?

adidahiya:
[docs preview](https://25317-71939872-gh.circle-artifacts.com/0/home/circleci/project/packages/docs-app/dist/index.html#core/components/file-input)

also one small problem: the button should expand to accomodate the text if possible, and have a max width at which it truncates (with `Classes.TEXT_OVERFLOW_ELLIPSIS`)

image

packages/core/src/components/forms/_file-input.scss
adidahiya: I still don't think this block is needed, it doesn't seem to do anything

packages/core/src/components/forms/file-input.md
```diff
@@ -3,6 +3,9 @@
 Use the standard `input type="file"` along with a `span` with class `@ns-file-upload-input`.
 Wrap that all in a `label` with class `@ns-file-input`.
```
adidahiya: can you please move this paragraph to the CSS section? right around L26

adidahiya:
[latest docs preview](https://circleci.com/gh/palantir/blueprint/25770/artifacts/0//home/circleci/project/packages/docs-app/dist/index.html)

packages/core/src/components/forms/_file-input.scss
thepocp: I added this to this block. Attr function has the fallback parameter, but it's not supported by browsers. How I can set style only for docs?
Screenshot 2019-06-18 at 23 45 07

adidahiya: `attr()` is supported for the `content` property by all the browsers we care about: https://caniuse.com/#feat=css-gencontent

not sure what you're trying to achieve here? this kind of selector _would_ be the right way to do docs-only styles, but there are a couple issues

- the styles should go in docs-app along with the other overrides; not in blueprint core
- the `data-example-id` selector is wrong

and I still think you don't need it overall

thepocp: I meant the second argument of the function attr (fallback)
If you don't use this style, the text is not displayed:
Screenshot 2019-06-19 at 00 13 30
I moved rule in docs styles

adidahiya: hmm, that's not good then... we don't want all users to have to add a style override just to get things working out of the box... and unfortunately I don't want to introduce a breaking change to the CSS API where it's _required_ to supply a `bp3-button-text` attribute in order to get the text to render...

can we try using multiple CSS statements to achieve the fallback? if not, we might have to delay merging this PR as it could be a breaking change...

```css
content: "Browse";
content: attr(#{$ns}-button-text);
```

thepocp: Maybe I wrote something wrong. Sorry for my English.
This rule is only needed for the CSS block in the documentation. In other places it isn't needed and everything works well.

thepocp: If you write like this, the second rule will overwrite the first. And there will be an empty string
```css
content: "Browse";
content: attr (#{$ns}-button-text);
```

adidahiya: Your English is fine, no need to apologize :)

> This rule is only needed for the CSS block in the documentation.

Unfortunately, this is not true for this library because the CSS-only styling API is part of Blueprint's public API (in addition to the JS components).

Some users render this markup to get a working file input with the word "Browse":

```html
<label class="bp3-file-input">
  <input type="file"/>
  <span class="bp3-file-upload-input">Choose file...</span>
</label>
```

We don't want to remove support for this, otherwise it would be a regression... even though I understand that it's a minor syntax migration to change the API to this instead:

```html
<label class="bp3-file-input">
  <input type="file"/>
  <span class="bp3-file-upload-input" bp3-button-text="Browse">Choose file...</span>
</label>
```

... it is still a _breaking change_.

adidahiya:
See my comments above about the CSS breaking change.

You can adjust PR, though, to make the implementation non-breaking. Try adding the new `attr()` style declaration with another class which is only added by the JS component (perhaps `.bp3-file-upload-input-text-content`) which overrides the CSS-only API (`.bp3-file-upload-input`).

packages/core/src/components/forms/_file-input.scss
thepocp: I solved the problem. The CSS is just the string `content: "Borwse"`. When creating a component, a new style is created.
But for the documentation I had to leave the style, because the new style is applied.

packages/core/src/components/forms/fileInput.tsx
```diff
@@ -135,6 +139,14 @@ export class FileInput extends React.PureComponent<IFileInputProps, {}> {
         );
     }
 
+    private createSheet = () => {
```
adidahiya: I don't want components to directly create new sheets like this... I think you can achieve the same result by conditionally adding a class name in the `render()` method... and style that class declaratively in `_file-input.scss`

thepocp: Thanks, I did too hard. Now everything seems fine

PR #3568, opened 2019-05-25
[core] Fix number of clicks required to use HTMLSelect inside ControlGroup in Firefox

### LINKED ISSUE #3554: HTMLSelect doesn't work correctly inside a ControlGroup in Firefox
#### Environment

- __Browser and OS versions__: Firefox 67

Minimal Repro
https://blueprintjs.com/docs/#core/components/control-group
https://codesandbox.io/s/3x54ryj0om

#### Steps to reproduce

1. Open the page in Firefox
2. Try click on the HTMLSelect to open it (The one in a ControlGroup)
3. Discover you need to click it twice instead of once when using Firefox

#### Actual behavior

It takes two clicks to open an HTMLSelect inside a ControlGroup when using Firefox (Opens in a single click in Chrome, or when not in a ControlGroup)

#### Expected behavior

Should only take one click to open.

ibk6:
Removed `position: relative;` which was there to create a new stacking context.

Not sure why this caused the issue but there is already `z-index` to create a stacking context, `position: relative` may not be needed.

### DIFF
--- a/packages/core/src/components/forms/_control-group.scss
+++ b/packages/core/src/components/forms/_control-group.scss
@@ -106,7 +106,6 @@ Styleguide control-group
 
     &:focus {
       // establish new stacking context so focus state covers neighbors
-      position: relative;
       z-index: index($control-group-stack, "button-focus");
     }
 
--- a/packages/core/src/components/forms/_control-group.scss
+++ b/packages/core/src/components/forms/_control-group.scss
@@ -100,12 +100,12 @@ Styleguide control-group
   .#{$ns}-button,
   .#{$ns}-html-select select,
   .#{$ns}-select select {
+    @include new-render-layer();
     z-index: index($control-group-stack, "button-default");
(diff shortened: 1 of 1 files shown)


packages/core/src/components/forms/_control-group.scss
```diff
@@ -106,7 +106,6 @@ Styleguide control-group
 
     &:focus {
       // establish new stacking context so focus state covers neighbors
-      position: relative;
```
adidahiya: hm, what do you mean there was already a new stacking context? I don't see one for these selectors (L100-102).

we probably want to try removing this line (and the commend above) and adding `@include new-render-layer();` right after the `.#{$ns}-select select {` line (L102)

ibk6: Got it!

danzel:
Thank you :)

PR #3591, opened 2019-06-12
Fix TagInput's Tags overflowing the container

n1xx1:
#### Checklist

- [ ] Includes tests
- [ ] Update documentation

Flexboxes default values for `min-width` is always `auto`. This means that that the size for that element will be at least its contents size.

This problem shows in the TagInput component where if the tag is too large it's going to overflow the container ignoring the parent width. The fix for that is just setting min-width to be 0 in the container (tag-input-values).

#### Screenshot

Before: 
image

After:
image


palantirtech:
Thanks for your interest in [palantir/blueprint](https://github.com/palantir/blueprint), @n1xx1! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

PR #3601, opened 2019-06-14
[core] Add controlled mode to PanelStack

invliD:
Allows working around #3173. I personally need this to start with multiple panels.

#### Checklist

- [x] Includes tests
- [x] Update documentation

Allow controlling the panel stack from the outside to have greater control over the exact panels rendered.


ailon:
Is this supposed to react to panels being added to the supplied through the `stack` prop? Can't seem to make it work. Or is this just for the initial `stack` of multiple panels?

adidahiya:
@ailon it's a pure component, so if you `push` elements into the `stack` array, it will not re-render. you will need to create a new cloned array to make the react component re-render

ailon:
Thanks, @adidahiya! That works.

PR #3608, opened 2019-06-17
Update abstract button props to match react's.

maclockard:
#### Checklist

- [x] ~Includes tests~ should be a compile time only change
- [x] Update documentation

React recently updated the typings for HTML button's `type` field from `string` to a more limited union. See: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/index.d.ts#L1848 for the current types. 

This is currently an issue for folks on newer versions of react typings since `IButtonProps` is no longer assignable to the props for `Button`.

### DIFF
--- a/packages/core/src/components/button/abstractButton.tsx
+++ b/packages/core/src/components/button/abstractButton.tsx
@@ -68,11 +68,11 @@ export interface IButtonProps extends IActionProps {
     small?: boolean;
 
     /**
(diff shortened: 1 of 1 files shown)


maclockard:
I think another option here would be to simply drop `type` from `IButtonProps` and leave it for the react typings. Trade off here being the documentation for how to properly use `Button` gets slightly worse, but the typings will work, regardless of patch version of react

PR #3613, opened 2019-06-19
[Datetime] Fix typing values into TimePicker's fields

### LINKED ISSUE #3581: Impossible to enter two-digit value in minutes field of DateTimePicker
#### Environment

[Documentation Example of Time Picker](https://blueprintjs.com/docs/#datetime/timepicker)

#### Steps to reproduce

1. Jump to the minutes in the example
2. Enter a two-digit value (i.e. 25)

#### Actual behavior

The value is 02. It is impossible to enter 25.

#### Expected behavior

A valid, two-digit number for the minutes should be enterable.

jacekjagiello:
#### Checklist

- [ ] Includes tests
- [ ] Update documentation

I couldn't find any way to write a test case for this. Changing input via `TestUtils.Simulate.change()` does not solve the problem since it enters a whole value. I've also tried with `TestUtils.Simulate.keyDown()` but no luck. I didn't have enough time to think about it, so I'll decide to open a PR. Maybe someone else has an idea how to test this?

fixed-time-picker

### DIFF
--- a/packages/datetime/src/timePicker.tsx
+++ b/packages/datetime/src/timePicker.tsx
@@ -263,28 +263,8 @@ export class TimePicker extends React.Component<ITimePickerProps, ITimePickerSta
     };
 
     private getInputChangeHandler = (unit: TimeUnit) => (e: React.SyntheticEvent<HTMLInputElement>) => {
-        const TWO_DIGITS = /^\d{0,2}$/;
-        const THREE_DIGITS = /^\d{0,3}$/;
         const text = getStringValueFromInputEvent(e);
-
-        let isValid = false;
-        switch (unit) {
-            case TimeUnit.HOUR_24:
-            case TimeUnit.HOUR_12:
-            case TimeUnit.MINUTE:
-            case TimeUnit.SECOND:
-                isValid = TWO_DIGITS.test(text);
-                break;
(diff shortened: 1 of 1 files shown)


packages/datetime/src/timePicker.tsx
```diff
@@ -263,28 +263,8 @@ export class TimePicker extends React.Component<ITimePickerProps, ITimePickerSta
...
-            case TimeUnit.MS:
-                isValid = THREE_DIGITS.test(text);
-                break;
-            default:
-                throw Error("Invalid TimeUnit");
-        }
-
-        if (isValid) {
-            this.updateTime(parseInt(text, 10), unit);
-        }
```
jacekjagiello: I think this is not needed since similar validation is done by `isTimeUnitValid()` at later point. I decided to remove it, tests passed and TimePicker start to work correctly.

adidahiya:
[docs preview from this PR](https://circleci.com/gh/palantir/blueprint/25788/artifacts/0//home/circleci/project/packages/docs-app/dist/index.html)

adidahiya:
I couldn't find a good way to test the keypress behavior; synthetic React events in the test environment don't propagate fully in such a way that a `keyDown` simulated event results in changing the input value...

PR #3628, opened 2019-06-27
[datetime] feat(DatePicker): add option to highlight current day

### LINKED ISSUE #3583: Datetime controls: possibility to highlight "today" in the calendar
#### Environment

- __Package version(s)__: v3
- __Browser and OS versions__: Chrome 74.0.3729.169

#### Feature request

possibility to show "today" highlighted in the various date pickers.
Even when the date field is not yet evaluated. 
The image below shows how I would like to see the date selector if today was June 7th 2019 and the "HighlightToday" property was active:
wanted
the image below shows how I would like to see the date picker if the "HighlightToday" property is not active:
actual

thanks!

tanmoyopenroot:
#### Checklist

- [x] Includes tests
- [x] Update documentation

Added `highlightCurrentDay` prop to toggle whether the current day should be highlighted in the calendar.
default prop value => `false`

### DIFF
--- a/packages/datetime/src/_datepicker.scss
+++ b/packages/datetime/src/_datepicker.scss
@@ -97,6 +97,10 @@ $header-margin: ($header-height - $pt-input-height) / 2 !default;
       color: $pt-text-color-disabled;
     }
 
+    &.DayPicker-Day--isToday {
+      outline: 1px solid $pt-divider-black;
+    }
+
     &:hover,
     &:focus {
       background: $datepicker-day-background-color-hover;
@@ -194,6 +198,10 @@ $header-margin: ($header-height - $pt-input-height) / 2 !default;
   .DayPicker-Day {
     &.DayPicker-Day--outside {
       color: $pt-dark-text-color-disabled;
+    }
+
+    &.DayPicker-Day--isToday {
+      outline: 1px solid $pt-dark-divider-white;
     }
 
     &:hover,
--- a/packages/datetime/src/_datepicker.scss
+++ b/packages/datetime/src/_datepicker.scss
@@ -98,7 +98,11 @@ $header-margin: ($header-height - $pt-input-height) / 2 !default;
     }
 
     &.DayPicker-Day--isToday {
-      outline: 1px solid $pt-divider-black;
+      span {
+        border: 1px solid $pt-divider-black;
+        border-radius: $pt-border-radius;
+        padding: 6px;
+      }
     }
 
     &:hover,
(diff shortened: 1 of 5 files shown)


adidahiya:
This is a good start... it's unfortunate that we can't use `border` because of the table layout... `outline` is likely the right approach

We need to make sure focus styles are still accessible / fit in with the rest of the styling. Check out the beginning and end of this gif:



You can try it out yourself with tab / arrow key interactions

tanmoyopenroot:
Done.

packages/datetime/src/_datepicker.scss
```diff
@@ -97,6 +97,14 @@ $header-margin: ($header-height - $pt-input-height) / 2 !default;
       color: $pt-text-color-disabled;
     }
 
+    &.DayPicker-Day--isToday {
+      span {
```
adidahiya: we try to avoid styling tag names. please create a new class instead... 

```ts
const DATEPICKER_DAY_WRAPPER = `${DATEPICKER}-day-wrapper`;
```

and if you make this a `div`, it will get the right block styling to take up the full date box

packages/datetime/src/_datepicker.scss
```diff
@@ -97,6 +97,14 @@ $header-margin: ($header-height - $pt-input-height) / 2 !default;
...
+        border: 1px solid $pt-divider-black;
+        border-radius: $pt-border-radius;
+        padding: 6px;
```
adidahiya: from my CSS tweaking in the inspector, I think this should be 7px.

otherwise, you currently get this:

image

adidahiya:
(meant to post this along with my review)

[latest docs preview](https://26000-71939872-gh.circle-artifacts.com/0/home/circleci/project/packages/docs-app/dist/index.html#datetime/datepicker)

yes, this is better, but the highlight box for single-digit days is not the right size:

image

packages/datetime/src/_datepicker.scss
tanmoyopenroot: Done

packages/datetime/src/_datepicker.scss
tanmoyopenroot: Done

adidahiya:
[docs preview](https://circleci.com/gh/palantir/blueprint/26139/artifacts/0//home/circleci/project/packages/docs-app/dist/index.html)

jmtame:
Great job @tanmoyopenroot

PR #3657, opened 2019-07-12
[core] docs(Toaster): fix position prop comment

maclockard:
According to https://github.com/palantir/blueprint/blob/06a186c90758bbdca604ed6d7bf639c3d05b1fa0/packages/core/src/components/toast/toaster.tsx#L91 only top and bottom are supported.


maclockard:
Ah, gotcha! sorry was confused. will update this to change that comment

maclockard:
@adidahiya this should be good to go now!

PR #3668, opened 2019-07-21
Shortcuts now check closeOnSelection to  decise shouldDismissPopover

### LINKED ISSUE #3338: DateRangePicker shortcuts incorrectly dismiss popovers

Yonben:
#### Checklist

- [x] Includes tests
- [ ] Update documentation

Correctly prevent DateRangePicker shortcuts from dismissing popovers

Components on the way to Shortcuts are already getting the `closeOnSelection` props. I added it to Shortcuts in order for it to know if it should pass `shouldDismissPopover` to the MenuItem (which used to manually get the className).
Also first real PR, so I feel code style etc.. needs some improvement and would love constructive inputs.

#### Screenshot
N/A

Thanks!

### DIFF
--- a/packages/datetime/src/dateRangePicker.tsx
+++ b/packages/datetime/src/dateRangePicker.tsx
@@ -108,6 +108,12 @@ export interface IDateRangePickerProps extends IDatePickerBaseProps, IProps {
      * If this prop is provided, the component acts in a controlled manner.
      */
     value?: DateRange;
+
+    /**
+     * Whether the calendar popover should close when a date range is fully selected.
+     * @default true
+     */
+    closeOnSelection?: boolean;
 }
 
 // leftView and rightView controls the DayPicker displayed month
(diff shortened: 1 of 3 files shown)


palantirtech:
Thanks for your interest in [palantir/blueprint](https://github.com/palantir/blueprint), @Yonben! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

Yonben:
@adidahiya I think that we should first fix the undesirable behaviour and then reiterate and discuss/change the API to fit the problem better. In the meantime, the component behaves incorrectly and that should be priority imo.

adidahiya:
> In the meantime, the component behaves incorrectly and that should be priority imo.

yes, but I don't want to do that if it means adding a new prop to the API. increasing the API surface area to fix the bug is not a good development tradeoff for this project.

PR #3688, opened 2019-08-05
[datetime] feat(DateRangePicker): add active menu styles to shortcuts

### LINKED ISSUE #3683: Using active menu style in DateRangeShortcut
#### Environment

- __Package version(s)__: @blueprintjs/core": 3.15.1, @blueprintjs/datetime": 3.9.1

#### Feature request

set active state for current DateRangeShortcut

#### Examples
image

tanmoyopenroot:
#### Checklist

- [x] Includes tests
- [ ] Update documentation

add `selectedShortcutIndex` and `onShortcutChange` props to `<DateRangePicker`

### DIFF
--- a/packages/datetime/src/dateRangePicker.tsx
+++ b/packages/datetime/src/dateRangePicker.tsx
@@ -100,9 +100,9 @@ export interface IDateRangePickerProps extends IDatePickerBaseProps, IProps {
 
     /**
      * Called when the `shortcuts` props is enabled and the user changes the shortcut.
-     * When triggered, it will pass the index of the selected `shortcut`.
+     * When triggered, it will pass the selected `shortcut` and its index which is optional.
      */
-    onShortcutChange?: (index: number) => void;
+    onShortcutChange?: (shortcut: IDateRangeShortcut, index?: number) => void;
 
     /**
      * Whether shortcuts to quickly select a range of dates are displayed or not.
@@ -115,8 +115,7 @@ export interface IDateRangePickerProps extends IDatePickerBaseProps, IProps {
 
     /**
      * The currently selected `shortcut`.
-     * It should be used only when `shortcuts` prop is provided
-     * @default -1
+     * If this prop is provided, the component acts in a controlled manner.
      */
     selectedShortcutIndex?: number;
 
@@ -140,6 +139,7 @@ export interface IDateRangePickerState {
     rightView?: MonthAndYear;
     value?: DateRange;
     time?: DateRange;
+    selectedShortcutIndex?: number;
 }
 
 export class DateRangePicker extends AbstractPureComponent<IDateRangePickerProps, IDateRangePickerState> {
@@ -150,7 +150,6 @@ export class DateRangePicker extends AbstractPureComponent<IDateRangePickerProps
         maxDate: getDefaultMaxDate(),
         minDate: getDefaultMinDate(),
         reverseMonthAndYearMenus: false,
-        selectedShortcutIndex: -1,
         shortcuts: true,
         singleMonthOnly: false,
         timePickerProps: {},
@@ -221,7 +220,14 @@ export class DateRangePicker extends AbstractPureComponent<IDateRangePickerProps
             !props.contiguousCalendarMonths && rightDate != null && !DateUtils.areSameMonth(initialMonth, rightDate)
                 ? MonthAndYear.fromDate(rightDate)
                 : leftView.getNextMonth();
-        this.state = { leftView, rightView, value, hoverValue: [null, null], time };
+        this.state = {
+            hoverValue: [null, null],
+            leftView,
+            rightView,
+            selectedShortcutIndex: this.props.selectedShortcutIndex || -1,
+            time,
+            value,
+        };
     }
 
     public render() {
@@ -259,6 +265,10 @@ export class DateRangePicker extends AbstractPureComponent<IDateRangePickerProps
                 nextProps.contiguousCalendarMonths,
             );
             this.setState(nextState);
+        }
+
+        if (this.props.selectedShortcutIndex !== nextProps.selectedShortcutIndex) {
+            this.setState({ selectedShortcutIndex: nextProps.selectedShortcutIndex });
         }
     }
 
@@ -303,7 +313,8 @@ export class DateRangePicker extends AbstractPureComponent<IDateRangePickerProps
             return null;
         }
 
-        const { allowSingleDayRange, maxDate, minDate, timePrecision, selectedShortcutIndex } = this.props;
+        const { selectedShortcutIndex } = this.state;
+        const { allowSingleDayRange, maxDate, minDate, timePrecision } = this.props;
         return [
             <Shortcuts
                 key="shortcuts"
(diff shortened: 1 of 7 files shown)


tanmoyopenroot:
> sorry for the delayed review... this mostly looks good but I don't think it needs to be controlled behavior... users should be able to pick up this new styling behavior for free without adding new props

Its ok nothing to worry about.
And about the changes requested, I will look into it.
And thanks for the feedback

packages/datetime/test/dateRangePickerTests.tsx
```diff
@@ -912,6 +913,43 @@ describe("<DateRangePicker>", () => {
...
+        it("shortcuts with active menu style", () => {
+            const selectedShortcut = 1;
+            const { wrapper } = render({ selectedShortcutIndex: selectedShortcut });
+
+            assert.isTrue(
+                wrapper
+                    .find(Shortcuts)
+                    .find(Menu)
+                    .find(`.${Classes.ACTIVE}`)
+                    .exists(),
```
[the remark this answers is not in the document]
tanmoyopenroot: Done

packages/datetime/test/dateRangeInputTests.tsx
```diff
@@ -399,6 +399,19 @@ describe("<DateRangeInput>", () => {
         expect(root.find(DateRangePicker).prop("shortcuts")).to.be.false;
     });
 
+    it("should set the selectedShortcutIndex state when when click on the shortcut", () => {
```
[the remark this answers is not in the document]
tanmoyopenroot: Done

packages/datetime/test/dateRangePickerTests.tsx
```diff
@@ -912,6 +913,43 @@ describe("<DateRangePicker>", () => {
             assert.isTrue(DateUtils.areSameDay(tomorrow, value[1]));
         });
 
+        it("shortcuts without active menu style", () => {
```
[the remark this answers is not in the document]
tanmoyopenroot: Done

packages/datetime/test/dateRangePickerTests.tsx
```diff
@@ -912,6 +913,43 @@ describe("<DateRangePicker>", () => {
...
+            assert.isFalse(
+                wrapper
+                    .find(Shortcuts)
+                    .find(Menu)
+                    .find(`.${Classes.ACTIVE}`)
+                    .exists(),
+            );
+        });
+
+        it("shortcuts with active menu style", () => {
```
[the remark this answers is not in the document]
tanmoyopenroot: Done

packages/docs-app/src/examples/datetime-examples/dateRangePickerExample.tsx
```diff
@@ -112,6 +114,7 @@ export class DateRangePickerExample extends React.PureComponent<IExampleProps, I
                     maxDate={maxDate}
                     minDate={minDate}
                     onChange={this.handleDateChange}
+                    onShortcutChange={this.handleShortcutChange}
```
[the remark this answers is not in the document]
tanmoyopenroot: Done

tanmoyopenroot:
>  I don't think it needs to be controlled behavior... users should be able to pick up this new styling behavior for free without adding new props

So now <DateRangePicker can be `controlled` and `uncontrolled` based on `selectedShortcutIndex` prop

packages/datetime/src/dateRangePicker.tsx
```diff
@@ -98,6 +98,12 @@ export interface IDateRangePickerProps extends IDatePickerBaseProps, IProps {
      */
     onHoverChange?: (hoveredDates: DateRange, hoveredDay: Date, hoveredBoundary: Boundary) => void;
 
+    /**
+     * Called when the `shortcuts` props is enabled and the user changes the shortcut.
+     * When triggered, it will pass the selected `shortcut` and its index which is optional.
+     */
+    onShortcutChange?: (shortcut: IDateRangeShortcut, index?: number) => void;
```
adidahiya: why is index optional? it seems like the Shortcuts component always passes it through the callback

tanmoyopenroot: I think usually when a `shortcut` is changed user would like the details of the clicked `shortcut` rather than the `index`.
So it's optional.
If it is required then the user can pass the optional `index` parameter.
Any thoughts?

adidahiya: but that's not how the typedef should be for the prop -- the library is _always_ passing `index`, so it shouldn't be marked as optional. a user can choose to assign a function which does or does not use the `index` argument, but they shouldn't have to deal with the case where `index` is possibly `undefined`.

if you make this line:
```suggestion
    onShortcutChange?: (shortcut: IDateRangeShortcut, index: number) => void;
```

then both of these are assignable as props by the user:
- `(shortcut: IDateRangeShortcut, index: number) => { ... }`
- `(shortcut: IDateRangeShortcut) => { ... }`

tanmoyopenroot: oh ok.
I get it now.
Thanks.

tanmoyopenroot: Done

packages/datetime/src/dateRangePicker.tsx
```diff
@@ -98,6 +98,12 @@ export interface IDateRangePickerProps extends IDatePickerBaseProps, IProps {
      */
     onHoverChange?: (hoveredDates: DateRange, hoveredDay: Date, hoveredBoundary: Boundary) => void;
 
+    /**
+     * Called when the `shortcuts` props is enabled and the user changes the shortcut.
+     * When triggered, it will pass the selected `shortcut` and its index which is optional.
```
adidahiya:
```suggestion
```

no need for this line, it's redundant with the typedef

packages/datetime/src/dateRangePicker.tsx
```diff
@@ -107,6 +113,12 @@ export interface IDateRangePickerProps extends IDatePickerBaseProps, IProps {
      */
     shortcuts?: boolean | IDateRangeShortcut[];
 
+    /**
+     * The currently selected `shortcut`.
```
adidahiya:
```suggestion
     * The currently selected shortcut.
```

adidahiya:
Just a couple minor docs suggestions and then this is ready to go! Thanks for the PR

packages/datetime/src/dateRangePicker.tsx
tanmoyopenroot: Done

packages/datetime/src/dateRangePicker.tsx
tanmoyopenroot: Done

ajmagic:
Could these same props be exposed to DateRangeInput as well?

ajmagic:
Can the same props be exposed on Data Range Input as well?

adidahiya:
@ajmagic what is your use case? what do you want to do with the new props? please open a new issue to discuss.

PR #3702, opened 2019-08-15
Refactor to React 16.3+ lifecycle methods

### LINKED ISSUE #3361: Support React Strict Mode
#### Environment

- __Package version(s)__: @blueprintjs/core: 3.13.0

#### Feature request
Support React Strict Mode to remove unsafe lifecycles, etc.
See: https://reactjs.org/docs/strict-mode.html

#### Examples
Currently when wrapping Blueprint components with `<React.StrictMode>`, several warnings are exposed as console errors:
```
Warning: findDOMNode is deprecated in StrictMode. findDOMNode was passed an instance of Blueprint3.ResizeSensor which is inside StrictMode. Instead, add a ref directly to the element you want to reference.
```
```
componentWillReceiveProps: Please update the following components to use static getDerivedStateFromProps instead: Blueprint3.Overlay, Blueprint3.Popover
```
```
componentWillReceiveProps: Please update the following components to use static getDerivedStateFromProps instead: Blueprint3.Spinner,
```
(There are probably others)

tobilen:
#### To Do:

- [x] Refactor `core` package
- [x] Refactor `datetime` package to work with updated `core` package
- [x] Remove deprecated lifecycles from `datetime`
- [x] Refactor `table` package to work with updated `core` package
- [x] Remove deprecated lifecycles from `table`
- [x] Refactor `select` package to work with updated `core` package
- [x] Remove deprecated lifecycles from `select`
- [x] Refactor `timezone` package to work with updated `core` package
- [x] Remove deprecated lifecycles from `timezone`
- [x] Add lifecycle polyfill

#### Checklist

- [x] Includes tests
~- [ ] Update documentation~

React 17 will deprecate several lifecycle methods. While there is a workaround to keep using them with the `UNSAFE_` prefix, this will not work in older React versions. Migrating to the new lifecycle methods also enables us to make use of the polyfill plugin provided by the react team ( https://github.com/reactjs/react-lifecycles-compat )

This PR:
* replaces lifecycle methods
* adds polyfill plugin to support older react versions

Since i am changing behavior of the basic building blocks of blueprint, feel free to test interactions with any and all components that either extend abstract components, or use these building blocks of the `core` package:
* popover
* alert
* multi slider
* numeric input
* hotkeys
* inputTag

`Table` and `QueryList` had complex `willReceiveProps` handlers, so checking them in-depth is probably a good idea as well


tobilen:
*bump*

adidahiya:
@tobilen thanks for the PR, I appreciate the effort put into this and we definitely want to use a polyfill to continue supporting React 15. I will take a closer look at the code changes this week.

packages/core/src/common/abstractComponent.ts
```diff
@@ -21,26 +21,39 @@ import { isNodeEnv } from "./utils";
...
 
-    constructor(props?: P, context?: any) {
+    protected constructor(props?: P, context?: any) {
         super(props, context);
         if (!isNodeEnv("production")) {
             this.validateProps(this.props);
         }
     }
 
-    public componentWillReceiveProps(nextProps: P & { children?: React.ReactNode }) {
```
adidahiya: I think we'll need to keep this method around for backwards compatibility, since `AbstractComponent` is exported from `@blueprintjs/core`. Its implementation can stay as-is, just marked as `@deprecated` in a JSDoc comment.

packages/core/src/common/abstractPureComponent.ts
```diff
@@ -14,14 +14,14 @@
  * limitations under the License.
  */
 
-import * as React from "react";
-import { isNodeEnv } from "./utils";
+import * as React from 'react';
```
adidahiya: we use double quotes, this will fail linting

packages/core/src/common/abstractPureComponent.ts
```diff
@@ -30,14 +30,14 @@ export abstract class AbstractPureComponent<P, S = {}> extends React.PureCompone
 
     constructor(props?: P, context?: any) {
         super(props, context);
-        if (!isNodeEnv("production")) {
+        if (!isNodeEnv('production')) {
             this.validateProps(this.props);
         }
     }
 
-    public componentWillReceiveProps(nextProps: P & { children?: React.ReactNode }) {
```
adidahiya: same here about keeping this deprecated method around for backwards compatibility

packages/core/src/components/alert/alert.tsx
```diff
@@ -126,6 +126,7 @@ export class Alert extends AbstractPureComponent<IAlertProps, {}> {
     };
 
     public static displayName = `${DISPLAYNAME_PREFIX}.Alert`;
+    public state = {};
```
adidahiya: why is this necessary?

packages/core/src/components/collapse/collapse.tsx
```diff
@@ -230,3 +241,5 @@ export class Collapse extends AbstractPureComponent<ICollapseProps, ICollapseSta
         }
     }
 }
+
+polyfill(Collapse);
```
adidahiya: It's probably safer to do the polyfill before exporting the component... side-effectful statements like this are code smells.

```ts
export const Collapse = polyfill(class extends AbstractPureComponent<...> {
```

packages/core/src/common/abstractComponent.ts
adidahiya: after looking at the react-lifecycles-compat documentation, it seems like it will throw an error if `componentWillReceiveProps` is defined. So I propose that we keep around the old `AbstractComponent` but make a new version of this class with a new name which includes the changes in this diff and is used internally in Blueprint components with `polyfill()`

packages/core/src/components/hotkeys/hotkeys.tsx
```diff
@@ -41,7 +41,7 @@ export interface IHotkeysProps extends IProps {
     tabIndex?: number;
 }
 
-export class Hotkeys extends AbstractPureComponent<IHotkeysProps, {}> {
+export class Hotkeys extends AbstractPureComponent<IHotkeysProps, {}, {}> {
```
adidahiya: is this necessary? shouldn't it be inferred?

packages/core/src/components/slider/multiSlider.tsx
```diff
@@ -150,6 +160,10 @@ export class MultiSlider extends AbstractPureComponent<IMultiSliderProps, ISlide
 
     private trackElement: HTMLElement | null;
 
+    constructor(props: IMultiSliderProps) {
```
adidahiya: looks like an unnecessary constructor implementation

packages/core/test/slider/multiSliderTests.tsx
```diff
@@ -215,6 +215,25 @@ describe("<MultiSlider>", () => {
     });
 
     describe("validation", () => {
+        let expectedErrors = 0;
+        let actualErrors = 0;
+        function onError(e: ErrorEvent) {
```
adidahiya: I'm curious why these changes were necessary...?

packages/core/test/tag-input/tagInputTests.tsx
```diff
@@ -148,13 +148,14 @@ describe("<TagInput>", () => {
...
             wrapper.simulate("blur");
+
             // Need setTimeout here to wait for focus to change after blur event
             setTimeout(() => {
                 assert.isTrue(onAdd.calledOnce);
                 assert.deepEqual(onAdd.args[0][0], [NEW_VALUE]);
                 assert.equal(onAdd.args[0][1], "blur");
                 done();
-            });
+            }, 50);
```
adidahiya: same here, this seems a little suspect... I feel like there's a semantic difference between setTimeout without a timeout value ("just wait until the next event loop") vs. using an explicit timeout

packages/table/src/headers/header.tsx
```diff
@@ -239,8 +239,11 @@ export class Header extends React.Component<IInternalHeaderProps, IHeaderState>
         this.state = { hasValidSelection: this.isSelectedRegionsControlledAndNonEmpty(props) };
     }
 
-    public componentWillReceiveProps(nextProps?: IInternalHeaderProps) {
-        this.setState({ hasValidSelection: this.isSelectedRegionsControlledAndNonEmpty(nextProps) });
+    public componentDidUpdate(_: IInternalHeaderProps, prevState: IHeaderState) {
+        const nextHasValidSection = this.isSelectedRegionsControlledAndNonEmpty(this.props);
+        if (prevState.hasValidSelection !== nextHasValidSection) {
```
adidahiya: nit: is this check necessary? isn't `this.setState` a no-op if the values are equal?

packages/table/src/interactions/resizable.tsx
```diff
@@ -97,12 +97,15 @@ export class Resizable extends React.PureComponent<IResizableProps, IResizeableS
...
-    public componentWillReceiveProps(nextProps: IResizableProps) {
-        const { size } = nextProps;
-        this.setState({
-            size,
-            unclampedSize: size,
-        });
+    public componentDidUpdate(prevProps: IResizableProps) {
+        const didSizePropChange = prevProps.size !== this.props.size;
+        const { size } = this.props;
+        if (didSizePropChange) {
```
adidahiya: same here, curious why the check is necessary. also if it is, it doesn't need to be stored in a variable, just write `prevProps.size !== size` inline

packages/table/src/table.tsx
```diff
@@ -918,24 +919,30 @@ export class Table extends AbstractComponent<ITableProps, ITableState> {
         this.maybeScrollTableIntoView();
     }
 
-    protected validateProps(props: ITableProps & { children: React.ReactNode }) {
-        const { children, columnWidths, numFrozenColumns, numFrozenRows, numRows, rowHeights } = props;
+    protected validateProps(prevProps: ITableProps, forceValidation: boolean = false) {
+        const { children, columnWidths, numFrozenColumns, numFrozenRows, numRows, rowHeights } = this.props;
         const numColumns = React.Children.count(children);
 
+        const shouldValidate =
```
adidahiya: would be nice to leave a comment about this change for future git blames and such

adidahiya:
@tobilen I think you inadvertently enabled CircleCI for your fork. This may not allow the build to proceed if you don't have enough parallelism in your build. See our contributing docs: https://github.com/palantir/blueprint/blob/develop/CONTRIBUTING.md

tobilen:
probably did that a couple of months ago, i'm sorry. i pushed the `stop building` button, did that fix it?

adidahiya:
@tobilen you probably need to push a new commit, or open a new PR

tobilen:
that should happen when i incorporate your change requests tomorrow 👍

packages/core/test/slider/multiSliderTests.tsx
tobilen: because the errors weren't asserted correctly. i took the issue and the implementation from https://github.com/facebook/react/issues/11098#issuecomment-355032539

packages/table/src/table.tsx
tobilen: what kind of comment would you expect here?

packages/core/src/common/abstractComponent.ts
tobilen: i've added `UNSAFE_` components for the abstract components as well as the ones that implement them. is that what you had in mind?

packages/core/src/components/alert/alert.tsx
tobilen: it isn't

packages/core/test/tag-input/tagInputTests.tsx
tobilen: see above

packages/table/src/headers/header.tsx
tobilen: it is. no noop

packages/core/src/common/abstractPureComponent.ts
```diff
@@ -30,14 +30,14 @@ export abstract class AbstractPureComponent<P, S = {}> extends React.PureCompone
 
     constructor(props?: P, context?: any) {
         super(props, context);
-        if (!isNodeEnv("production")) {
+        if (!isNodeEnv('production')) {
```
adidahiya: wrong quotemarks, this will fail linting

packages/core/src/common/abstractPureComponent.ts
```diff
@@ -30,14 +30,14 @@ export abstract class AbstractPureComponent<P, S = {}> extends React.PureCompone
...
             this.validateProps(this.props);
         }
     }
 
-    public componentWillReceiveProps(nextProps: P & { children?: React.ReactNode }) {
-        if (!isNodeEnv("production")) {
-            this.validateProps(nextProps);
+    public componentDidUpdate(_: P, __: S, ___: SS) {
+        if (!isNodeEnv('production')) {
+            this.validateProps(this.props);
```
adidahiya: why is this not using `nextProps` anymore?

packages/core/src/common/unsafe_abstractComponent.ts
```diff
@@ -0,0 +1,87 @@
...
+import * as React from "react";
+import { isNodeEnv } from "./utils";
+
+/**
+ * An abstract component that Blueprint components can extend
+ * in order to add some common functionality like runtime props validation.
+ * This version will not work with React 17 upwards and only exists for legacy purposes
+ */
+/* tslint:disable-next-line:class-name */
+export abstract class UNSAFE_AbstractComponent<P, S> extends React.Component<P, S> {
```
adidahiya: this isn't exactly what I was hoping for... in order to make this change without breaking backwards compatibility, the exported symbol `AbstractComponent` in `@blueprintjs/core` must behave identically to the currently released one. you should make a new `AbstractComponentWithNewLifecycle` component and use that in the new components you are polyfilling.

packages/core/src/components/collapse/collapse.tsx
```diff
@@ -105,7 +106,12 @@ export enum AnimationStates {
     CLOSED,
 }
 
-export class Collapse extends AbstractPureComponent<ICollapseProps, ICollapseState> {
+export interface ICollapseSnapshot {
+    animationState?: AnimationStates;
+    height?: string;
+}
+
+class CollapseWithoutPolyfill extends AbstractPureComponent<ICollapseProps, ICollapseState, ICollapseSnapshot> {
```
adidahiya: did you try the approach I suggested of polyfilling inline with `const Collapse = polyfill(class extends ...)`? I'd like to avoid having multiple symbols here, to discourage anyone from using the un-polyfilled versions (like you've done in some test suites in this PR).

packages/core/src/components/forms/unsafe_controls.tsx
```diff
@@ -0,0 +1,141 @@
+/*
```
adidahiya: these components are unnecessary. if the polyfill is doing its job right, why would anyone need the unsafe versions? we should not expand Blueprint's API surface area by adding these components which duplicate so much code.

packages/table/src/headers/header.tsx
```diff
@@ -239,8 +239,11 @@ export class Header extends React.Component<IInternalHeaderProps, IHeaderState>
...
+            this.setState({ hasValidSelection: nextHasValidSection });
```
adidahiya: you seem to have reverted the commit which removed the change check. can you explain why please?

packages/table/src/table.tsx
```diff
@@ -520,6 +520,7 @@ export class Table extends AbstractComponent<ITableProps, ITableState> {
 
     public constructor(props: ITableProps, context?: any) {
         super(props, context);
+        this.validateProps(props, true);
```
adidahiya: why is this necessary? `AbstractComponent` should be doing this.

packages/table/src/table.tsx
```diff
@@ -883,6 +775,7 @@ export class Table extends AbstractComponent<ITableProps, ITableState> {
      * ClientRects.
      */
     public componentDidMount() {
+        this.validateProps(this.props, true);
```
adidahiya: same here, `AbstractComponent` should be responsible for this

packages/table/src/table.tsx
adidahiya: explain your reasoning for the behavior change, what led you to add this code?

packages/table/test/tableTests.tsx
```diff
@@ -622,15 +623,15 @@ describe("<Table>", function(this) {
 
             const table2 = mount(<Table numFrozenColumns={1} />);
             expect(table2.state("numFrozenColumnsClamped")).to.equal(0);
-            expect(consoleWarn.callCount).to.equal(1, "warned 1");
+            expect(consoleWarn.callCount).to.equal(2, "warned 1");
```
adidahiya: these callCount changes throughout this suite are a little concerning to me; why do our expectations about the table behavior have to change in this PR?

tobilen:
> can you please allow me write access to your fork? I may need to push some changes directly to get this PR over the line
sure, done. but this is currently not mergable (i readded a WIP to the title). i just tried some stuff to get the CI green.

packages/table/src/headers/header.tsx
tobilen: see https://github.com/palantir/blueprint/pull/3702#discussion_r326727681
removing the checks changes behavior (tests break)

packages/table/test/tableTests.tsx
tobilen: this is not done. i plan on removing the `validateProps` call in the constructor that leads to the double warnings

packages/core/src/components/forms/unsafe_controls.tsx
tobilen: yes, i don't like them either. this is what i figured you meant with https://github.com/palantir/blueprint/pull/3702#discussion_r325780179. i'll remove them

packages/core/src/components/collapse/collapse.tsx
tobilen: that syntax breaks several typechecks. i suggest reverting the multiple export stuff and let you apply the change you'd like to see?

packages/table/src/table.tsx
tobilen: see https://github.com/palantir/blueprint/pull/3702#discussion_r326826457

packages/core/src/common/abstractPureComponent.ts
tobilen: `componentWillReceiveProps` provides the next set of props as argument, but `componentDidUpdate` provides the previous props. we want to validate the props that have just been applied, not the old ones

packages/core/src/common/unsafe_abstractComponent.ts
tobilen: i've done this, although i do not think it is a good idea. it will propagate usage of the old `AbstractComponent` by external consumers and cause internal renaming work down the line.

tobilen:
@adidahiya all the feedback should be incorporated now, feel free to make any changes you deem necessary

packages/table/src/table.tsx
tobilen: added a comment

adidahiya:
I pushed some commits to the branch. Please don't force push (for this, or other PRs in Blueprint)

ptgamr:
Awesome guys ... Thank you very much for the effort!

PR #3722, opened 2019-09-03
Prevent select active item after key down on tag remove

### LINKED ISSUE #3674: MultiSelect adds first element from list again if when item removed using "Enter Key"
#### Environment

- __Package version(s)__: 3.16.2
- __Browser and OS versions__: Chrome and MacOS

If possible, link to a minimal repro (fork [this code sandbox](https://codesandbox.io/s/nko3k41y60)): 
delete

#### Steps to reproduce

1. Add multiple tags in the multi-select field. (Except the first element on the list)
2. Use the keyboard Tab to navigate to any tag close icon.
3. Press enter

#### Actual behavior

The selected tag gets removed but the first element from the dropdown list gets inserted as a new tag.

#### Expected behavior

Only the selected tag should be removed

#### Possible solution

I think this is happening because if the enter key is pressed on a blank multiselect field, the first element of the dropdown is selected.

chuyachia:
#### Checklist

- [ ] Includes tests
- [ ] Update documentation

Add a state in multiSelect to track if key was pressed on the remove button of the tags in tagInput. If yes, do not call handleKeyUp in queryList which causes active item to be selected.

### DIFF
--- a/packages/select/src/components/select/multiSelect.tsx
+++ b/packages/select/src/components/select/multiSelect.tsx
@@ -145,10 +145,7 @@ export class MultiSelect<T> extends React.PureComponent<IMultiSelectProps<T>, IM
                 popoverClassName={classNames(Classes.MULTISELECT_POPOVER, popoverProps.popoverClassName)}
                 onOpened={this.handlePopoverOpened}
             >
-                <div
-                    onKeyDown={this.getTargetKeyDownHandler(handleKeyDown)}
-                    onKeyUp={handleTagInKeyUp}
-                >
+                <div onKeyDown={this.getTargetKeyDownHandler(handleKeyDown)} onKeyUp={handleTagInKeyUp}>
                     <TagInput
                         placeholder={placeholder}
                         {...tagInputProps}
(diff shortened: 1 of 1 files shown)


palantirtech:
Thanks for your interest in [palantir/blueprint](https://github.com/palantir/blueprint), @chuyachia! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

adidahiya:
are you able to sign the CLA @chuyachia?

chuyachia:
Done. What do you think of this approach?

packages/select/src/components/select/multiSelect.tsx
```diff
@@ -196,6 +202,13 @@ export class MultiSelect<T> extends React.PureComponent<IMultiSelectProps<T>, IM
     ) => {
         return (e: React.KeyboardEvent<HTMLElement>) => {
             const { which } = e;
+            const htmlTarget = e.target as HTMLElement;
+            if (htmlTarget.className.includes("remove")) {
```
adidahiya: we shouldn't use "magic" string constants like this. instead we should use predictable variables

adidahiya:
sorry for the delay @chuyachia, this PR was a good start but there were a number of changes I wanted to make to the implementation so I went ahead and started my own PR so I can release this fix in 3.19. thanks for your PR.

PR #3724, opened 2019-09-03
[select] Default autoComplete for text input in <Suggest> to "off"

skurger:
#### Fixes #0000

No issue yet

#### Checklist

- [ ] Includes tests
- [ ] Update documentation

Please advise if these items are needed.  inputProps are already documented and this commit just sets a default.

Set the default autocomplete HTML attribute to "off" so that browser drop downs don't render in front of the component menu options.

The default value set by this commit

### DIFF
--- a/packages/select/src/components/select/suggest.tsx
+++ b/packages/select/src/components/select/suggest.tsx
@@ -153,7 +153,7 @@ export class Suggest<T> extends React.PureComponent<ISuggestProps<T>, ISuggestSt
         const { fill, inputProps = {}, popoverProps = {} } = this.props;
         const { isOpen, selectedItem } = this.state;
         const { handleKeyDown, handleKeyUp } = listProps;
-        const { placeholder = "Search..." } = inputProps;
+        const { autoComplete = "off", placeholder = "Search..." } = inputProps;
 
         const selectedItemText = selectedItem ? this.props.inputValueRenderer(selectedItem) : "";
         // placeholder shows selected item while open.
@@ -183,6 +183,7 @@ export class Suggest<T> extends React.PureComponent<ISuggestProps<T>, ISuggestSt
                 onOpened={this.handlePopoverOpened}
             >
                 <InputGroup
+                    autoComplete={autoComplete}
                     disabled={this.props.disabled}
                     {...inputProps}
                     inputRef={this.refHandlers.input}


palantirtech:
Thanks for your interest in [palantir/blueprint](https://github.com/palantir/blueprint), @skurger! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

adidahiya:
What does the component look like before and after this change? Generally it's better to file an issue first before sending a PR (or at the same time), even if you are ready to contribute, so that we can discuss the issue more directly.

skurger:
I don't have a screenshot ready but can generate one later if needed. The component looks as it does currently with the default set.  The way that I reproduced it was spreading props from a Formik control into the input props of <Suggest>. One of these props is "value".  Since it is a text input under the hood in the component the browser will show the autocomplete list of previous values.  This would make it so users would not see that if input props are passed that happen to contain value.

adidahiya:
> The way that I reproduced it was spreading props from a Formik control into the input props of <Suggest>. One of these props is "value".

Ok, but the component ignores `inputProps.value` and sets its own right here:

https://github.com/palantir/blueprint/blob/dbb5bf48a243024decd5cbd07d92079320a2dde6/packages/select/src/components/select/suggest.tsx#L194

I think it would help to see a minimal repro of the bug before proceeding... try using the code sandbox in the issue template.

skurger:
It is very easy to reproduce in Chrome if you create a form and pass the "name" attribute to "inputProps" for Suggest.

https://codesandbox.io/s/react-example-o7gvh?fontsize=14

Select a value and click submit.  The browser autocomplete menu should render over the Suggest Popover.  I can verify this on MacOS and Windows Chrome.

This is my MacOS Chrome version

Version 76.0.3809.132 (Official Build) (64-bit)

skurger:
I just noticed you said to use the sandbox in the template but overlooked that initially.  Sorry about that.  If this change is accepted I'd be happy to look at the other components in the select package to see if they need the same logic applied.  `<MultiSelect>` may need the same changes.

adidahiya:
I have the same version of chrome and can't repro it:



Firefox is the same

skurger:
Hmm, I'm able to reproduce as well as a coworker that found this bug. I just clicked the sandbox above and in Chrome I get this.


blueprint-select

skurger:
Any thoughts on this?

PR #3732, opened 2019-09-10
[table] fix: column resize calculation on header double click

### LINKED ISSUE #3731: Table.enableColumnReordering causes ColumnHeader.handleResizeDoubleClick to calculate wrong column width
Hello.

Looks like `Table.enableColumnReordering` causes `ColumnHeader.handleResizeDoubleClick` to calculate wrong column width.

#### Environment

- __Package version(s)__: @blueprintjs/table@3.7.0, @blueprintjs/core@3.18.0
- __Browser and OS versions__: firefox 69.0, chromium 76.0.3809.100, ubuntu 16.04

Link to a minimal repro: https://codesandbox.io/s/blueprint-sandbox-m93im

#### Steps to reproduce

1. Create component with `Table.enableColumnReordering={true}` and resizable columns.
1. Set column name to `Some column name`
1. Its more reproducable, when Column name is longer then cells data
1. Doubleclick right border of the column’s header

#### Actual behavior

1. The width of the column is adjusted like there is somethins like `drag-handle-verticalSome column name` in it.

#### Expected behavior

1. The width of the column shuold be adjusted like there is `Some column name` in it, ignoring `drag-handle-vertical` element.

#### Possible solution

Looks like `Header.maybeRenderReorderHandle` method [here](https://github.com/palantir/blueprint/blob/06a186c90758bbdca604ed6d7bf639c3d05b1fa0/packages/table/src/headers/header.tsx#L380) misses `bp-table-text-no-measure` class.

This leads to incorrect calculation at `ColumnHeader.handleResizeDoubleClick` method, as `drag-handle-vertical` text appears in dom element's textContent.

Adding `bp-table-text-no-measure` class should fix the bug.

zwerg44:
#### Checklist

- [ ] Includes tests
- [ ] Update documentation

- fixes `ColumnHeader.handleResizeDoubleClick` with enabled ColumnReordering and menuRenderer
- added `bp-table-text-no-measure` class to header's maybeRenderReorderHandle
- added `bp-table-text-no-measure` class to columnHeaderCell's maybeRenderDropdownMenu
- modified **packages/table/src/common/utils.ts** `measureTextContentWithExclusions` logic

- I searched through the code, `measureTextContentWithExclusions` is used only at `handleResizeDoubleClick`, so I assumed that it would be ok to modify it
- I modified `measureTextContentWithExclusions` as it is possible to have both ColumnReordering and menuRenderer icons inside header
- I removed dom manipulations at `measureTextContentWithExclusions`, and replaced it with metrics computation
- added `EXCLUDED_ICON_PLACEHOLDER_WIDTH` constant to add spacing to auto-resize width at the place of that icons
- `CLASSNAME_EXCLUDED_FROM_TEXT_MEASUREMENT` is exported and used at `header.tsx` and `columnHeaderCell.tsx`

### DIFF
--- a/packages/table/src/common/utils.ts
+++ b/packages/table/src/common/utils.ts
@@ -14,7 +14,10 @@
  * limitations under the License.
  */
 
-const CLASSNAME_EXCLUDED_FROM_TEXT_MEASUREMENT = "bp-table-text-no-measure";
+// used to exclude icons from column header measure
+export const CLASSNAME_EXCLUDED_FROM_TEXT_MEASUREMENT = "bp-table-text-no-measure";
+// supposed width of the icons placeholder
+const EXCLUDED_ICON_PLACEHOLDER_WIDTH = 16;
 
 /**
  * Since Firefox doesn't provide a computed "font" property, we manually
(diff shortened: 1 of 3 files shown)


palantirtech:
Thanks for your interest in [palantir/blueprint](https://github.com/palantir/blueprint), @zwerg44! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

packages/table/src/common/utils.ts
```diff
@@ -14,7 +14,10 @@
  * limitations under the License.
  */
 
-const CLASSNAME_EXCLUDED_FROM_TEXT_MEASUREMENT = "bp-table-text-no-measure";
+// used to exclude icons from column header measure
+export const CLASSNAME_EXCLUDED_FROM_TEXT_MEASUREMENT = "bp-table-text-no-measure";
+// supposed width of the icons placeholder
+const EXCLUDED_ICON_PLACEHOLDER_WIDTH = 16;
```
adidahiya:
```suggestion
const EXCLUDED_ICON_PLACEHOLDER_WIDTH = Icon.SIZE_STANDARD;
```

packages/table/src/common/utils.ts
```diff
@@ -14,7 +14,10 @@
  * limitations under the License.
  */
```
adidahiya:
```suggestion

import { Icon } from "@blueprintjs/core";

```

PR #3742, opened 2019-09-18
Fix a controlled EditableText bug

chiubaka:
Fixes a bug where EditableText controlled value cannot be reset to undefined or null.

The comments in this file state that the point of mirroring the `props.value` to `state.value` is to maintain that value while the user
is editing. However, if the user is not editing, the edited line in question as is does not allow for the value of the component to be
reset to something undefined or null for emptiness.

#### Checklist

- [ ] Includes tests
- [ ] Update documentation

### DIFF
--- a/packages/core/src/components/editable-text/editableText.tsx
+++ b/packages/core/src/components/editable-text/editableText.tsx
@@ -170,7 +170,7 @@ export class EditableText extends AbstractPureComponent<IEditableTextProps, IEdi
 
     public render() {
         const { disabled, multiline } = this.props;
-        const value = this.props.value == null ? this.state.value : this.props.value;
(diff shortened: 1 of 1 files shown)


palantirtech:
Thanks for your interest in [palantir/blueprint](https://github.com/palantir/blueprint), @chiubaka! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

adidahiya:
You're failing a unit test:

image


I think you'll need to abstract the logic for getting the value from `this.props` into a small private helper method, as it is used in the constructor as well.

In addition, can you please add a unit test for this behavior change?

adidahiya:
superseded by #3924

PR #3768, opened 2019-10-06
[core] feat(PanelStack): add renderCurrentPanelOnly prop

### LINKED ISSUE #3667: PanelStack doesn't allow shifted panel components to keep their state
#### Feature request

Right now it appears that pushing a new panel on the stack causes the previously active panel to no longer be rendered which, because of the way React works, means it loses any internal state it has. When you navigate back/pop you'll get the old panel component back, but without any state it had built up. In many use cases this is problematic.

This would presumably be a breaking change, unless opt-in via some sort of flag e.g. `keepPanelState: true`

Demo: https://stackblitz.com/edit/react-fzveef?file=index.js

radekmie:
#### Checklist

- [x] Includes tests
- [x] Update documentation

A new `renderCurrentPanelOnly` prop, which is `true` by default to preserve the backward compatibility, which allows the panels to preserve their state while being hidden.

My current implementation does not work _yet_ though - do read the code comments.

Nothing specific.

#### Screenshot

Screenshot from 2019-10-06 01-54-29

### DIFF
--- a/packages/core/src/components/panel-stack/_panel-stack.scss
+++ b/packages/core/src/components/panel-stack/_panel-stack.scss
@@ -13,7 +13,6 @@
   display: flex;
   flex-shrink: 0;
   align-items: center;
-  z-index: 1;
   box-shadow: 0 1px $pt-divider-black;
   height: $pt-grid-size * 3;
 
@@ -48,6 +47,7 @@
   @include position-all(absolute, 0);
   display: flex;
   flex-direction: column;
+  z-index: 1;
 
   // border between panels, visible during transition
   margin-right: -1px;
@@ -58,6 +58,10 @@
 
   .#{$ns}-dark & {
     background-color: $dark-gray4;
+  }
+
+  &:nth-last-child(n + 4) {
+    display: none;
   }
 }
 
--- a/packages/core/src/components/panel-stack/panelStack.tsx
+++ b/packages/core/src/components/panel-stack/panelStack.tsx
@@ -47,6 +47,12 @@ export interface IPanelStackProps extends IProps {
     onOpen?: (addedPanel: IPanel) => void;
 
     /**
+     * If false, PanelStack will render all panels in the stack to the DOM, allowing their React component trees to maintain state as a user navigates through the stack. Panels other than the currently active one will be invisible.
+     * @default true
+     */
+    renderCurrentPanelOnly?: boolean;
+
+    /**
      * Whether to show the header with the "back" button in each panel.
      * @default true
      */
@@ -93,7 +99,7 @@ export class PanelStack extends AbstractPureComponent2<IPanelStackProps, IPanelS
         );
         return (
             <TransitionGroup className={classes} component="div">
-                {this.renderCurrentPanel()}
+                {this.renderPanels()}
             </TransitionGroup>
         );
     }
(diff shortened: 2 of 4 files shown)


palantirtech:
Thanks for your interest in [palantir/blueprint](https://github.com/palantir/blueprint), @radekmie! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

packages/core/src/components/panel-stack/panelStack.tsx
```diff
@@ -110,25 +116,35 @@ export class PanelStack extends AbstractPureComponent2<IPanelStackProps, IPanelS
...
+        const panelViews = panelsToRender.map(this.renderPanel).reverse();
+        return panelViews;
+    }
+
+    private renderPanel = (panel: IPanel, index: number) => {
+        const { showPanelHeader = true } = this.props;
+        const { stack } = this.state;
+        const active = index === 0;
+        const layer = stack.length - index;
+        const key = `${layer}-${active}`;
```
radekmie: The `active` part is required for `CSSTransition` to properly animate the bottom panel to move in while _popping_ and moving out while _pushing_. It changes the `key` though, so the `PanelView` is rerendered anyway, so no state is preserved. Removing it makes it work, the state is preserved, but the bottom panel is not animated at all.

I'm not sure how to tackle this issue.

packages/core/src/components/panel-stack/_panel-stack.scss
```diff
@@ -59,6 +59,10 @@
   .#{$ns}-dark & {
     background-color: $dark-gray4;
   }
+
+  &:nth-last-child(n + 4) {
```
radekmie: _All but last 3_. Why 3? When one removes the top panel, the bottom two should be visible to animate properly. The rest can be hidden as a layering optimization.

adidahiya: > The rest can be hidden as a layering optimization.

Can you provide any evidence that we actually need this optimization?

packages/core/src/components/panel-stack/panelStack.tsx
```diff
@@ -46,6 +46,12 @@ export interface IPanelStackProps extends IProps {
      */
     onOpen?: (addedPanel: IPanel) => void;
 
+    /**
+     * If false, PanelStack will render all panels in the stack to the DOM, allowing their React component trees to maintain state as a user navigates through the stack. Panels other than the currently active one will be invisible.
+     * @default true
+     */
+    renderCurrentPanelOnly?: boolean;
```
adidahiya: let's change this name to mirror a similar prop for the `Tabs` component... `renderActivePanelOnly`

packages/core/src/components/panel-stack/panelStack.tsx
adidahiya: I'm confused, what does key have to do with CSSTransition? I don't see it in the docs http://reactcommunity.org/react-transition-group/css-transition

packages/core/src/components/panel-stack/_panel-stack.scss
radekmie: Well, it's not really _needed_ - I wanted to make it in align with the `Tabs` implementation ([JSX](https://github.com/palantir/blueprint/blob/274ef9fb153e1879551e1e9aaddd2ca3f03c775c/packages/core/src/components/tabs/tabs.tsx#L295), [SCSS](https://github.com/palantir/blueprint/blob/274ef9fb153e1879551e1e9aaddd2ca3f03c775c/packages/core/src/components/tabs/_tabs.scss#L145-L147)).

packages/core/src/components/panel-stack/panelStack.tsx
radekmie: The thing is that with `renderCurrentPanelOnly={false}` we would keep all the `CSSTransition`s rendered, therefore they would not trigger the enter event as they were _entered_. To force the enter event, I wanted to change the `key` (just like the current implementation), but `stack.length` is no longer OK, as the length changes with the, well, stack length, and a single panel should not rerender as long as it's hidden.

This `key` contains two parts: first one, `stack.length - index` is constant (and unique) for each panel, second one, `active` changes only when the panel becomes or stops being active.

zachkirsch:
@radekmie I would love this feature - are you still working on it? Would be happy to take it over but don't want to repeat work you've already done.

radekmie:
@zachkirsch Well, I haven't done anything since the last review as there was no response (yet). If you have an idea on how to actually fix the problem - feel free to pick it up.

packages/core/src/components/panel-stack/_panel-stack.scss
```diff
@@ -13,7 +13,6 @@
   display: flex;
   flex-shrink: 0;
   align-items: center;
-  z-index: 1;
```
invliD, line 16: Why was this removed? This breaks the box-shadow from the next line when the panel view's content has a background.

radekmie: @invliD It was quite a while ago, but it seems it got moved to `.#{$ns}-panel-stack-view` below. If it's a problem at the moment, I think you should file a new issue about that.

invliD: I did see it got added to the view, but that doesn't explain why it was removed here. I added it here for exactly this reason when I wrote the panel stack; guess I'll just add it back…

PR #3770, opened 2019-10-06
Ability to set max number of active toasts

jamesgiu:
I've used BlueprintJS for work projects and this little idea has been mentioned a few times at work so I thought I'd see what you guys think. Also doubles as a Hacktoberfest contribution!

#### Checklist

- [x] Includes tests
- [x] Update documentation

Ability to set a maximum number of active toasts being displayed at once. Any new toast effectively 'pushes out' the oldest one.

Is 5 a suitable default limit? Or should default be no limit (as it is right now in BlueprintJS)?
Does it make sense for a user to specify '0' as an argument? I've covered for the use case, just in case

#### Screenshot
ezgif-1-12f9cee6fc3a

### DIFF
--- a/packages/core/src/components/toast/toaster.tsx
+++ b/packages/core/src/components/toast/toaster.tsx
@@ -88,6 +88,14 @@ export interface IToasterProps extends IProps {
      * @default Position.TOP
      */
     position?: ToasterPosition;
+
+    /**
+     * The maximum number of active toasts that can be displayed at once.
+     *
+     * When the limit is about to be exceeded, the oldest active toast is removed.
+     * @default 5
+     */
+    maxToasts?: number;
 }
 
 export interface IToasterState {
@@ -101,6 +109,7 @@ export class Toaster extends AbstractPureComponent2<IToasterProps, IToasterState
     public static defaultProps: IToasterProps = {
         autoFocus: false,
         canEscapeKeyClear: true,
+        maxToasts: 5,
         position: Position.TOP,
         usePortal: true,
     };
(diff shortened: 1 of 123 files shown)


palantirtech:
Thanks for your interest in [palantir/blueprint](https://github.com/palantir/blueprint), @jamesgiu! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

packages/core/src/components/toast/toaster.tsx
```diff
@@ -101,6 +109,7 @@ export class Toaster extends AbstractPureComponent2<IToasterProps, IToasterState
     public static defaultProps: IToasterProps = {
         autoFocus: false,
         canEscapeKeyClear: true,
+        maxToasts: 5,
```
adidahiya: this is technically a breaking change in behavior. I think you should not set a default, and if `maxToasts === undefined`, then do not enforce a limit.

packages/core/src/components/toast/toaster.tsx
```diff
@@ -133,6 +142,12 @@ export class Toaster extends AbstractPureComponent2<IToasterProps, IToasterState
     private toastId = 0;
 
     public show(props: IToastProps, key?: string) {
+        // show no toasts if max toasts is less than 1.
+        if (this.props.maxToasts < 1) {
```
adidahiya: this validation should happen in `validateProps` (invoked via the super class `AbstractPureComponent2`). if `props.maxToasts < 1` I think you should fail loudly and throw a validation error (some other components do this).

jamesgiu:
Thanks for your review @adidahiya!

adidahiya:
thanks for the PR @jamesgiu!

PR #3787, opened 2019-10-21
[core] fix(PanelStack): do not trigger unintended animations

marvinsum:
#### Checklist

- [ ] Includes tests
- [ ] Update documentation

Fix an issue where `<PanelStack />` could animate in the wrong direction.

`<PanelStack />` should only update its internal state via `componentDidUpdate()` when the `stack` length changes. Currently, pushing an item onto `stack` sets `direction: push`. As the `<PanelStack />` begins to animate, updating the `stack` without changing its length (i.e. updating a property in one of its `<IPanel />`s) will set `direction: pop`, leading to an unintended animation change.


sgarnotel:
With this PR, panel is not updated with stack props update (from API call for example).
Is it a good idea to only test stack length ?

sgarnotel:
Please review #3875

PR #3802, opened 2019-10-30
[DateTime] Add TimePicker input customisation

### LINKED ISSUE #3801: [DateTime] TimePicker inputs are not customisable
#### Environment

- __Package version(s)__: 3.14.0
- __Browser and OS versions__: -

#### Feature request

I'd like to have more control over the input elements the TimePicker renders.

Use Case 1: CSS-in-JS styling
In its current implementation, customisation of the TimePicker input components can only occur by conventional css styling. When using CSS-in-JS solutions like `emotion` or `styled-components`, styles are directly coupled to a given component, meaning an application has a `StyledInput` component that implements their styling. I want to use my `StyledInput` within the TimePicker.

Use Case 2: Custom event handlers
I'd like to provide my own handlers for native DOM events like the user bluring one of the TimePicker's input elements

Use Case 3: Additional attributes
I'd like to provide additional html attributes to the input elements (like `title` and `aria-label`).

tobilen:
#### Checklist

- [x] Includes tests
- [x] Update documentation

This PR introduces two changes to the `TimePicker` component:
* Adding an `inputProps` prop that is handed down to all input elements rendered by the `TimePicker` (analogous to  https://github.com/palantir/blueprint/blob/develop/packages/datetime/src/dateInput.tsx#L89 )
* Adding an `inputComponent` prop that enables users to pass their own `input` implementations to the `TimePicker` to be used by it instead of pure input elements.

I was unsure about which tests of the `TimePicker` test suite i would want to repeat for a version using a custom input element. Realistically, i think the typing is strict enough to ensure that any element passed to `inputComponent` adheres to the same interface as the native input element, but i am open to suggestions.

### DIFF
--- a/packages/datetime/src/timePicker.tsx
+++ b/packages/datetime/src/timePicker.tsx
@@ -148,6 +148,10 @@ export class TimePicker extends React.Component<ITimePickerProps, ITimePickerSta
 
     public static displayName = `${DISPLAYNAME_PREFIX}.TimePicker`;
 
+    private static renderDivider(text = ":") {
+        return <span className={Classes.TIMEPICKER_DIVIDER_TEXT}>{text}</span>;
+    }
+
     public constructor(props?: ITimePickerProps, context?: any) {
         super(props, context);
 
@@ -180,12 +184,12 @@ export class TimePicker extends React.Component<ITimePickerProps, ITimePickerSta
                 </div>
                 <div className={Classes.TIMEPICKER_INPUT_ROW}>
                     {this.renderInput(Classes.TIMEPICKER_HOUR, hourUnit, this.state.hourText)}
-                    {this.renderDivider()}
+                    {TimePicker.renderDivider()}
                     {this.renderInput(Classes.TIMEPICKER_MINUTE, TimeUnit.MINUTE, this.state.minuteText)}
-                    {shouldRenderSeconds && this.renderDivider()}
+                    {shouldRenderSeconds && TimePicker.renderDivider()}
                     {shouldRenderSeconds &&
                         this.renderInput(Classes.TIMEPICKER_SECOND, TimeUnit.SECOND, this.state.secondText)}
-                    {shouldRenderMilliseconds && this.renderDivider(".")}
+                    {shouldRenderMilliseconds && TimePicker.renderDivider(".")}
                     {shouldRenderMilliseconds &&
                         this.renderInput(Classes.TIMEPICKER_MILLISECOND, TimeUnit.MS, this.state.millisecondText)}
                 </div>
@@ -234,10 +238,6 @@ export class TimePicker extends React.Component<ITimePickerProps, ITimePickerSta
                 <Icon icon={isDirectionUp ? "chevron-up" : "chevron-down"} />
             </span>
         );
-    }
-
-    private renderDivider(text = ":") {
-        return <span className={Classes.TIMEPICKER_DIVIDER_TEXT}>{text}</span>;
     }
 
     private renderInput(className: string, unit: TimeUnit, value: string) {
--- a/packages/datetime/src/timePicker.tsx
+++ b/packages/datetime/src/timePicker.tsx
@@ -62,6 +62,11 @@ export interface ITimePickerProps extends IProps {
     disabled?: boolean;
 
     /**
+     * Props passed through to each rendered input element.
+     */
+    inputProps?: React.InputHTMLAttributes<HTMLInputElement>;
+
+    /**
      * Callback invoked when the user changes the time.
      */
     onChange?: (newTime: Date) => void;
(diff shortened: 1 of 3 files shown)


packages/datetime/src/timePicker.tsx
```diff
@@ -61,6 +61,17 @@ export interface ITimePickerProps extends IProps {
...
+    /**
+     * Props passed through to each rendered input element.
+     */
+    inputProps?: React.InputHTMLAttributes<HTMLInputElement>;
+
+    /**
+     * Custom input component that will be used instead of a pure input element.
+     * Needs to have the same interface as a normal input.
+     */
+    inputComponent?: React.ComponentType<React.InputHTMLAttributes<HTMLInputElement>>;
```
adidahiya: usually we make a "renderX" prop instead of asking for a component... so how about we change this to `renderInput: (props: React.InputHTMLAttributes<HTMLInputElement>) => JSX.Element`

packages/datetime/src/timePicker.tsx
```diff
@@ -285,11 +299,17 @@ export class TimePicker extends React.Component<ITimePickerProps, ITimePickerSta
                 this.setState({ millisecondText: text });
                 break;
         }
+        if (this.props.inputProps) {
```
adidahiya: please avoid implicit type coercion like this, instead use `if (this.props.inputProps != null)`

adidahiya: actually, how about abstracting this repeated check into a helper method?

```ts
private maybeInvokeInputEventHandler(handlerName: "onChange" | "onBlur" | "onKeyDown" | "onFocus") {
  if (this.props.inputProps != null) {
    BlueprintUtils.safeInvoke<React.SyntheticEvent<HTMLInputElement>, void>(this.props.inputProps[handlerName], e);
  }
}
```

packages/datetime/src/timePicker.tsx
tobilen: of course we can do it like this, although i believe this is a different (and more far reaching) change.

the way i would expect a passable render function to behave is different from a simple component. More specifically, i would expect a render function to receive all relevant props and states that get handled by the default  method and hand off implementation completely to the user. My solution was a bit smaller in scope.

So the interface for a `renderInput` would look like this:
```tsx
const renderInputProps = {
  disabled: boolean;
  inputProps: React.InputHTMLAttributes<HTMLInputElement>;
  value: string;
  onBlur: (e: React.SyntheticEvent<HTMLInputElement>) => void,
  onChange: (e: React.SyntheticEvent<HTMLInputElement>) => void,
  onFocus: (e: React.SyntheticEvent<HTMLInputElement>) => void,
  onKeyDown: (e: React.KeyboardEvent<HTMLInputElement>) => void,
}
// ...
renderInput:  (renderProps: renderInputProps) => React.ComponentType
```
implementation would then be in the `render` method, where we either call `this.props.renderInput(/* ... */)` or fall back to `this.renderInput(/* ... */)`. 

While providing more control, this definitely complicates usage and as such i would not advocate for it. However, anything other than a complete replacement for the actual `renderInput` method of the component would be confusing.

packages/datetime/src/timePicker.tsx
tobilen: sure, good suggestion 👍

tobilen:
*bump* :)

tobilen:
@adidahiya do you need any more input?

tobilen:
@adidahiya can i do anything to facilitate a merge?

spaceneedle2019:
Any suggestions or news here?

tobilen:
@adidahiya still waiting for feedback here. anything i can do?

packages/datetime/src/timePicker.tsx
```diff
@@ -225,28 +240,27 @@ export class TimePicker extends React.Component<ITimePickerProps, ITimePickerSta
...
+            ),
+            disabled,
+            onBlur: this.getInputBlurHandler(unit),
+            onChange: this.getInputChangeHandler(unit),
+            onFocus: this.handleFocus,
+            onKeyDown: this.getInputKeyDownHandler(unit),
+            value,
+        };
+
+        return CustomInputComponent ? <CustomInputComponent {...props} /> : <input {...props} />;
```
adidahiya: I thought about this some more and I'm still not convinced or super comfortable with adding this kind of customization. We don't accept a custom component like this in any other part of our API, and I don't think #3801 presents a strong enough argument for us to start doing that here. This kind of API puts the burden on the user to properly call `onBlur`, `onChange`, etc. at the right time inside their `CustomInputComponent` and could lead to confusing bugs.

tobilen:
hey @adidahiya , sorry for the delayed response.

I've solved my most urgent requirement of css-in-js styling with a utility package https://github.com/tobilen/react-styledown. It lets me css classnames generated in styled-components to blueprint (or other 3rd party components).

I'm afraid i don't have the time to rework this branch in the next few weeks. If we refactor our TimePicker and stick with blueprint for the implementation, i'll go the `inputProps` route, but for now i'll close the PR. I will leave the fork alive if anybody wants to pick up from here though.

Thanks for all your input 👋

PR #3832, opened 2019-11-10
[select] fix(QueryList): trim whitespace around query input text

### LINKED ISSUE #3765: Leading spaces in free text input in dropdowns are not displayed
Users can accidentally create entries in dropdowns that have leading spaces (e.g. “ SMITH” instead of “SMITH”), but the leading space is not shown in the dropdown thus creating confusion about what the created value is, and 

dropdown

#### Environment

- __Package version(s)__: 
- __Browser and OS versions__: Chrome 77.0.3865.90, Windows 10

If possible, link to a minimal repro (fork [this code sandbox](https://codesandbox.io/s/blueprint-sandbox-et9xy)): 

#### Steps to reproduce

1.	Create a form
2.	Put a dropdown property
3.	Allow users to create new values
4.	Type 'Maxime' and ' Maxime' (with a leading space).

#### Actual behavior
The two values (incorrectly) look the same in the dropdown. See screenshot

#### Expected behavior
Either trim the value or at least show any leading spaces when displaying.

adidahiya:
#### Checklist

- [x] Includes tests
- [x] Update documentation

QueryList now ignores any whitespace around the _input_ text when reporting changes. Leading/trailing whitespace around controlled query text (supplied through a `query` prop) is still preserved.


adidahiya:
🙅‍♂ this is a bad approach, it prevents entering any spaces into the input at all

adidahiya:
seems to work fine after merging the latest `develop`... not sure why, the commits between then and now don't immediately suggest something else which made it work 🤷‍♂️

adidahiya:
just kidding... it's not fixed. it's only MultiSelect which is working as intended now

PR #3837, opened 2019-11-12
Fix Numeric input round value at first render

raphdpx:

Numeric Input round the value with the getSanitizedValue function to the precision of minorStepsize if the value is declared, if not it takes the Stepsize value. Previously, the value was round to precision 0 when the input was rendered.

Numeric input render, value with decimals and decimals stepsize / minorstepsize.

### DIFF
--- a/packages/core/src/components/forms/numericInput.tsx
+++ b/packages/core/src/components/forms/numericInput.tsx
@@ -30,7 +30,7 @@ import {
     MaybeElement,
     Position,
     removeNonHTMLProps,
-    Utils,
+    Utils
 } from "../../common";
 import * as Errors from "../../common/errors";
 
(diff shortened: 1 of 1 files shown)


palantirtech:
Thanks for your interest in [palantir/blueprint](https://github.com/palantir/blueprint), @raphdpx! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

adidahiya:
In general I recommend you first file a bug report using the issue template so that we can understand the exact problem you're having. Also for changes like this I would like to see unit tests added.

adidahiya:
superseded by https://github.com/palantir/blueprint/pull/3894

PR #3841, opened 2019-11-13
[core] feat(EditableText): add alwaysRenderInput experimental feature

### LINKED ISSUE #3839: Difficult to implement custom copy/paste clipboard into EditableText
#### Environment

- __Package version(s)__: any version of core
- __Browser and OS versions__: any browser

Minimal repro: https://codesandbox.io/s/blueprint-sandbox-8umsu?fontsize=14

#### Steps to reproduce

1. Focus the editable text area
1. Right click to open the context menu

#### Actual behavior

Text area is unfocussed, making it tricky to implement custom paste-from-clipboard behavior

#### Expected behavior

It's easy to implement such behavior

#### Possible solution

1. Keep the text area focussed so that a paste action works normally.
2. Provide an `onContextMenuOpen` hook to `ContextMenuTarget`-decorated components so that they can re-focus the text area. This still has the downside of temporarily un-focussing the text area when the context menu is triggered.

adidahiya:
#### Checklist

- [ ] Includes tests ("experimental" 🙃)
- [x] Update documentation

Add new prop `alwaysRenderInput` to make the component render an input which can, in theory, retain focus across `isEditing={true|false}` interactions.

Does this work in practice to solve the linked bug

Note that, at least right now, it doesn't work well with Blueprint's ContextMenu.

### DIFF
--- a/packages/core/src/components/editable-text/editableText.tsx
+++ b/packages/core/src/components/editable-text/editableText.tsx
@@ -148,20 +148,25 @@ export class EditableText extends AbstractPureComponent2<IEditableTextProps, IEd
         type: "text",
     };
 
+    private inputElement?: HTMLInputElement | HTMLTextAreaElement;
     private valueElement: HTMLSpanElement;
     private refHandlers = {
         content: (spanElement: HTMLSpanElement) => {
             this.valueElement = spanElement;
         },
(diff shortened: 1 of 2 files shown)


adidahiya:
confirmed this works on 3.17.2, so this is good to go

Issue #3799, opened 2019-10-29
Yarn install fails in Node 12


nicholas-l:
#### Environment

- __Package version(s)__: develop branch
- __Browser and OS versions__: Ubuntu 19.10, Node 13

If possible, link to a minimal repro (fork [this code sandbox](https://codesandbox.io/s/blueprint-sandbox-et9xy)): 

#### Steps to reproduce

1. Clone git repo
1. Run yarn install

#### Actual behavior

Deps install fails due to tree-sitter-typescript failing to build. I am not sure if that package is actually the problem or the version of nan or node-gyp being pulled in by yarn hoisting. Is anyone else having this problem?

<details><summary>Build Log</summary>

```
[4/4] Rebuilding all packages...
[-/8] ⢀ waiting...


Error: /home/nicholas/Development/blueprint/node_modules/tree-sitter-typescript: Command failed.
Exit code: 1
Command: node-gyp rebuild
Arguments: 
Directory: /home/nicholas/Development/blueprint/node_modules/tree-sitter-typescript
Output:
gyp info it worked if it ends with ok
gyp info using node-gyp@3.8.0
gyp info using node@13.0.1 | linux | x64
gyp info spawn /usr/bin/python2
gyp info spawn args [
gyp info spawn args   '/home/nicholas/Development/blueprint/node_modules/node-gyp/gyp/gyp_main.py',
gyp info spawn args   'binding.gyp',
gyp info spawn args   '-f',
gyp info spawn args   'make',
gyp info spawn args   '-I',
gyp info spawn args   '/home/nicholas/Development/blueprint/node_modules/tree-sitter-typescript/build/config.gypi',
gyp info spawn args   '-I',
gyp info spawn args   '/home/nicholas/Development/blueprint/node_modules/node-gyp/addon.gypi',
gyp info spawn args   '-I',
gyp info spawn args   '/home/nicholas/.node-gyp/13.0.1/include/node/common.gypi',
gyp info spawn args   '-Dlibrary=shared_library',
gyp info spawn args   '-Dvisibility=default',
gyp info spawn args   '-Dnode_root_dir=/home/nicholas/.node-gyp/13.0.1',
gyp info spawn args   '-Dnode_gyp_dir=/home/nicholas/Development/blueprint/node_modules/node-gyp',
gyp info spawn args   '-Dnode_lib_file=/home/nicholas/.node-gyp/13.0.1/<(target_arch)/node.lib',
[...]

PR #3842, opened 2019-11-13
[3.17.x] [core] feat(EditableText): add alwaysRenderInput experimental feature

adidahiya:
Same as https://github.com/palantir/blueprint/pull/3841 but on a 3.17.x release branch


adidahiya:
N.B. the build will not pass because of #3799

PR #3862, opened 2019-11-19
[core] feat: high contrast CSS with color adjustments to pass WCAG requirements

adamstorr:
Modifying the intent colors for various UI elements to pass contrast requirements. Testing used the [Color Contrast Analyser](https://developer.paciellogroup.com/resources/contrastanalyser/) from The Paciello Group.

Note: The variable `$orange1` for `$pt-intent-warning` doesn't communicate "warning" as well as `$orange3` did. This may require a slight redesign of how we display alerts/tooltips, or require us to choose a different range of colors for `$orange`.

I changed which color variables are used for intents.

#### Buttons
image
image
image
image

#### Toasts
image
image
image
image

#### Tooltips
image
image
image
image

#### Progress bars
image
image
image
image


adidahiya:
Updated PR to not use these colors by default, but the modified CSS is available in a new file `@blueprintjs/core/lib/css/accessibility/blueprint-hi-contrast.css`. Verified that when you import it in the docs you get this

image

packages/docs-app/src/index.scss
```diff
@@ -8,7 +8,7 @@ Licensed under the Apache License, Version 2.0.
 @import "~normalize.css/normalize.css";
 
 // import these first because they're the base layer
-@import "~@blueprintjs/core/lib/css/blueprint.css";
```
adidahiya: this should be reverted

PR #3867, opened 2019-11-26
[core] fix: Explicitly declare dom4 side effects

ericjeney:

This change explicitly declares the [dom4 side effect](https://github.com/palantir/blueprint/blob/3eef70e0b8b1bcce9de0836dbe2e28deff3ee8d9/packages/core/src/components/index.ts#L21) that is necessary for Blueprint to work in IE11 so that Webpack doesn't tree-shake it away.

Without this change, Webpack was stripping dom4 out of my build, which resulted in my app working fine in modern browsers but throwing errors such as this one in IE11:

```
Object doesn't support property or method 'closest'
```

I've tested this change locally by editing the `package.json` file in my `node_modules/@blueprintjs/core` directory and it works! IE11 no longer throws errors when the build includes this.

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -9,7 +9,10 @@
   "style": "lib/css/blueprint.css",
   "unpkg": "dist/core.bundle.js",
   "sideEffects": [
-    "*.css"
+    "*.css",
+    "lib/esm/components/index.js",
+    "lib/esnext/components/index.js",
+    "lib/cjs/components/index.js"
   ],
   "bin": {
     "upgrade-blueprint-2.0.0-rename": "./scripts/upgrade-blueprint-2.0.0-rename.sh",


adidahiya:
@ericjeney did you enable CI for your fork? the checks won't get reported here if you do that (see contributing docs)

adidahiya:
hm, ok, I think I slightly misunderstood how webpack does tree shaking. I assumed that if anything exported from `components/index.js` is used, then that module would be included, but according to [these docs](https://webpack.js.org/guides/tree-shaking/#clarifying-tree-shaking-and-sideeffects) that is not the case. your change looks correct.

ericjeney:
Whoops. Still had CircleCI enabled from my last commit, which was back when it was necessary. Fixed now.

adidahiya:
@ericjeney released in core 3.22.0

PR #3879, opened 2019-12-05
Fix PanelStack

### LINKED ISSUE #3875: PanelStack totally broken after 3.17.2
#### Environment

- __Package version(s)__: @blueprintjs/core@3.22.0 - npm 6.13.1 - node 13.2.0
- __Browser and OS versions__: Chromium Version 78.0.3904.108 (Build officiel) Arch Linux (64 bits)

#### Steps to reproduce

1. Build a component with PanelStack
```jsx
<div className="Panel">
     <PanelStack
        className="Panel-stack"
        stack={this.state.panels}
        onOpen={(panel) => this.handleOpen(panel)}
        onClose={this.handleClose}
     />
 </div>
```
The state element `this.state.panels` contains only one element
2. Update `this.state.panels` keeping the same `title`, `component` (update just `props`)

The PanelStack is not updated

#### Actual behavior

With @blueprintjs/core@3.22.0, there is no update

#### Expected behavior

Same behavior of @blueprintjs/core@3.17.2

#### Possible solution

I thing [PR #3787](https://github.com/palantir/blueprint/pull/3787) is part of the problem, but reverse this commit is not enough

marvinsum:
#### Checklist

- [ ] Includes tests
- [ ] Update documentation

Fix an issue where the PanelStack would not update when its props were changed.

In #3787 a change was made to only update the PanelStack if its stack length changed. Before that, a change to the stack would update the internal state of the animation and stack. This caused issues when 1) there were changes to the stack (but not its length) and 2) the component was animating, since it could trigger an animation in the opposite direction.

Unfortunately, it did not take into account that the stack could change without changing its length. This PR uses two different conditions to check if the stack has changed and if the animation direction needs to be updated.

This fixes an issue introduced in #3787.

### DIFF
--- a/packages/core/src/components/panel-stack/panelStack.tsx
+++ b/packages/core/src/components/panel-stack/panelStack.tsx
@@ -76,14 +76,17 @@ export class PanelStack extends AbstractPureComponent2<IPanelStackProps, IPanelS
 
     public componentDidUpdate(prevProps: IPanelStackProps, _prevState: IPanelStackState, _snapshot: {}) {
         super.componentDidUpdate(prevProps, _prevState, _snapshot);
+
+        // Always update local stack if stack prop changes
+        if (this.props.stack !== prevProps.stack && prevProps.stack != null) {
+            this.setState({ stack: this.props.stack.slice().reverse() });
+        }
+
+        // Only update animation direction if stack length changes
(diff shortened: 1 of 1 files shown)


sgarnotel:
Thanks, it works like a charm

PR #3887, opened 2019-12-10
Use ESLint to run TSlint

johnwiseheart:

As a step towards converting blueprint to use ESLint, start by using the `@typescript-eslint/tslint/config` rule from `@typescript-eslint` to run tslint inside of eslint. Following this, we can slowly move and/or rewrite rules while maintaining reasonable-sized PRs.

We should double check this is actually still runnning all the rules.


packages/eslint-config/package.json
```diff
@@ -0,0 +1,18 @@
+{
+    "name": "@blueprintjs/eslint-config",
+    "version": "1.9.0",
```
johnwiseheart: Feels like we should start this at 0.1.0 again.

johnwiseheart:
Hmm, it seems like eslint isn't passing through the `fix` status when calling tslint. [Opened an issue](https://github.com/typescript-eslint/typescript-eslint/issues/1340) against `typescript-eslint` - we should wait to merge this until that gets resolved.

johnwiseheart:
Fixed and merged https://github.com/typescript-eslint/typescript-eslint/pull/1342, just need to wait for them to release and then I'll update the PR.

johnwiseheart:
Updated to the version which includes the fixer, so this should be good to go now. Not sure if its worth looking at perf at all. Didn't notice it being particularly slower/faster (it should be about the same, since eslint is just invoking tslint).

packages/core/package.json
```diff
@@ -31,10 +31,10 @@
     "dist:css": "css-dist lib/css/*.css",
     "dist:variables": "generate-css-variables common/_colors.scss common/_color-aliases.scss common/_variables.scss",
     "dist:verify": "assert-package-layout",
-    "lint": "run-p lint:scss lint:ts",
+    "lint": "run-p lint:scss lint:es",
     "lint:scss": "sass-lint",
-    "lint:ts": "ts-lint",
-    "lint-fix": "ts-lint --fix",
```
adidahiya: I think this syntax was intentional since we don't intend to run this command with `npm-run-all "lint:*"`

CONTRIBUTING.md
```diff
@@ -39,7 +39,7 @@ A typical contributor workflow looks like this:
     - Linting is best handled by your editor for real-time feedback (see
       [Editor integration](https://github.com/palantir/blueprint/wiki/Editor-integration)). Run
       `yarn lint` to be 100% safe.
-    - TypeScript lint errors can often be automatically fixed by TSLint. Run lint fixes with `yarn lint-fix`.
+    - TypeScript lint errors can often be automatically fixed by ESLint. Run lint fixes with `yarn lint-fix`.
```
adidahiya: `yarn lint-fix` doesn't match what's in the next file. I would revert the script name change and keep it as `lint-fix`

packages/eslint-config/index.js
```diff
@@ -0,0 +1,41 @@
...
+    plugins: [
+        "@typescript-eslint",
+        "@typescript-eslint/tslint",
+    ],
+    parser: "@typescript-eslint/parser",
+    parserOptions: {
+        sourceType: "module",
+        ecmaFeatures: {
+            jsx: true,
+
```
adidahiya: nit: extra newline

packages/eslint-config/index.js
```diff
@@ -0,0 +1,41 @@
...
+        },
+        project: "**/tsconfig.json",
+    },
+    rules: {
+        // run the tslint rules which are not yet converted (run inside eslint)
+        "@typescript-eslint/tslint/config": ["error", {
+          "lintFile": "../../tslint.json"
```
adidahiya: nit: inconsistent indentation

packages/node-build-scripts/es-lint.js
```diff
@@ -0,0 +1,48 @@
...
+ * @fileoverview Runs ESLint, with support for generating JUnit report
+ */
+
+// @ts-check
+"use strict";
+
+const path = require("path");
+const fs = require("fs");
+const { junitReportPath } = require("./utils");
+const { spawn, execSync } = require('child_process');
```
adidahiya: nit: quotemark style

packages/node-build-scripts/es-lint.js
```diff
@@ -0,0 +1,48 @@
...
+const gitRoot = execSync("git rev-parse --show-toplevel").toString().trim();
+const commandLineOptions = ["-c", path.join(gitRoot, "./.eslintrc.json"), "--color", "-f", format];
+if (process.argv.includes("--fix")) {
+    commandLineOptions.push("--fix")
+}
+
+const fileGlob = "{src, test}/**/*.tsx";
+const absoluteFileGlob = path.resolve(process.cwd(), fileGlob);
+const anyFilesToLint = glob.sync(absoluteFileGlob)
+if (anyFilesToLint.length === 0) {
```
adidahiya: curious why we need to add this... does eslint complain if there are no files?

packages/node-build-scripts/es-lint.js
```diff
@@ -0,0 +1,48 @@
...
+    console.log(`Not running ESLint because no files match the glob "${fileGlob}"`)
+    process.exit();
+}
+
+const eslint = spawn("eslint", [...commandLineOptions, absoluteFileGlob]);
+
+eslint.stdout.pipe(outputStream);
+eslint.stderr.pipe(process.stderr);
+
+eslint.on('close', code => {
```
adidahiya: nit: quotemark style

adidahiya:
Linting seems to be working fine, as well as vscode integration (I disabled my TSLint extension and enabled ESLint) and auto-fixing. This error shows up at the top of every TS file, suggesting that semantic rules will not work... but I don't think we have any of those enabled in this repo:

image

johnwiseheart:
Hmm, I wonder if its applying rules in node_modules somehow, I wouldn't expect `node_modules/tsconfig.base.json` to ever be extended by something 🤔

PR #3888, opened 2019-12-10
Invalidate grid before rendering table when dimensions change

Nifdee:
### Checklist

- [x] Includes tests
- [ ] ~~Update documentation~~

Due to the react lifecycle changes in https://github.com/palantir/blueprint/pull/3702, the `Table` component would call `cellRenderer` on outdated row / column indices. The `grid` needs to be invalidated as soon as the component props change, only checking this in `componentDidUpdate` is too late.


packages/table/src/table.tsx
```diff
@@ -810,6 +810,10 @@ export class Table extends AbstractComponent2<ITableProps, ITableState, ITableSn
             enableColumnInteractionBar,
         } = this.props;
         const { horizontalGuides, numFrozenColumnsClamped, numFrozenRowsClamped, verticalGuides } = this.state;
+        if (!this.gridDimensionsMatchProps()) {
```
adidahiya: I feel like this should go into `componentDidUpdate` or another lifecycle method... I'm poking around at different places to move this check

PR #3913, opened 2020-01-02
[core] feat: add intent prop to Slider and RangeSlider

### LINKED ISSUE #3497: Allow non-multi sliders to set their intent or background color
#### Environment

- __Browser and OS versions__: Chrome, current

#### Feature request

The slider is a really nice component. It would be great if the default `<Slider>` element had an `intent` or another way to set the background color of the lefthand side of the slider.

#### Examples

`<Slider intent={...} />`

or

`<Slider progressClassName={...} />` which would get applied to the current `bp3-slider-progress` element

The latter provides a lot more flexibility, but the ergonomics of `intent` is nice as well.

jdpaterson:
#### Checklist

- [ ] Includes tests
- [ ] Update documentation

Implements intent prop for Slider and RangeSlider by setting ISliderBaseProps to extend IIntentProps.


packages/core/src/components/slider/rangeSlider.tsx
```diff
@@ -55,7 +55,11 @@ export class RangeSlider extends AbstractPureComponent2<IRangeSliderProps> {
         const { value, ...props } = this.props;
         return (
             <MultiSlider {...props}>
-                <MultiSlider.Handle value={value[RangeIndex.START]} type="start" intentAfter={Intent.PRIMARY} />
+                <MultiSlider.Handle
+                    value={value[RangeIndex.START]}
+                    type="start"
+                    intentAfter={props.intent || Intent.PRIMARY}
```
adidahiya: we try to avoid implicit type coercion like this. instead, please set the default intent to `Intent.PRIMARY` inside `public static defaultProps`. then this line can just be `={props.intent}`

packages/core/src/components/slider/slider.tsx
```diff
@@ -53,12 +53,13 @@ export class Slider extends AbstractPureComponent2<ISliderProps> {
 
     public render() {
         const { initialValue, value, onChange, onRelease, ...props } = this.props;
+        const intent = props.intent || Intent.PRIMARY;
```
adidahiya: same here, please use `public static defaultProps`

packages/core/src/components/slider/rangeSlider.tsx
jdpaterson: this change has been made

packages/core/src/components/slider/slider.tsx
jdpaterson: this change has also now been made

jdpaterson:
I accidentally pushed some code I was working on for something else, so I removed it and force pushed. If the force push is an issue I can create a new PR.

packages/core/src/components/slider/multiSlider.tsx
```diff
@@ -20,7 +20,7 @@ import { polyfill } from "react-lifecycles-compat";
 
 import { AbstractPureComponent2, Classes, Intent } from "../../common";
 import * as Errors from "../../common/errors";
-import { DISPLAYNAME_PREFIX, IProps } from "../../common/props";
+import { DISPLAYNAME_PREFIX, IProps, IIntentProps } from "../../common/props";
```
adidahiya: this is failing lint

jdpaterson: lint is fixed and merged to develop

PR #3917, opened 2020-01-09
chore: convert custom tslint rules to eslint

johnwiseheart:

Convert the tslint custom rules to eslint.

Its been a while since I made these changes, we should probably check they all still work and are linked up correctly.

I'm not sure how I feel about the IconComponents rule - may be worth removing/updating? Its unclear to me why you would want to enforce anything other than `<Icon icon={IconClasses.ICON_NAME} />`


packages/eslint-config/tslint.json
```diff
@@ -10,6 +10,10 @@
     // TODO: test dependencies are only declared at root but used in all packages.
     "no-implicit-dependencies": false,
     // TODO: Currently, a lot of chai assertions are not written as proper statements.
-    "no-unused-expression": false
+    "no-unused-expression": false,
+    // Turn off the blueprint tslint rules since we have replaced them with eslint rules
+    "blueprint-classes-constants": false,
```
johnwiseheart: Turn off the tslint rules now that we have the eslint ones

packages/eslint-plugin-blueprint/README.md
```diff
@@ -0,0 +1,114 @@
+<img height="204" src="https://cloud.githubusercontent.com/assets/464822/20228152/d3f36dc2-a804-11e6-80ff-51ada2d13ea7.png">
```
johnwiseheart: Need to update this to explain how to use eslint plugin

packages/eslint-plugin-blueprint/src/index.ts
```diff
@@ -0,0 +1,32 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import rules from "./rules";
+
+/**
+ * Enable Blueprint-specific TSLint rules defined in this package.
```
johnwiseheart:
```suggestion
 * Enable Blueprint-specific ESLint rules defined in this package.
```

johnwiseheart:
Unclear why the tests are failing in CI and not locally - the plugin module isn't being found correctly. Might be able to look into this some more early next week but feel free to poke if you have time.

adidahiya:
I'm updating & fixing up this PR now.

Apparently the icon-components rule was added so we might allow tree shaking of icons in the future (see https://github.com/palantir/blueprint/pull/2394, https://github.com/palantir/blueprint/pull/2356):

> it's an opt-in decision to ensure tree shaking of icons is possible.

so let's leave it in for now.

adidahiya:
`yarn test` wasn't actually getting run in CI for the new package. fixing that here: https://github.com/palantir/blueprint/pull/3938

Issue #2817, opened 2018-08-17
[Sketch] provide Symbols for components


romanr:
Continuing from #2668.   
Sketch file does not represent UI as symbols. Symbols in Sketch are like components in React.   

Imagine if controls in Blueprint were developed as functions in one monolithic javascript file. Then in order to use a component the developer had to open a large javascript file, copy relevant code and paste into their application.

TranTrongKh:
hi

newstart346:
Hi

giladgray:
@romanr yes we are aware and actively working on this.

equinusocio:
@giladgray Please make good use of symbols. All colors should be use a master simbol which, if changed, should affect all ui components. Also, remember to use anchor points inside fluid components and sub elements.

Let me know if you need help about that.

giladgray:
@romanr @equinusocio FYSA this is a _huge_ amount of work and we all have full time jobs over here so it'll take some time to symbol-ize the entire library. don't think there's much opportunity for OSS contributions here unfortunately. watch this issue for updates.

vkorobov-boostlabs:
I could propose a version of symbolizing everything for your review in March. What do you think?

adidahiya:
@pkwi is this still on our intended roadmap for the sketch file? if so, I think @vkorobov-boostlabs can go for it

pkwi:
@vkorobov-boostlabs @adidahiya yeah, that would be great!

vkorobov-boostlabs:
@pkwi @adidahiya I made a fork and working on it.

equinusocio:
To work on sketch files together there are some tools like:

https://www.abstract.com/
https://plantapp.io/
https://kactus.io/

vkorobov-boostlabs:
Hey guys, quick update. Working on symbols. Buttons are done including buttons with icons (left, right and both). 
![image](https://user-images.githubusercontent.com/47153854/55409147-74a05800-552f-11e9-8835-c8d7c5c3c534.png)
Structured good:
![image](https://user-images.githubusercontent.com/47153854/55409184-88e45500-552f-11e9-9f90-5109f0f13cd1.png)
I'm doing that in the fork here https://github.com/boostlabscom/blueprint

vkorobov-boostlabs:
Did 60% of symbolization. 
Here is what's left to do:
- Forms
- Dropdown menus
- Lists & trees
- Navigation bar
- Breadcrumbs
- Dialogs & alerts
[...]

PR #3933, opened 2020-01-20
[resources] add icons sketch library

estevesd:
Add a separate Sketch Library for Icons.
16/20px, dark/Light versions are available for easy of use in sketch
Nested components allow overrides with shared styles.
Icons are organised by categories like on the docs

Is it useful for everyone? or only me?

#### Screenshot
blueprint-icons-library


palantirtech:
Thanks for your interest in [palantir/blueprint](https://github.com/palantir/blueprint), @estevesd! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

estevesd:
based on discussions in https://github.com/palantir/blueprint/issues/2817
the icons have been added in the kit instead of a separate library.
shall we keep it like this or still do a separate library only for icons ?

aycai:
I think for now keeping it in the kit file (as is) works. I'm wary of perf, but it seems to work fine so far.

I was going to say that you don't need separate symbols for dark and light, but checking your file in #2817 looks like you fixed that :) 

Changes in there look good as well 👍

PR #3950, opened 2020-02-02
[core] fix(PanelStack): rename prop and other fixes

adidahiya:
- rename renderCurrentPanelOnly -> renderActivePanelOnly
- add to docs example to demonstrate the effect of the new prop


adidahiya:
still not quite working:

PR #3973, opened 2020-02-20
fix right element padding adjusting on update

### LINKED ISSUE #3965: InputGroup rightElementWidth no longer gets recomputed if rightElement changes
#### Environment

- __Package version(s)__: @blueprintjs/core 3.23.1
- __Browser and OS versions__: Windows

#### Steps to reproduce

1. Create a InputGroup that has no rightElement on mount
2. Update prop of InputGroup to include a rightElement

#### Actual behavior

The InputGroup does not resize to adjust its padding-right for the rightElement, which causes overlapping.

#### Expected behavior

The inputGroup recomputes its padding-right to reflect the new rightElement prop.

#### Possible solution

I believe this bug is caused by a change [here](https://github.com/palantir/blueprint/pull/3813), which seemed to be an attempt to fix some sort of infinite loop scenario.  The fix ended up removing any re-calculation via `componentDidUpdate`, which means the right padding is effectively _locked in_ after initial render.  I think perhaps we should introduce the `componentDidUpdate`, but only recompute the rightElementWidth if the `rightElement` prop changed.

https://github.com/palantir/blueprint/pull/3813
https://github.com/palantir/blueprint/commit/66b16dd5d1833fd4e6da21423fee8dc9a62aecd2

lexistein:
#### Checklist

- [ ] Includes tests
- [ ] Update documentation

Reverted a change that stopped updating input field padding when right element changed. Infinite update is fixed by a check to see if the rightElement prop changed.

#### Screenshot
Before
bpbad

After
bp

### DIFF
--- a/packages/core/src/components/forms/inputGroup.tsx
+++ b/packages/core/src/components/forms/inputGroup.tsx
@@ -129,13 +129,11 @@ export class InputGroup extends AbstractPureComponent2<IInputGroupProps & HTMLIn
     }
 
     public componentDidMount() {
-        if (this.rightElement != null) {
-            const { clientWidth } = this.rightElement;
-            // small threshold to prevent infinite loops
-            if (Math.abs(clientWidth - this.state.rightElementWidth) > 2) {
-                this.setState({ rightElementWidth: clientWidth });
-            }
-        }
+        this.updateInputWidth();
+    }
+
+    public componentDidUpdate() {
+        this.updateInputWidth();
     }
 
     private maybeRenderRightElement() {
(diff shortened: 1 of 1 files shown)


palantirtech:
Thanks for your interest in [palantir/blueprint](https://github.com/palantir/blueprint), @lexistein! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

packages/core/src/components/forms/inputGroup.tsx
```diff
@@ -129,12 +129,12 @@ export class InputGroup extends AbstractPureComponent2<IInputGroupProps & HTMLIn
     }
 
     public componentDidMount() {
-        if (this.rightElement != null) {
-            const { clientWidth } = this.rightElement;
-            // small threshold to prevent infinite loops
-            if (Math.abs(clientWidth - this.state.rightElementWidth) > 2) {
```
adidahiya: huh this code was pretty wonky anyway. thanks for the fix!

PR #3974, opened 2020-02-21
set button user select to none

lexistein:
#### Checklist

- [ ] Includes tests
- [ ] Update documentation

Button text is currently selectable on right click. This PR simply sets user-select to none.

#### Screenshot
Before:
image

### DIFF
--- a/packages/core/src/components/button/_common.scss
+++ b/packages/core/src/components/button/_common.scss
@@ -109,6 +109,7 @@ $button-intents: (
   vertical-align: middle;
   text-align: left;
   font-size: $pt-font-size;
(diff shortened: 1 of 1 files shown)


adidahiya:
no, I don't want to interfere with this default browser behavior. I don't see a strong justification for this change

lexistein:
Then why is it done for tree nodes, file inputs, table headers, etc?

adidahiya:
Those components are more complex and support selection and other features... Button is intentionally very simple and basically just renders markup. You can easily apply this user-select styling customization in your own app if you prefer to apply it to Buttons across the board.

PR #3993, opened 2020-02-29
[core] feat(TagInput): handle delete to remove items

grancalavera:
#### Checklist

- [x] Includes tests
- [x] Update documentation

Allows users to press `DELETE` on their keyboards to delete an item from `<TagInput />`, which in turn is used by the `Select` family of components.

Handling `DELETE` to remove item mimics the way `DELETE` works in a text editor, where an user can delete from the beginning of the line forward. Reaching the end of the line, in our case passing the last selected item in `<TagInput />` will stop deleting. Moving the selection to the first item and deleting from that point onwards repeatedly will delete all the selected items.

- Maybe `DELETE` was not implemented on purpose, if that's the case then ignore this PR.
- Ensure new behavior is in the spirit of Blueprint
- Ensure this change is compatible with other components, since I know other components depend on `<TagInput />`

### DIFF
--- a/packages/core/src/components/tag-input/tag-input.md
+++ b/packages/core/src/components/tag-input/tag-input.md
@@ -25,7 +25,7 @@ new items. A `separator` prop is supported to allow multiple items to be added
 at once; the default splits on commas and newlines.
 
 **Tags can be removed** by clicking their <span class="@ns-icon-standard @ns-icon-cross"></span>
-buttons, or by pressing <kbd>backspace</kbd> repeatedly.
+buttons, or by pressing either <kbd>backspace</kbd> or <kbd>delete</kbd> repeatedly. Pressing <kbd>delete</kbd> mimics the behaviour of deleting in a text editor, where trying to delete at the end of the last line will do nothing.
 Arrow keys can also be used to focus on a particular tag before removing it. The
 cursor must be at the beginning of the text input for these interactions.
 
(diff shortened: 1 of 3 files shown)


adidahiya:
[docs preview artifact index.html](https://30370-71939872-gh.circle-artifacts.com/0/packages/docs-app/dist/index.html)

packages/core/src/components/tag-input/tag-input.md
```diff
@@ -25,7 +25,7 @@ new items. A `separator` prop is supported to allow multiple items to be added
 at once; the default splits on commas and newlines.
 
 **Tags can be removed** by clicking their <span class="@ns-icon-standard @ns-icon-cross"></span>
-buttons, or by pressing <kbd>backspace</kbd> repeatedly.
+buttons, or by pressing either <kbd>backspace</kbd> or <kbd>delete</kbd> repeatedly. Pressing <kbd>delete</kbd> mimics the behaviour of deleting in a text editor, where trying to delete at the end of the last line will do nothing.
```
adidahiya:
```suggestion
buttons, or by pressing either <kbd>backspace</kbd> or <kbd>delete</kbd> repeatedly.
Pressing <kbd>delete</kbd> mimics the behavior of deleting in a text editor, where trying to delete at the end of the line will do nothing.
```

grancalavera:
@adidahiya thanks for your feedback! I've added it to the PR.

PR #4005, opened 2020-03-07
Solved bug when using fixed position in a child of a Collapse node

### LINKED ISSUE #4004: Fixed position on child of Collapse doesnt work
#### Environment

- __Package version(s)__: 3.23.1
- __Browser and OS versions__: Chrome / Windows 10

If possible, link to a minimal repro (fork [this code sandbox](https://codesandbox.io/s/blueprint-sandbox-et9xy)): https://codesandbox.io/s/blueprintjs-collapse-fixed-position-iiu13

#### Steps to reproduce

1. Open the collapse
2. Move your cursor around the grey area inside the collapse
3. Scroll down

#### Actual behavior

As you can see, the rectangle which is supposed to follow the cursor (and works as intended in the green area outside of the collapse), doesnt follow it properly.

#### Possible solution

The issue can be fixed by disabling the translateY(0px) in the CSS of the collapse body, when the collapse is uncollapsed.

A more detailed explanation of this issue can be found here (basically, any transform even of 0px, will make the fixed position unusable for any child of the component) : 

https://achrafkassioui.com/blog/position-fixed-and-CSS-transforms/

A possible fix is simply to not define any transform in the CSS when the collapse component is fully uncollapse and there is no need for a transform anymore.

Annotation 2020-03-07 121823

Lumyo:
#### Checklist

- [ ] Includes tests
- [ ] Update documentation

Removed transform in the CSS of the Collapse component, when the Collapse is fully uncollapsed, instead of having a transform of 0.

This enable the use of fixed position to be able to have a component following the cursor, on children of the collapse.

### DIFF
--- a/packages/core/src/components/collapse/collapse.tsx
+++ b/packages/core/src/components/collapse/collapse.tsx
@@ -176,7 +176,9 @@ export class Collapse extends AbstractPureComponent2<ICollapseProps, ICollapseSt
 
         const contentsStyle = {
             // only use heightWhenOpen while closing
-            transform: displayWithTransform ? "translateY(0)" : `translateY(-${this.state.heightWhenOpen}px)`,
+            // If translate is equal to 0, we use a none transform, to avoid having problem if trying to use
+            // fixed position in a child of this node
+            transform: displayWithTransform ? "none" : `translateY(-${this.state.heightWhenOpen}px)`,
             // transitions don't work with height: auto
             transition: isAutoHeight ? "none" : undefined,
         };


adidahiya:
[docs preview link](https://30438-71939872-gh.circle-artifacts.com/0/packages/docs-app/dist/index.html#core/components/collapse)

Lumyo:
I found out that I could achieve what I wanted to do, by embedding the fixed element inside an overlay, so I am closing this PR.

PR #4034, opened 2020-04-01
[core] Stop propagation on remove tag

### LINKED ISSUE #3310: MultiSelect tagInputProps: How to avoid input focus when onRemove triggered?
#### Environment

- __Package version(s)__: 2

#### Question
Is it possible to prevent input focus when onRemove triggered?  Thanks in advance!

grancalavera:
#### Checklist

- [ ] Includes tests
- [ ] Update documentation
- [x] Update example

Fixes #3310 by stopping event propagation when removing tags from `InputTag`.

This PR assumes event propagation isn't required by anything else. Another approach would be to pass the event to consumers of `TagInput`, but unless the event is required internally for something else, this might not be necessary.

#### Screenshot

tag-input-stop-propagation

### DIFF
--- a/packages/core/src/components/tag-input/tagInput.tsx
+++ b/packages/core/src/components/tag-input/tagInput.tsx
@@ -436,6 +436,7 @@ export class TagInput extends AbstractPureComponent2<ITagInputProps, ITagInputSt
     };
 
     private handleRemoveTag = (event: React.MouseEvent<HTMLSpanElement>) => {
+        event.stopPropagation();
         // using data attribute to simplify callback logic -- one handler for all children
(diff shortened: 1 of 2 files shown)


grancalavera:
It might be also worth fixing the example so that event propagation is stopped on click on `rightElement` :thinking:

grancalavera:
@adidahiya what about this?

> Another approach would be to pass the event to consumers of TagInput

Something else we could try is handling the event propagation in `MultiSelect`, and assuming we can know the identity of the event target, ignore click events originating from removing a tag. Not sure if that's even possible, but I'm happy to explore that path.

What do you think?

PR #4050, opened 2020-04-08
[icons] feat: Added virus icon

jbcolin18:
#### Checklist

- [ ] Includes tests
- [ ] Update documentation

Added virus icon to icon library approved by other Palantir designers. I didn't add legacy file types.

Icons should render as below screenshot.

#### Screenshot

Screen Shot 2020-04-08 at 1 33 12 PM


palantirtech:
Thanks for your interest in [palantir/blueprint](https://github.com/palantir/blueprint), @jbcolin18! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

adidahiya:
The icon doesn't show up correctly with the `<Icon>` react component. I think it's because multiple paths are used. In theory, this should be ok (we should be able to render such SVGs into the DOM), but in practice, it looks like all other icons in Blueprint are drawn with a single `<path>`, even for disjoint shapes:

layout-grid:

image

virus:

image

is there any way you can configure the export from Sketch to only draw one `<path>` in the icon contents?

adidahiya:
Actually this could be a problem in the way we optimize svgs with `svgo`... i'm taking a closer look

packages/icons/resources/icons/icons.json
```diff
@@ -3400,5 +3400,11 @@
     "tags": "science, beaker, chemistry, mixing, scientist",
     "group": "interface",
     "content": "\\e90e"
+  },
+  {
+    "displayName": "Virus",
+    "iconName": "virus",
+    "tags": "virus, doctor, medical, examination, disease, pandemic, sick",
+    "group": "interface"
   }
```
adidahiya, line 3409: I removed the "content" field here because it was unused, and I changed the build scripts to allow for this omittance

PR #4060, opened 2020-04-12
Fix failure to run eslint on win platform

justinbhopper:
Fixes errors when running `lerna run lint` on a win-platform.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -4,4 +4,4 @@ lib
 fixtures
 coverage
 __snapshots__
-generated
+**/src/generated
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,6 +1,7 @@
 node_modules
 dist
+lib
 fixtures
 coverage
 __snapshots__
-src/generated
+generated
--- a/.npmrc
+++ b/.npmrc
@@ -1,0 +1,2 @@
+registry=https://registry.npmjs.org
+always-auth=true
--- a/.npmrc
+++ b/.npmrc
@@ -1,1 +1,0 @@
-registry=https://registry.npmjs.org
--- a/.npmrc
+++ b/.npmrc
@@ -1,2 +1,1 @@
 registry=https://registry.npmjs.org
-always-auth=true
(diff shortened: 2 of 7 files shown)


.npmrc
```diff
@@ -0,0 +1 @@
+registry=https://registry.npmjs.org
```
adidahiya, line 1: we use yarn, so I don't think this is necessary

.eslintignore
```diff
@@ -1,6 +1,7 @@
 node_modules
 dist
+lib
 fixtures
 coverage
 __snapshots__
-src/generated
+generated
```
adidahiya, line 7: why this change? are there other generated folders?

.npmrc
justinbhopper: yarn respects this file too, but you are right it is not necessary assuming users have not overridden their default registry locally.  I had overridden mine, but have recently fixed that.  Will remove

.eslintignore
justinbhopper: For some reason, eslint was linting the generated folders.  I couldn't figure out why, because the path `src/generated` seems correct.  I had assumed it was due to some win-platform thing.  I will retest and see if I can avoid this change or improve the path.

justinbhopper: I modified the path to be `**/src/generated` and it appears to work.  I can only surmise that `src/generated` was being interpretted as an absolute path from the root.  According to the docs and examples I found online, using ** for wildcard nested directories is necessary.

packages/eslint-config/index.js
```diff
@@ -30,6 +30,7 @@ module.exports = {
     parser: "@typescript-eslint/parser",
     parserOptions: {
         sourceType: "module",
+        createDefaultProgram: true,
```
adidahiya, line 33: this ended up causing a huge performance hit (see https://github.com/typescript-eslint/typescript-eslint/tree/master/packages/parser#parseroptionscreatedefaultprogram), so I am reverting this change in https://github.com/palantir/blueprint/pull/4078

PR #4093, opened 2020-04-25
Add support for callback ref and ref hook

### LINKED ISSUE #4064: [core] Button and InputGroup ref prop types not assignable with React hooks
#### Environment

- __Browser and OS versions__: Google Chrome 80.0.3987.149, MacOS 10.14.6

#### Question

What is the correct way of using ElementRef for a Button component?
When I do the following, the `normalButtonRef` is set but the `bpButtonRef` is not.

```javascript
import { Button } from '@blueprintjs/core';

const MyComponent = (props) => {
  const normalButtonRef = React.useRef(null);
  const bpButtonRef = React.useRef(null);
  
React.useEffect(() => {
    console.log(normalButtonRef); // {current: HTMLButtonElement}
    console.log(bpButtonRef); // {current: null}
  }, []);

return (
  <div>
    <button ref={normalButtonRef} />
    <Button elementRef={bpButtonRef} />
  <div>
);
```

<ins>Edit</ins>: typos in example code

tanmoyopenroot:
#### Checklist

- [x] Includes tests
- [x] Update documentation

Add support for `callback ref` and `ref hook` for `Button`, `Anchor Button` and `InputGroup`

Typing for the props(`elementRef`, `inputRef`)

### DIFF
--- a/packages/core/src/components/button/abstractButton.tsx
+++ b/packages/core/src/components/button/abstractButton.tsx
@@ -86,13 +86,10 @@ export abstract class AbstractButton<H extends React.HTMLAttributes<any>> extend
         isActive: false,
     };
 
-    protected getButtonRefHandler = (): (Utils.IRefCallback<any> | Utils.IRefObject<any>) => {
-        const elementRef = this.props.elementRef as (Utils.IRefCallback<HTMLElement> | Utils.IRefObject<HTMLElement>);
+    protected getButtonRefHandler = (): Utils.IRefCallback<any> | Utils.IRefObject<any> => {
+        const elementRef = this.props.elementRef as Utils.IRefCallback<HTMLElement> | Utils.IRefObject<HTMLElement>;
 
-        if (
-            elementRef &&
-            !Utils.isFunction(elementRef)
-        ) {
+        if (elementRef && !Utils.isFunction(elementRef)) {
             this.buttonRef = elementRef;
 
             return elementRef;
@@ -107,7 +104,7 @@ export abstract class AbstractButton<H extends React.HTMLAttributes<any>> extend
 
     protected buttonRefCallback = (ref: HTMLElement | Utils.IRefObject<HTMLElement>) => {
         this.buttonRef = ref;
-    }
+    };
 
     protected buttonRef: HTMLElement | Utils.IRefObject<HTMLElement>;
     protected refHandlers = {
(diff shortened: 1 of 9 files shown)


tanmoyopenroot:
Currently `test-react-15` is failing as `hooks` are not available.
Is there any way we can detect `react version` to skip those test?

packages/core/src/components/button/abstractButton.tsx
```diff
@@ -39,7 +39,7 @@ export interface IButtonProps extends IActionProps {
     alignText?: Alignment;
 
     /** A ref handler that receives the native HTML element backing this component. */
-    elementRef?: (ref: HTMLElement | null) => any;
+    elementRef?: React.Ref<HTMLElement>;
```
adidahiya, line 42: I don't love this approach because this type reference will not compile with `@types/react@^15` (the latest version of which was published in October 2019).

There may be some React 15 users of Blueprint who use those typings (albeit I have not tested this myself, whether Blueprint compiles with `@types/react@^15`, but judging by a quick search of our references to react types across the monorepo, I think it does)

You should try to make this change without referencing the new `React.Ref` type, maybe with a new function in a utils file which helps you get the ref across React versions:

```ts
// interface which is assignable to React.Ref<T>, but does not require us to depend on @types/react@^16
export interface IRefObject<T> {
    readonly current: T | null;
}

export function getRef<T>(ref: any) {
    return isFunction(ref) ? ref as (r: T | null) => any
        : typeof ref.current !== "undefined" ? ref as IRefObject<T>
        : null;
}
```

packages/core/test/buttons/buttonTests.tsx
```diff
@@ -103,6 +103,62 @@ function buttonTestSuite(component: React.ComponentClass<any>, tagName: string)
...
+            );
+
+            // wait for the whole lifecycle to run
+            setTimeout(() => {
+                assert.equal(elementRef, (wrapper.instance() as any).buttonRef);
+                done();
+            }, 0);
+        });
+
+        it("matches buttonRef with elementRef using useRef", done => {
```
adidahiya, line 143: wrap this whole `it` in a:

```ts
if (typeof React.useRef !== "undefined") {
    ...
}
```

packages/core/src/components/button/abstractButton.tsx
tanmoyopenroot: Done

packages/core/test/buttons/buttonTests.tsx
tanmoyopenroot: Done

adidahiya:
[docs preview](https://31898-71939872-gh.circle-artifacts.com/0/packages/docs-app/dist/index.html)

adidahiya:
thanks @tanmoyopenroot!

PR #4099, opened 2020-04-27
Add vertical padding to "Tag" component for large style

### LINKED ISSUE #4037: Tag element has no Y padding when "large" and "multiline"
#### Environment

If possible, link to a minimal repro (fork [this code sandbox](https://codesandbox.io/s/blueprint-sandbox-et9xy)): https://codesandbox.io/s/festive-wing-j7s4e

#### Steps to reproduce

1. Create a Tag element
2. Set it to be "large" and "multiline"
3. Set a long text as its content

#### Actual behavior

The Tag element has no top and bottom padding

#### Expected behavior

The Tag element should have top and bottom margins, with same size as the left and right ones.

Eusbolh:
#### Checklist

- [ ] Includes tests
- [ ] Update documentation

Add vertical padding to "Tag" component for large style

Ratio between horizontal and vertical padding value should be decided. For default style, the ratio  between horizontal and vertical padding is 3 to 2. However, I followed 2 to 1 ratio for large style which is the current ratio for single line large tags.

#### Screenshot

There is no visual change for single line content.

For multiline content, the old version is:
image
The updated version for multiline content is:
image

### DIFF
--- a/packages/core/src/components/tag/_common.scss
+++ b/packages/core/src/components/tag/_common.scss
@@ -12,6 +12,7 @@ $tag-margin: ($pt-input-height - $tag-height) / 2 !default;
 
 $tag-height-large: $pt-grid-size * 3 !default;
 $tag-line-height-large: $pt-icon-size-large !default;
+$tag-padding-large-top: ($tag-height-large - $tag-line-height-large) / 2 !default;
 $tag-padding-large: ($tag-height-large - $tag-line-height-large) !default;
 
 $tag-icon-spacing: ($tag-height - 12px) / 2 !default;
@@ -70,7 +71,7 @@ $tag-round-adjustment: 2px !default;
   line-height: $tag-line-height-large;
   min-height: $tag-height-large;
   min-width: $tag-height-large;
-  padding: 0 $tag-padding-large;
+  padding: $tag-padding-large-top $tag-padding-large;
 
   &.#{$ns}-round {
     padding-left: $tag-padding-large + $tag-round-adjustment;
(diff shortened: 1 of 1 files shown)


palantirtech:
Thanks for your interest in [palantir/blueprint](https://github.com/palantir/blueprint), @Eusbolh! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

packages/core/src/components/tag/_common.scss
```diff
@@ -70,7 +71,7 @@ $tag-round-adjustment: 2px !default;
   line-height: $tag-line-height-large;
   min-height: $tag-height-large;
   min-width: $tag-height-large;
-  padding: 0 $tag-padding-large;
+  padding: $tag-padding-large-top $tag-padding-large;
```
adidahiya, line 74: can you just use `($tag-padding-large / 2)` here instead? I'd prefer that over adding a new Sass variable

packages/core/src/components/tag/_common.scss
```diff
@@ -12,6 +12,7 @@ $tag-margin: ($pt-input-height - $tag-height) / 2 !default;
 
 $tag-height-large: $pt-grid-size * 3 !default;
 $tag-line-height-large: $pt-icon-size-large !default;
+$tag-padding-large-top: ($tag-height-large - $tag-line-height-large) / 2 !default;
```
adidahiya, line 15:
```suggestion
```

packages/core/src/components/tag/_common.scss
adidahiya, line 74:
```suggestion
  padding: ($tag-padding-large / 2) $tag-padding-large;
```

PR #4108, opened 2020-05-03
Adding resetOnSelect behavior to created item

### LINKED ISSUE #4103: @blueprintjs/select: creating an item remove the value in the input
#### Environment

- __Package version(s)__: https://codesandbox.io/s/blueprint-sandbox-i6ox9?file=/package.json

Chrome Version 81.0.4044.113, Mint 

  Operating System: Linux Mint 19.1
            Kernel: Linux 4.15.0-64-generic
      Architecture: x86-64

If possible, link to a minimal repro (fork [this code sandbox](https://codesandbox.io/s/blueprint-sandbox-et9xy)): https://codesandbox.io/s/blueprint-sandbox-i6ox9?file=/src/index.tsx

#### Steps to reproduce

1. type h
2. select hello
3. click outside of the input
4. click in the input: you should be able to modify the query 'hello' 
5. remove hello and type sugar
6. press enter
7. click outside of the input
8. click in the input: the query was removed and we can't modify 'sugar' 

#### Actual behavior

It remove the value in the input when we create a new item so we can't modify it. It shouldn't. 

#### Expected behavior

It should do the same behavior as if I wasn't creating a new item. 

------------

Currently I just want an autocomplete input but the design of the suggest is too way far from what I want for the input behavior. (I want something like https://twitter.github.io/typeahead.js/)

I don't understand why the input is not correlated with the suggester. Why the default behavior is not just like typeahead, the expected behavior for most of us I think  :thinking: ?

Thank's for reading, I hope I am just stupid and you will have a solution haha :)

alexisvisco:
#### Fixes #4103 

After investigating the source code I found an issue with the `resetOnSelect` when there is an item creation. The current behavior when an item is created is by default reset the query event if `resetOnSelect` is set to false

This PR change the behavior when an item is created and if `resetOnSelect` is true. 
 If `resetOnSelect` is false the query is not reset.

----------

The code change is minimal so the review should be fast :)


palantirtech:
Thanks for your interest in [palantir/blueprint](https://github.com/palantir/blueprint), @alexisvisco! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

adidahiya:
This seems like a reasonable change, would you mind adding a unit test @alexisvisco?

FYSA @cmslewis

cmslewis:
@alexisvisco Are you seeing this bug in any of the higher-order components that use `QueryList` (i.e., `Select`, `Suggest`, `MultiSelect`, `Omnibar`) or just when using `QueryList` directly? Code change looks reasonable, but would be helpful to see a repro if you have one.

alexisvisco:
This bug is in all components that use `QueryList` that allow to change `resetOnSelect` and allow creating new item. Again this bug is only when you are creating new item, you create it, but in a UX perspective you are selecting it because there is no differences (a part from the fact that you are creating an item which is transparent for the user).

@cmslewis instead of a repo I have made a video that demonstrate the change between the actual production blueprint and my change https://youtu.be/C2m8WvYxpqw

Can you give me some hint to create tests please ? I don't known how to trigger a item selection / creation. I have looked your tests in the `queryListTests.tsx` but I don't really understand how it works. 

Thanks you for your reviewing. :)

adidahiya:
@alexisvisco sorry I didn't mention the new `yarn format` script in the dev docs -- adding that here https://github.com/palantir/blueprint/pull/4118

adidahiya:
friendly bump on this @alexisvisco

alexisvisco:
I don't understand, the yarn format that I am running don't indicate any errors ... Can you help me ?

adidahiya:
merged develop into your branch. here's the latest [docs preview build](https://33031-71939872-gh.circle-artifacts.com/0/packages/docs-app/dist/index.html)

PR #4121, opened 2020-05-12
Update MultiSelect Popover target-click work-around for controlled usage

thasner:
MultiSelect is backed by Popover, which closes itself whenever the `target` is clicked:
https://github.com/palantir/blueprint/blob/a98ac28dc5079a728fa40d8c4d31faa98b569001/packages/core/src/components/popover/popover.tsx#L518-L528

For a MultiSelect, the `target` is the search input.
Uncontrolled usage is protected by a check against the `activeElement` before changing state. Controlled usage (specifically, the call to `onInteraction`) does not have this protection.

### DIFF
--- a/packages/select/src/components/select/multiSelect.tsx
+++ b/packages/select/src/components/select/multiSelect.tsx
@@ -184,11 +184,12 @@ export class MultiSelect<T> extends React.PureComponent<IMultiSelectProps<T>, IM
             if (this.input != null && this.input !== document.activeElement) {
                 // the input is no longer focused so we can close the popover
                 this.setState({ isOpen: false });
+                Utils.safeInvokeMember(this.props.popoverProps, "onInteraction", false);
             } else if (!this.props.openOnKeyDown) {
                 // open the popover when focusing the tag input
                 this.setState({ isOpen: true });
+                Utils.safeInvokeMember(this.props.popoverProps, "onInteraction", true);
             }
-            Utils.safeInvokeMember(this.props.popoverProps, "onInteraction", nextOpenState);
         });
 
     private handlePopoverOpened = (node: HTMLElement) => {


adidahiya:
@thasner did you by chance enable Circle for your fork? That will prevent CI from running here, so you should avoid that. I think if you stop building and push a new commit it should trigger the build again.

This change sounds reasonable but can you point out to me what code this is referring to?

> Uncontrolled usage is protected by a check against the activeElement before changing state.

packages/select/src/components/select/multiSelect.tsx
```diff
@@ -184,11 +184,12 @@ export class MultiSelect<T> extends React.PureComponent<IMultiSelectProps<T>, IM
             if (this.input != null && this.input !== document.activeElement) {
```
thasner, line 184: @adidahiya for uncontrolled usage, `nextOpenState` is entirely ignored.
Regardless of the value of `nextOpenState`, we close the popover if the input is not active.

Before this change, we'd still forward the (potentially incorrect) value of `nextOpenState` to those with controlled usage (through `onInteraction`).
After this change, we only call `onInteraction` when we'd also call `setState`, and we call with the literal boolean values, indifferent of the `nextOpenState` value.

packages/select/src/components/select/multiSelect.tsx
```diff
@@ -184,11 +184,12 @@ export class MultiSelect<T> extends React.PureComponent<IMultiSelectProps<T>, IM
...
                 // the input is no longer focused so we can close the popover
                 this.setState({ isOpen: false });
+                Utils.safeInvokeMember(this.props.popoverProps, "onInteraction", false);
             } else if (!this.props.openOnKeyDown) {
```
thasner, line 188: separately, I don't understand what this code is doing.
`if (!this.props.openOnKeyDown)`, we always set `isOpen: true`, regardless of the value of `nextOpenState`. I don't understand that. I'd think this would make any Popover with `!this.props.openOnKeyDown` be impossible to close

thasner: I'd love to change this code entirely to be:
```
if (this.input != null && this.input === document.activeElement) {
    // the input is active, we stay open regardless of `nextOpenState`
    this.setState({ isOpen: true });
    Utils.safeInvokeMember(this.props.popoverProps, "onInteraction", true);
} else {
    Utils.safeInvokeMember(this.props.popoverProps, "onInteraction", false);
}
```

adidahiya: that code looks good to me, and it seems to pass unit tests on my end, do you want to push it up or should I add a commit?

adidahiya:
closing in favor of https://github.com/palantir/blueprint/pull/4133, but I need to do some more investigation here.. the "open on key down" behavior is still not quite right

PR #4122, opened 2020-05-12
Add control of hasBackdrop prop in Dialog

alxmiron:
#### Checklist

- [ ] Includes tests
- [x] Update documentation

Sometimes I want to reuse Dialog component but without a backdrop.  
I even don't understand why dialog is required to have it 

#### Screenshot

image

### DIFF
--- a/packages/core/src/components/dialog/dialog.tsx
+++ b/packages/core/src/components/dialog/dialog.tsx
@@ -33,12 +33,6 @@ export interface IDialogProps extends IOverlayableProps, IBackdropProps, IProps 
     isOpen: boolean;
 
     /**
-     * Dialog always has a backdrop so this prop is excluded from the public API.
-     * @internal
-     */
-    hasBackdrop?: boolean;
-
-    /**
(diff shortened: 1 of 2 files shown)


adidahiya:
This is an intentional design choice in the design system. Dialogs are commonly used components, and we want to enforce consistent visual affordances provided by the dialog backdrop. If you need a custom overlay solution where you remove the backdrop, you are welcome to use `Overlay` directly.

cnrudd:
The documentation here:
https://blueprintjs.com/docs/#core/components/dialog
should explicitly point out this "you will always have a backdrop on your Dialog" design decision.

The `backdropProps`  prop documentation on this Dialog page misleads developers to the opposite conclusion because the 
`interface IBackdropProps` documentation features `hasBackdrop`.

It's not until developers dig into the code that they can see that the `hasBackdrop` prop is permanently set to true on the Dialog's Overlay.  This should be pointed out in the docs.

The CSS example here on the same page: https://blueprintjs.com/docs/#core/components/dialog.css shows a much less important backdrop or none at all, and this adds to the misdirection.

PR #4133, opened 2020-05-19
[select] docs(MultiSelect): clarify openOnKeyDown prop

adidahiya:
supersedes https://github.com/palantir/blueprint/pull/4121

@thasner


packages/select/src/components/select/multiSelect.tsx
```diff
@@ -178,17 +178,17 @@ export class MultiSelect<T> extends React.PureComponent<IMultiSelectProps<T>, IM
...
         // deferring to rAF to get properly updated document.activeElement
         requestAnimationFrame(() => {
-            if (this.input != null && this.input !== document.activeElement) {
-                // the input is no longer focused so we can close the popover
-                this.setState({ isOpen: false });
-            } else if (!this.props.openOnKeyDown) {
-                // open the popover when focusing the tag input
+            if (this.input != null && this.input === document.activeElement) {
+                // the input is active, we stay open regardless of `nextOpenState`
                 this.setState({ isOpen: true });
```
thasner, line 186: hmmmm, after looking at the docs for `openOnKeyDown`, I'm realizing this might not be right https://blueprintjs.com/docs/#select/multi-select

I think if the input is still focused, we definitely don't close (so the below branch is correct).
However, in this branch, I think we only open if `!this.props.openOnKeyDown`. I assume somewhere else there's a keydown handler that opens the popover?

Maybe the original change in https://github.com/palantir/blueprint/pull/4121 was correct
(assuming the keydown handler also calls `onInteraction`)

adidahiya:
I looked more closely at this and decided that the original code was correct; the prop documentation was a little unclear though. I've updated the documentation and code to be more explicit and clear about the behavior. I've also filed a related issue about a small inconsistency in the API: https://github.com/palantir/blueprint/issues/4152

adidahiya:
@thasner ^

packages/select/src/components/select/multiSelect.tsx
```diff
@@ -178,17 +188,21 @@ export class MultiSelect<T> extends React.PureComponent<IMultiSelectProps<T>, IM
...
+                // input is no longer focused, we should close the popover
                 this.setState({ isOpen: false });
             } else if (!this.props.openOnKeyDown) {
-                // open the popover when focusing the tag input
+                // we should open immediately on click focus events
                 this.setState({ isOpen: true });
             }
-            Utils.safeInvokeMember(this.props.popoverProps, "onInteraction", nextOpenState);
+
+            this.props.popoverProps?.onInteraction?.(nextOpenState);
```
thasner, line 205: why is `onInteraction` not always treated the same as `setState`?

adidahiya: there are some code paths here when `setState` does not get called. for example when the input gets clicked and `openOnKeyDown={true}`. to be "correct" according to the behavior of Popovers in general, we should still call `onInteraction`.

PR #4137, opened 2020-05-20
 [select] fix(MultiSelect): create new item with IME 

### LINKED ISSUE #4128: [select/query-list] "Enter key key-up" hack for new item creation interacts badly with IMEs (Japanese, Chinese, etc.)
As explained at https://github.com/palantir/blueprint/blob/develop/packages/select/src/components/query-list/queryList.tsx#L478.-, key-up instead of key-down is used to detect the pressing of an enter key and trigger the creation of a new item in QueryList (and components like Multiselect that are built on top of it).

This works fine for English, but when using an IME to enter Japanese or Chinese text (https://en.wikipedia.org/wiki/Input_method), the enter key is pressed to confirm the character(s) to be input from a list of options.

The operating system intercepts the enter key key-down event and prevents it from propagating to the application, but the key-up event is still fired in this case.

As a result, users typing Japanese or Chinese text inadvertently fire off a "create new item" action whenever they confirm IME input. This makes the component near-unusable.

Is there another workaround that could be applied here? For what it's worth, I found that simply moving the enter key handler from handleKeyUp to handleKeyDown achieved the desired behavior for Japanese and Chinese and didn't seem to have any adverse effects on the Multiselect component, but considering the comment "if we were to process enter on keydown, then Button would click itself on keyup and the popvoer would re-open out of our control" (which I don't fully understand, since I don't know what Button is being referenced here), I have to assume that there is a bad consequence here that I'm not sure how to reproduce.

adidahiya:
#### Fixes #4128

thanks to @AaronJRubin for the proposal in #4134... this is a slightly different implementation which avoids using `this.state` since this boolean flag shouldn't cause a re-render of the component

#### Checklist

- [x] Includes tests
- N/A ~Update documentation~

### DIFF
--- a/packages/select/src/components/query-list/queryList.tsx
+++ b/packages/select/src/components/query-list/queryList.tsx
@@ -166,6 +166,25 @@ export class QueryList<T> extends AbstractComponent2<IQueryListProps<T>, IQueryL
      */
     private expectedNextActiveItem: T | ICreateNewItem | null = null;
 
+    /**
+     * Flag which is set to true while in between a "keydown" event and its
+     * corresponding "keyup" event.
+     *
+     * N.B. This is distinct from a "keypress" event, which we are in
+     * uninterested in since we care about all keys, not just those which
+     * produce character outputs.
+     *
+     * When entering text via an IME (https://en.wikipedia.org/wiki/Input_method),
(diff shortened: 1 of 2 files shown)


AaronJRubin:
I confirm that this resolves the IME issue!

PR #4143, opened 2020-05-21
`react-transition-group` is no longer externalized

thasner:
https://github.com/palantir/blueprint/commit/a866049f956eb416bc65bec96537985267e61b58

### DIFF
--- a/packages/docs-app/src/getting-started.md
+++ b/packages/docs-app/src/getting-started.md
@@ -140,7 +140,7 @@ Blueprint supports the venerable [unpkg CDN](https://unpkg.com). Each package pr
 library on the `Blueprint` global variable: `Blueprint.Core`, `Blueprint.Datetime`, etc.
 
 These bundles _do not include_ external dependencies; your application will need to ensure that
-`normalize.css`, `classnames`, `dom4`, `react`, `react-dom`, `react-transition-group`, `popper.js`, and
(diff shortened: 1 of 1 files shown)


palantirtech:
Thanks for your interest in [palantir/blueprint](https://github.com/palantir/blueprint), @thasner! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

adidahiya:
making something a "regular" dependency is not directly related to externalization. you can see this module is still being externalized in webpack-build-scripts: https://github.com/palantir/blueprint/blob/4b3e69c5600ec5251ae33637ae90302dd22ee9de/packages/webpack-build-scripts/externals.js#L32

thasner:
@adidahiya if BP is externalizing these libs, why would BP remove the peer dependency?

adidahiya:
By "remove the peer dependency", I assume you mean "change it to a regular dependency". It was done in response to this comment: https://github.com/reactjs/react-transition-group/issues/32. Multiple versions of react-transition-group can be installed, but multiple versions of React should not. I guess that means we don't need to externalize react-transition-group really. But the UMD modules are not an important part of this library; they were just added for convenience for environments like Code Sandbox. We don't use them in any production software.

PR #4159, opened 2020-05-29
Allow users to control which window object is used

### LINKED ISSUE #3972: FR: components should be able to operate in child window documents
currently, some components like (e.g. [NumericInput](https://github.com/palantir/blueprint/blob/develop/packages/core/src/components/forms/numericInput.tsx#L388)) reference `window.document` implicitly, which can may not be the document of the window where the component is being rendered in the case of React portal techniques like [this one](https://medium.com/hackernoon/using-a-react-16-portal-to-do-something-cool-2a2d627b0202). we should look into supporting this use case.

possible solution: define a blueprint React context which includes an optional field which sets `this.document` on all components

TylerU:
#### Fixes https://github.com/palantir/blueprint/issues/3972

Let parent components put a `windowOverride` on the context which will be used when we access `document`.  This isn't comprehensive. I fixed places that were breaking or likely to break in my use case and places where it was easy to extend the fix. 

#### Checklist

- [ ] Includes tests
- [ ] Update documentation


palantirtech:
Thanks for your interest in [palantir/blueprint](https://github.com/palantir/blueprint), @TylerU! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

packages/table/src/headers/columnHeaderCell.tsx
```diff
@@ -33,6 +33,10 @@ import { columnInteractionBarContextTypes, IColumnInteractionBarContextTypes } f
 import { LoadableContent } from "../common/loadableContent";
 import { CLASSNAME_EXCLUDED_FROM_TEXT_MEASUREMENT } from "../common/utils";
 import { HeaderCell, IHeaderCellProps } from "./headerCell";
+import {
+    IWindowOverrideContext,
+    WINDOW_OVERRIDE_REACT_CONTEXT_TYPES,
+} from "@blueprintjs/core/lib/esm/common/windowOverrideContext";
```
TylerU: fix this

packages/core/src/common/abstractPureComponent2.ts
```diff
@@ -16,12 +16,15 @@
...
 /**
  * An abstract component that Blueprint components can extend
  * in order to add some common functionality like runtime props validation.
  */
 export abstract class AbstractPureComponent2<P, S = {}, SS = {}> extends React.PureComponent<P, S, SS> {
+    public static contextTypes = WINDOW_OVERRIDE_REACT_CONTEXT_TYPES;
+
     // unsafe lifecycle method
     public componentWillUpdate: never;
     public componentWillReceiveProps: never;
```
TylerU: I think I had a reason for not declaring context's type. I'm not sure of the best way to handle the way inheritance works here. Do we want children to have the same context type? Should we even do the contextTypes thing?

adidahiya:
pulled your branch, looking into this now

adidahiya:
I considered using the newer `globalThis` API instead of `window` to accomplish this, but ultimately decided against it. Fulfilling `window` is more specific than `globalThis`, and we are not particularly concerned with working in multiple environments (Node, web worker, regular window) here.

I pushed a commit which adds support for this new window override in more of the components, but some parts of Blueprint's public API will still break w/r/t #3972. These may be addressed in separate future PRs:

- HotkeysDialog
- HotkeysTarget
- ContextMenu.show
- various table components

adidahiya:
[docs PR preview](https://32952-71939872-gh.circle-artifacts.com/0/packages/docs-app/dist/index.html)

adidahiya:
This looks ok overall but now I'm trying to write a good unit test and it's a little tricky to create a fake window. If you are able to write a test, feel free to push it.

PR #4160, opened 2020-05-29
[core] fix(NumericInput): enforce numeric chars only with IME

AaronJRubin:
Close loophole where IMEs could be used to inject invalid characters, and handle Japanese full-width number input.

#### Checklist

- [ ] Includes tests
- [ ] Update documentation

Whether I should be using preventDefault or any other HTML event API things in my handler (doesn't seem necessary, but maybe there are edge cases?).

#### Screenshot

Before:

buggy-numeric-input

After:

fixed-numeric-input

### DIFF
--- a/packages/core/src/components/forms/numericInput.tsx
+++ b/packages/core/src/components/forms/numericInput.tsx
@@ -41,7 +41,6 @@ import { InputGroup } from "./inputGroup";
 import {
     clampValue,
     getValueOrEmptyValue,
-    isFloatingPointNumericCharacter,
     isValidNumericKeyboardEvent,
     isValueNumeric,
     sanitizeNumericInput,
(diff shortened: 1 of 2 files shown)


adidahiya:
[docs preview](https://32785-71939872-gh.circle-artifacts.com/0/packages/docs-app/dist/index.html#core/components/numeric-input)

this change mostly looks good, thanks for the PR @AaronJRubin 

I haven't checked out this branch, and this is the first time I'm seeing the `onCompositionEnd` event... I wonder if there's an easy way to provide user feedback that the non-numeric characters they entered via IME are invalid (perhaps with a "danger" intent on the input)? This is what currently happens in that case (at the end, I hit ENTER and the characters go away, potentially surprising the user):



... as opposed to non-IME text input, where `allowNumericCharactersOnly` prevents those characters from ever showing up in the input _at all_ (they are rejected upon keystroke).

If it's an easy change to add to this PR then let's do it.

---

semi-relatedly, I found this open React issue about IME but it seems like we're working around it: https://github.com/facebook/react/issues/3926

AaronJRubin:
Turned out to be quite the easy change indeed!

intent-danger

PR #4176, opened 2020-06-12
Right element for EditableText

### LINKED ISSUE #2185: [EditableText] Add support for rightElements in EditableText
Title says it all.

ilhanadiyaman:
#### Checklist

- [x] Includes tests
- [x] Update documentation

For some use cases, it's helpful to have right element in the EditableText. This PR enables to render right element in EditableText by using `rightElement` prop similar to InputGroup.

- `updateInputDimensions` function to update the padding
- `maybeRenderRightElement` function and the way how it handles the visibility of the element and focus (there might be better way to handle this)

#### Screenshot

Screen Shot 2020-06-12 at 03 11 07

### DIFF
--- a/packages/core/src/components/editable-text/editableText.tsx
+++ b/packages/core/src/components/editable-text/editableText.tsx
@@ -87,6 +87,12 @@ export interface IEditableTextProps extends IIntentProps, IProps {
     placeholder?: string;
 
     /**
+     * Element to render on the right side of the input.
+     * For best results, use a minimal button, tag, or small spinner.
+     */
+    rightElement?: JSX.Element;
+
+    /**
      * Whether the entire text field should be selected on focus.
      * If `false`, the cursor is placed at the end of the text.
      * This prop is ignored on inputs with type other then text, search, url, tel and password. See https://html.spec.whatwg.org/multipage/input.html#do-not-apply for details.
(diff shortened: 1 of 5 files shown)


palantirtech:
Thanks for your interest in [palantir/blueprint](https://github.com/palantir/blueprint), @ilhanadiyaman! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

adidahiya:
Hi @ilhanadiyaman thanks for the PR. Since https://github.com/palantir/blueprint/issues/2185 is quite an old request, I'd like to re-evalaute it before proceeding further. Can you explain your context and use case for this new feature?

ilhanadiyaman:
Hi @adidahiya, I put the simplest case in the screenshot which helps you to clear the filled text with one click. I assume, most of the time, right element will be a simple button which triggers a popover or menu. But we have couple of other use cases in our dashboard:

- Displaying word/character entered in the editable text. Kind of like Twitter input where it counts remained characters.
- Helping user to enter emojis while they are editing the content.
- Helping user to enter more specific input where it's not possible to memorize such as simple code blocks(we have a campaign editor where user can enter templating variables..)

Even though most of these options are replaceable by combining other components, it's more handy to show available options only when user edits the text itself. Having separate button which connects editable text events would be harder to handle.

packages/core/src/components/editable-text/editableText.tsx
```diff
@@ -397,6 +424,33 @@ export class EditableText extends AbstractPureComponent2<IEditableTextProps, IEd
...
+                onClick={this.handleFocus}
+                ref={this.refHandlers.rightElement}
+            >
+                {rightElement}
+            </span>
+        );
+    }
+
+    private handleMouseEnter = () => {
+        this.setState({ isRightElementVisible: true });
```
adidahiya, line 447: what is the UX logic here? I was sort of expecting the right element to become visible once mousing into the whole editable text, not just the right element. also, why are we focusing the input at this time?

adidahiya (changes requested):
sorry for the delayed review, can you please merge `develop`? there will be some new strict null checks compile errors you'll have to address

PR #4198, opened 2020-06-29
fix TextArea inputRef update event

### LINKED ISSUE #4072: TextArea does not call updated inputRef prop
#### Environment

- __Package version(s)__: @blueprintjs/core 3.25.0

- __Demo__: https://codesandbox.io/s/funny-cloud-wtpiy

#### Steps to reproduce

1. render a `<TextArea />` with  `inputRef` set to anything
2. update the `inputRef` prop with something different

#### Actual behavior

The updated `inputRef` prop will not be called. This is inconsistent with how other components like `InputGroup` and built-in components like `<input>` behave on `inputRef` changes.

#### Expected behavior

The updated `inputRef` prop will be called with the current DOM element.

#### Possible solution

Maybe implement `componentDidUpdate`:

``` javascript
componentDidUpdate(prevProps) {
    if (this.props.inputRef != null && prevProps.inputRef !== this.props.inputRef) {
       this.props.inputRef(this.internalTextAreaRef);
    }
}
```

Arbalester:
#### Checklist

- [X] Includes tests
- [ ] Update documentation

TextArea should call updated inputRef prop with the current DOM element.

### DIFF
--- a/packages/core/src/components/forms/textArea.tsx
+++ b/packages/core/src/components/forms/textArea.tsx
@@ -66,6 +66,15 @@ export class TextArea extends AbstractPureComponent2<ITextAreaProps, ITextAreaSt
             });
         }
     }
+    public componentDidUpdate(prevProps: ITextAreaProps) {
+        if (
+            prevProps.inputRef &&
+            this.props.inputRef &&
+            prevProps.inputRef.toString() !== this.props.inputRef.toString()
+        ) {
+            this.props.inputRef(this.internalTextAreaRef);
+        }
+    }
     public render() {
         const { className, fill, inputRef, intent, large, small, growVertically, ...htmlProps } = this.props;
 
--- a/packages/core/test/forms/textAreaTests.tsx
+++ b/packages/core/test/forms/textAreaTests.tsx
@@ -65,4 +65,16 @@ describe("<TextArea>", () => {
         const scrollHeightInPixels = `${(textarea.getDOMNode() as HTMLElement).scrollHeight}px`;
         assert.equal((textarea.getDOMNode() as HTMLElement).style.height, scrollHeightInPixels);
     });
+    it("updates on ref change", () => {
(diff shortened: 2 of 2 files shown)


palantirtech:

adidahiya (approved):
thanks @Arbalester

trixn86 (commented):
@Arbalester 

Why are the `inputRef`s compared as their string representations (`prevProps.inputRef.toString() !== this.props.inputRef.toString()`) and not simply by reference equality? Is this intended to prevent it from updating when the user did not correctly memorize the callback? Feels like a hack to be honest. And it is also not be the behaviour one would expect. Reacts built-in elements handle that differently (https://codesandbox.io/s/thirsty-panini-oc3uu?file=/src/App.js).In my opinion the developer using the ref is responsible for correctly memorizing his/her ref callbacks. If I wanted to `console.log` the ref on every render cycle I couldn't do it just because the code of my function stays the same even when it's a new function.

Any even more problematic: If you pass a ref created by the `useRef()` hook, it will **always** be equal as it's an object whose string representation is simply `[object Object]`.

Arbalester:
> @Arbalester
> 
> Why are the `inputRef`s compared as their string representations (`prevProps.inputRef.toString() !== this.props.inputRef.toString()`) and not simply by reference equality? Is this intended to prevent it from updating when the user did not correctly memorize the callback? Feels like a hack to be honest. And it is also not be the behaviour one would expect. Reacts built-in elements handle that differently (https://codesandbox.io/s/thirsty-panini-oc3uu?file=/src/App.js).In my opinion the developer using the ref is responsible for correctly memorizing his/her ref callbacks. If I wanted to `console.log` the ref on every render cycle I couldn't do it just because the code of my function stays the same even when it's a new function.
> 
> Any even more problematic: If you pass a ref created by the `useRef()` hook, it will **always** be equal as it's an object whose string representation is simply `[object Object]`.

Yeah, my bad. Made a fix for this.

PR #4131, opened 2020-05-19
Fix unexpected uncontrolled behaviour on <NumericInput>

benjaminbours:
#### Checklist

- [x] Includes tests
- [x] Update documentation (I don't think I need to in this case)

#### Changes proposed in this pull request:
A quick fix to have the expected controlled component behaviour with NumericInput

#### Reviewers should focus on:
The problem itself. The implementation is probably not what you want because it can be a breaking change for applications using this component for a while without noticing this unexpected behaviour.

#### Screenshot
I do not have a screenshot for this, sorry.

PR #4201, opened 2020-06-30
Improve NumericInput and AbstractButton internals

### LINKED ISSUE #4200: NumericInput events run incorrectly when using min or max props
#### Environment

- __Package version(s)__: 3.28 / 3.29
- __Operating System__: Linux
- __Browser name and version__:  Firefox 77 / Chromium 81

#### Code Sandbox

Link to a minimal repro (fork [this code sandbox](https://codesandbox.io/s/blueprint-sandbox-et9xy)): https://codesandbox.io/s/blueprint-numericinput-vxkuu

#### Steps to reproduce

1. Hold arrow button until limit is reached, then do the same with opposite button, repeating a couple of times will
   make all events to not work properly (button clicks set more than 1 step at once, and keyboard up and down arrow don't stop until limit is reached, even if tapped once)

#### Actual behavior

Button & keyboard events don't work properly

#### Expected behavior

Behaviour shouldn't not change when usin min/max

Edit: Seems I found the cause of the issue, this bug doesn't happen if you release the mouse before reaching the limit, or upon reaching the limit, outside of the disabled button. I've checked the source and it seems stopContinuousChange is not firing when releasing the mouse on the disabled button

ejose19:
#### Checklist

- [X] Includes tests
- [X] Update documentation (No changes to docs)

- ~~Add a check to each iteration of handleContinuousChange, since mouseup event won't fire if the mouse is released on the disabled element (after it reached min/max limit) current behaviour keeps running it indefinitely when using min/max props and releasing the mouseup after described event.~~ Fixed using AnchorButton

- ~~Allow accessing class instance during onValueChange callback, this is necessary as if doing any conditional logic to allow a value, one needs to update the internal value to keep it in sync with given value prop (see [my comment on this issue for more details](https://github.com/palantir/blueprint/pull/4131#issuecomment-651435032))~~ Removed, another PR improves controlled mode

- Add ~~componentDidUpdate~~ getDerivedStateFromProps to AbstractButton, related to the 1st issue, its handleKeyUp is never fired if the button is disabled, so internal state is still "active" while the button is disabled (which is not correct) displaying inconsistent UI to the user when using along other components (For example NumericInput)

Mentioned changes

#### Screenshot

N/A

### DIFF
--- a/packages/core/src/components/button/abstractButton.tsx
+++ b/packages/core/src/components/button/abstractButton.tsx
@@ -134,6 +134,16 @@ export abstract class AbstractButton<H extends React.HTMLAttributes<HTMLElement>
         };
     }
 
+    // A disabled element cannot be active, there're are cases where
+    // this situation can happen (NumericInput buttons), hence the need
+    // to check if disabled prop has been passed and change isActive
+    // accordingly
+    public componentDidUpdate() {
+        if (this.state.isActive && this.props.disabled) {
+            this.setState({ isActive: false });
+        }
+    }
+
     // we're casting as `any` to get around a somewhat opaque safeInvoke error
     // that "Type argument candidate 'KeyboardEvent<T>' is not a valid type
     // argument because it is not a supertype of candidate
@@ -149,6 +159,8 @@ export abstract class AbstractButton<H extends React.HTMLAttributes<HTMLElement>
         this.props.onKeyDown?.(e);
     };
 
+    // Keyup event won't fire on a disabled element, so componentDidUpdate
+    // will take care of those cases and set isActive to false
     protected handleKeyUp = (e: React.KeyboardEvent<any>) => {
         if (Keys.isKeyboardClick(e.which)) {
             this.setState({ isActive: false });
--- a/packages/core/src/components/button/abstractButton.tsx
+++ b/packages/core/src/components/button/abstractButton.tsx
@@ -134,7 +134,7 @@ export abstract class AbstractButton<H extends React.HTMLAttributes<HTMLElement>
         };
     }
 
-    // A disabled element cannot be active, there're are cases where
+    // A disabled element cannot be active, but there are cases where
     // this situation can happen (NumericInput buttons), hence the need
     // to check if disabled prop has been passed and change isActive
     // accordingly
--- a/packages/core/src/components/button/abstractButton.tsx
+++ b/packages/core/src/components/button/abstractButton.tsx
@@ -163,8 +163,6 @@ export abstract class AbstractButton<H extends React.HTMLAttributes<HTMLElement>
         this.props.onKeyDown?.(e);
     };
 
-    // Keyup event won't fire on a disabled element, so componentDidUpdate
-    // will take care of those cases and set isActive to false
     protected handleKeyUp = (e: React.KeyboardEvent<any>) => {
         if (Keys.isKeyboardClick(e.which)) {
             this.setState({ isActive: false });
--- a/packages/core/src/components/button/abstractButton.tsx
+++ b/packages/core/src/components/button/abstractButton.tsx
@@ -105,7 +105,7 @@ export abstract class AbstractButton<H extends React.HTMLAttributes<HTMLElement>
     // this situation can happen (NumericInput buttons), hence the need
     // to check if disabled prop has been passed and change isActive
     // accordingly
-    protected static getDerivedStateFromProps(props: IButtonProps, state: IButtonState) {
+    public static getDerivedStateFromProps(props: IButtonProps, state: IButtonState) {
         if (state.isActive && props.disabled) {
             return {
                 isActive: false,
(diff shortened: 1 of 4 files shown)


palantirtech:
Thanks for your interest in [palantir/blueprint](https://github.com/palantir/blueprint), @ejose19! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

packages/core/src/components/button/abstractButton.tsx
```diff
@@ -134,6 +134,16 @@ export abstract class AbstractButton<H extends React.HTMLAttributes<HTMLElement>
         };
     }
 
+    // A disabled element cannot be active, there're are cases where
```
```suggestion
    // A disabled element cannot be active, but there are cases where
```

packages/core/src/components/button/abstractButton.tsx
```diff
@@ -134,6 +134,16 @@ export abstract class AbstractButton<H extends React.HTMLAttributes<HTMLElement>
...
+    // this situation can happen (NumericInput buttons), hence the need
+    // to check if disabled prop has been passed and change isActive
+    // accordingly
+    public componentDidUpdate() {
```
adidahiya: I think this should be implemented inside `static getDerivedStateFromProps` instead:

```ts
    protected static getDerivedStateFromProps(props: IButtonProps, state: IButtonState) {
        if (state.isActive && props.disabled) {
            return {
                isActive: false,
            };
        }

        return null;
    }
```

adidahiya (changes requested):
> Allow accessing class instance during onValueChange callback, this is necessary as if doing any conditional logic to allow a value, one needs to update the internal value to keep it in sync with given value prop

I'm not yet convinced that this is necessary. What kind of conditional logic are you doing? Can we support that use case some other way instead of providing access to the component's internals? I don't want to increase the size of the public API this way.

---

Also, I think this whole issue might be easier solved by simply replacing the `Button` components rendered by `NumericInput` with `AnchorButton` components... they are able to receive events when `disabled=true`

ejose19:
@adidahiya Regarding the instance access, if there's any other way then great, but I would like to know how you would do that. Here's the issue: Controlled mode, my value is set to 1, and NumericInput sets its internal state to '1', now I only want integers in a range, so I add a conditional logic to onValueChange callback to apply the new value to my state,  this theorically works good if the user is either using the buttons or tapping only numbers, but there are 2 cases: (pressing the letter 'e' is allowed, and pasting content into the field) first action is allowed and internal value is stored as '1e' but my state rejected it and it's 1, now they're not in sync and buttons won't display limits properly. On paste if text pasted is invalid (letters only), then internal state is VALUE_EMPTY and my state is still 1. So either we need to add a callback function like ("isValid") to determine if internal state should change, or as I'm  doing in this patch, let me change the internal state manually after I do my checks.

packages/core/src/components/button/abstractButton.tsx
ejose19: If this is better than componentDidUpdate then no issues, I will change it.

ejose19 (commented):
Applied both recommendations and I confirm they work (Using AnchorButton on numericInput and getDerivedStateFromProps on abstractButton)

adidahiya:
@ejose19 regarding controlled mode... check out the PR I just opened #4211... that should get you sorted. For this PR please just stick to fixing #4200. I'll merge my PR today and then you can test it.

packages/core/src/components/forms/numericInput.tsx
```diff
@@ -151,7 +151,12 @@ export interface INumericInputProps extends IIntentProps, IProps {
     onButtonClick?(valueAsNumber: number, valueAsString: string): void;
 
     /** The callback invoked when the value changes due to typing, arrow keys, or button clicks. */
-    onValueChange?(valueAsNumber: number, valueAsString: string, inputElement: HTMLInputElement | null): void;
+    onValueChange?(
+        valueAsNumber: number,
+        valueAsString: string,
+        inputElement: HTMLInputElement | null,
+        instance: NumericInput,
```
adidahiya: please revert this change unrelated to #4200

adidahiya (changes requested):

packages/core/src/components/button/abstractButton.tsx
adidahiya: It's better because it's a more natural place to manage state updates, it gets called before `render`. Check out the React lifecycle documentation to see what kinds of behavior is recommended in the various methods

packages/core/src/components/forms/numericInput.tsx
ejose19: Tried your fork and apparently it works correctly, so I'm removing instance from this PR and will try again on official version once this and the other PR is merged

adidahiya (approved):

adidahiya:
thanks for testing my branch. I'll merge yours first so you don't have to deal with conflicts

ejose19:
@adidahiya Since there's a side effect caused by using AnchorButton, should we use the original idea instead?

> Add a check to each iteration of handleContinuousChange, since mouseup event won't fire if the mouse is released on the disabled element (after it reached min/max limit) current behaviour keeps running it indefinitely when using min/max props and releasing the mouseup after described event.

adidahiya:
@ejose19 ok, yeah, I would be open to a change that switches to the original idea using Button instead

PR #4207, opened 2020-07-03
ensure null values do not cause numeric input to throw

jonnywildey:
#### Checklist

- [x] Includes tests
- [ ] Update documentation

Currently if null is passed in as a value to the Numeric input, an `null is not an object (evaluating 'value.toString')` error occurs. 
This is due to the `getValueOrEmptyValue` utility function treating null as a string.

Although null isn't an acceptable value for the numeric input (according to the types) it should probably not throw an error when a null is passed in.

I think the current behaviour is unintentional, probably due to the ambiguity of whether a null value would default to an empty string in the `string = ""` argument. Unfortunately is does not. See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Default_parameters

I have purposely not updated any types, I'm not too sure if that was the correct approach?

#### Screenshot

Current error: 
Screenshot 2020-07-03 at 12 04 41

### DIFF
--- a/packages/core/src/components/forms/numericInputUtils.ts
+++ b/packages/core/src/components/forms/numericInputUtils.ts
@@ -25,7 +25,7 @@ export function clampValue(value: number, min?: number, max?: number) {
 }
 
 export function getValueOrEmptyValue(value: number | string = "") {
-    return value.toString();
+    return value === null ? "" : value.toString();
 }
 
 /** Returns `true` if the string represents a valid numeric value, like "1e6". */
--- a/packages/core/test/controls/numericInputTests.tsx
+++ b/packages/core/test/controls/numericInputTests.tsx
@@ -183,6 +183,12 @@ describe("<NumericInput>", () => {
             expect(value).to.equal("10");
         });
 
+        it("accepts a null value", () => {
+            const component = mount(<NumericInput value={null} />);
+            const value = component.state().value;
+            expect(value).to.equal("");
+        });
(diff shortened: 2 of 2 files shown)


palantirtech:
Thanks for your interest in [palantir/blueprint](https://github.com/palantir/blueprint), @jonnywildey! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

jonnywildey:
> Thanks for your interest in [palantir/blueprint](https://github.com/palantir/blueprint), @jonnywildey! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

Hey, it doesn't seem like this is updating. Have signed the CLA

Screenshot 2020-07-06 at 10 52 35

adidahiya (commented):
I don't think we should add this kind of runtime check. As a TypeScript-first library, we rely on correct typings to prevent this class of bugs. If you enable `strictNullChecks` or `strict: true` in your TS project which uses Blueprint, you wouldn't be able to send `null` to this prop (it would cause a compile error).

If we add guards for nulls in this part of the API, then to be consistent we would need to add them in many other places. For the reason stated above, and to keep the component implementation concise, I'd like to avoid doing that.

PR #4208, opened 2020-07-03
fix button click from keyboard

### LINKED ISSUE #4170: Button onClick is triggered twice from the keyboard
#### Environment

- __Package version(s)__: 
- __Operating System__: Ubuntu 20.04 LTS
- __Browser name and version__: 77.0.1+build1-0ubuntu0.20.04.1 and 77.0.1 (64-bit) (windows)

#### Code Sandbox

[Link to a minimal repro](https://codesandbox.io/s/blueprint-sandbox-f968c?file=/src/index.tsx)

#### Steps to reproduce

1. Tab onto the button and hit the spacebar.

#### Actual behavior

The word 'hello' prints to the console twice.

#### Expected behavior

The word 'hello' should only print once. This is the correct behavior, which happens on chromium based browsers, and also when mouse clicking on Firefox.

Arbalester:
#### Checklist

- [ ] Includes tests
- [ ] Update documentation

Button's onClick event should fire once from the keyboard.


palantirtech:

adidahiya (changes requested):
`preventDefault()` (and more generally, code which blocks default browser behavior) can lead to mysterious bugs when events are not propagated correctly. For example, this might break hotkeys tied to a container element when the button is focused.

Is there another way we can address the underlying issue?

Arbalester:
> `preventDefault()` (and more generally, code which blocks default browser behavior) can lead to mysterious bugs when events are not propagated correctly. For example, this might break hotkeys tied to a container element when the button is focused.
> 
> Is there another way we can address the underlying issue?

In this fix `preventDefault()` is only for events when keys  `enter` or `space` pressed (for method `handleKeyUp` as already done in `handleKeyDown`). Could you create a reproduction of the issue that you described?

ejose19:
@adidahiya I believe this is the way to go, since Firefox does things differently (`preventDefault` in `handleKeyDown` doesn't work as in other browsers). I was about to submit a PR with: https://github.com/palantir/blueprint/blob/d0c6dfb66d05d44d6fc796ce1c578fc4a1744e1d/packages/core/src/components/button/abstractButton.tsx#L175-L177

but if any other browser suffers from the same issue, it won't be enough. By calling `preventDefault` at least we're sure only one `onClick` will be called through `this.buttonRef?.click();`, and @Arbalester point stands, this is only called for `enter` or `space` keys, as done in `handleKeyDown`, so there shouldn't be any surprises.

adidahiya:
This seems to have been fixed, either through some Blueprint changes or in Firefox itself via https://bugzilla.mozilla.org/show_bug.cgi?id=1487102

PR #4210, opened 2020-07-06
Update toast.md

ghost:

[`Button`s] to [`Buttons`], the tiniest things bother me :+1: 

Things like this for users sanity ;)

### DIFF
--- a/packages/core/src/components/toast/toast.md
+++ b/packages/core/src/components/toast/toast.md
@@ -17,7 +17,7 @@ Similarly, focusing the toast (for example, by hitting the `tab` key) halts the 
 
 You can add one additional action button to a toast. You might use this to provide an undo button, for example.
 
-You can also apply the same visual intent styles to `Toast`s that you can to [`Button`s](#core/components/button.css).
+You can also apply the same visual intent styles to `Toast`s that you can to [`Buttons-`](#core/components/button.css).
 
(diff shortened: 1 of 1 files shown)


palantirtech:
Thanks for your interest in [palantir/blueprint](https://github.com/palantir/blueprint), @milkysheets! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

packages/core/src/components/toast/toast.md
```diff
@@ -17,7 +17,7 @@ Similarly, focusing the toast (for example, by hitting the `tab` key) halts the
 
 You can add one additional action button to a toast. You might use this to provide an undo button, for example.
 
-You can also apply the same visual intent styles to `Toast`s that you can to [`Button`s](#core/components/button.css).
+You can also apply the same visual intent styles to `Toast`s that you can to [`Buttons-`](#core/components/button.css).
```
adidahiya, line 20: the original code was correct, albeit a little awkward. we use this kind of syntax in many places in the docs, where the name of the component is emphasized with code backticks. `Buttons` doesn't make sense in this style because that is not a component (there are some pluralized components, though, like `Tabs`).

also there is a stray `-` here

PR #4211, opened 2020-07-07
[core] fix(NumericInput): more robust controlled mode

### LINKED ISSUE #2186: NumericInput with `value` set does not behave as Controlled component
### Bug report

- __Browser and OS versions__: Ubuntu 16.04, Chrome 64.0.3282.186

#### Steps to reproduce

1. Create a basic NumericInput component. It can be as simple as this:
```
<NumericInput
  value={1.0}
  onValueChange={v => console.log(v)}
/>
```
2. Type in a different value.
3. Notice that the value does change.

#### Actual behavior

The component freely changes value despite the `value` property being set to a constant.

#### Expected behavior

The component's value shall never change unless the `value` property is changed.

### LINKED ISSUE #3553: NumericInput controlled mode not actually controlled..
Hey there,

I know this has already been asked for before, but I'd like to clamp a NumericInput to integers. I don't want commas, or periods. I want the input to only respond to integer changes. Seems simple enough.

Unfortunately I think that some of the internal logic of the numeric input field make it difficult to actually have it be fully controlled. 
I'm using the controlled mode, but even if I coerce the value to not include those chars, the input still acts in a slightly uncontrolled manner. For example, I would expect this NumericInput to be totally stuck on "hello":

```ts
<NumericInput onValueChange={() => {}} value={"hello"} />
```

As you can see in this codesandbox, that is not the case. You can change it all you want. 
https://codesandbox.io/s/4qm5p6my14

Am I missing something here? The current "controlled" behavior seems very uncontrolled. 

I need a truly controlled component if I want to implement my custom integer-only input.

Thanks!

adidahiya:
#### Checklist

- [x] Includes tests
  - [x] need some more tests for the new controlled functionality and `defaultValue` prop
- [x] Update documentation

- Refactor NumericInput component's controlled mode to no longer change `state.value`
- New `defaultValue` prop behaves as you would expect for a controlled component

if there are enough unit tests

### DIFF
--- a/packages/core/src/common/errors.ts
+++ b/packages/core/src/common/errors.ts
@@ -51,6 +51,8 @@ export const NUMERIC_INPUT_MAJOR_STEP_SIZE_NON_POSITIVE =
 export const NUMERIC_INPUT_STEP_SIZE_NON_POSITIVE =
     ns + ` <NumericInput> requires stepSize to be strictly greater than zero.`;
 export const NUMERIC_INPUT_STEP_SIZE_NULL = ns + ` <NumericInput> requires stepSize to be defined.`;
+export const NUMERIC_INPUT_CONTROLLED_VALUE_INVALID =
+    ns + ` <NumericInput> controlled value prop does not adhere to stepSize, min, and/or max constraints.`;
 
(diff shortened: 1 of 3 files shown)


adidahiya:
fyi @benjaminbours @tnrich

PR #4214, opened 2020-07-07
[core] fix(NumericInput): disabled prop prevents button interactions

### LINKED ISSUE #4213: NumericInput fill prop doesn't work
discovered while testing https://github.com/palantir/blueprint/pull/4211

#### Environment

- __Package version(s)__: develop & 3.29.0
- __Operating System__: macOS
- __Browser name and version__: Firefox

#### Steps to reproduce

1. Enable fill option in NumericInput docs

#### Actual behavior

image

#### Expected behavior

control group should be full width

ejose19:
#### Fixes #4213

Also fixes a bug where disabled element allowed interacting with the buttonHandlers (due to now using AnchorButton)

#### Checklist

- [X] Includes tests
- [X] Update documentation

- Add missing class to sass to handle control-group fill
- Don't execute handleButtonClick if component is disabled

### DIFF
--- a/packages/core/src/components/forms/_control-group.scss
+++ b/packages/core/src/components/forms/_control-group.scss
@@ -229,6 +229,9 @@ Styleguide control-group
 
   Styleguide control-group-fill
   */
+  &.#{$ns}-fill {
+    width: 100%;
+  }
 
   > .#{$ns}-fill {
     flex: 1 1 auto;
--- a/packages/core/src/components/forms/_control-group.scss
+++ b/packages/core/src/components/forms/_control-group.scss
@@ -229,9 +229,6 @@ Styleguide control-group
 
   Styleguide control-group-fill
   */
-  &.#{$ns}-fill {
-    width: 100%;
-  }
 
   > .#{$ns}-fill {
     flex: 1 1 auto;
--- a/packages/core/src/components/forms/numericInput.tsx
+++ b/packages/core/src/components/forms/numericInput.tsx
@@ -405,15 +405,15 @@ export class NumericInput extends AbstractPureComponent2<HTMLInputProps & INumer
         return {
             // keydown is fired repeatedly when held so it's implicitly continuous
             onKeyDown: evt => {
-                if (Keys.isKeyboardClick(evt.keyCode)) {
-                    if (this.props.disabled) return;
+                if (!this.props.disabled && Keys.isKeyboardClick(evt.keyCode)) {
                     this.handleButtonClick(evt, direction);
                 }
             },
             onMouseDown: evt => {
-                if (this.props.disabled) return;
-                this.handleButtonClick(evt, direction);
-                this.startContinuousChange();
+                if (!this.props.disabled) {
+                    this.handleButtonClick(evt, direction);
+                    this.startContinuousChange();
+                }
             },
         };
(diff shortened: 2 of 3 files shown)


packages/core/src/components/forms/numericInput.tsx
```diff
@@ -406,10 +406,12 @@ export class NumericInput extends AbstractPureComponent2<HTMLInputProps & INumer
             // keydown is fired repeatedly when held so it's implicitly continuous
             onKeyDown: evt => {
                 if (Keys.isKeyboardClick(evt.keyCode)) {
+                    if (this.props.disabled) return;
```
adidahiya: I am surprised we don't have a lint rule for this... I swear we used to... please use `{ }` braces for all if statements.

adidahiya: also I think this check should go outside the `if (Keys.isKeyboardClick)` statement... or just be merged into the same conditional

packages/core/test/controls/numericInputTests.tsx
```diff
@@ -1015,6 +1016,34 @@ describe("<NumericInput>", () => {
...
+            const decrementButton = component.find(AnchorButton).last();
+            decrementButton.simulate("mousedown");
+            decrementButton.simulate("mousedown", { altKey: true });
+            decrementButton.simulate("keyDown", SPACE_KEYSTROKE);
+            decrementButton.simulate("keyDown", { ...SPACE_KEYSTROKE, altKey: true });
+
+            expect(handleButtonClickSpy.notCalled).to.be.true;
+        });
+
+        it("must set fill class on controlGroup element when fill is enabled", () => {
```
adidahiya: we don't need this test, it's tautological and doesn't add much value

packages/core/src/components/forms/_control-group.scss
```diff
@@ -229,6 +229,9 @@ Styleguide control-group
 
   Styleguide control-group-fill
   */
+  &.#{$ns}-fill {
```
adidahiya: while looking into this bug I found some strange code which I don't think should exist. we have special logic in the control group example to make the "fill" option work on the page... this logic is outside the Blueprint API so it's misleading for developers testing the component. please remove these lines in this PR: https://github.com/palantir/blueprint/blob/842b2bb43f43ad11b71cdec86fff410d1a14474c/packages/docs-app/src/examples/core-examples/controlGroupExample.tsx#L46-L49

adidahiya (changes requested):
thanks for the contribution. in the future I would appreciate separate PRs for unrelated issues/features.

ejose19:
@adidahiya Sure, adjusted files. Will take your notes in consideration for next PRs

adidahiya (approved):

ejose19:
@adidahiya Is there any estimate where all the NumericInput changes will be released on npm?

adidahiya:
@ejose19 this was released. you can see which releases a commit is included in here:

image

ejose19:
@adidahiya If it's not much trouble, can you tell me where do I see exactly that? I checked all "tabs" on current page but I can't find that.

ejose19:
@adidahiya Thanks, didn't know about that feature.

PR #4218, opened 2020-07-09
[core] fix(Overlay): handle document click/focus in Shadow DOM

### LINKED ISSUE #4220: Overlay component in Shadow DOM is closed even if I click inside it
#### Environment

- __Package version(s)__: @blueprintjs/core v3.29.0
- __Operating System__: Windows 10
- __Browser name and version__: Chrome 83.0.4103.106

#### Code Sandbox

Link to a minimal repro (fork [this code sandbox](https://codesandbox.io/s/blueprint-sandbox-et9xy)): https://codesandbox.io/s/blueprint-sandbox-n2131

I modified two places:

1. In `SelectExample.tsx` I added `popoverProps={{ usePortal: false }}` on `<FilmSelect>` to ensure the popover render in the Shadow DOM。
2. In `index.tsx`, I moved blueprintjs CSS into the Shadow DOM.

#### Steps to reproduce

1. Click the Select Component
2. Click the filter input in the popover

#### Actual behavior

Popover closed.

#### Expected behavior

It should stay open.

#### Possible solution

https://github.com/palantir/blueprint/blob/5ad90950bab5e9fbab1a90c02c310968ac8c0e8b/packages/core/src/components/overlay/overlay.tsx#L426-L443

Line 428 used `e.target`, but if an event is triggered in a Shadow DOM, the `event.target` is the Shadow host, not the element actually trigger the event, so `elem.contains(eventTarget)` (in line 436) is alaways `false`.

We need check `event.composed` ([MDN link](https://developer.mozilla.org/en-US/docs/Web/API/Event/composed)) if this event is triggered in a Shadow DOM, and use `event.composedPath()` ([MDN link](https://developer.mozilla.org/en-US/docs/Web/API/Event/composedPath)) to get the event bubbles path, and the first item is the actually element that trigger the event in a Shadow DOM.

lmk123:
#### Checklist

- [ ] Includes tests
- [ ] Update documentation

I use blueprint in Shadow DOM, and the Overlay component is closed even if I click inside it. This PR fix this problem.

I searched the place where `event.target` was used in the entire code base, it seems that only Overlay component used `event.target` in the event registered in the `document`.

### DIFF
--- a/packages/core/src/components/overlay/overlay.tsx
+++ b/packages/core/src/components/overlay/overlay.tsx
@@ -443,11 +443,12 @@ export class Overlay extends AbstractPureComponent2<IOverlayProps, IOverlayState
     };
 
     private handleDocumentFocus = (e: FocusEvent) => {
+        const eventTarget = e.composed ? e.composedPath()[0] : e.target
         if (
             this.props.enforceFocus &&
             this.containerElement != null &&
-            e.target instanceof Node &&
-            !this.containerElement.contains(e.target as HTMLElement)
+            eventTarget instanceof Node &&
+            !this.containerElement.contains(eventTarget as HTMLElement)
         ) {
             // prevent default focus behavior (sometimes auto-scrolls the page)
             e.preventDefault();
--- a/packages/core/src/components/overlay/overlay.tsx
+++ b/packages/core/src/components/overlay/overlay.tsx
@@ -425,6 +425,7 @@ export class Overlay extends AbstractPureComponent2<IOverlayProps, IOverlayState
 
     private handleDocumentClick = (e: MouseEvent) => {
         const { canOutsideClickClose, isOpen, onClose } = this.props;
+        // get the actually target even if we are in an open mode Shadow DOM
         const eventTarget = (e.composed ? e.composedPath()[0] : e.target) as HTMLElement;
 
         const stackIndex = Overlay.openStack.indexOf(this);
@@ -443,6 +444,7 @@ export class Overlay extends AbstractPureComponent2<IOverlayProps, IOverlayState
     };
 
     private handleDocumentFocus = (e: FocusEvent) => {
+        // get the actually target even if we are in an open mode Shadow DOM
         const eventTarget = e.composed ? e.composedPath()[0] : e.target;
         if (
             this.props.enforceFocus &&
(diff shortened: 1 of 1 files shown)


palantirtech:
Thanks for your interest in [palantir/blueprint](https://github.com/palantir/blueprint), @lmk123! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

packages/core/src/components/overlay/overlay.tsx
```diff
@@ -425,7 +425,7 @@ export class Overlay extends AbstractPureComponent2<IOverlayProps, IOverlayState
 
     private handleDocumentClick = (e: MouseEvent) => {
         const { canOutsideClickClose, isOpen, onClose } = this.props;
-        const eventTarget = e.target as HTMLElement;
+        const eventTarget = (e.composed ? e.composedPath()[0] : e.target) as HTMLElement;
```
adidahiya, line 429: is `e.composedPath()` always guaranteed to be non-empty? can you add some code comments to explain why we are adding this special logic here and on L446?

adidahiya (changes requested):
any chance you could show me a code sandbox demonstrating the issue you're trying to fix? generally before submitting PRs to Blueprint I ask contributors to file an issue and fill out the issue template.

also, could you add a unit test for this behavior?

lmk123:
> any chance you could show me a code sandbox demonstrating the issue you're trying to fix? generally before submitting PRs to Blueprint I ask contributors to file an issue and fill out the issue template.

I posted an issue just now: https://github.com/palantir/blueprint/issues/4220

> also, could you add a unit test for this behavior?

When I tried to add unit tests yesterday, I encountered some problems when installing `windows-build-tools`. I will try again later and submit unit tests.

packages/core/src/components/overlay/overlay.tsx
lmk123: > is `e.composedPath()` always guaranteed to be non-empty?

Yes. It returns _an array of EventTarget objects representing the objects on which an event listener will be invoked_ (from [MDN](https://developer.mozilla.org/en-US/docs/Web/API/Event/composedPath#Return_value)), the first item is always equals to the `event.target`, or if this event is triggered in an open mode Shadow DOM, the first item is the actually element that trigger the event. There are [some examples in MDN](https://developer.mozilla.org/en-US/docs/Web/API/Event/composedPath#Examples).

>  can you add some code comments to explain why we are adding this special logic here and on L446?

Done. A more detailed explanation can be found in #4220

lmk123:
I ran into a problem when running `yarn verify`: #4221

adidahiya (approved):

adidahiya:
Change looks good, can you try to add a unit test in a separate PR please?

lmk123:
> can you try to add a unit test in a separate PR please?

I will do this in the next few days.

PR #4228, opened 2020-07-16
fix numericinput in controlgroup

### LINKED ISSUE #4036: NumericInput has rounded corners inside ControlGroup
#### Environment

If possible, link to a minimal repro (fork [this code sandbox](https://codesandbox.io/s/blueprint-sandbox-et9xy)): https://codesandbox.io/s/throbbing-forest-vc64k

#### Steps to reproduce

- Create a ControlGroup with a few elements inside and a NumericInput in the middle

#### Actual behavior

The NumericInput has rounded corners.

#### Expected behavior

The NumericInput should have no rounded corners. If I put an InputGroup instead of a NumericInput for example, the InputGroup has no rounded corners.

Arbalester:
#### Checklist

- [ ] Includes tests
- [ ] Update documentation

Fix` border-radius` for NumericInput component inside ControlGroup


palantirtech:

packages/core/src/components/forms/_control-group.scss
```diff
@@ -273,4 +273,43 @@ Styleguide control-group
       border-radius: 0 0 $pt-border-radius $pt-border-radius;
     }
   }
+
+  /* fix for NumericInput inside ControlGroup */
```
adidahiya, line 277: I think this implementation should match how we handle `.bp3-input` inside control group, see line 67.

the weird thing here is that NumericInput renders a ControlGroup, so we end up nesting control groups. this is typically not supported, but in this case I think it's worth fixing.

there's another issue here with z-indices, which we should also fix in this implementation:

image


since this is a little more involved fix, I'll submit a PR with my proposed changes

adidahiya (changes requested):

adidahiya:
see https://github.com/palantir/blueprint/pull/4238

PR #4244, opened 2020-07-23
Make Select close on ESC provided that Filterable is false

### LINKED ISSUE #4243: Select: cannot close on escape provided that Filterable props is false
#### Environment

- __Package version(s)__: fill this out
- __Operating System__: Ubuntu 18.04.4 LTS
- __Browser name and version__: Chrome 84

#### Steps to reproduce

1. Set Filterable props to false as in Blueprint select page
1. Open the select and press ESC key

#### Actual behavior
The select is not closing

#### Expected behavior
The select should close
#### Possible solution
[see fork](https://github.com/snony14/blueprint.git)

snony14:
#### Checklist

- [ ] Includes tests
- [ ] Update documentation

Add a keydown callback that closes the `popover` on esc key pressed down regardless of Filterable props state

### DIFF
--- a/packages/select/src/components/select/select.tsx
+++ b/packages/select/src/components/select/select.tsx
@@ -152,12 +152,12 @@ export class Select<T> extends React.PureComponent<ISelectProps<T>, ISelectState
                 onClosing={this.handlePopoverClosing}
             >
                 <div
-                    onKeyDown={this.state.isOpen ? handleKeyDown : this.handleTargetKeyDown}
+                    onKeyDown={this.state.isOpen ? this.onHandleKeyDown(handleKeyDown) : this.handleTargetKeyDown}
                     onKeyUp={this.state.isOpen ? handleKeyUp : undefined}
                 >
                     {this.props.children}
                 </div>
-                <div onKeyDown={handleKeyDown} onKeyUp={handleKeyUp}>
+                <div onKeyDown={this.onHandleKeyDown(handleKeyDown)} onKeyUp={handleKeyUp}>
                     {filterable ? input : undefined}
                     {listProps.itemList}
                 </div>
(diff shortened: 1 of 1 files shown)


palantirtech:
Thanks for your interest in [palantir/blueprint](https://github.com/palantir/blueprint), @snony14! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

snony14:
@adidahiya , please let me know if there is something I missed or is uncleared. Thanks.

adidahiya (changes requested):
There is already key handler logic present in Overlay [here](https://github.com/palantir/blueprint/blob/9358c4e6029c6b9612cbb21bd5530271385a5bc9/packages/core/src/components/overlay/overlay.tsx#L462) which should handle closing the popover rendered by Select. So in order to solve this problem you shouldn't be adding a new handler, but rather making sure that the Select popover contents get focussed when it's open.

adidahiya:
closing as stale, happy to re-open if you push an update

rfbowen:
> There is already key handler logic present in Overlay [here](https://github.com/palantir/blueprint/blob/9358c4e6029c6b9612cbb21bd5530271385a5bc9/packages/core/src/components/overlay/overlay.tsx#L462) which should handle closing the popover rendered by Select. So in order to solve this problem you shouldn't be adding a new handler, but rather making sure that the Select popover contents get focussed when it's open.

I ran into this bug and PR recently after encountering the same on a project I'm working on. I have somewhat mixed feelings about how we should go about attacking it.

Select includes key handling functionality that is preserved whether focused on the Popover itself or on the child element of the Select. This makes for a fairly convenient experience, and seems intentional. However, the escape key handler we're discussing here is buried within Overlay, and thus creates a bit of a division in terms of how we're binding our handlers. Up/Down get bound to popover AND child, whereas Escape is only bound to the popover.

https://github.com/palantir/blueprint/blob/develop/packages/select/src/components/select/select.tsx#L163-L172
https://github.com/palantir/blueprint/blob/develop/packages/core/src/components/overlay/overlay.tsx#L270

I was able to hack around this by adding an onOpened handler to add a tabindex and .focus() on the resulting menu, but it feels a bit hacky. It seems like Select could benefit from this handler being included on the child as well.

@adidahiya

PR #4266, opened 2020-08-06
[core] fix(InputGroup): async controlled updates

adidahiya:
#### Checklist

- [x] Includes tests
- [ ] Update documentation

Add a new private component, `AsyncControllableInput`, which works around the problems in the linked React issue, alleviating some bugs we've experienced with IME

Test this in a real product and make sure it doesn't break anything

#### Screenshot

N/A

### DIFF
--- a/packages/core/src/components/forms/asyncControllableInput.tsx
+++ b/packages/core/src/components/forms/asyncControllableInput.tsx
@@ -1,0 +1,102 @@
+/* !
+ * Copyright 2020 Palantir Technologies, Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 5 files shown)


rynorris:
@adidahiya I've tested this PR against the product where we originally found the issue, and it does appear to fix it.

However, unless I'm mistaken, we're still relying on the state update [here](https://github.com/palantir/blueprint/pull/4266/files#diff-f39910a9e6e127ef326a3556b3db0d2aR99) being synchronous.  In practice, given how small and simple this component is, it seems likely this state update _will_ always be synchronous (hence why it appears to fix the issue), but I don't think that's guaranteed by React at all.

I'm not sure if there's a better fix though, and maybe we're ok with it just being significantly less vulnerable to the issue than before?

adidahiya:
I tried fixing the tests but I don't see an obvious reason why they would fail only in React 15. I will come back to this later this week, but if you have any ideas I would welcome them in the meantime

adidahiya:
🤦 I had forgot to apply the react-lifecycles-compat polyfill to make `getDerivedStateFromProps` work in React 15...

PR #4268, opened 2020-08-07
[core] fix(EditableText): add small width buffer to prevent crash

### LINKED ISSUE #4267: Edge case in EditableText causes infinite loop crash
#### Environment

- __Package version(s)__: @blueprintjs/core 3.9.0
- __Operating System__: MacOS
- __Browser name and version__: Chrome 84.0.4147.89

#### Code Sandbox

https://codesandbox.io/s/reproduce-editabletext-bug-2nisb

#### Steps to reproduce

1. Create an `EditableText` component.
2. Set a `minWidth={17}` prop on it.
3. Type in "aa"

#### Actual behavior

The component crashes with the following error:
```
Maximum update depth exceeded. This can happen when a component repeatedly calls setState inside componentWillUpdate or componentDidUpdate. React limits the number of nested updates to prevent infinite loops.
```

editable-text-bug

#### Expected behavior

No crash.

#### Possible solution

It seems that the call to `updateInputDimensions()` [here](https://github.com/palantir/blueprint/blob/develop/packages/core/src/components/editable-text/editableText.tsx#L266) for some reason causes an infinite update loop specifically when `minWidth` is set, and entering text with very specific width.  Removing this call prevents the crash (but also prevents the component resizing of course)

rynorris:
#### Fixes #4267

I've tested this PR against the application where we originally noticed the crash, and confirmed it does fix the problem.

#### Checklist

- [ ] Includes tests
- [ ] Update documentation

Add a small buffer to scrollWidth in EditableText to prevent the infinite loop described in #4267 
Before this PR we only added the buffer for IE and Edge.  This PR makes it so we add the buffer for all browsers.

Is this really the "right" fix?  Or is this just a patch over a deeper issue?

### DIFF
--- a/packages/core/src/components/editable-text/editableText.tsx
+++ b/packages/core/src/components/editable-text/editableText.tsx
@@ -128,7 +128,7 @@ export interface IEditableTextState {
     value?: string;
 }
 
-const BUFFER_WIDTH_EDGE = 5;
+const BUFFER_WIDTH_DEFAULT = 5;
 const BUFFER_WIDTH_IE = 30;
 
 @polyfill
@@ -258,6 +258,7 @@ export class EditableText extends AbstractPureComponent2<IEditableTextProps, IEd
         if (this.props.disabled || (this.props.disabled == null && prevProps.disabled)) {
             state.isEditing = false;
         }
+
         this.setState(state);
 
         if (this.state.isEditing && !prevState.isEditing) {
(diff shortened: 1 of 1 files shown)


adidahiya:
[docs preview](https://34238-71939872-gh.circle-artifacts.com/0/packages/docs-app/dist/index.html#core/components/editable-text)

adidahiya (approved):
yeah, kind of a patchy fix, but seems harmless enough. I tested that it doesn't break existing behavior

PR #4286, opened 2020-08-20
added prop validation for multiSlider and rangeSlider to prevent out of bounds values

### LINKED ISSUE #4285: Slider Handle goes out of bounds when provided values lower/higher than min/max
#### Environment

- __Package version(s)__: 3.25.0
- __Operating System__: Ubuntu 18.04
- __Browser name and version__: Chrome 84

#### Code Sandbox

[sandbox](https://codesandbox.io/s/blueprint-sandbox-forked-gqlde) 

#### Steps to reproduce

1. Create a slider (also works for multislider)
2. Provide min/max props
3. Set the value to be greater than max value

#### Actual behavior

The handle goes out of bounds and floats to the side

#### Expected behavior

Should filter the value under the hood

#### Possible solution

`.filter(value => value > min] && value < max)` under the hood

This specifically came up because I'm trying to sync a slider with a d3 graph which zooms. I think everything works pretty nicely except for this out of bounds issue. I understand that this wouldn't happen under normal use of the slider where you would only be able to set the value prop of the slider by dragging and setting on the slider, but I also don't think this would be difficult/performance heavy.

jdleo:
#### Checklist

- [x] Includes tests
- [x] Update documentation

As referenced in issue #4285 , a user can supply out-of-bounds value and put the `MultiSlider` and `RangeSlider` components into bad states. I proposed a logical guard, and created a new validation check in `validateProps`, following as the other components had done.

That props are validated, and you cannot supply values that lie outside of the `min, max` range.

#### Screenshot

N/A

### DIFF
--- a/packages/core/src/components/slider/multiSlider.tsx
+++ b/packages/core/src/components/slider/multiSlider.tsx
@@ -210,6 +210,20 @@ export class MultiSlider extends AbstractPureComponent2<IMultiSliderProps, ISlid
         if (anyInvalidChildren) {
             throw new Error(Errors.MULTISLIDER_INVALID_CHILD);
         }
+
+        // don't allow supplied values to be outside of min/max range
+        let anyOutOfBoundsChildren = false;
+        React.Children.forEach(props.children, (child: Handle) => {
+            const { value, min, max } = child.props;
+
+            // make sure this child's values are within parent range and individual range
+            if (value < min || value > max || value < props.min || value > props.max) {
+                anyOutOfBoundsChildren = true;
+            }
+        });
+        if (anyOutOfBoundsChildren) {
+            throw new Error(Errors.MULTISLIDER_OUT_OF_BOUNDS);
+        }
     }
 
     private formatLabel(value: number): React.ReactChild {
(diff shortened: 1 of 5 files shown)


palantirtech:
Thanks for your interest in [palantir/blueprint](https://github.com/palantir/blueprint), @jdleo! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

adidahiya:
@jdleo this looks like a step in the right direction, any reason why you closed the PR?

jdleo:
> @jdleo this looks like a step in the right direction, any reason why you closed the PR?

Hey! Sorry about that, I've actually picked this to be my first contribution on this repo, and I had to write tests for it, I wrote a test for `expectPropValidationError` as it seemed fitting. Will re-submit PR now 👍

packages/core/src/components/slider/multiSlider.tsx
```diff
@@ -210,6 +210,20 @@ export class MultiSlider extends AbstractPureComponent2<IMultiSliderProps, ISlid
...
+        React.Children.forEach(props.children, (child: Handle) => {
+            const { value, min, max } = child.props;
+
+            // make sure this child's values are within parent range and individual range
+            if (value < min || value > max || value < props.min || value > props.max) {
+                anyOutOfBoundsChildren = true;
+            }
+        });
+        if (anyOutOfBoundsChildren) {
+            throw new Error(Errors.MULTISLIDER_OUT_OF_BOUNDS);
```
adidahiya, line 225: this is potentially a breaking change and it can result in apps breaking if they are relying on silent failures of the component in out-of-bounds cases. so I think we should introduce this check with `console.error` instead of throwing an error.

adidahiya (changes requested):
sorry for the late review! let's be a little gentler in rolling out this API change

adidahiya:
closing for now, happy to re-open with an updated PR

PR #4293, opened 2020-08-27
chore: enable useDefineForClassFields compiler option

### LINKED ISSUE #3957: chore: enable TS 3.7 useDefineForClassFields flag
https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#the-usedefineforclassfields-flag-and-the-declare-property-modifier

adidahiya:


adidahiya:
I think this is less relevant with TS 4.0

PR #4306, opened 2020-09-03
Allows labelStepSize to receive an array of specific label positions

cybae0804:
#### Checklist

- [x] Includes tests
- [x] Update documentation

Allows labelStepSize to receive an array of specific label positions.

First contribution, not sure if I updated the docs properly.


J-Vin:
Array of #'s, tests, updates AND organized ☑! How could it possibly not be good? ☺ I do have a question about deleting. I am not real clear on that process and wonder what creates that step. Are they typically duplicates?

PR #4308, opened 2020-09-03
[core] feat: allow sliders to specify custom label values

cybae0804:
#### Checklist

- [x] Includes tests
- [x] Update documentation

Allows labelStepSize to receive an array of specific label positions.

First contribution, not sure if I updated the docs properly.


packages/core/src/components/slider/multiSlider.tsx
```diff
@@ -41,10 +41,11 @@ export interface ISliderBaseProps extends IProps, IIntentProps {
     disabled?: boolean;
 
     /**
-     * Increment between successive labels. Must be greater than zero.
+     * As a number, increment between successive labels. Must be greater than zero.
+     * As an array of numbers, places label that corresponds to each entry.
      * @default 1
      */
-    labelStepSize?: number;
+    labelStepSize?: number | number[];
```
adidahiya: I think `labelStepSize` no longer makes sense as a name for this prop if it's going to be an array of explicit labels. Let's instead add this feature with a new prop called `labelValues: number[]` and add validation inside `validateProps` to make it mutually exclusive w/ `labelStepSize` (see [here](https://github.com/palantir/blueprint/blob/ba5c9b56e03642c8168fba55e1766fd3ec864e32/packages/core/src/components/forms/radioGroup.tsx#L87-L89) for an example of that kind of check in another component).

adidahiya (changes requested):

packages/core/src/components/slider/multiSlider.tsx
cybae0804: So currently, `labelStepSize` defaults to 1. And if I were to add a new prop called `labelValues` with the default of `null`, and they are meant to be mutually exclusive, does that mean the user should manually set the `labelStepSize` to `null` in order to use `labelValues`?

cybae0804: I guess this would be fixed by allowing the `labelValues` to override `labelStepSize` (while still throwing a warning indicating they should be mutually exclusive?). Should I use this implementation?

adidahiya: a slightly different suggestion: remove the default value of `labelStepSize` in `defaultProps`, instead implement that default value dynamically in a private helper method `getLabelValues()`. then you can still validate for mutual exclusivity properly

cybae0804:
Updated as requested. 

Some things to note

- The default for `labelStepSize` is set to inferred 1 in the docs, because as requested, it's dynamically set. 
- The default for `labelStepSize` and `labelValues` are set to `null`, but in hindsight maybe it would be fine to leave it as `undefined`. Let me know if I need to change that.
- The original loop has been significantly simplified to work off of just the label values, and I tested pretty extensively and it all seems to work.

packages/core/src/components/slider/multiSlider.tsx
```diff
@@ -42,9 +42,15 @@ export interface ISliderBaseProps extends IProps, IIntentProps {
 
     /**
      * Increment between successive labels. Must be greater than zero.
-     * @default 1
+     * @default inferred 1
+     */
+    labelStepSize?: number | null;
```
adidahiya: as you just suggested, I would like this to use `undefined` instead of `null`. there's no need to introduce _two_ kinds of "non-values" unless absolutely necessary.

```suggestion
    labelStepSize?: number;
```

packages/core/src/components/slider/multiSlider.tsx
```diff
@@ -42,9 +42,15 @@ export interface ISliderBaseProps extends IProps, IIntentProps {
...
+
+    /**
+     * Array of specific values for the label placement. This prop is mutually exclusive with
+     * `labelStepSize`.
      */
-    labelStepSize?: number;
+    labelValues?: number[] | null;
```
adidahiya:
```suggestion
    labelValues?: number[];
```

packages/core/src/components/slider/multiSlider.tsx
```diff
@@ -42,9 +42,15 @@ export interface ISliderBaseProps extends IProps, IIntentProps {
 
     /**
      * Increment between successive labels. Must be greater than zero.
-     * @default 1
+     * @default inferred 1
```
adidahiya:
```suggestion
     * @default 1 (if labelStepSize is undefined)
```

packages/core/src/components/slider/multiSlider.tsx
```diff
@@ -117,7 +123,8 @@ export interface ISliderState {
 export class MultiSlider extends AbstractPureComponent2<IMultiSliderProps, ISliderState> {
     public static defaultSliderProps: ISliderBaseProps = {
         disabled: false,
-        labelStepSize: 1,
+        labelStepSize: null,
+        labelValues: null,
```
adidahiya:
```suggestion
```

packages/core/src/components/slider/multiSlider.tsx
```diff
@@ -196,7 +203,10 @@ export class MultiSlider extends AbstractPureComponent2<IMultiSliderProps, ISlid
         if (props.stepSize! <= 0) {
             throw new Error(Errors.SLIDER_ZERO_STEP);
         }
-        if (props.labelStepSize! <= 0) {
+        if (props.labelStepSize !== null && props.labelValues !== null) {
```
adidahiya:
```suggestion
        if (props.labelStepSize !== undefined && props.labelValues !== undefined) {
```

packages/core/src/components/slider/multiSlider.tsx
```diff
@@ -196,7 +203,10 @@ export class MultiSlider extends AbstractPureComponent2<IMultiSliderProps, ISlid
...
+            throw new Error(Errors.MULTISLIDER_WARN_LABEL_STEP_SIZE_LABEL_VALUES_MUTEX);
+        }
+        if (props.labelStepSize !== null && props.labelStepSize! <= 0) {
```
adidahiya:
```suggestion
        if (props.labelStepSize !== undefined && props.labelStepSize <= 0) {
```

packages/core/src/components/slider/multiSlider.tsx
```diff
@@ -419,6 +423,19 @@ export class MultiSlider extends AbstractPureComponent2<IMultiSliderProps, ISlid
         });
     };
 
+    private getLabelValues() {
+        const { labelStepSize, labelValues, min, max } = this.props;
+        let values: number[] = [];
+        if (labelValues) values = labelValues;
+        else {
```
adidahiya: 🤦 I really need to re-enable the lint rule for this... but please ensure that all if/else/for/while/etc... clauses are blocks enclosed by curlys

```suggestion
        if (labelValues !== undefined) {
            values = labelValues;
        } else {
```

packages/core/test/slider/multiSliderTests.tsx
```diff
@@ -170,12 +170,24 @@ describe("<MultiSlider>", () => {
...
+            assertLabelCount(wrapper, 6);
+        });
+
         it("renders label for value and for each labelStepSize", () => {
             // [0  10  20  30  40  50]
             const wrapper = renderSlider({ min: 0, max: 50, labelStepSize: 10 });
             assertLabelCount(wrapper, 6);
         });
 
+        it("renders label for value and for each labelValues entry", () => {
```
adidahiya:
```suggestion
        it("renders labels provided in labelValues prop", () => {
```

packages/core/test/slider/multiSliderTests.tsx
```diff
@@ -188,6 +200,12 @@ describe("<MultiSlider>", () => {
             assert.strictEqual(wrapper.find(`.${Classes.SLIDER}-axis`).text(), "0#10#20#30#40#50#");
         });
 
+        it("renders result of labelRenderer() in each label with labelValues", () => {
```
adidahiya, line 203: thx for adding this test 👍

adidahiya (changes requested):
looking good, just minor comments

cybae0804:
Updated with all your suggestions

adidahiya (approved):

PR #4309, opened 2020-09-03
[core] fix(Button): always exit active state if blurred

### LINKED ISSUE #3977: Button can become stuck in pressed state
#### Environment

- __Package version(s)__:   3.23
- __Browser and OS versions__:  Google Chrome v80.0 on MacOS Catalina v10.15.3

If possible, link to a minimal repro (fork [this code sandbox](https://codesandbox.io/s/blueprint-sandbox-et9xy)): https://codesandbox.io/s/blueprint-sandbox-1ojtw

#### Steps to reproduce

1. Use Tab to focus the first button.

2. Press and hold Enter.

3. Press Tab to move focus to the second button.

4. Release Enter.

#### Actual behavior

The button will be displayed as pressed even though the user is no longer interacting with it.

image

Note that the first two buttons are pressed.

#### Expected behavior

Buttons should only be shown as pressed when the user is pressing them or they are passed `active: true` as a prop.

#### Possible solution

The cause is definitely the key event handlers in [abstractButton.tsx](https://github.com/palantir/blueprint/blob/develop/packages/core/src/components/button/abstractButton.tsx). The first button gets the keyDown event and sets its `isActive` state to true, but the second button gets the `keyUp` event so the first button never resets its `isActive` state.

I don't have any ideas for a solution at this time.

#### Other notes

This seems to be reproducible on any button. It was first noticed with buttons inside ButtonGroups in https://github.com/palantir/blueprint/issues/3971, but the problem is with the underlying Button component.

jaamison:

Add an onBlur handler passed to the button component that unsets the local component `isActive` state on DOM blur events if necessaary.

### DIFF
--- a/packages/core/src/components/button/abstractButton.tsx
+++ b/packages/core/src/components/button/abstractButton.tsx
@@ -141,6 +141,7 @@ export abstract class AbstractButton<H extends React.HTMLAttributes<HTMLElement>
         return {
             className,
             disabled,
+            onBlur: this.handleBlur,
             onClick: disabled ? undefined : this.props.onClick,
             onKeyDown: this.handleKeyDown,
             onKeyUp: this.handleKeyUp,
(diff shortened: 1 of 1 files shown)


packages/core/src/components/button/abstractButton.tsx
```diff
@@ -176,6 +177,11 @@ export abstract class AbstractButton<H extends React.HTMLAttributes<HTMLElement>
         this.props.onKeyUp?.(e);
     };
 
+    protected handleBlur = (e: React.FocusEvent<any>) => {
+        if (this.state.isActive) this.setState({ isActive: false });
```
```suggestion
        if (this.state.isActive) {
            this.setState({ isActive: false });
        }
```

we used to have a lint rule for this 😕

adidahiya (changes requested):

adidahiya (approved):

PR #4314, opened 2020-09-06
[datetime] feat: support highlightCurrentDay prop in more components

ejose19:
Add missing highlightCurrentDay prop for dateInput (which already supported it), and port support to dateRangePicker/dateRangeInput

### DIFF
--- a/packages/datetime/src/dateInput.tsx
+++ b/packages/datetime/src/dateInput.tsx
@@ -91,6 +91,12 @@ export interface IDateInputProps extends IDatePickerBaseProps, IDateFormatProps,
     fill?: boolean;
 
     /**
+     * Whether the current day should be highlighted in the calendar.
+     * @default false
+     */
+    highlightCurrentDay?: boolean;
+
+    /**
      * Props to pass to the [input group](#core/components/text-inputs.input-group).
      * `disabled` and `value` will be ignored in favor of the top-level props on this component.
      * `type` is fixed to "text" and `ref` is not supported; use `inputRef` instead.
--- a/packages/datetime/src/dateInput.tsx
+++ b/packages/datetime/src/dateInput.tsx
@@ -16,7 +16,6 @@
 
 import classNames from "classnames";
 import * as React from "react";
-import { DayPickerProps } from "react-day-picker";
 import { polyfill } from "react-lifecycles-compat";
 
 import {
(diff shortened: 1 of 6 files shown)


J-Vin:
How would current day highlight determine the difference globally?

ejose19:
@J-Vin Care to explain what do you mean exactly?

packages/datetime/src/dateRangePicker.tsx
```diff
@@ -295,6 +301,29 @@ export class DateRangePicker extends AbstractPureComponent2<IDateRangePickerProp
         }
     }
 
+    private isToday = (date: Date) => DateUtils.areSameDay(date, new Date());
```
adidahiya: can we make this a new util function in `DateUtils`? that way we can share it with `datePicker.tsx` and we can remove the private class method in both places.

packages/datetime/src/dateRangePicker.tsx
```diff
@@ -80,6 +80,12 @@ export interface IDateRangePickerProps extends IDatePickerBaseProps, IProps {
      */
     defaultValue?: DateRange;
```
adidahiya: please make sure to add to the JSDoc comment for the `dayPickerProps` prop, just like in `datePicker.tsx`:


```
     * In case of supplying your owner `renderDay` function, make sure to apply the appropriate
     * CSS wrapper class to obtain default Blueprint styling.
     * eg.
     * `<div className={Classes.DATEPICKER_DAY_WRAPPER}>{CONTENT_HERE}</div>`
```

packages/datetime/src/dateInput.tsx
```diff
@@ -90,6 +90,12 @@ export interface IDateInputProps extends IDatePickerBaseProps, IDateFormatProps,
      */
     fill?: boolean;
 
+    /**
+     * Whether the current day should be highlighted in the calendar.
+     * @default false
+     */
+    highlightCurrentDay?: boolean;
```
adidahiya: since we're adding support to all day picker components, let's move this to `IDatePickerBaseProps`, so it gets inherited by all prop interfaces

adidahiya (changes requested):

ejose19:
@adidahiya I've applied your reviews and also moved dayPickerProps to IDatePickerBaseProps since that one is also being used in all day picker components.

adidahiya (approved):
nice, thanks @ejose19

PR #4316, opened 2020-09-07
[datetime] fix(DateInput): Improve blur handling to avoid unexpected popover closures

### LINKED ISSUE #3603: DatePicker popover close unexpectedly in Chrome when moving to min/max month
#### Steps to reproduce

In the dateInput component
1. set the start value to null or delete the selected date
2. use the arrows to change month
3. if your max date is, for example, 2019-12-31 when you are on November 2019, if you click the next month button the date picker close unexpectedly

ejose19:

The poposed solution is to avoid the case where minDate/maxDate is set, and using the change month arrows makes the popover to close on chrome browser, upon debugging, the issue is that document.activeElement is returning "body" instead of the change month button, which makes ```handlePopoverBlur``` set up execute incorrectly and subsequently calling ```handleClosePopover``` which causes the described issue.

While this solution works correctly for that case, it makes "Popover closes when first day of the month is blurred" test fail, and after evaluating both use cases, it seems the min/max date has more weight, and so losing that "feature" is worth it.

Here's the [sandbox](https://codesandbox.io/s/old-butterfly-spcxx) (open in chrome) and without selecting any date, move 2 months before/after and notice how the popover closes incorrectly.

### DIFF
--- a/packages/datetime/src/common/classes.ts
+++ b/packages/datetime/src/common/classes.ts
@@ -36,7 +36,7 @@ export const DATEPICKER_FOOTER = `${DATEPICKER}-footer`;
 export const DATEPICKER_MONTH_SELECT = `${DATEPICKER}-month-select`;
 export const DATEPICKER_YEAR_SELECT = `${DATEPICKER}-year-select`;
 export const DATEPICKER_NAVBAR = `${DATEPICKER}-navbar`;
-export const DATEPICKER_NAVBUTTON = `${DATEPICKER}-NavButton`;
+export const DATEPICKER_NAVBUTTON = `DayPicker-NavButton`;
 export const DATEPICKER_TIMEPICKER_WRAPPER = `${DATEPICKER}-timepicker-wrapper`;
 
 export const DATERANGEPICKER = `${NS}-daterangepicker`;
--- a/packages/datetime/src/common/classes.ts
+++ b/packages/datetime/src/common/classes.ts
@@ -36,6 +36,7 @@ export const DATEPICKER_FOOTER = `${DATEPICKER}-footer`;
 export const DATEPICKER_MONTH_SELECT = `${DATEPICKER}-month-select`;
 export const DATEPICKER_YEAR_SELECT = `${DATEPICKER}-year-select`;
 export const DATEPICKER_NAVBAR = `${DATEPICKER}-navbar`;
+export const DATEPICKER_NAVBUTTON = `${DATEPICKER}-NavButton`;
 export const DATEPICKER_TIMEPICKER_WRAPPER = `${DATEPICKER}-timepicker-wrapper`;
 
 export const DATERANGEPICKER = `${NS}-daterangepicker`;
--- a/packages/datetime/src/dateInput.tsx
+++ b/packages/datetime/src/dateInput.tsx
@@ -207,13 +207,13 @@ export class DateInput extends AbstractPureComponent2<IDateInputProps, IDateInpu
         const dateValue = isDateValid(value) ? value : null;
         const dayPickerProps: DayPickerProps = {
             ...this.props.dayPickerProps,
-            // If the user tabs on first day of the month, close the Popover
+            // If the user tabs on a DayPicker Day element and the lastElementInPopover is also a DayPicker Day
+            // element, the popover should be closed
             onDayKeyDown: (day, modifiers, e) => {
-                if (day.getDate() === 1 && e.key === "Tab") {
+                if (e.key === "Tab" && !e.shiftKey && this.lastElementInPopover.classList.contains("DayPicker-Day")) {
                     this.setState({ isOpen: false });
-                } else {
-                    this.props.dayPickerProps.onDayKeyDown?.(day, modifiers, e);
                 }
+                this.props.dayPickerProps.onDayKeyDown?.(day, modifiers, e);
             },
             // dom elements for the updated month is not available when
             // onMonthChange is called. setTimeout is necessary to wait
@@ -421,16 +421,23 @@ export class DateInput extends AbstractPureComponent2<IDateInputProps, IDateInpu
             // Support IE11 (#2924)
             relatedTarget = document.activeElement as HTMLElement;
         }
+        const eventTarget = e.target as HTMLElement;
         // Beware: this.popoverContentEl is sometimes null under Chrome
         if (
             relatedTarget == null ||
-            (this.popoverContentEl != null &&
-                !this.popoverContentEl.contains(relatedTarget) &&
-                // Exclude events that leads to relatedTarget being body, like changing
-                // months (either by change month buttons or using arrow keys on Day element)
-                relatedTarget !== document.body)
+            (this.popoverContentEl != null && !this.popoverContentEl.contains(relatedTarget))
         ) {
-            this.handleClosePopover();
+            // Exclude the following blur operations that makes "body" the activeElement
+            // and would close the Popover unexpectedly
+            // - On disabled change months buttons
+            // - DayPicker day elements, their "blur" will be managed at its own onKeyDown
+            // event listener for improving UX
+            const isChangeMonthEvt = eventTarget.classList.contains("DayPicker-NavButton");
+            const isChangeMonthButtonDisabled = isChangeMonthEvt && (eventTarget as HTMLButtonElement).disabled;
+            const isDayPickerDayEvt = eventTarget.classList.contains("DayPicker-Day");
+            if (!isChangeMonthButtonDisabled && !isDayPickerDayEvt) {
+                this.handleClosePopover();
+            }
         } else if (relatedTarget != null) {
             this.unregisterPopoverBlurHandler();
             this.lastElementInPopover = relatedTarget;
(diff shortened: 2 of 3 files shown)


adidahiya (changes requested):
see my comment https://github.com/palantir/blueprint/issues/3603#issuecomment-689126068

ejose19:
@adidahiya I think by the point we reach ```handlePopoverBlur``` it's too late to do something already. However at ```registerPopoverBlurHandler``` we can pass an argument to check if ```onMonthChange``` called this function, and which arrow triggered the event (left or right). Then since ```lastTabbableElement``` will be undefined, if the flag is set we assign a better candidate for ```lastTabbableElement```.

However what happens if theoretically a month change leads the popover to a state without any tabbable element? Besides from that let me know what you think about this idea.

ejose19:
@adidahiya I've refactored this PR, now it should cover all use cases and even improve existing function-ability which had another unreported bug.

To reproduce the other bug, on a DateInput will nul value, place the active element on the first day of the month, and then press left arrow key, notice how the popover closes. This is because we're making ```handlePopoverBlur``` fire when DayPicker Day elements lose blur, but this can be a normal operation if the user is just navigating using the arrow buttons.

I also removed the step where it makes ```lastTabbableElement``` the document.activeElement, which is incorrect. An example of this mistake is having the dateinput value null, the time element being shown, and navigating using the keyboard arrows, when changing months it will report the active DayPicker Day element as ```lastTabbableElement```, but in practice it's still the milliseconds input.

packages/datetime/src/dateInput.tsx
```diff
@@ -205,8 +205,16 @@ export class DateInput extends AbstractPureComponent2<IDateInputProps, IDateInpu
         const { value, valueString } = this.state;
         const dateString = this.state.isInputFocused ? valueString : getFormattedDateString(value, this.props);
         const dateValue = isDateValid(value) ? value : null;
-        const dayPickerProps = {
+        const dayPickerProps: DayPickerProps = {
             ...this.props.dayPickerProps,
+            // If the user tabs on a DayPicker Day element and the lastElementInPopover is also a DayPicker Day
+            // element, the popover should be closed
+            onDayKeyDown: (day, modifiers, e) => {
+                if (e.key === "Tab" && !e.shiftKey && this.lastElementInPopover.classList.contains("DayPicker-Day")) {
```
adidahiya:
```suggestion
                if (e.key === "Tab" && !e.shiftKey && this.lastElementInPopover?.classList.contains(Classes.DATEPICKER_DAY)) {
```

packages/datetime/src/dateInput.tsx
```diff
@@ -205,8 +205,16 @@ export class DateInput extends AbstractPureComponent2<IDateInputProps, IDateInpu
         const { value, valueString } = this.state;
         const dateString = this.state.isInputFocused ? valueString : getFormattedDateString(value, this.props);
         const dateValue = isDateValid(value) ? value : null;
-        const dayPickerProps = {
+        const dayPickerProps: DayPickerProps = {
             ...this.props.dayPickerProps,
+            // If the user tabs on a DayPicker Day element and the lastElementInPopover is also a DayPicker Day
+            // element, the popover should be closed
```
adidahiya, line 213: I don't quite follow this comment, why is this the case?

packages/datetime/src/dateInput.tsx
```diff
@@ -413,12 +421,23 @@ export class DateInput extends AbstractPureComponent2<IDateInputProps, IDateInpu
...
         if (
             relatedTarget == null ||
             (this.popoverContentEl != null && !this.popoverContentEl.contains(relatedTarget))
         ) {
-            this.handleClosePopover();
+            // Exclude the following blur operations that makes "body" the activeElement
+            // and would close the Popover unexpectedly
+            // - On disabled change months buttons
+            // - DayPicker day elements, their "blur" will be managed at its own onKeyDown
+            // event listener for improving UX
```
adidahiya: what does this comment line refer to? it feels a little out of place

packages/datetime/src/dateInput.tsx
```diff
@@ -413,12 +421,23 @@ export class DateInput extends AbstractPureComponent2<IDateInputProps, IDateInpu
...
+            const isChangeMonthEvt = eventTarget.classList.contains("DayPicker-NavButton");
```
adidahiya:
```suggestion
            const isChangeMonthEvt = eventTarget.classList.contains(Classes.DATEPICKER_NAVBUTTON);
```

you would need to add this class to `common/classes.ts` in this package

packages/datetime/src/dateInput.tsx
```diff
@@ -413,12 +421,23 @@ export class DateInput extends AbstractPureComponent2<IDateInputProps, IDateInpu
...
+            const isChangeMonthButtonDisabled = isChangeMonthEvt && (eventTarget as HTMLButtonElement).disabled;
+            const isDayPickerDayEvt = eventTarget.classList.contains("DayPicker-Day");
```
adidahiya:
```suggestion
            const isDayPickerDayEvt = eventTarget.classList.contains(Classes.DATEPICKER_DAY);
```

adidahiya (changes requested):
[docs preview](https://35010-71939872-gh.circle-artifacts.com/0/packages/docs-app/dist/index.html#datetime/dateinput)

this does appear to fix the issue, so the UX changes are ok

packages/datetime/src/dateInput.tsx
ejose19: From your https://github.com/palantir/blueprint/issues/3603#issuecomment-689126068
> but I would still like to retain the functionality added in #2093 which closes the DateInput popover after a user TABs out of the last tabbable element in the popover.

Which is basically a remark with that comment, if you feel it's redundant I can remove it.

packages/datetime/src/dateInput.tsx
ejose19: Actually one needs to read the entire comment to understand it.

```ts
// Exclude the following blur operations that makes "body" the activeElement
// and would close the Popover unexpectedly
// - On disabled change months buttons
// - DayPicker day elements, their "blur" will be managed at its own onKeyDown
// event listener for improving UX
const isChangeMonthEvt = eventTarget.classList.contains(Classes.DATEPICKER_NAV_BUTTON);
const isChangeMonthButtonDisabled = isChangeMonthEvt && (eventTarget as HTMLButtonElement).disabled;
const isDayPickerDayEvt = eventTarget.classList.contains(Classes.DATEPICKER_DAY);
if (!isChangeMonthButtonDisabled && !isDayPickerDayEvt) {
    this.handleClosePopover();
}
```

So the excluded blur events are ```isChangeMonthButtonDisabled``` and ```isDayPickerDayEvt``` as you can see in the if condition. Any idea to make it more clear?

packages/datetime/src/common/classes.ts
```diff
@@ -36,6 +36,7 @@ export const DATEPICKER_FOOTER = `${DATEPICKER}-footer`;
 export const DATEPICKER_MONTH_SELECT = `${DATEPICKER}-month-select`;
 export const DATEPICKER_YEAR_SELECT = `${DATEPICKER}-year-select`;
 export const DATEPICKER_NAVBAR = `${DATEPICKER}-navbar`;
+export const DATEPICKER_NAVBUTTON = `${DATEPICKER}-NavButton`;
```
adidahiya: actually it's 
```suggestion
export const DATEPICKER_NAVBUTTON = `DayPicker-NavButton`;
```

sorry, I know the "date" vs "day" thing is confusing in this file

packages/datetime/src/dateInput.tsx
adidahiya: but what about "...and the lastElementInPopover is also a DayPicker Day element"? why is this guard necessary? the comment leaves me wondering what the other possibilities are (I assume it refers to cases where a time picker is rendered, making that the last tabbable element?)

packages/datetime/src/dateInput.tsx
adidahiya: oh, sorry, I wasn't reading the last line as a continuation of the 2nd bullet. in that case just remove these words, `onKeyDown` is enough info:

```suggestion
```

packages/datetime/src/dateInput.tsx
```diff
@@ -205,8 +205,20 @@ export class DateInput extends AbstractPureComponent2<IDateInputProps, IDateInpu
         const { value, valueString } = this.state;
         const dateString = this.state.isInputFocused ? valueString : getFormattedDateString(value, this.props);
         const dateValue = isDateValid(value) ? value : null;
-        const dayPickerProps = {
+        const dayPickerProps: DayPickerProps = {
             ...this.props.dayPickerProps,
+            // If the user tabs on a DayPicker Day element and the lastElementInPopover is also a DayPicker Day
```
adidahiya: I think this could be a little more clear, making it explicit that it's TAB being pressed while the day is focussed, not the TAB moving focus _onto_ the day element

```suggestion
            // If the user presses the TAB key on a DayPicker Day element and the lastElementInPopover is also a DayPicker Day
```

packages/datetime/src/dateInput.tsx
ejose19: Yes, I tried all combinations and that extra guards need to be in place for when there's any tabbable element below the calendar (like time picker, but it would cover any new element added in the future)

adidahiya: ok, then that begs the question... which code is responsible for closing the popover when there is a tabbable element (like TimePicker) below the calendar? your implementation seems to work fine, but I want to make sure the code comments are clear and comprehensible for the next person who reads this code. I'm pulling your branch now to figure this out for myself

ejose19: ```registerPopoverBlurHandler``` along ```handlePopoverBlur``` is handling that (like originally implemented), just excluding the 2 cases mentioned in the comment. It's actually "agnostic" regarding if the last tabbable element is below the calendar or not, that state variable holds the last tabbable element inside the popover and listens for blur events.

Actually, now that you mention it, I found that this code https://github.com/palantir/blueprint/blob/408a8a73c4b49f1642abadcd0cde3e113d82e055/packages/datetime/src/dateInput.tsx#L446-L447 is also incorrect. They are storing the "last tabbed element", not "last tabbable element" which will cause an issue if the user shift+tab and then tab. I will fix that part to use the same formula as ```registerPopoverBlurHandler``` (also maybe a documentation to ```lastElementInPopover``` meaning "last element in popover that is tabbable, and the one that triggers closure if the user press tabs on it"

adidahiya:
[docs preview build](https://35250-71939872-gh.circle-artifacts.com/0/packages/docs-app/dist/index.html#datetime/dateinput)

confirmed in Chrome, FF, IE11 👍 

thanks @ejose19

adidahiya (approved):

PR #4325, opened 2020-09-09
[core] fix(ControlGroup): normalize fill style

### LINKED ISSUE #4213: NumericInput fill prop doesn't work

ejose19:

Normalize fill for ControlGroup like the rest of the components, so it works as intended.

### DIFF
--- a/packages/core/src/components/forms/_control-group.scss
+++ b/packages/core/src/components/forms/_control-group.scss
@@ -272,6 +272,9 @@ Styleguide control-group
 
   Styleguide control-group-fill
   */
+  &.#{$ns}-fill {
+    width: 100%;
+  }
 
   > .#{$ns}-fill {
     flex: 1 1 auto;
--- a/packages/docs-app/src/examples/core-examples/controlGroupExample.tsx
+++ b/packages/docs-app/src/examples/core-examples/controlGroupExample.tsx
@@ -43,14 +43,9 @@ export class ControlGroupExample extends React.PureComponent<IExampleProps, ICon
             </>
         );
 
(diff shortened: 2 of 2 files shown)


ejose19:
Maybe an edit to https://blueprintjs.com/docs/#core/components/control-group.flex-layout is required, as it's stating:
>Enable the fill prop on a control group to make all controls expand equally to fill the available space.

However, that was never the case as seen in the docs, using fill on ControlGroup distributes the space proportionally to the original child sizes. True equality can be obtained if all the childs also have fill set.

adidahiya (approved):

adidahiya:
made a documentation tweak here: https://github.com/palantir/blueprint/commit/13ff0ae9ec122a9c284df9efff70e8073a0c9e2c

ejose19:
@adidahiya Ok that change is good, but shouldn't the other line be modified too?

from this:
> Enable the fill prop on a control group to make all controls expand equally to fill the available space.

to this:
> Enable the fill prop on a control group to make all controls fill the available space expanding proportionally to their original sizes.

adidahiya:
I think the existing documentation can be interpreted as correct; "expand equally" (although somewhat ambiguous) can mean that they grow proportionally, i.e. `flex-grow: 1` for all elements

PR #4334, opened 2020-09-16
[core] feat(Text): add title prop

ejose19:
#### Checklist

- [X] Includes tests
- [X] Update documentation

Adds title prop to Text element and adjust all dependant components so they can pass that property.

### DIFF
--- a/packages/core/src/components/panel-stack/panelView.tsx
+++ b/packages/core/src/components/panel-stack/panelView.tsx
@@ -42,6 +42,11 @@ export interface IPanelViewProps {
 
     /** Whether to show the header with the "back" button. */
     showHeader: boolean;
+
+    /**
+     * HTML title to be passed to the <Text> component
+     */
+    textTitle?: string;
 }
 
 @polyfill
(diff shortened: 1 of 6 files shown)


packages/core/src/components/tag/tag.tsx
```diff
@@ -97,6 +97,11 @@ export interface ITagProps extends IProps, IIntentProps, React.HTMLAttributes<HT
      * @default false
      */
     round?: boolean;
+
+    /**
+     * HTML title to be passed to the <Text> component
+     */
+    textTitle?: string;
```
adidahiya:
```suggestion
    htmlTitle?: string;
```

packages/core/src/components/panel-stack/panelView.tsx
```diff
@@ -42,6 +42,11 @@ export interface IPanelViewProps {
 
     /** Whether to show the header with the "back" button. */
     showHeader: boolean;
+
+    /**
+     * HTML title to be passed to the <Text> component
+     */
+    textTitle?: string;
```
adidahiya: I think this should be a property of `IPanel`, as `htmlTitle`. then you can render it in the back button as well (`maybeRenderBack()`)

adidahiya (changes requested):

adidahiya (approved):
thanks @ejose19

Issue #4272, opened 2020-08-10
PanelStack: onOpen, onClose and initialPanel prop should use type argument P


cecton:
#### Environment

- __Package version(s)__: 3.30.0
- __Browser and OS versions__: Firefox 79, Arch Linux

#### Feature request

The arguments `addedPanel` and `removedPanel` of the functions `onOpen` and `onClose` on the interface [`IPanelStackProps`](https://blueprintjs.com/docs/#core/components/panel-stack.props) don't allow setting their type argument. Therefore it is set to `{}`.

This is very inconvenient because then the type `IPanel<{}>` must be used for the state. Could you add a type parameter `P` to `IPanelStackProps` and use it in those methods? (You probably want to use it for the `initialPanel` too.)

#### Examples

The following code doesn't compile:

```tsx
function Menubar(props: MenubarProps) {
  const [panels, setPanels] = useState<IPanel<MenubarProps>[]>([ // the type of the state is invalid
    {
      component: Settings,
      props,
      title: "Settings",
    },
  ]);

  return (
    <PanelStack
      className="Menubar"
      initialPanel={panels[0]}
      onOpen={(new_) => setPanels([new_, ...panels])}
      onClose={() => setPanels(panels.slice(1))}
    />
  );
}
```

This is the error:

```
TypeScript error in /home/cecile/repos/third-i-frontend/src/Menubar.tsx(39,35):
Argument of type '(IPanel<MenubarProps> | IPanel<{}>)[]' is not assignable to parameter of type 'SetStateAction<IPanel<MenubarProps>[]>'.
  Type '(IPanel<MenubarProps> | IPanel<{}>)[]' is not assignable to type 'IPanel<MenubarProps>[]'.
    Type 'IPanel<MenubarProps> | IPanel<{}>' is not assignable to type 'IPanel<MenubarProps>'.
      Type 'IPanel<{}>' is not assignable to type 'IPanel<MenubarProps>'.
        Types of property 'component' are incompatible.
          Type 'ComponentType<IPanelProps>' is not assignable to type 'ComponentType<PhotoMode & Network & IPanelProps>'.
            Type 'ComponentClass<IPanelProps, any>' is not assignable to type 'ComponentType<PhotoMode & Network & IPanelProps>'.
              Type 'ComponentClass<IPanelProps, any>' is not assignable to type 'ComponentClass<PhotoMode & Network & IPanelProps, any>'.
[...]

PR #4346, opened 2020-09-28
[core] chore: enable strict null checks

adidahiya:

### DIFF
--- a/packages/core/src/components/panel-stack/panelView.tsx
+++ b/packages/core/src/components/panel-stack/panelView.tsx
@@ -26,13 +26,13 @@ export interface IPanelViewProps {
      * Callback invoked when the user presses the back button or a panel invokes
      * the `closePanel()` injected prop method.
      */
-    onClose: (removedPanel: IPanel) => void;
+    onClose: (removedPanel: IPanel<any>) => void;
 
     /**
      * Callback invoked when a panel invokes the `openPanel(panel)` injected
      * prop method.
      */
-    onOpen: (addedPanel: IPanel) => void;
+    onOpen: (addedPanel: IPanel<any>) => void;
 
     /** The panel to be displayed. */
     panel: IPanel;
--- a/packages/core/src/components/popover/popover.tsx
+++ b/packages/core/src/components/popover/popover.tsx
@@ -520,8 +520,8 @@ export class Popover extends AbstractPureComponent2<IPopoverProps, IPopoverState
         }
     };
 
-    private handleOverlayClose = (e: React.SyntheticEvent<HTMLElement>) => {
-        if (this.targetElement === null) {
+    private handleOverlayClose = (e?: React.SyntheticEvent<HTMLElement>) => {
+        if (this.targetElement === null || e === undefined) {
             return;
         }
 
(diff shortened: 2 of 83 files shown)


packages/core/src/components/panel-stack/panelView.tsx
```diff
@@ -26,13 +26,13 @@ export interface IPanelViewProps {
      * Callback invoked when the user presses the back button or a panel invokes
      * the `closePanel()` injected prop method.
      */
-    onClose: (removedPanel: IPanel) => void;
+    onClose: (removedPanel: IPanel<any>) => void;
```
adidahiya, line 29: related: https://github.com/palantir/blueprint/issues/4272

packages/core/src/components/popover/popover.tsx
```diff
@@ -549,18 +558,17 @@ export class Popover extends AbstractPureComponent2<IPopoverProps, IPopoverState
...
 
     private isArrowEnabled() {
-        const {
-            minimal,
-            modifiers: { arrow },
-        } = this.props;
+        const { minimal, modifiers } = this.props;
         // omitting `arrow` from `modifiers` uses Popper default, which does show an arrow.
-        return !minimal && (arrow == null || arrow.enabled);
+        return !minimal && modifiers?.arrow?.enabled;
```
adidahiya: this looks wrong -- the statements are not functionally equivalent. this can evaluate to `undefined`

packages/core/src/components/portal/portal.tsx
```diff
@@ -99,7 +104,11 @@ export class Portal extends React.Component<IPortalProps, IPortalState> {
 
     public componentDidUpdate(prevProps: IPortalProps) {
         // update className prop on portal DOM element
-        if (this.portalElement != null && prevProps.className !== this.props.className) {
+        if (
+            this.portalElement != null &&
+            prevProps.className !== undefined &&
```
adidahiya: this check might be overly narrow - if `prevProps.className === undefined`, we still might want to call `maybeAddClass()` in this block

packages/core/src/components/slider/handle.tsx
```diff
@@ -67,28 +66,24 @@ export class Handle extends AbstractPureComponent2<IInternalHandleProps, IHandle
...
     public render() {
-        const { className, disabled, label, min, tickSizeRatio, value, vertical } = this.props;
+        const { className, disabled, label } = this.props;
         const { isMoving } = this.state;
 
         // The handle midpoint of RangeSlider is actually shifted by a margin to
         // be on the edge of the visible handle element. Because the midpoint
         // calculation does not take this margin into account, we instead
         // measure the long side (which is equal to the short side plus the
         // margin).
```
adidahiya: this comment needs to be moved to the new helper function `getStyleProperties()`

packages/core/src/components/popover/popover.tsx
adidahiya: ✅

packages/core/src/components/portal/portal.tsx
adidahiya: ✅

packages/core/src/components/slider/handle.tsx
adidahiya: ✅

adidahiya:
clicked through most of the docs examples and everything seems to work fine 🎉

Issue #3833, opened 2019-11-11
[Table] prop Table.columnWidths breaks column resizing


zwerg44:
Hello.

Looks like setting initial column width with `Table.columnWidths` breaks column resizing starting from @blueprintjs/table@3.8.1.
Version 3.8.0 - seams working.
 
#### Environment

- __Package version(s)__: @blueprintjs/table@3.8.1, @blueprintjs/core@3.19.1
- __Browser and OS versions__: firefox 70.0.1, chromium 78.0.3904.70, ubuntu 16.04

Link to a minimal repro: https://codesandbox.io/s/blueprint-sandbox-l4l11
#### Steps to reproduce

1. Create component with defined `Table.columnWidths` and resizable columns.
1. Try to resize column with mouse.

#### Actual behavior

The layout looks broken like this
![image](https://user-images.githubusercontent.com/7299028/68590626-16adf880-04a0-11ea-9c81-143069d0b568.png)

#### Possible solution

Don't have solution, I rolled back to package previous version.

anurag060:
Did anyone find a work arround for this issue?
Hi @zwerg44 ... you got this fixed?

zwerg44:
Hi @anurag060 
No, I rolled back versions, I still at:
    "@blueprintjs/core": "3.18.1",
    "@blueprintjs/datetime": "3.13.0",
    "@blueprintjs/select": "3.10.0",
    "@blueprintjs/table": "3.7.1",

bheklilr:
I've also experienced this issue.  I attempted to work around it by keeping up with the column widths in a state variable and using the `onColumnWidthChanged` handler, but I happened to notice that when the table has to horizontally scroll I get a weird glitch on resize.  Any columns to the right of the one that was just resized are blanked out and completely white (pic attached).  The cells still exist and I can select them with a click.  As soon as I scroll horizontally it renders properly.  Sometimes it seems that react will just trigger a render anyway and it'll come back.

Glad to know there is a workaround to pin at 3.8.0 for now.

![image](https://user-images.githubusercontent.com/1035817/80138042-d0d67f80-8569-11ea-8ff5-bbecb61028f9.png)

bsdis:
I also have this issue. Any plan of fixing it?

adolfogarcia:
Was this issue ever fixed?

alecf:
I am seeing this in 3.8.17

aamirbutt:
[...]

Issue #3779, opened 2019-10-16
[Table] only 4 columns render on table mounted if table horizontal overflow


PeacePan:
#### Environment

- __Package version(s)__: @blueprintjs/table: 3.8.1, @blueprintjs/core: 3.19.1
- __Browser and OS versions__: Chrome 76.0.3809.100, MacOS 10.14.6

reproduce: https://codesandbox.io/s/blueprint-sandbox-4envh

#### Steps to reproduce

1. create a Blueprintjs Table
1. set many columns and let table overflow
1. default rowNum set to 0
1. setTimeout and update rowNum
1. only render 4 columns on table mounted

#### Actual behavior

![ezgif com-video-to-gif](https://user-images.githubusercontent.com/12136439/66899049-e607aa00-f02c-11e9-9637-2a15781a1c08.gif)

#### Expected behavior

table columns render expected.

#### Possible solution

The `viewportRect.left` when table mounted is `undefined`

then I scroll the table, the `viewportRect.left` will set to `number`

PeacePan:
Already fix via #3771 ?

alecf:
seeing this as well. Related?
 * https://github.com/palantir/blueprint/issues/3774
 * https://github.com/palantir/blueprint/issues/1712

ramirs:
+1 still happening with the following @blueprint deps

{
  "@blueprintjs/core": "^3.19.1",
  "@blueprintjs/icons": "^3.10.0",
  "@blueprintjs/select": "^3.11.0",
  "@blueprintjs/table": "^3.8.1",
}

If it helps, I recently migrated from these older versions of @blueprint, I rolled back to remove this rendering error and it's fine.
{
  "@blueprintjs/core": "3.15.1",
  "@blueprintjs/icons": "3.9.0",
  "@blueprintjs/select": "3.8.1",
  "@blueprintjs/table": "3.5.0",
}

wcyoyo44:
> +1 still happening with the following blueprint deps
> 
> {
> "@blueprintjs/core": "^3.19.1",
> "@blueprintjs/icons": "^3.10.0",
> "@blueprintjs/select": "^3.11.0",
> "@blueprintjs/table": "^3.8.1",
> }
> 
> If it helps, I recently migrated from these older versions of blueprint, I rolled back to remove this rendering error and it's fine.
> {
> "@blueprintjs/core": "3.15.1",
> "@blueprintjs/icons": "3.9.0",
[...]

PR #4356, opened 2020-10-03
[table] refactor: replace force-hardware-acceleration mixin with "will-change: transform"

### LINKED ISSUE #859: Delete force-hardware-acceleration() mixin when Chrome will-change bug fixed
Reminder to move to `will-change: transform` and delete our the `force-hardware-acceleration` mixin once Chrome's blurry-text bug is fixed:

- **Demo:** https://www.youtube.com/watch?v=iSvUlSpIbNk
- **Chome bug:** https://bugs.chromium.org/p/chromium/issues/detail?id=596382

As of this writing, the mixin is used in `table.scss` to improve scrolling performance on big `Table`s.

cemreyavuz:
#### Checklist

- [ ] Includes tests
- [ ] Update documentation

Delete force-hardware-acceleration mixin and replace the places that this mixin is used with `will-change: transform`

In the mentioned issue, it is said that this mixin is only used in `table.scss`. However, it turned out that it is also used in `_quadrants.scss`. Changes on that file should also be reviewed.

#### Screenshot

Because it is a refactor, there isn't any screenshot.

### DIFF
--- a/packages/core/src/common/_mixins.scss
+++ b/packages/core/src/common/_mixins.scss
@@ -99,30 +99,8 @@ $pt-dark-intent-text-colors: (
   @return $number;
 }
 
-@mixin force-hardware-acceleration() {
-  // CSS animations are typically smoother when they run on the GPU. most browsers trigger GPU
-  // ("hardware") acceleration automatically when certain CSS rules are applied (like `transform`).
-  //
-  // the transform rule below has no visual effect, but it achieves the following:
-  //   - creates a new stacking context
-  //   - creates a new containing block for fixed-position descendants
-  //   - "tricks" the browser into using hardware acceleration
-  //
-  // `will-change: transform` does the same thing, but it suffers from a Chrome bug that could make
-  // the styled element blurry when transform'd (e.g. this happens when a Blueprint Table is in a
-  // Popover that animates open). we should move to `will-change` and delete this mixin once that
-  // bug is fixed (tracking in issue #859).
-  //
-  // see the following YouTube video and Chrome bug report for more:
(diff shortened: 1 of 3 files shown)


packages/core/src/common/_mixins.scss
```diff
@@ -99,30 +99,8 @@ $pt-dark-intent-text-colors: (
   @return $number;
 }
 
-@mixin force-hardware-acceleration() {
```
adidahiya, line 102: this is a slight break for any Blueprint consumer who's reaching in to import `@blueprintjs/core/src/common/mixins` in Sass code, but that's not really part of our public API and a search across GitHub reveals that the break won't have any real impact, so I think it's ok.

adidahiya (approved):

alecf:
@adidahiya I'm glad to see the use of standard css, but one effect this has is that those of us who are still using version 3.7.x of `@blueprintjs/table` (due to issues like https://github.com/palantir/blueprint/issues/3833 and https://github.com/palantir/blueprint/issues/3779 ) can no longer upgrade `@blueprintjs/core`! Any chance you would release a 3.7.2 of `@blueprintjs/table` or give a timeline on fixing some of those table bugs?

adidahiya:
@alecf is this a critical CSS fix for you? I can give you a rough timeline of EOY (end of year) for the table 3.8.x issues milestone

MartinDawson:
@adidahiya Hi adidahiya, is their an timeline on the https://github.com/palantir/blueprint/issues/3833 fix? This seems to be a problem for quite a few people.

Thanks for all the great work!

PR #4358, opened 2020-10-06
feat: new icon - layer outline

MichalPrzybys-TomTom:
Changes proposed in this pull request:

New icon: layer outline


palantirtech:
Thanks for your interest in [palantir/blueprint](https://github.com/palantir/blueprint), @MichalPrzybys-TomTom! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

adidahiya:
Hi @MichalPrzybys-TomTom, as a general rule we don't take icon requests from external contributors, but this one _might_ fit into our design system since it's pretty close to the other layer icons. @aycai what do you think?

MichalPrzybys-TomTom:
> Hi @MichalPrzybys-TomTom, as a general rule we don't take icon requests from external contributors, but this one _might_ fit into our design system since it's pretty close to the other layer icons. @aycai what do you think?

There are several tools using Blueprint that I'm working on and there's high probability that there will be more icons, matching style with the ones you provide. Should I bother with PRs in the future?

aycai:
First, @MichalPrzybys-TomTom thanks for the submission!

@adidahiya I think it's reasonable to accept external icon submissions, but they will go through the same review process, which means possibly needing to expose some form of guidelines to help streamline the process. I'd also need to port resources internally, which won't take too much time but will be an added step. To that end I think external submissions are valid, but will be slow and on a case-by-case basis for now.

@MichalPrzybys-TomTom As for reviewing the icons themselves:
- The premise seems reasonable; a layer outline as a concept is pretty straightforward.
- +1 for keeping the outline the same as the original layer icon.
- Generally, default line thickness should be 2px, which I understand can get tricky when the lines are curved/diagonal. Here the thickness is not determined, but the corners were on the left and right sides and the thickness ended up being around 1px. Which this might look alright when zoomed in while looking at the icon, it quickly looks inconsistent when placed next to other icons as too thin (especially with a simple shape as the layer outline).
- I generally avoid hard-drawing vector points unless I absolutely have to e.g. optically set points. What I try to do first is draw the line, then use thickness and convert that resulting shape to an outline. I noticed (16px screenshot for reference) that there are some wonky vector points in the corners, likely from drawing the outlines manually.
Screen Shot 2020-10-09 at 3 41 40 PM

- A trick I use when going from line to thickness to outlines is redrawing the vector to be the center of the line thickness, in this case 14x8 pixels. Then when I add the 2px thickness it pads out to the desired outer pixel size (in this case 16x10 pixels). You'll notice that it matches the outline of the original layer icon as well (I suspect this was how it was drawn, but I didn't originally design the icon).
Screen Shot 2020-10-09 at 3 51 46 PM

- In the suggestion above, I think the lack of round corners on the inside edges is acceptable - this is a 16px icon, and an acute corner on an outline 10px wide will be minimally noticeable compared to the cornering on the outside edges (added a rasterized screenshot for reference):
Screen Shot 2020-10-09 at 4 04 58 PM


Let me know what you think, and if the suggestions make sense! Happy to take a look at the additional icons, and hope the guidance here can apply there as well.

packages/icons/resources/icons/icons.json
```diff
@@ -3538,5 +3538,11 @@
...
         "tags": "array, timestamp",
         "group": "miscellaneous"
-    }
+    },
+    {
+        "displayName": "Layer Outline",
+        "iconName": "layer-outline",
+        "tags": "zone, level, outline",
+        "group": "interface"
+    },
```
adidahiya: build is failing due to this trailing comma, which is not permitted in JSON at the end of a list (yes, it's annoying, since JS _does_ permit it 🤷‍♂️ )

MichalPrzybys-TomTom:
@aycai I've redone icons based on your suggestions. 
16px: layer-outline
20px: layer-outline-20

adidahiya:
@MichalPrzybys-TomTom it looks like you have enabled CircleCI on your fork of the blueprint repo. Can you try disabling the project on app.circleci.com, closing this PR, and opening a new one so that it builds in the palantir CI pipeline? See [CONTRIBUTING.md](https://github.com/palantir/blueprint/blob/develop/CONTRIBUTING.md)

PR #4363, opened 2020-10-07
[select] fix: force Select to respect query prop

### LINKED ISSUE #4362: [select] Select component doesn't respect the query prop
#### Environment

- __Package version(s)__: @blueprintjs/select (3.14.2)
- __Operating System__: Windows Version 10.0.18363 Build 18363
- __Browser name and version__: Google Chrome 85.0.4183.121

#### Code Sandbox

https://codesandbox.io/s/select-component-doesnt-respect-query-prop-dlih7

#### Steps to reproduce

1. Import the Select component from @blueprintjs/select and render
2. Pass the query prop to a Select component
3. Click the rendered Select component and update the text input

#### Actual behavior

Text input is updated.

#### Expected behavior

As mentioned in the documentation of Select component, the expected behavior is that the text input is not changed and controlled via query prop.

> **inputProps**
> Props to spread to the query InputGroup. Use query and onQueryChange instead of inputProps.value and inputProps.onChange to control this input.

#### Possible solution

Updating line 134 in `select.tsx` from:
```
value={listProps.query}
```

to this:
```
value={this.props.query || listProps.query}
```

solves the problem. It seems like the query prop is never directly passed to the input group. QueryList component is the one who controls the query of InputGroup in Select component. It initializes its internal query state with the query prop, but from that point onward, it doesn't respect the query prop of Select component and send its internal state to InputGroup.

cemreyavuz:
#### Checklist

- [ ] Includes tests
- [ ] Update documentation

As mentioned in the documentation of Select component, query prop can be used to control the InputGroup in Select component:

> **inputProps**
> Props to spread to the query InputGroup. Use query and onQueryChange instead of inputProps.value and inputProps.onChange to control this input.

However, as the code sandbox I provided in #4362, query prop acts like an initial value to it and doesn't control the InputGroup. This is because the one that controls the InputGroup is actually QueryList component. It initializes its internal state with the query prop that is passed to Select component. However, from that point onward, it keeps its internal query state and passes that internal state to the InputGroup.

What I have changed is that no matter what the internal query state of QueryList, the query prop of Select component is passed to the InputGroup if it is provided.

### DIFF
--- a/packages/select/src/components/select/select.tsx
+++ b/packages/select/src/components/select/select.tsx
@@ -131,7 +131,7 @@ export class Select<T> extends AbstractPureComponent2<ISelectProps<T>, ISelectSt
                 {...inputProps}
                 inputRef={this.refHandlers.input}
                 onChange={listProps.handleQueryChange}
-                value={listProps.query}
+                value={this.props.query || listProps.query}
             />
         );
 


adidahiya:
Discussion in https://github.com/palantir/blueprint/issues/4362

PR #4364, opened 2020-10-07
[timezone] fix: pass query and onQueryChange props to override Select's own behavior

### LINKED ISSUE #4353: [TimezonePicker] 'inputProps' does not override search behavior.
#### Steps to reproduce
```javascript
        <TimezonePicker
          ...
          inputProps={{
            value: this.searchQuery,
            onChange: this.handlerWhichShouldBeCalledButItIsnt
          }}
        >
```
#### Actual behavior

Callback for search input change is not called.

#### Expected behavior

According to docs callback for search input change should override Select's behavior and should be called.

Screenshot 2020-10-01 at 00 13 34

#### Possible solution

The `onQueryChange` prop which you pass to `Select` in `timezonePicker.tsx` is just:
```javascript
private handleQueryChange = (query: string) => this.setState({ query });
```
No mentions about `inputProps.onChange`.

And as I understand in `Select` `onQueryChange` has some priority over `inputProps.onChange`.
Thus my callback which I pass to `TimezonePicker` is not called.

But it's just an assumption.

Please let me know if I do something wrong...

cemreyavuz:
#### Checklist

- [ ] Includes tests
- [ ] Update documentation

As mentioned in the documentation of Select component, query and onQueryChanges props should be used instead of inputProps.value and inputProps.onChange to control the InputGroup in Select component:
> **inputProps**
> Props to spread to the query InputGroup. Use query and onQueryChange instead of inputProps.value and inputProps.onChange to control this input.

Because current implementation is directly passing inputProps to Select component, Select component doesn't use these props for controlling the InputGroup. Instead it uses the QueryList's internal query state. What I have done with this commit is that now TimezonePicker passes inputProps.value as query to Select component and calls inputProps.onChange when ChangeEvent is fired from Select component.

As I mentioned in #4362, actually Select component has a bug as well. Currently it doesn't respect the query prop and I opened another PR for that issue as well: #4363. After #4362 is resolved, this implementation of TimezonePicker will be working fine. However, if this PR is merged before #4362 is resolved, only inputProps.onChange will be working and inputProps.value will continue to be ignored.

### DIFF
--- a/packages/timezone/src/components/timezone-picker/timezonePicker.tsx
+++ b/packages/timezone/src/components/timezone-picker/timezonePicker.tsx
@@ -163,6 +163,7 @@ export class TimezonePicker extends AbstractPureComponent2<ITimezonePickerProps,
                 inputProps={finalInputProps}
                 disabled={disabled}
                 onQueryChange={this.handleQueryChange}
+                query={this.props.inputProps.value}
             >
                 {children != null ? children : this.renderButton()}
             </TypedSelect>
@@ -224,5 +225,10 @@ export class TimezonePicker extends AbstractPureComponent2<ITimezonePickerProps,
 
     private handleItemSelect = (timezone: ITimezoneItem) => this.props.onChange?.(timezone.timezone);
 
-    private handleQueryChange = (query: string) => this.setState({ query });
+    private handleQueryChange = (query: string, event: React.ChangeEvent<HTMLInputElement>) => {
+        if (this.props.inputProps.onChange) {
+            this.props.inputProps.onChange(event);
+        }
+        this.setState({ query });
+    };
 }


adidahiya:
On second thought, this API feels weird diverging from Select. Since a TimezonePicker is a Select under the hood, we should try to make it feel more like a Select in its query control behavior. Something like:

```ts
export interface ITimezonePickerProps extends IProps, Pick<IListItemsProps<ITimezoneItem>, "query" | "onQueryChange"> {
    ...
}
```

Then we'd have to make sure `props.query` could override `state.query` inside the component implementation.

This would technically be a a slight "break", but since `inputProps.value` and `inputProps.onChange` never actually worked, it's an acceptable break.

PR #4374, opened 2020-10-19
Feature request - tooltip minimal prop and flexibility over arrow

jamesgiu:
#### Checklist

- [x] Includes tests
- [x] Update documentation (auto-generated)

Flexibility over rendering the arrow was a use case required for our own project - however, I also noticed that Popover's minimal prop exhibited this same behaviour so I thought I'd keep it consistent. 

If `minimal` isn't quite right, I'd be happy to change this PR to just be a prop that controls the rendering of the arrow, as that was of importance to us in our use case. 

Whether `minimal` is necessary for tooltip given one can pass in Classes.MINIMAL via popoverProps. If not, propose prop solely controls rendering of the arrow.

#### Screenshot

image

Thank you for the consideration! Big fan of BlueprintJS

### DIFF
--- a/packages/core/src/components/popover/popoverSharedProps.ts
+++ b/packages/core/src/components/popover/popoverSharedProps.ts
@@ -96,6 +96,14 @@ export interface IPopoverSharedProps extends IOverlayableProps, IProps {
     isOpen?: boolean;
 
     /**
+     * Whether to apply minimal styles to this tooltip, which includes removing
+     * the arrow and adding `Classes.MINIMAL` to minimize and accelerate the
+     * transitions.
+     * @default false
+     */
+    minimal?: boolean;
+
+    /**
      * Popper modifier options, passed directly to internal Popper instance. See
      * https://popper.js.org/docs/modifiers/ for complete
      * details.
--- a/packages/core/src/components/popover/popoverSharedProps.ts
+++ b/packages/core/src/components/popover/popoverSharedProps.ts
@@ -96,9 +96,8 @@ export interface IPopoverSharedProps extends IOverlayableProps, IProps {
     isOpen?: boolean;
 
     /**
-     * Whether to apply minimal styles to this tooltip, which includes removing
-     * the arrow and adding `Classes.MINIMAL` to minimize and accelerate the
-     * transitions.
+     * Whether to apply minimal styling to this popover or tooltip. Minimal popovers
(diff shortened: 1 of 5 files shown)


packages/core/src/components/tooltip/tooltip.tsx
```diff
@@ -61,6 +61,14 @@ export interface ITooltipProps extends IPopoverSharedProps, IIntentProps {
...
      */
     transitionDuration?: number;
+
+    /**
+     * Whether to apply minimal styles to this tooltip, which includes removing
+     * the arrow and adding `Classes.MINIMAL` to minimize and accelerate the
+     * transitions.
+     * @default false
+     */
+    minimal?: boolean;
```
adidahiya: this should be moved from `IPopoverProps` in `popover.tsx` to `IPopoverSharedProps` in `popoverSharedProps.ts`. Then you don't have to add it here.

adidahiya (changes requested):
In general, I am ready to add this feature to Tooltip. I think we've heard of enough use cases for it that it makes sense.

Can you please update the docs example in `tooltipExample.tsx` to demonstrate usage of this new feature?

packages/core/src/components/tooltip/tooltip.tsx
jamesgiu: Hey @adidahiya - thanks for your feedback! Commit [ab11371fb3fe26d67c63f6e0934a869d023689b0](https://github.com/palantir/blueprint/pull/4374/commits/ab11371fb3fe26d67c63f6e0934a869d023689b0) addresses your suggestions.

packages/core/src/components/popover/popoverSharedProps.ts
```diff
@@ -95,6 +95,14 @@ export interface IPopoverSharedProps extends IOverlayableProps, IProps {
      */
     isOpen?: boolean;
 
+    /**
+     * Whether to apply minimal styles to this tooltip, which includes removing
+     * the arrow and adding `Classes.MINIMAL` to minimize and accelerate the
+     * transitions.
```
adidahiya:
```suggestion
     * Whether to apply minimal styling to this popover or tooltip. Minimal popovers
     * do not have an arrow pointing to their target and use a subtler animation.
```

adidahiya (changes requested):
minor docs language suggestion

adidahiya (approved):

Issue #4298, opened 2020-08-29
Version 3.31.0 breaks InputGroup with redux form


tgreen7:
#### Environment

- __Package version(s)__: 3.31.0

#### Code Sandbox

working on 3.30.1
https://codesandbox.io/s/blueprint-sandbox-forked-n039t?file=/src/index.tsx

![working](https://user-images.githubusercontent.com/10415101/91646196-88403e80-ea01-11ea-96fe-a74dab8eb468.gif)

broken on 3.31.0
https://codesandbox.io/s/blueprint-sandbox-forked-327mz?file=/src/index.tsx

![broken](https://user-images.githubusercontent.com/10415101/91646197-8b3b2f00-ea01-11ea-917e-e65048b93f24.gif)

#### Steps to reproduce

1. Type in the input field in the sandbox
2. Move cursor to beginning of the input and type again. Cursor jumps to the end of the input on version 3.31.0

#### Possible solution

Version 3.31.0 added the asyncControllableInput

tnrich:
Ya it seems like https://github.com/palantir/blueprint/pull/4266 changed how async updates fired. This is definitely an issue for us

adidahiya:
@tnrich are you also using `redux-form`? curious if you are able to demonstrate a repro without using any external libraries

tnrich:
@adidahiya yes also using redux-form. I haven't tried to replicate outside of a redux-form context

adidahiya:
It looks like these issues: https://github.com/facebook/react/issues/955 and https://github.com/facebook/react/issues/14904. I'm going to try the workaround in https://github.com/facebook/react/issues/14904#issuecomment-522213842:

> What I've done to workaround is to make the <input> uncontrolled (change value to defaultValue) with a onChange prop.

tnrich:
Thanks @adidahiya hopefully that works! Let us know when we can test out the new version.

adidahiya:
Update: using `defaultValue` all the time doesn't work, it is incompatible with a controlled API for InputGroup / NumericInput. Investigating other approaches.

tnrich:
Thanks @adidahiya let us know how it goes!

adidahiya:
[...]

PR #4383, opened 2020-10-21
[core] fix(InputGroup): make async control optional

### LINKED ISSUE #4375: [InputGroup] Displayed value does not match value of 'value' prop
#### Environment

- __Package version(s)__: @blueprintjs/core: 3.34.0
- __Operating System__: Mac
- __Browser name and version__: Chrome latest

#### Steps to reproduce

My onChange function is designed to limit the entered characters to only 10 numbers:

```js
<InputGroup
    disabled={disabled}
    leftIcon={leftIcon}
    placeholder={placeholder}
    rightElement={rightElement}
    type={type}
    name={name}
    id={id}
    onChange={(e) => {
        e.target.value = e.target.value.replace(/[^0-9]/g, '').substr(0, 10);
        onChange(e);
    }}
    onBlur={onBlur}
    value={value}
/>
```

#### Actual behavior

Any non-number characters and characters after length 10 are properly not stored in 'value' but still display in the input field.

#### Expected behavior

The displayed text in the input field should match the value of the 'value' prop. This works in @blueprintjs/core 3.33.0.

adidahiya:
#### Checklist

- [x] Includes tests
- [x] Update documentation

Using the suggestion from https://github.com/palantir/blueprint/issues/4298#issuecomment-697604936, this change makes "async control" of InputGroup _optional_, after it was enabled across the board by default in #4266 (released in @blueprintjs/core@3.31.0).

I originally wanted to support all possible `<input>` use cases out-of-the-box by default with good UX for both synchronous `value` updates and async ones (see https://github.com/palantir/blueprint/pull/4262, https://github.com/palantir/blueprint/issues/4298). This turned out to be too much of a hassle, so I am now opting to make the complex async state handling behavior opt-in using a new prop called `asyncControl`.

`asyncControl={true}` should be used in places where applications expect that the state update to a controlled `<InputGroup>` will be asynchronous, either by explicit design in their state management logic, or through a library like redux-form. I think it makes sense to hide this complexity (and potential source of bugs) behind a flag; see the docs about this:

> This is not broadly encouraged (a value returned from `onChange` should be sent back to the component as a controlled `value` synchronously)

Fix the linked issue, and ensure no regression

### DIFF
--- a/packages/core/src/components/forms/asyncControllableInput.tsx
+++ b/packages/core/src/components/forms/asyncControllableInput.tsx
@@ -109,7 +109,7 @@ export class AsyncControllableInput extends React.PureComponent<
     }
 
     public render() {
-        const { isComposing, hasPendingUpdate: pendingUpdate, value, nextValue } = this.state;
+        const { isComposing, hasPendingUpdate, value, nextValue } = this.state;
         const { inputRef, ...restProps } = this.props;
         return (
             <input
@@ -118,7 +118,7 @@ export class AsyncControllableInput extends React.PureComponent<
                 // render the pending value even if it is not confirmed by a parent's async controlled update
                 // so that the cursor does not jump to the end of input as reported in
                 // https://github.com/palantir/blueprint/issues/4298
-                value={isComposing || pendingUpdate ? nextValue : value}
+                value={isComposing || hasPendingUpdate ? nextValue : value}
                 onCompositionStart={this.handleCompositionStart}
                 onCompositionEnd={this.handleCompositionEnd}
                 onChange={this.handleChange}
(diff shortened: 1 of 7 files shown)


johnwiseheart (approved):
Looks reasonable, nothing jumps out

styu:
technically this is a break for folks that were relying on the new async behavior, but now have to add the new prop to turn it on. what do we want to do about that?

adidahiya:
@styu yeah it is an API break for anyone who has upgraded to the range `v3.31.0–3.34.0`. I think that's ok since relying on async updates is kind of an edge case for this component, and I'd rather make the default / majority use case work properly.

@tgreen7 @tnrich @ejose19 heads up this API break is likely coming in the next minor release

adidahiya:
heads up @rynorris as well

Issue #4149, opened 2020-05-24
Overlay uses legacy context API through react-transition-group


tanveersingh926:
While using the Dialog, I am getting below warning in the console.

```
index.js:1 Warning: Legacy context API has been detected within a strict-mode tree.

The old API will be supported in all 16.x releases, but applications using it should migrate to the new version.

Please update the following components: TransitionGroup

Learn more about this warning here: https://fb.me/react-legacy-context
    in TransitionGroup (created by Blueprint3.Overlay)
    in Blueprint3.Overlay (created by Blueprint3.Dialog)
    in Blueprint3.Dialog (at Dialog.tsx:16)
    in Dialog (at ViewFormulationDialog.tsx:24)
    in ViewFormulationDialog (at SearchResults.tsx:65)
    in SearchResults (at SearchFormulation.tsx:42)
    in div (created by Blueprint3.Card)
    in Blueprint3.Card (at SearchFormulation.tsx:17)
    in SearchFormulation (at Formulation.tsx:21)
    in Formulation (at AppRouter.tsx:20)
    in div (at Navbar.tsx:63)
    in _Navbar (at AppRouter.tsx:19)
    in Route (at AppRouter.tsx:15)
    in Switch (at AppRouter.tsx:14)
    in Router (at AppRouter.tsx:13)
    in AppRouter (at App.tsx:7)
    in App (at src/index.tsx:9)
    in StrictMode (at src/index.tsx:8) 

```

adidahiya:
I don't think this will be resolved until the next major version of Blueprint, since react-transition-group 3.0 requires react ^16: https://github.com/reactjs/react-transition-group/blob/master/CHANGELOG.md#300-2019-04-15

The next major version of Blueprint will work with v16 and v17, at which point we can upgrade to the latest react-transition-group.

tanveersingh926:
Thanks, @adidahiya. Do you have any idea about the release date?

aymenoppa:
when you'll fix it? is there any workaround except using blueprint CSS

adidahiya:
I hope to start work on the next major version (Blueprint v4.0) before the end of the year, but no guarantees it will ship before 2021. React 17 isn't even out yet...

adidahiya:
fixed by #4545, released in 4.0.0-alpha.0

adidahiya:
Not fixed in v5.0, we still need to migrate to using the `nodeRef` feature of react-transition-group.

adidahiya:
[...]

Issue #3979, opened 2020-02-26
Removing findDOMNode


mormahr:
**Update**:
I think my analysis about how this can be solved is wrong. See my [comment](https://github.com/palantir/blueprint/issues/3979#issuecomment-705573953).

---

As part of the move to strict and concurrent mode support, I'd like to tackle removing usages of `findDOMNode`. I analyzed the scope of changes needed in the first-party components and possible implications on third-party consumers.

### Refs (preferred)
Refs are probably the standard way to get access to an HTML node. Instead of explicitly wiring refs through the consuming component, they can often be set through `cloneElement`.
**Downsides:**
- Can reference JSX elements instead of DOM nodes

**Workarounds:**
- Add a warning in the documentation and print a runtime warning
- Some components already only allow a single DOM node as a child (e.g. `ResizeSensor`)
  - Can clone the element and attach a ref that way

### Wrapper element
**Downsides:**
- Can cause problems with DOM nesting and layout
- Possible performance overhead

**Workarounds:**
- `display: contents;` (doesn't work all most usages)
- Provide `tagName`, etc. properties

### ResizeSensor
ResizeSensor uses `findDOMNode` to attach a ResizeObserver to the child node. It requires a single DOM node child and can use `cloneElement` to attach a ref. I don't know about the performance implications of `cloneElement` (resize sensor explicitly mentions this in the source comments), but `ContextMenuTarget` also uses `cloneElement`. My understanding is that `cloneElement` is pretty cheap. I don't even think it's slower than `findDOMNode`, so I'd say `cloneElement` and attaching a ref is the way to go here.

### ContextMenuTarget
The element returned by `findDOMNode` is used to check if dark mode is enabled or not, by calling a helper method. `ContextMenuTarget` already requires a DOM node and uses `cloneElement` to attach the event-listener so that we can attach a ref in the same call.

### Overlay
If I understand this correctly `TransitionGroup` creates a container by instantiating the component passed to it. As `Overlay` requests a `div`, this `div` is the element we want a reference to. I currently don't know how to get a ref to this `div`. Afaik `TransitionGroup` doesn't expose a reference. This needs some further investigation.

## Conclusion
Without considering `Overlay`, I'd say this is doable without breaking changes. I would split this into three separate tasks.

1. Using `cloneElement` in `ContextMenuTarget` to attach a ref is a non-breaking change. 
2. Using `cloneElement` in `ResizeSensor` to attach a ref is a non-breaking change.
[...]

Issue #4023, opened 2020-03-24
Upgrade to Popper.js v2.0


adidahiya:
[v2 docs](https://popper.js.org/docs/v2/)

LoiKos:
You may consider using [React-Popper v2](https://popper.js.org/react-popper/) i think it would be easier for you

adidahiya:
We're [already using](https://github.com/palantir/blueprint/blob/f896892084105939d5ee4af6fd1c3c4e27608625/packages/core/src/components/popover/popover.tsx#L21) react-popper.

alex-kinokon:
Will this be in 4.0?

adidahiya:
@proteriax yes, likely, but I don't have a concrete timeline for 4.0 yet

justinbhopper:
@adidahiya ~Are there plans to transition @blueprintjs/core and @blueprintjs/select to use Popover.js v2.0?  I would have thought the v4.0 milestone would be a good opportunity to make the switch, since it would undoubtedly be a breaking change to the API.~

Ignore this comment.  I see this is exactly what has been done on the v4 branch.

PR #4403, opened 2020-11-09
feat: allow React 17 peer dependency

### LINKED ISSUE #4391: React 17 support
React 17 released already, but peerDependencies contains `"react": "^15.3.0 || 16"`, so there are npm warnings after React is updated

adidahiya:

Allow React 17 as an NPM peer dependency for all packages.

Despite Blueprint's usage of some legacy APIs (https://github.com/palantir/blueprint/issues/4149, https://github.com/palantir/blueprint/issues/3979, etc), React 17's [breaking changes](https://reactjs.org/blog/2020/08/10/react-v17-rc.html#other-breaking-changes) don't seem to affect our components, so I think it's ok to expand the allowed version ranges.

This PR does not change the repo to build with React 17 or test it out in CI workflows. If there are issues with React 17 compatibility in the future, we can flag those issues and work on resolving them incrementally.

#### Screenshot

Tested it out locally with a Yarn resolution set to `17.0.1`:

image

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -59,8 +59,8 @@
         "tslib": "~1.13.0"
     },
     "peerDependencies": {
-        "react": "^15.3.0 || 16",
-        "react-dom": "^15.3.0 || 16"
+        "react": "^15.3.0 || 16 || 17",
+        "react-dom": "^15.3.0 || 16 || 17"
     },
     "devDependencies": {
         "@blueprintjs/karma-build-scripts": "^1.0.3",
--- a/packages/datetime/package.json
+++ b/packages/datetime/package.json
@@ -42,8 +42,8 @@
         "tslib": "~1.13.0"
     },
     "peerDependencies": {
-        "react": "^15.3.0 || 16",
-        "react-dom": "^15.3.0 || 16"
+        "react": "^15.3.0 || 16 || 17",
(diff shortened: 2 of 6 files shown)


jasongrout:
FYI, there still is a dependency on react-popper ^1:

https://github.com/palantir/blueprint/blob/241b20d5b45a62ef90c91d795f78d429adbda47f/packages/core/package.json#L56

react-popper 1 has a peer dependency on react 16, so installing @blueprintjs/core still ends up with a warning for react 17 from it.

react-popper ^2 now has updated to have a peer dependency allowing react 17: https://github.com/popperjs/react-popper/pull/390

jasongrout:
> react-popper ^2 now has updated to have a peer dependency allowing react 17: [popperjs/react-popper#390](https://github.com/popperjs/react-popper/pull/390)

It looks like https://github.com/palantir/blueprint/pull/4488 upgrades to react-popper 2 (fixing https://github.com/palantir/blueprint/issues/4023). Unfortunately, it seems that change will be released in an upcoming release, and doesn't fix the react 17 peer dependency compatibility problem with the current blueprint.

For those running into npm install errors with npm v7, installing with `npm install --legacy-peer-deps` lets the install proceed instead of erroring out.

PR #4414, opened 2020-11-11
[core] fix(Tag): restore 30px height for removable tags

### LINKED ISSUE #4397: Large tags have unusually large padding
#### Environment

- __Package version(s)__: 3.31.0
- __Operating System__: irrelevant
- __Browser name and version__: same

#### Code Sandbox

https://blueprintjs.com/docs/#core/components/tag (enable the "Large" switch)

#### Steps to reproduce

1. Render a `<Tag large={true} />`

#### Actual behavior

The tag has a height of 36px.

#### Expected behavior

The tag has a height of 30px.

Most inline UI components have a height divisible by `$pt-grid-size` (10px). Regular tags are 20px high, buttons and menu items are and large tags used to be 30px, large buttons are 40px, etc.
Large tags used to have a very similar layout to regular buttons, allowing trivial side-by-side use.
With the changes in #4099 the added vertical padding makes large tags 36px high, which (1) looks odd, there's an unusual amount of vertical padding, and (2) removes the large tag from the divisible-by-10px sizes of most comparable components, making them largely unusable alongside those components.

#### Possible solution

Revert #4099, or make that padding explicitly opt-in.

cc @kmblake

adidahiya:
#### Fixes #4397

Regression was introduced by #4099

### DIFF
--- a/packages/core/src/components/tag/_common.scss
+++ b/packages/core/src/components/tag/_common.scss
@@ -192,8 +192,7 @@ $tag-round-adjustment: 2px !default;
   .#{$ns}-large & {
     /* stylelint-disable-next-line declaration-no-important */
     margin-right: -$tag-padding-large !important;
-    padding: ($tag-height-large - $pt-icon-size-large) / 2;
-    padding-left: 0;
+    padding: 0 ($tag-padding-large / 2) 0 0;
 
     &:empty::before {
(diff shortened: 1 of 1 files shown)


invliD (approved):
Oh sweet! I hadn't noticed this only happens for removable tags, but those are indeed what I saw affected.

PR #4416, opened 2020-11-11
[icons] feat: add 'font-display: swap' to font-face mixing

ejose19:

According to https://web.dev/avoid-invisible-text/ this PR makes UX better since text display is not delayed until the font is downloaded.


adidahiya:
Why would we want to display a fallback font for icons? That doesn't really make sense to me. What would that look like? Can you show screenshots/videos/gifs demonstrating the difference and why this is an improvement?

adidahiya:
Ping @ejose19 on the above question ^

ejose19:
@adidahiya Mostly to comply with the Core Web Vitals tools, if you have a page with mixed icons and texts, you will get this warning on Lighthouse

https://webdev.imgix.net/font-display/font-display.png

And as in my first comment, here's the source: https://web.dev/font-display/

Those tools are a big part of the ecosystem, even CRA added webvitals to the bundle, and I don't think we want to generate errors on those tools to consumers of this library.

Regarding on how I noticed the issue, on previous versions on firefox & chromium, I would get an invisible text (not icons) until the font was loaded (had to use the throttle in devtools to delay the font loading), and when manually patching the ```font-display: swap``` the issue was no longer present. However currently I tried to reproduce with and without the patch and couldn't see a difference, so I don't know if the behavior regarding font loading was changed.

adidahiya:
Yeah, I checked out that link, and I agree that it's not great to generate errors in webvitals. But I'm still not convinced.

Here's what icon font loading looks like with your change on a throttled connection (ignore the long delay at the beginning of the capture):



Here's what it looks like on blueprintjs.com:



I don't see an improvement in the first gif.

I keep coming back to the _semantics_ of this change: what are we actually displaying as a fallback? _There is no meaningful content to display when the icon font is not loaded_. This is probably true for all icon fonts (regular fonts, of course, are a different story).

Instead of `font-display: swap`, we might want to consider `font-display: block` instead? Since the font is crucial to display the icon.

ejose19:
@adidahiya Can you tell me what changes you did exactly on the repo to test the behavior? I would like to do some testing regarding the combinations and error messages.

adidahiya:
I didn't edit any code, I just went to the PR build preview for this branch [here](https://36705-71939872-gh.circle-artifacts.com/0/packages/docs-app/dist/index.html)

you can access that build on any PR in this repo by going to the "deploy-preview" job

image

... and clicking docs-app/dist/index.html in the Artifacts tab

image

I compared that with the latest documentation on blueprintjs.com.

ejose19:
@adidahiya Ah, thought you were testing locally. Yes I was aware of that but rather want to test it locally so I can see the differences between font-display options.

I was able to test the differences between ```font-display: swap``` and ```font-display: block```, both doesn't show any errors on lighthouse as opposed to not specifying a ```font-display``` property. However the difference is that ```block``` shows an invisible icon while ```swap``` shows the ```not found``` character (the box).

While the ```block``` seems more clean, in case the font fails to load the user won't have any visible cue that there's an icon, while the ```swap``` showing the box at least let the user know that something is there (which accompanied with a proper html title, at least the user would know what that icon does)

Ultimately it's mostly taking a decision regarding using ```swap``` which would act like it does right now but without triggering errors, or changing to ```block``` with the considerations above.

adidahiya:
@ejose19 I'm willing to accept this change for v4, where I have refactored the icons build pipeline. I changed the target branch for your PR. Can you please merge `origin/v4` and resolve conflicts?

ejose19:
@adidahiya Where I'm supposed to put that option now? The file in original PR is gone in v4

adidahiya:
@ejose19 oops, my bad, you are right. The font-face declaration is now auto-generated by fantasticon, the template for it is defined here: https://github.com/tancredi/fantasticon/blob/master/templates/css.hbs. Can you ask what the fantasticon folks think about this change? We do have the option of using our own `.css.hbs` template to add the `font-display` property ourselves here, but I'd like to hear what they think first, if it's a good idea to upstream the enhancement.

adidahiya:
Closing for now due to unresolved discussion, but happy to re-open if you want to discuss again.

PR #4422, opened 2020-11-17
[core] feat(Drawer): return focus to active element after closing

### LINKED ISSUE #3592: Focus back to element when drawer close
#### Environment

- __Browser and OS versions__: Firefox, Arch Linux

#### Feature request

Focus return to calling element when opening drawer like element.

#### Examples

When the user clicks on the menu button the drawer opens and when the user presses the `esc`. the focus of the application when somewhere else rather than the menu button.
It can be better accessbility who use the keyboard.

adidahiya:
Fixes #3592

Difficult to test because `document.activeElement` doesn't seem to update as expected in enzyme...

### DIFF
--- a/packages/core/src/components/drawer/drawer.tsx
+++ b/packages/core/src/components/drawer/drawer.tsx
@@ -108,6 +108,8 @@ export class Drawer extends AbstractPureComponent2<IDrawerProps> {
     public static readonly SIZE_STANDARD = "50%";
     public static readonly SIZE_LARGE = "90%";
 
+    private lastActiveElementBeforeOpened: Element | null | undefined;
+
     public render() {
         // eslint-disable-next-line deprecation/deprecation
         const { size, style, position, vertical } = this.props;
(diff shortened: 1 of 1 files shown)


adidahiya:
works in Chrome & FF

Nantris:
Is some way to disable this available? We manage focus ourselves in our app, but this change prevents us from doing that, because it immediately undoes whatever we do and focuses the last `document.activeElement`.

Nantris:
@adidahiya is there any way to circumvent this? If you know one, it would really help us a lot!

adidahiya:
@Slapbox no, but we can add a prop which enables this behavior, something like `returnFocusOnClose` where it's set to `true` by default

adidahiya:
@Slapbox see https://github.com/palantir/blueprint/pull/4620

Nantris:
Wow that was crazy fast to have a PR opened! Thanks for your attention on this @adidahiya!

Issue #4032, opened 2020-04-01
Compilation issue with newer version of SASS


vkbansal:
When trying to import to blueprint using SASS, error is encountered.

#### Environment
- __Package version(s)__:
  - @blueprintjs/core: 3.24.0
  - webpack: 4.42.1
  - sass: 1.26.3
  - sass-loader: 8.0.2
- __Browser and OS versions__:  MacOS Catalina 10.15.4

#### Steps to reproduce

1. Create a SASS file and add `@import "~@blueprintjs/core/src/blueprint";`
2. Try compiling it using webpack.

#### Actual behavior

```
ModuleBuildError: Module build failed (from ./node_modules/sass-loader/dist/cjs.js):
SassError: compound selectors may no longer be extended.
Consider `@extend .bp3-menu-item, :hover` instead.
See http://bit.ly/ExtendCompound for details.

   ╷
19 │       @extend .#{$ns}-menu-item:hover;
   │               ^^^^^^^^^^^^^^^^^^^^^^^
   ╵
  node_modules/@blueprintjs/core/src/components/menu/_submenu.scss 19:15  root stylesheet
```

#### Expected behavior

There should be no error.

adidahiya:
From my quick googling it seems like you can work around this by using `node-sass` instead of `sass` (Dart-Sass). This feature has not been deprecated in `node-sass` yet. Longer term it looks like we'll need to refactor some of this Sass code.

vkbansal:
I tried with `node-sass` too. Same error.

DTrejo:
If you add a `, `, it compiles for `sass` + `parcel` at least.
Note that I've yet to actually look at my UI to confirm it looks correct.

```diff
-@extend .#{$ns}-menu-item:hover;
+@extend .#{$ns}-menu-item, :hover;
```

omeid:
node-sass is deprecated in favour of sass (dart-sass) in Node 14. So this needs addressing sooner than later.

nickensoul:
Hi everyone!

Just remind that `node-sass` is [officially deprecated now](https://twitter.com/sasscss/status/1320820104561524736?s=21). We're using `node-sass` for now in our project, but will glad to know any plans and news about Blueprint's migration to `dart-sass`.
Thanks!

stolenng-investing:
Hey, 
Having this issue also, any plans on resolving it ? i can try also ?

adidahiya:
[...]

PR #4429, opened 2020-11-19
[menu] - @extend menu-item:hover not working on dart-css

stolenng-investing:
Hey,
I'm not sure if what i did is correct, but if you could direct to me to right solution here ? i want to use `menu.scss` and it breaks me because of us using dart-css :)

I tested it and seems good:
image

Also seems all CI passes - related to https://github.com/palantir/blueprint/issues/4032

### DIFF
--- a/packages/core/src/components/menu/_submenu.scss
+++ b/packages/core/src/components/menu/_submenu.scss
@@ -17,6 +17,7 @@
       // keep a trail of hovered items as submenus are opened
       /* stylelint-disable-next-line scss/at-extend-no-missing-placeholder */
       @extend .#{$ns}-menu-item;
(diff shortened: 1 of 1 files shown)


palantirtech:
Thanks for your interest in [palantir/blueprint](https://github.com/palantir/blueprint), @stolenng-investing! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

adidahiya:
we're not compatible with dart-sass. I'd rather do a holistic change to update the sass compiler rather than small fixes like this. follow #4032 for updates on that.

PR #4438, opened 2020-11-30
refactor(all): normalize ref handling

ejose19:
#### Checklist

- [ ] Includes tests
- [X] Update documentation

- Normalize ref usage so components accept both ref object and callback, instead of just a callback. 
- Centralize the ref handling logic so it can be reused in all components.
- Use exact HTML element type (ie ```HTMLDivElement```) on ref typing (instead of ```HTMLElement```) when they're known.


packages/core/src/common/refs.ts
```diff
@@ -38,3 +38,32 @@ export function getRef<T = HTMLElement>(ref: T | IRefObject<T> | null) {
...
+    }
+}
+
+export function refHandler<T extends HTMLElement>(
+    refProp: IRef<T> | undefined | null,
+    refTargetParent: Record<string, any>,
+    refTargetKey: string,
+) {
+    if (!(refTargetKey in refTargetParent)) {
+        throw new Error(`${refTargetKey} does not exists in ${Object.keys(refTargetParent)}`);
```
adidahiya: I really don't like that we've moved an error from compilation time that is guarded by type checking to one that is thrown at runtime. That goes against the goals of using TypeScript. I would be open to another solution which has a small drawback of requiring all ref fields to be public:

```ts
export function refHandler<T extends HTMLElement, K extends string>(
    refProp: IRef<T> | undefined | null,
    refTargetParent: { [k in K]: T | IRefObject<T> | null },
    refTargetKey: K,
) {
    if (isRefObject<T>(refProp)) {
        refTargetParent[refTargetKey] = refProp;
        return refProp;
    }
    return (ref: T | null) => {
        refTargetParent[refTargetKey] = ref;
        if (isRefCallback(refProp)) {
            refProp(ref);
        }
    };
}
```

example usage in Checkbox:

```ts
    public input: HTMLInputElement | null = null;
    private handleInputRef: IRef<HTMLInputElement> = refHandler(this.props.inputRef, this, "input");
```

note the type annotation is required on `handleInputRef` because otherwise the inferred type has a cyclical reference which cannot be computed.

adidahiya (changes requested):

packages/core/src/common/refs.ts
ejose19: Yes, I originally wanted to do refTargetKey as ```keyof T``` but Typescript doesn't expose private properties on ```keyof```. Didn't made ref public as thought original implementation would be preferred (since users are not modifying the classes and the tests would fail if an invalid key was given), however I don't mind changing refs to public.

adidahiya: The developer experience concern (it's 2021, we should not allow typos in code uncaught by the compiler lol) outweighs the private encapsulation benefit (getting a reference to the class instance is nonstandard usage anyway). Go ahead with changing them to public

adidahiya:
The new implementation looks fine, can you resolve the conflicts please?

ejose19:
@adidahiya Done

adidahiya (approved):
looks good! I clicked around some components in the latest deploy preview and nothing seems to be broken. thanks @ejose19

adidahiya:
@ejose19 heads up I'm cleaning up some of this code here https://github.com/palantir/blueprint/pull/4475

PR #4058, opened 2020-04-12
Stepper

justinbhopper:
Adds a new components `Stepper`, `Step`, and `StepConnector`.  A stepper is a wizard-like progress indicator.  

**This PR is not complete** due to an [issue with compiling docs-data](https://github.com/palantir/blueprint/issues/3062#issuecomment-439363356) on a windows machine.  

I added an example and started work on the docs page, but all of this was done blindly and very may have issues.  Help is very much appreciated.

#### Screenshot

remarkable ai (1)

PR #4462, opened 2020-12-17
Add a multistep dialog component

ycvfu:
This PR adds a multistep dialog component.

#### Checklist

- [X] Includes tests
- [X] Update documentation

This PR aims to create a new multistep dialog component that enables the users to move through a sequence of steps in a dialog, each of which presents a different panel. It is implemented as a wrapper around Dialog.

This PR is an updated version of https://github.com/palantir/blueprint/pull/4455 from a forked repository. I am temporarily keeping https://github.com/palantir/blueprint/pull/4455 open as other contributors may be looking at it (@kmblake are you planning on updating the old PR/referencing the old PR, or is it okay for me to close it?).

The dialog should present to the user a summary of steps on the left and the main panel associated with the currently selected panel on the right. When the user is on the first step, there should only be a 'next' or 'submit' footer button as appropriate. When the user is on the last step, there should only be a 'submit' footer button and potentially a 'back' footer button. When the user is on any page in between, they should see a back and next button.

The user also has the option of navigating between viewed steps on the left panel by clicking on a step. Only steps that have been previously accessed via clicking the "next" footer button can be clicked on. The user can select any pre-existing steps to see the panel.

All regular functionalities from Dialog should also be present.

#### Screenshot
image


packages/core/src/components/dialog/dialog.md
```diff
@@ -35,3 +42,18 @@ whenever possible, as they automatically generate some of this markup.
...
+
+@## Multistep Dialog
+
+@reactExample MultistepDialogExample
+
+`MultistepDialog` is a wrapper around `Dialog` that maps each sequential step to a view provided by the developer.
+
+The children you provide to this component are rendered as contents inside the
+`Classes.DIALOG` element. Typically, you will want to render a panel with
+`Classes.DIALOG_BODY` that contains the body content for each step.
```
kmblake, line 56: I think we should explain a little more about the `Step` interface similar to the docs for `Tab`. Also probably worth clarifying that _only_ children of type `Step` will be rendered.

packages/core/src/components/dialog/multistepDialog.tsx
```diff
@@ -0,0 +1,236 @@
...
+    IActionProps &
+    IElementRefProps<any>;
+
+type StepElement = React.ReactElement<IStepProps & { children: React.ReactNode }>;
+
+export interface IMultistepDialogProps extends IDialogProps {
+    /**
+     * Props for the button to display on the final step. To illustrate, this may be a submit button.
+     */
+    finalButtonProps?: IFinalButtonProps;
```
kmblake: I know there was some earlier discussion about not using `Partial<IButtonProps>` here because we wanted to prevent users from changing things like button size, but it's a bit difficult to interpret what these `IFinalButtonProps` actually includes based on the type declaration. I would be in favor of `Partial<IButtonProps>` or at least adding some more clarity in the comment.

packages/core/src/components/dialog/multistepDialog.tsx
```diff
@@ -0,0 +1,236 @@
...
+    }
+
+    /** Filters children to only `<Step>`s */
+    private getStepChildren(props: IMultistepDialogProps & { children?: React.ReactNode } = this.props) {
+        return React.Children.toArray(props.children).filter(isStepElement);
+    }
+}
+
+function isNextButtonEnabled(step: StepElement) {
+    return step.props.nextButtonEnabledByDefault !== undefined ? step.props.nextButtonEnabledByDefault : true;
```
kmblake:
```suggestion
    return step.props.nextButtonEnabledByDefault ?? true;
```

packages/core/src/components/dialog/multistepDialog.tsx
```diff
@@ -0,0 +1,236 @@
...
+    }
+
+    private getDialogStyle() {
+        return { paddingBottom: PADDING_BOTTOM, minWidth: MIN_WIDTH, ...this.props.style };
+    }
+
+    private renderLeftPanel() {
+        return (
+            <div className={Classes.MULTISTEP_DIALOG_LEFT_PANEL}>
+                {React.Children.toArray(this.props.children).filter(isStepElement).map(this.renderStep)}
```
kmblake:
```suggestion
                {this.getStepChildren().map(this.renderStep)}
```

packages/core/src/components/dialog/multistepDialog.tsx
```diff
@@ -0,0 +1,236 @@
...
+            </div>
+        );
+    }
+
+    private renderStep = (step: StepElement, index: number) => {
+        const stepNumber = index + 1;
+        const hasBeenViewed = this.state.lastViewedIndex >= index;
+        const currentlySelected = this.state.selectedIndex === index;
+        return (
+            <div className={classNames(Classes.STEP_CONTAINER, hasBeenViewed && Classes.ACTIVE)} key={index}>
```
kmblake:
```suggestion
            <div className={classNames(Classes.STEP_CONTAINER, { [Classes.ACTIVE]: hasBeenViewed })} key={index}>
```
Looks like other places in BP use this syntax for conditional classes but defer to @adidahiya

packages/core/src/components/dialog/multistepDialog.tsx
```diff
@@ -0,0 +1,236 @@
...
+            });
+        };
+    }
+
+    private getFinalButton() {
+        if (this.props.finalButtonProps !== undefined) {
+            return <Button key="final" {...this.props.finalButtonProps} />;
+        } else {
+            return <Button intent="danger" key="final" onClick={this.props.onClose} text="Submit" />;
+        }
```
kmblake:
```suggestion
return <Button intent="primary" key="final" onClick={this.props.onClose} text="Submit" {...this.props.finalButtonProps} />
```
Similar to how `buttonProps` are handled in `TimezonePicker`

Also, don't think the submit button should default to intent "danger"

packages/core/src/components/dialog/step.tsx
```diff
@@ -0,0 +1,72 @@
...
+export interface IStepProps extends IProps, Omit<HTMLDivProps, "id" | "title" | "onClick"> {
+    /**
+     * Unique identifier used to control which step is selected.
+     */
+    id: StepId;
+
+    /**
+     * Indicates whether the next button should be enabled by default on this step
+     */
+    nextButtonEnabledByDefault?: boolean;
```
kmblake: Wondering if we should just expose `nextButtonProps` on `MultistepDialog` and make the next button fully controlled. Then we wouldn't need `nextButtonEnabledByDefault` or `updateDialog`, and `Step` would be a purely presentational component that could just render its children. If the next button needs to be disabled at the start of certain steps, the client would do that in their change handler for MultistepDialog's `onChange`. Since the parent component of `MultistepDialog` is ultimately responsible for keeping track of what data has been entered in the dialog, it might make sense to just allow it to control when the next button is enabled as well.

packages/core/src/components/dialog/_multistep-dialog.scss
```diff
@@ -0,0 +1,87 @@
...
+}
+
+.#{$ns}-step {
+  align-items: center;
+  background-color: $light-gray5;
+  border-radius: $step-radius;
+  cursor: not-allowed;
+  display: flex;
+  margin: 4px;
+  padding: 6px;
```
kmblake: Nit: why not just 10px of padding? I think would expect these to highlight full-bleed like rows in a table.

packages/core/src/components/dialog/_multistep-dialog.scss
```diff
@@ -0,0 +1,87 @@
...
+
+$dialog-border-radius: $pt-border-radius * 2 !default;
+$step-radius: $pt-border-radius * 2 !default;
+
+.#{$ns}-multistep-dialog-panels {
+  display: flex;
+}
+
+.#{$ns}-multistep-dialog-left-panel {
+  border-radius: 0 $dialog-border-radius 0 0;
```
kmblake: Do we need this given that this div doesn't have a background color?

kmblake (commented):
Thanks for contributing! I've provided some initial comments but ultimately defer to @adidahiya. Also @ycvfu if you want to offload this PR just let me know--I'm happy to own any additional changes to get it over the line after the holidays.

packages/core/src/components/dialog/_multistep-dialog.scss
ycvfu: I think Carol's original designs wanted only a part of the button to be highlighted when we hover over it. The margin enables us to prevent that space from being highlighted as well. 
 
image

Not sure if this is the still the case, but if not, I can change it to 10px padding.

packages/core/src/components/dialog/multistepDialog.tsx
ycvfu: Hmm, the only concern I have is that we would need to overwrite props like onClick `<Button intent="primary" key="final" text="Submit" {...this.props.finalButtonProps} onClick={this.getCombinedOnCloseProps}/>` in order to ensure that the props we want to enforce don't get overwritten. This seems a little messy to me, especially when the user shouldn't need to use those props to begin with. I will add a better comment for this prop for now, and if we still want to change this to use Partial, I can do that.

packages/core/src/components/dialog/step.tsx
ycvfu: Hmm, maybe we can expose a disableNext prop in the MultistepDialog instead of exposing nextButtonProps. I don't think that the user should be able to control any aspect of this button except its enabled state. If the user could change the button design such as the text, intent, etc. it would enable them to deviate from the Blueprint design, and I can't think of a case where this would be desirable (since we want to standardize this). For instance, enabling them to change a button on step 3 out of 5 to show "submit" would break the flow and cause confusion.

packages/core/src/components/dialog/dialog.md
ycvfu: Sounds good, I added info under line 54.

packages/core/src/components/dialog/step.tsx
kmblake: Sounds good, I think just exposing `disableNext` on `MultistepDialog` is sufficient! In that case, we should remove this prop from `Step`, right?

packages/core/src/components/dialog/multistepDialog.tsx
```diff
@@ -0,0 +1,228 @@
...
+        return <Button intent="primary" key="final" text="Submit" {...this.props.finalButtonProps} />;
+    }
+
+    /** Filters children to only `<Step>`s */
+    private getStepChildren(props: IMultistepDialogProps & { children?: React.ReactNode } = this.props) {
+        return React.Children.toArray(props.children).filter(isStepElement);
+    }
+}
+
+function isNextButtonEnabled(step: StepElement) {
```
kmblake: No longer needed if we're exposing the `disableNext` prop

packages/core/src/components/dialog/multistepDialog.tsx
```diff
@@ -0,0 +1,228 @@
...
+        );
+    }
+
+    public componentDidUpdate(prevProps: IMultistepDialogProps) {
+        if (!prevProps.isOpen && this.props.isOpen) {
+            this.setState(this.getDefaultState());
+        }
+    }
+
+    private getDefaultState() {
```
kmblake: If we only need to define `lastViewedIndex` and `selectedIndex`, probably a little clearer to just define those in an `INITIAL_STATE` const and then set state to `INITIAL_STATE` in the constructor and `componentDidUpdate`

packages/core/src/components/dialog/multistepDialog.tsx
```diff
@@ -0,0 +1,228 @@
...
+        if (this.state.selectedIndex === this.getStepChildren().length - 1) {
+            buttons.push(this.getFinalButton());
+        } else {
+            buttons.push(
+                <Button
+                    disabled={this.props.disableNext}
+                    intent="primary"
+                    key="next"
+                    onClick={this.handleClickNext()}
+                    text="Next"
```
kmblake, line 174: @adidahiya how does BP handle localization? It seems like for most components the client supplies all of the copy. Should that also be the case here (at least making it possible to override the button text)?

packages/core/src/components/dialog/step.tsx
```diff
@@ -0,0 +1,72 @@
...
+import classNames from "classnames";
+import * as React from "react";
+import { polyfill } from "react-lifecycles-compat";
+
+import { AbstractPureComponent2, Classes } from "../../common";
+import { DISPLAYNAME_PREFIX, HTMLDivProps, IProps } from "../../common/props";
+
+export type StepId = string | number;
+
+export interface IMultistepDialogPanelProps {
```
kmblake: No longer used?

kmblake (commented):
Looking good, mostly just comments on completing the transition to make `disableNext` controlled at the dialog level rather than handled by each step.

packages/core/src/components/dialog/step.tsx
ycvfu: Oh my bad, I will go back and clean up the step component.

packages/core/src/components/dialog/step.tsx
ycvfu: Yeah! My bad, I missed that. I fixed it in the latest revision.

packages/core/src/components/dialog/_multistep-dialog.scss
```diff
@@ -0,0 +1,86 @@
...
+  &.#{$ns}-active {
+    background-color: $white;
+  }
+}
+
+.#{$ns}-step {
+  align-items: center;
+  background-color: $light-gray5;
+  border-radius: $step-radius;
+  cursor: not-allowed;
```
adidahiya, line 71: I don't think this is strictly necessary, it can just be grayed out and have a normal pointer with no hover state. Also, what about previous steps? Can't you click on those?

packages/core/src/components/dialog/multistepDialog.tsx
```diff
@@ -0,0 +1,220 @@
...
+    public static displayName = `${DISPLAYNAME_PREFIX}.MultistepDialog`;
+
+    public static defaultProps: Partial<IMultistepDialogProps> = {
+        canOutsideClickClose: true,
+        isOpen: false,
+    };
+
+    public constructor(props: IMultistepDialogProps) {
+        super(props);
+        this.state = INITIAL_STATE;
```
adidahiya: nit: instead of using a constructor, we can just write

```ts
public state: IMultistepDialogState = {
    lastViewedIndex: 0,
    selectedIndex: 0,
};
```

packages/core/src/components/dialog/multistepDialog.tsx
```diff
@@ -0,0 +1,220 @@
...
+                    {this.renderLeftPanel()}
+                    {this.maybeRenderRightPanel()}
+                </div>
+            </Dialog>
+        );
+    }
+
+    public componentDidUpdate(prevProps: IMultistepDialogProps) {
+        if (!prevProps.isOpen && this.props.isOpen) {
+            this.setState(INITIAL_STATE);
```
adidahiya, line 98: Do we always want to reset to initial state when a dialog is re-opened? Let's document this design decision somewhere in the docs. Maybe this behavior should be controlled by a prop in the future.

packages/core/src/components/dialog/multistepDialog.tsx
```diff
@@ -0,0 +1,220 @@
...
+                        {step.props.title}
+                    </div>
+                </div>
+            </div>
+        );
+    };
+
+    private handleClickStep = (index: number) => {
+        if (index > this.state.lastViewedIndex) {
+            return undefined;
```
adidahiya: nit: this code path doesn't need to return a value; I would either `return;` here or reverse the condition to `if (index <= this.state.lastViewedIndex)`

packages/core/src/components/dialog/multistepDialog.tsx
adidahiya: Yeah, I think we need to add a prop to override the button text for i18n. See my comment on `IMultistepDialogProps`.

packages/core/src/components/dialog/multistepDialog.tsx
```diff
@@ -0,0 +1,220 @@
...
+                    onClick={this.handleClickNext()}
+                    text="Next"
+                />,
+            );
+        }
+
+        return buttons;
+    }
+
+    private handleClickBack() {
```
adidahiya: nit: naming semantics are slightly off, this should be `getBackClickHandler` since it returns a function

packages/core/src/components/dialog/multistepDialog.tsx
```diff
@@ -0,0 +1,220 @@
...
+        return this.handleChangeStep(this.state.selectedIndex - 1);
+    }
+
+    private handleClickNext() {
```
adidahiya: same here, `getNextClickHandler`

packages/core/src/components/dialog/multistepDialog.tsx
```diff
@@ -0,0 +1,220 @@
...
+        return this.handleChangeStep(this.state.selectedIndex + 1);
+    }
+
+    private handleChangeStep(index: number) {
```
adidahiya: same here, `getStepChangeHandler`

packages/core/src/components/dialog/multistepDialog.tsx
```diff
@@ -0,0 +1,220 @@
...
+import { Dialog, IDialogProps } from "./dialog";
+import { IStepProps, Step, StepId } from "./step";
+
+type StepElement = React.ReactElement<IStepProps & { children: React.ReactNode }>;
+
+export interface IMultistepDialogProps extends IDialogProps {
+    /**
+     * Whether to disable the next button in the panel footer.
+     */
+    disableNext?: boolean;
```
adidahiya: I propose making this a little more flexible and solving the problem of the button text mentioned below with a props object here.

```
nextButtonProps?: Partial<Pick<IButtonProps, "disabled" | "text">>;
```

packages/core/src/components/dialog/multistepDialog.tsx
```diff
@@ -0,0 +1,220 @@
...
+                this.props.onChange(newStepId, prevStepId, event);
+            }
+            this.setState({
+                lastViewedIndex: Math.max(this.state.lastViewedIndex, index),
+                selectedIndex: index,
+            });
+        };
+    }
+
+    private getFinalButton() {
```
adidahiya: nit: `renderFinalButton()` since this returns JSX

packages/core/src/components/dialog/step.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+export interface IStepProps extends IProps, Omit<HTMLDivProps, "id" | "title" | "onClick"> {
+    /**
+     * Unique identifier used to identify which step is selected.
+     */
+    id: StepId;
+
+    /**
+     * Panel content, rendered by the parent `MultistepDialog` when this step is active.
+     * If omitted, no panel will be rendered for this step.
```
adidahiya: what does this mean / look like? why wouldn't there be a panel?

packages/core/src/components/dialog/step.tsx
```diff
@@ -0,0 +1,63 @@
...
+    panelClassName?: string;
+
+    /**
+     * Content of step title element, rendered in a list left of the active panel.
+     */
+    title?: React.ReactNode;
+}
+
+@polyfill
+export class Step extends AbstractPureComponent2<IStepProps> {
```
adidahiya: I think we need to be more explicit about the naming of this component and its CSS classes. They should all be prefixed with "dialog" because they correspond to MultiStepDialog. This helps disambiguate from other potential "step" components, like those proposed in #4058 / #260.

packages/core/src/components/dialog/multistepDialog.tsx
```diff
@@ -0,0 +1,220 @@
...
+    }
+
+    private renderStep = (step: StepElement, index: number) => {
+        const stepNumber = index + 1;
+        const hasBeenViewed = this.state.lastViewedIndex >= index;
+        const currentlySelected = this.state.selectedIndex === index;
+        return (
+            <div className={classNames(Classes.STEP_CONTAINER, { [Classes.ACTIVE]: hasBeenViewed })} key={index}>
+                <div
+                    className={classNames(Classes.STEP, { [Classes.ACTIVE]: hasBeenViewed })}
```
adidahiya: do we need all 4 active classes? I think we can remove this one and the one on L113 and change the Sass to be something like:

```sass
.step {
  .active & {
    ...
  }
}
```

packages/core/src/components/dialog/_multistep-dialog.scss
adidahiya: On further review, I realized the semantics of "active" here are a little different from what I expected. That's fine, I think what you have works. but I think it would help to leave a code comment along the lines of "by default, steps are inactive until they are visited"

adidahiya (changes requested):
[PR preview build](https://38167-71939872-gh.circle-artifacts.com/0/packages/docs-app/dist/index.html#core/components/dialog.multistep-dialog)

Looking great so far! The docs example is currently pretty simple, but we can flesh it out more in a future PR.

It's a little unfortunate how the header icon is misaligned with the step icons here. If we're going to keep the padding this way, we should change the header icon in the example to something that's not a circle.

image

packages/core/src/components/dialog/multistepDialog.tsx
ycvfu: Hmm, yeah I made the assumption here that if we closed a dialog or clicked away from it, we would want any progress would be lost. We can expose a prop like "resetOnClose" to enable the user to control this.

packages/core/src/components/dialog/_multistep-dialog.scss
```diff
@@ -0,0 +1,87 @@
...
+}
+
+.#{$ns}-dialog-step {
+  align-items: center;
+  background-color: $light-gray5;
+  border-radius: $step-radius;
+  cursor: not-allowed;
+  display: flex;
+  margin: 4px;
+  padding: 6px 14px;
```
ycvfu, line 74: Changed padding to better align with the title icon.

adidahiya:
Latest changes look good! But it looks like we need to add dark theme support:

image

adidahiya (approved):
looks good! thanks for finishing this up @ycvfu 🎉

PR #4463, opened 2020-12-18
fix dateInput crash in dev mode

### LINKED ISSUE #4251: DateInput default settings can easily crash application in dev mode
Summarizing a collection of grievances in https://github.com/palantir/blueprint/issues/3394, https://github.com/palantir/blueprint/issues/2965, https://github.com/palantir/blueprint/issues/3407, and other issues:

#### Environment

- __Package version(s)__: datetime 3.18.5
- __Operating System__: macOS
- __Browser name and version__: any

#### Code Sandbox

From @barrymay in https://github.com/palantir/blueprint/issues/3394#issuecomment-665406086:

> https://codesandbox.io/s/angry-morse-hn3rm?file=/src/App.js
>
> To see the problem:
> - Enter 1/1/1999. Hit 'enter'

#### Actual behavior

Page crashes during `validateProps()` lifecycle method

#### Expected behavior

Page does not crash, instead we get some nicer out-of-the-box error handling like the docs example:

#### Possible solution

The developer has provided a valid `parseDate` function, but since they are using the default values of `minDate` and `maxDate` set by the component, they are not able to properly validate whether the entered date is out of bounds. Another option would be to wrap the component in an ErrorBoundary, but that's clunky and heavy-handed.

In general I think users should be setting their own `minDate` and `maxDate` bounds. But since we have already set the defaults in the API I think we should improve the developer experience here without breaking the API.

IshinoJun:
Fixes https://github.com/palantir/blueprint/issues/4251

#### Checklist

- [ ] Includes tests
- [ ] Update documentation

In dev mode, entering a value outside the range of minDate and maxDate causes the page to crash. So I fixed it to warn level.

#### Screenshot
8a496a952ffda8b9fd846025d471f65c

### DIFF
--- a/packages/datetime/src/datePicker.tsx
+++ b/packages/datetime/src/datePicker.tsx
@@ -212,7 +212,7 @@ export class DatePicker extends AbstractPureComponent2<IDatePickerProps, IDatePi
         }
 
         if (value != null && !DateUtils.isDayInRange(value, [minDate, maxDate])) {
-            throw new Error(Errors.DATEPICKER_VALUE_INVALID);
+            console.warn(Errors.DATEPICKER_VALUE_INVALID);
         }
     }
 


palantirtech:
Thanks for your interest in [palantir/blueprint](https://github.com/palantir/blueprint), @IshinoJun! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

adidahiya (changes requested):
This doesn't really fix the issue properly as I mentioned in the "expected behavior" part of #4251. Also it makes the different checks in `validateProps` inconsistent where some are exceptions and others are console errors. You can already work around #4251 by using your own `minDate` and `maxDate` and proper `parseDate` implementation, so I don't want to accept a band-aid fix like this current PR.

PR #4471, opened 2021-01-05
[core] feat(Slider): labelRenderer option to allow different output in handle tooltip

### LINKED ISSUE #3578: Allow the slider tooltip render to be separated from the label renderer
#### Environment

Package version: @blueprintjs/core 3.15.1
Browser: Chrome 74+

#### Feature request

Allow a separate tooltip renderer on the slider, instead of just using the label renderer. The label renderer is a great fallback but sometimes the tooltip needs to be different.

#### Examples

We use a slider for stepping through data over a range of date/times - anything from the last hour to the last 30 days.

When the date range spans over multiple days, we render labels on the hour; but for midnight, we render the new date instead. Something like this -

image

The problem though, is that when the current value passes the date boundary, the tooltip shows the date instead of the time (due to the way that labels for the slider are generated):

image

We could get around this if either we had a separate renderer for the tooltips, OR if the renderer knew the "context" about where the label was being applied (that is, if we knew that the label was going to be attached as a tooltip instead of as a label on the slider itself, we would calculate the label's value differently.)

naughton:
#### Checklist

- [x] Includes tests
- [x] Update documentation
can update docs if it looks like this PR will be accepted.

add a boolean (always true) to formatLabel callback, that indicates the value should be formatted for the handle, not the axis.

#### Screenshot

image

Here on this Date range selector, we would like to add the DAY of the month to the handle, but not to the axis.

### DIFF
--- a/packages/core/src/components/slider/multiSlider.tsx
+++ b/packages/core/src/components/slider/multiSlider.tsx
@@ -97,9 +97,12 @@ export interface ISliderBaseProps extends IProps, IIntentProps {
      * If `true`, labels will use number value formatted to `labelPrecision` decimal places.
      * If `false`, labels will not be shown.
      *
+     * The callback is provided a numeric value and optional rendering options, which include:
+     * - isHandleTooltip: whether this label is being rendered within a handle tooltip
+     *
      * @default true
      */
-    labelRenderer?: boolean | ((value: number, isHandle?: boolean) => string | JSX.Element);
+    labelRenderer?: boolean | ((value: number, opts?: { isHandleTooltip: boolean }) => string | JSX.Element);
 
     /**
      * Whether to show the slider in a vertical orientation.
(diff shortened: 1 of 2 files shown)


palantirtech:
Thanks for your interest in [palantir/blueprint](https://github.com/palantir/blueprint), @naughton! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

packages/core/src/components/slider/multiSlider.tsx
```diff
@@ -99,7 +99,7 @@ export interface ISliderBaseProps extends IProps, IIntentProps {
      *
      * @default true
      */
-    labelRenderer?: boolean | ((value: number) => string | JSX.Element);
+    labelRenderer?: boolean | ((value: number, isHandle?: boolean) => string | JSX.Element);
```
adidahiya: for a bit of forwards compatibility if we add more options here:

```suggestion
    labelRenderer?: boolean | ((value: number, opts?: { isHandleTooltip: boolean; }) => string | JSX.Element);
```

packages/core/src/components/slider/multiSlider.tsx
```diff
@@ -99,7 +99,7 @@ export interface ISliderBaseProps extends IProps, IIntentProps {
      *
```
adidahiya, line 99: we need some more docs here about the new feature...

```suggestion
     *
     * The callback is provided a numeric value and optional rendering options, which include:
     * - isHandleTooltip: whether this label is being rendered within a handle tooltip
     *
```

adidahiya (changes requested):
thanks for the PR. left some minor comments

adidahiya (approved):

PR #4480, opened 2021-01-13
add loading state to alerts

nnchang:
I ran into a situation where I wanted to make an alert that had a clear load state (specifically the one supplied by `<Button>`). Because `<Alert>` doesn't give allow much control over the buttons, I wound up implementing that solution using a dialog and styling it to look like an alert. I think it would be nice for `<Alert>` to have this functionality.

#### Checklist

- [ ] Includes tests
- [x] Update documentation

Add a loading state to `<Alert>`.

I'm not familiar with the testing library, so I haven't yet added tests. I can dig into that if maintainers find this to be an acceptable suggestion.

#### Screenshot

alert loading state

### DIFF
--- a/packages/core/src/components/alert/alert.tsx
+++ b/packages/core/src/components/alert/alert.tsx
@@ -74,6 +74,15 @@ export interface IAlertProps extends IOverlayLifecycleProps, IProps {
      */
     isOpen: boolean;
 
+
+    /**
+     * If set to `true`, the confirm button will be set to its loading state. The cancel button, if
+     * visible, will be disabled.
+     *
+     * @default false
+     */
+    loading?: boolean;
+
     /**
      * CSS styles to apply to the alert.
      */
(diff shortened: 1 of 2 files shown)


nnchang:
side note: as a new contributor, I was surprised that the `yarn dev` command as suggested in the CONTRIBUTING.md took me straight to the table app and didn't seem to have the core package building. Looking at [package.json](https://github.com/palantir/blueprint/blob/a37f79c95c5b377a9793b222a142df7f5bff2ce7/package.json#L17), I think this is expected behavior. I'm happy to make a PR to change the docs if appropriate.

adidahiya:
good point on `yarn dev` from root. I made a change in https://github.com/palantir/blueprint/commit/842b2bb43f43ad11b71cdec86fff410d1a14474c but forgot to reverse the negation in the glob. I'll fix that.

it looks like you may have enabled CircleCI for your fork, which is not encouraged -- we want your PR to build on our org's Circle instance to report status checks correctly.

as for the new prop itself... I am generally very conservative about what we add to `Alert` but in this case I think it's an elegant addition and I support it :+1:

nnchang:
cool! I will close this PR and try to reopen such that circle will run correctly.

PR #4513, opened 2021-02-01
[v4] load icons asynchronously

### LINKED ISSUE #2193: A way to only bundle icons you actually use

adidahiya:
#### Checklist

- [x] Includes tests
- [x] Update documentation

- All packages now build to a `"es2020"` [module target](https://www.typescriptlang.org/tsconfig#module). This is nearly identical to the previous `"es2015"` target, but now has support for dynamic `import()` expressions ([MDN docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import#dynamic_imports)).
  - ⚠️ You will have to ensure that your build system is able to transpile / handle this new ES syntax. If you are using modern browsers, modern Node.js, modern Webpack, and/or [`@babel/plugin-syntax-dynamic-import`](https://babeljs.io/docs/en/babel-plugin-syntax-dynamic-import), this should require no action on your part. In other words, most modern tooling in the JS ecosystem does support this new feature.
- Expose a singleton class API `{ Icons } from "@blueprintjs/icons"` to load specific icons from the icon set
- Expose individual icon SVG component modules, e.g. `import { Download } from "@blueprintjs/icons"`
- `<Icon>` component's static SVG path string loader has been replaced by a loader that uses a dynamic `import()` expression
- Because of the dynamic import change, I have changed all BP component source code to use static icon module imports instead of the string literal API. This means they will work out of the box without any user configuration.

- Developer experience of icon loader
- Performance of dynamic import


switz:
I'm seeing every icon imported into my nextjs app on 4.0.0. Is there a specific API I should be using to treeshake them dynamically? https://nextjs.org/docs/advanced-features/dynamic-import

Before I was overriding the generated/svgPaths file. I generally specify my icons via: `<Icon icon="caret-down" />` but I can switch to explicit imports if that helps.

image

adidahiya:
@switz can you open a new issue and we can try to debug from there? you shouldn't need to configure anything if you're using webpack (which Next.js does use under the hood); this bit of code should be doing the dynamic import for you: https://github.com/palantir/blueprint/blob/184cab89dbe2d1ee9765c12141daea9f67a35076/packages/icons/src/iconLoader.ts#L36-L49

alex-kinokon:
@adidahiya Is this change reversed? `import { Download } from "@blueprintjs/icons"` no longer works for `4.0.0-beta.2`.

adidahiya:
@proteriax yes, see my comment https://github.com/palantir/blueprint/issues/4645#issuecomment-944411944

PR #4522, opened 2021-02-08
chore: upgrade to webpack v5

adidahiya:

- Upgrade webpack to v5 and dependent packages
- Upgrade karma to v6, update required config accordingly
- Remove deprecated awesome-typescript-loader, use ts-loader instead
- Remove very old karma-mocha-reporter, use [karma-helpful-reporter](https://github.com/whyboris/karma-helpful-reporter) instead
- Remove deprecated karma-remap-coverage, this should be integrated into nyc (latest iteration of istanbul) now
- Enable HMR

#### Upgrade notes

Dependency changelogs / upgrade guides:

- https://webpack.js.org/migrate/5/
- https://github.com/karma-runner/karma/blob/master/CHANGELOG.md
- https://github.com/ryanclark/karma-webpack/blob/master/CHANGELOG.md

Issues I ran into / worked around:

- https://github.com/pmmmwh/react-refresh-webpack-plugin/issues/289
- https://github.com/TypeStrong/ts-loader/issues/405#issuecomment-689804673
- karma-mocha-reporter wasn't getting registered, and after much karma plugin fiddling, I decided to replace it with a more up-to-date reporter.
- I was getting `process` undefined errors, which I expected with webpack 5, but in the case of the karma bundle using `webpack.EnvironmentPlugin` or `webpack.DefinePlugin` was not sufficient, I had to use [`webpack.ProvidePlugin`](https://webpack.js.org/plugins/provide-plugin/)
- New webpack warnings apparently flag some type exports that might not work when TS is compiled with `isolatedModules: true`. I don't think we care about isolated modules compilation. These warnings are a bit noisy, but I've left them there for this PR. Might revisit in a future PR. See https://github.com/webpack/webpack/issues/7378


adidahiya:
:shipit:

PR #4556, opened 2021-03-03
[core] fix(Icon): accept standard HTML attributes

### LINKED ISSUE #4500: Icon component cannot accept ARIA `role` prop
#### Environment

- __Package version(s)__: 3.38
- __Operating System__: RHEL7 (not relevant)
- __Browser name and version__: Firefox 78 (not relevant)

#### Code Sandbox

N/A

#### Steps to reproduce

1. Try to render `<Icon icon="cross" role="image" />`

#### Actual behavior

TypeScript fails to compile because the `Icon` component does not accept a `role` prop.

#### Expected behavior

The `Icon` component would pass through `role` to the underlying `<svg>` or `<span>`

#### Possible solution

`Icon` is currently defined as using `React.DOMAttributes<HTMLElement>>`:

https://github.com/palantir/blueprint/blob/26351ca79595ebaf633c57b1beadc37a9f2d62b6/packages/core/src/components/icon/icon.tsx#L90

Per the React TS types, `role` is part of the `HTMLAttributes` interface, which extends `DOMAttributes`:

```ts
    interface HTMLAttributes<T> extends AriaAttributes, DOMAttributes<T> {
        // React-specific Attributes
        defaultChecked?: boolean;
        defaultValue?: string | number | ReadonlyArray<string>;
        suppressContentEditableWarning?: boolean;
        suppressHydrationWarning?: boolean;

        // Standard HTML Attributes
        accessKey?: string;
        // omit

        // WAI-ARIA
        role?: string;

        // omit
    }
```

Perhaps it should be switched over to use `HTMLAttributes` instead?

ChristianIvicevic:
#### Checklist

- [X] Includes tests
- [ ] Update documentation

This PR adds standard default HTML attributes to the core Icon component. The motivation was the inability to set a role for icons and with the proposed solution in the linked issue this is now possible.

Unfortunately the `IIconProps` type already declares a `title` property which collides with the definition in `HTMLAttributes` due to deviating types. In order to prevent issue in the `HTMLSelect` component I decided to omit the latter. This fix is open for dicussion.

### DIFF
--- a/packages/core/src/components/icon/icon.tsx
+++ b/packages/core/src/components/icon/icon.tsx
@@ -87,7 +87,9 @@ export interface IIconProps extends IIntentProps, IProps {
 }
 
 @polyfill
-export class Icon extends AbstractPureComponent2<IIconProps & React.DOMAttributes<HTMLElement>> {
+export class Icon extends AbstractPureComponent2<
+    IIconProps & Omit<React.HTMLAttributes<HTMLElement>, "title"> & React.DOMAttributes<HTMLElement>
+> {
     public static displayName = `${DISPLAYNAME_PREFIX}.Icon`;
 
     public static readonly SIZE_STANDARD = 16;
--- a/packages/core/src/components/icon/icon.tsx
+++ b/packages/core/src/components/icon/icon.tsx
@@ -87,9 +87,7 @@ export interface IIconProps extends IIntentProps, IProps {
 }
 
 @polyfill
-export class Icon extends AbstractPureComponent2<
-    IIconProps & Omit<React.HTMLAttributes<HTMLElement>, "title"> & React.DOMAttributes<HTMLElement>
-> {
+export class Icon extends AbstractPureComponent2<IIconProps & Omit<React.HTMLAttributes<HTMLElement>, "title">> {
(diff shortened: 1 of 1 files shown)


packages/core/src/components/icon/icon.tsx
```diff
@@ -87,7 +87,9 @@ export interface IIconProps extends IIntentProps, IProps {
 }
 
 @polyfill
-export class Icon extends AbstractPureComponent2<IIconProps & React.DOMAttributes<HTMLElement>> {
+export class Icon extends AbstractPureComponent2<
+    IIconProps & Omit<React.HTMLAttributes<HTMLElement>, "title"> & React.DOMAttributes<HTMLElement>
```
adidahiya: we don't need `React.DOMAttributes` anymore, since `React.HTMLAttributes` extends it.

adidahiya (changes requested):

packages/core/src/components/icon/icon.tsx
ChristianIvicevic: Ah this is my bad, didn't see it in the first place.

adidahiya (approved):

PR #4588, opened 2021-03-15
[v4] fix(ContextMenu2): position correctly inside Collapse containers

### LINKED ISSUE #4537: ContextMenu2 with Tree - wrong open position
#### Environment

- __Package version(s)__: @blueprintjs/popover2 0.3.0 (other versions below in codesandbox)
- __Operating System__: Windows 10
- __Browser name and version__: Chrome 88.0.4324.182

#### Code Sandbox

Link to a minimal repro: [https://codesandbox.io/s/blueprint-sandbox-forked-l8i73?file=/src/TreeView.tsx](https://codesandbox.io/s/blueprint-sandbox-forked-l8i73?file=/src/TreeView.tsx)

#### Steps to reproduce

1. Make ContextMenu2 on label in Tree, in child nodes
2. Open ContextMenu2

#### Actual behavior

ContextMenu2 opens in wrong place. Seems to go deeper and deeper with nested Tree items. Root level is correct.

#### Expected behavior

ContextMenu2 to open at mouse position.

#### Possible solution

Maybe I am using this wrong because I see there is an contextMenu event on Tree but idk how I could connect it to ContextMenu2.

adidahiya:
#### Checklist

- [ ] Includes tests
- [x] Update documentation

- Refactor Tree example as a function component with cleaner state handling logic (using a React reducer hook)
- Add a repro case for #4537 in the refactored Tree example
- Add `positioningStrategy` prop for Popover2
- Add `popoverClassName` prop for ContextMenu2
- Fix `ContextMenu2`'s styles to use their own classes instead of `ContextMenu`'s, add the relevant Sass stylesheet
- Create a new utility class `.bp4-fixed-positioning-containing-block` which is used to give hints to child / ancestor elements which use `position: fixed` and need to know about their containing block for proper layout calculations. [See MDN docs](https://developer.mozilla.org/en-US/docs/Web/CSS/Containing_block#identifying_the_containing_block)
- Use this new utility class in `Collapse` and `ContextMenu2` to fix #4537 

Overall design of the bugfix using the utility class

### DIFF
--- a/packages/docs-app/src/examples/core-examples/treeExample.tsx
+++ b/packages/docs-app/src/examples/core-examples/treeExample.tsx
@@ -18,7 +18,7 @@ import React, { useCallback, useReducer } from "react";
 
 import { Classes, Icon, Intent, TreeNodeInfo, Tree } from "@blueprintjs/core";
 import { Example, ExampleProps } from "@blueprintjs/docs-theme";
-import { Tooltip2 } from "@blueprintjs/popover2";
+import { Classes as Popover2Classes, ContextMenu2, Tooltip2 } from "@blueprintjs/popover2";
 
 type NodePath = number[];
 
@@ -108,16 +108,22 @@ const INITIAL_STATE: TreeNodeInfo[] = [
         id: 0,
         hasCaret: true,
         icon: "folder-close",
-        label: "Folder 0",
+        label: (
+            <ContextMenu2 popoverClassName={Popover2Classes.POPOVER2_CONTENT_SIZING} content={<div>Hello there!</div>}>
+                Folder 0
+            </ContextMenu2>
+        ),
     },
     {
         id: 1,
         icon: "folder-close",
         isExpanded: true,
         label: (
-            <Tooltip2 content="I'm a folder <3" placement="right">
-                Folder 1
-            </Tooltip2>
+            <ContextMenu2 popoverClassName={Popover2Classes.POPOVER2_CONTENT_SIZING} content={<div>Hello there!</div>}>
+                <Tooltip2 content="I'm a folder <3" placement="right">
+                    Folder 1
+                </Tooltip2>
+            </ContextMenu2>
         ),
         childNodes: [
             {
@@ -140,9 +146,14 @@ const INITIAL_STATE: TreeNodeInfo[] = [
                 hasCaret: true,
                 icon: "folder-close",
                 label: (
-                    <Tooltip2 content="foo" placement="right">
-                        Folder 2
-                    </Tooltip2>
+                    <ContextMenu2
+                        popoverClassName={Popover2Classes.POPOVER2_CONTENT_SIZING}
+                        content={<div>Hello there!</div>}
+                    >
+                        <Tooltip2 content="foo" placement="right">
+                            Folder 2
+                        </Tooltip2>
+                    </ContextMenu2>
                 ),
                 childNodes: [
                     { id: 5, label: "No-Icon Item" },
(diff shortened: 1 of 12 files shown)


packages/popover2/src/contextMenu2.tsx
```diff
@@ -51,22 +53,28 @@ export const ContextMenu2: React.FC<ContextMenu2Props> = ({
...
 
-        e.preventDefault();
-        const newTargetOffset = { left: e.clientX, top: e.clientY };
-        setTargetOffset(newTargetOffset);
-        setIsOpen(true);
-    }, []);
+            e.preventDefault();
+
+            const { left, top } = getContainingBlockOffset(containerRef.current);
+            setTargetOffset({ left: e.clientX - left, top: e.clientY - top });
```
adidahiya, line 73: this is the crux of the bugfix. when there is a containing block which is not the document root, we need to subtract its client bounding rect left/top in the generated target offset.

packages/docs-app/src/examples/core-examples/treeExample.tsx
```diff
@@ -14,82 +14,116 @@
...
-    nodes: TreeNodeInfo[];
+type NodePath = number[];
+
+type TreeAction =
+    | { type: "SET_IS_EXPANDED"; payload: { path: NodePath; isExpanded: boolean } }
+    | { type: "DESELECT_ALL" }
+    | { type: "SET_IS_SELECTED"; payload: { path: NodePath; isSelected: boolean } };
+
+function forEachNode(nodes: TreeNodeInfo[], callback: (node: TreeNodeInfo) => void) {
+    if (nodes == null) {
```
p-szm: `nodes` are `TreeNodeInfo[]` so they're never null (unless `strictNullChecks` is off?)

packages/docs-app/src/examples/core-examples/treeExample.tsx
```diff
@@ -14,82 +14,116 @@
...
-                    onNodeExpand={this.handleNodeExpand}
-                    className={Classes.ELEVATION_0}
-                />
-            </Example>
-        );
+function treeExampleReducer(state: TreeNodeInfo[], action: TreeAction) {
+    switch (action.type) {
+        case "DESELECT_ALL":
+            const newState1 = [...state];
+            forEachNode(newState1, node => (node.isSelected = false));
```
p-szm, line 50: Doesn't this mutate the non-root nodes since you only did a shallow copy?

packages/popover2/src/contextMenu2.tsx
```diff
@@ -76,26 +84,27 @@ export const ContextMenu2: React.FC<ContextMenu2Props> = ({
...
 
-    const targetRef = useRef<HTMLDivElement>();
+    const targetRef = useRef<HTMLDivElement>(null);
     const renderTarget = useCallback(
         ({ ref }: Popover2TargetProps) => (
             <div
-                className={Classes.CONTEXT_MENU2_POPOVER_TARGET}
+                className={Classes.CONTEXT_MENU2_POPOVER2_TARGET}
                 style={targetOffset}
+                // style={{ transform: `translate(${targetOffset.left}px, ${targetOffset.top}px)` }}
```
p-szm: should this be removed?

packages/docs-app/src/examples/core-examples/treeExample.tsx
adidahiya: yep, good catch, I noticed this too, I will have to do a deep copy for this to be pure / correct

packages/popover2/src/contextMenu2.tsx
adidahiya: 😅 yes, errant comment

packages/docs-app/src/examples/core-examples/treeExample.tsx
adidahiya: hm I think you're right, and it's a copy-pasta remnant of old code from when `strictNullChecks` was off

adidahiya: actually, strictNullChecks is off for docs-app. will track fixing that separately

Issue #4521, opened 2021-02-08
Tooltip2+Button selection tabindex issue


dlech:
Follow up to #4502

#### Environment

- __Package version(s)__: @blueprintjs/popover2@0.2.0
- __Operating System__: Ubuntu 20.04
- __Browser name and version__:  Firefox 85.0

#### Code Sandbox

Link to a minimal repro (fork [this code sandbox](https://codesandbox.io/s/blueprint-sandbox-et9xy)): https://codesandbox.io/s/blueprint-sandbox-forked-146cf?file=/src/CoreExample.tsx

#### Steps to reproduce

1. click on a blank part of the browser
1. press tab on the keyboard to focus the buttons
1. notice that the selection indicator is rendered behind the buttons instead of in front
2. try pressing spacebar to activate the button

#### Actual behavior

The selection indicator is behind the neighboring buttons. Also, the highlight looks different from the normal button highlight like it is selecting the tooltip rather than the button. And more importantly, the button can't be activated with the keyboard by pressing space.

![image](https://user-images.githubusercontent.com/963645/107263845-dd986700-6a07-11eb-9e6b-1176ecebfd17.png)

#### Expected behavior

The UI should show the normal button selection highlight instead and should respond to keyboard activation.

![image](https://user-images.githubusercontent.com/963645/107264237-5c8d9f80-6a08-11eb-940f-66b8f6b78715.png)

seve:
Comparing the DOM between Popover1 and 2, it looks like the tabindex is attributed to the wrong element:

![image (2)](https://user-images.githubusercontent.com/8716829/111192004-db30ac00-8575-11eb-9e74-e00ffa83a177.png)
![image (3)](https://user-images.githubusercontent.com/8716829/111192017-de2b9c80-8575-11eb-9d6e-30022a44b5db.png)

PR #4589, opened 2021-03-16
[v4] migrate non-core packages from Popover to Popover2

adidahiya:
Includes a bugfix for https://github.com/palantir/blueprint/issues/4521, which I will extract into a separate PR for the 3.x `develop` branch.

Comparison screenshots of tab focus indicator:

Before:

After:

### DIFF
--- a/packages/timezone/test/timezonePickerTests.tsx
+++ b/packages/timezone/test/timezonePickerTests.tsx
@@ -20,16 +20,8 @@ import moment from "moment-timezone";
 import React from "react";
 import sinon from "sinon";
 
-import {
-    Button,
-    ButtonProps,
-    InputGroupProps,
-    InputGroup,
-    PopoverProps,
-    MenuItem,
-    Popover,
-    Position,
-} from "@blueprintjs/core";
+import { Button, ButtonProps, InputGroupProps, InputGroup, MenuItem } from "@blueprintjs/core";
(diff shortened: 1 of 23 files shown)


packages/timezone/test/timezonePickerTests.tsx
```diff
@@ -236,24 +231,15 @@ describe("<TimezonePicker>", () => {
...
-    }
-
-    function findInputGroup(timezonePicker: TimezonePickerShallowWrapper) {
-        return findQueryList(timezonePicker).shallow().find(InputGroup);
+    function findPopover(timezonePicker: TimezonePickerWrapper) {
+        return timezonePicker.find(Popover2);
     }
 
-    function clickFirstMenuItem(timezonePicker: TimezonePickerShallowWrapper): void {
-        findQueryList(timezonePicker).shallow().find(MenuItem).first().simulate("click");
```
adidahiya, line 257: note: this was a bit of a tricky fix; for some reason this click wasn't working on the MenuItem, I found it better to select for an HTML element with a class name and target that directly rather than selecting a custom React component. I think this advice generalizes to all kinds of unit tests where we are not inspecting the component props/state (which... is another conversation, we probably shouldn't be relying on those kinds of implementation details in tests)

adidahiya (commented):
clicked around in preview build, found no regressions from develop for select / datetime / timezone / table column header

PR #4597, opened 2021-03-19
Remove inner shadow on focused inputs

CPerinet:
#### Checklist

- ~~[ ] Includes tests~~
- ~~[ ] Update documentation~~

- Remove inner shadow when an input is focused

The rational here is that the inner shadow is mostly useful to:
- Identify the component as an input
- Have it stand out
But in the case when the field has been focused by a user or is auto-focused, it seems unnecessary / redundant given we already have a strong visual indication with the blue shadow.
We can find the same behaviour on Github for example

Screenshot 2021-03-19 at 13 01 57
Screenshot 2021-03-19 at 13 01 59

- Make sure the change is not affecting other components that should not be

#### Screenshot

Screenshot 2021-03-19 at 11 59 15

### DIFF
--- a/packages/core/src/components/forms/_common.scss
+++ b/packages/core/src/components/forms/_common.scss
@@ -102,7 +102,7 @@ $control-group-stack: (
 
   &:focus,
   &.#{$ns}-active {
-    box-shadow: input-transition-shadow($input-shadow-color-focus, true), $input-box-shadow-focus;
+    box-shadow: input-transition-shadow($input-shadow-color-focus, true);
   }
 
   &[type="search"],


invliD (commented):
This change makes the input appear larger when focused, because the blue focus outline is _outside_ of the regular grey border, and by removing the inner shadow the grey border is completely gone when focused.

CPerinet:
@invliD That's correct — it's not a regression though, already is the case that inputs looks bigger when focused. Would it makes sense to try and fix it in this PR? That would probably required adding a different focus mixin to accommodate both behaviours. (focus outside border, focus over border)

invliD:
I wouldn't necessarily say that. Before this PR, the focus style mostly adds the outline (+glow), everything inside stays mostly the same. With this PR, everything grey turns white.

What are you trying to achieve with this change? Just because the focus outline makes it obvious the user is looking at an input, doesn't really mean there must not be anything else showing the same thing.

CPerinet:
1. That's correct, I meant that the PR does not make the input looks bigger than it used to be. Removing the gray inside does not enlarge it.

2. I would disagree there, showing redundant visual clues on a component that is such widely spread hurts the overall visual balance of the page. Our UIs are so dense already, we shall thrive to trim unnecessary noise.

aycai:
To carry over my thoughts from an internal thread:
- Changing the input border color on focus seems reasonable (where previous versions I saw added the blue border outside the existing grey border). That noise seems reasonable to cut.
- Removing the inner shadow (not referring to the inset border) I agree looks cleaner, but flattens an otherwise inset-appearing element, which is a new paradigm in Blueprint. It's not just the static states, but how the input transitions between the two (where the removal of the inner shadow is most apparent and jarring). 

Based on other chats I believe the intent of this change is broader than a visual "fix" here, which is a separate conversation. Given that, I'd opt to hold off on this change (which I believe is unnecessary) and focus on pushing forward that broader effort.

PR #4603, opened 2021-03-23
[popover2] feat: add position prop for improved backcompat

adidahiya:
#### Checklist

- [ ] Includes tests
- [ ] Update documentation

Improve Popover2's backwards compatibility with Popover. Originally I planned to move towards popper.js semantics in Popover2 with the upgrade to v2, where "placement" replaces "position". In practice, this distinction doesn't seem very meaningful and just makes the upgrade harder. Also there is some value in abstracting away low-level library positioning implementation details at the Blueprint API level here. So, this PR has these changes:

- [core] feat: un-deprecate the `position` prop for `<Popover>`
- [popover2] feat: add `position` prop to set the popper.js placement using an API familiar from Blueprint v3

No regressions

### DIFF
--- a/packages/core/src/components/popover/popover.tsx
+++ b/packages/core/src/components/popover/popover.tsx
@@ -116,8 +116,7 @@ export class Popover extends AbstractPureComponent2<IPopoverProps, IPopoverState
         minimal: false,
         modifiers: {},
         openOnTargetFocus: true,
-        // N.B. we don't set a default for `placement` here because that would override
-        // the deprecated `position` prop
+        // N.B. we don't set a default for `placement` here because that would override the `position` prop
         position: "auto",
         targetTagName: "span",
         transitionDuration: 300,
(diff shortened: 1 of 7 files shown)


packages/core/src/components/popover/popover.tsx
```diff
@@ -116,8 +116,7 @@ export class Popover extends AbstractPureComponent2<IPopoverProps, IPopoverState
         minimal: false,
         modifiers: {},
         openOnTargetFocus: true,
-        // N.B. we don't set a default for `placement` here because that would override
-        // the deprecated `position` prop
+        // N.B. we don't set a default for `placement` here because that would override the `position` prop
         position: "auto",
```
adidahiya: this will cause errors when `placement` is specified, which should be valid. we should configure this default inline during the render phase, not here as a static default prop

packages/popover2/src/popover2.tsx
```diff
@@ -122,7 +123,8 @@ export class Popover2<T> extends AbstractPureComponent2<IPopover2Props<T>, IPopo
         interactionKind: Popover2InteractionKind.CLICK,
         minimal: false,
         openOnTargetFocus: true,
-        placement: "auto",
+        // N.B. we don't set a default for `placement` here because that would override the `position` prop
+        position: "auto",
```
adidahiya: same problem here. it can be seen producing an error in Popover2Example

PR #4604, opened 2021-03-23
[popover2] docs: remove "content as second child" JSDoc

zachkirsch:
#### Checklist

- [ ] Includes tests
- [ ] Update documentation

Remove `This can instead be provided as the second element in children (first is target).` from Popover2's `content` prop, since Popover2 only accepts a single child (`target`).


adidahiya (approved):
thanks @zachkirsch!

PR #4610, opened 2021-03-26
[core] feat(MultistepDialog): allow custom next/back button props per step

### LINKED ISSUE #4579: add different nextButtonProps for each Dialog in MultistepDialog
#### Question

I have a MultistepDialog with several sub-dialogs in it, and I want to custom each dialog with different code to define if the nextButton should be `disabled` or not. For now, the only API `nextButtonProps` can just set the global Props for all nextButtons. 

something like this
Screen Shot 2021-03-11 at 10 52 47 PM

tanmoyopenroot:
#### Checklist

- [x] Includes tests
- [x] Update documentation

Provide a way to pass a callback to `backButtonProps` and `nextButtonProps`

### DIFF
--- a/packages/core/src/components/dialog/multistepDialog.tsx
+++ b/packages/core/src/components/dialog/multistepDialog.tsx
@@ -20,17 +20,19 @@ import { polyfill } from "react-lifecycles-compat";
 
 import { AbstractPureComponent2, Classes, Utils } from "../../common";
 import { DISPLAYNAME_PREFIX } from "../../common/props";
+import { isFunction } from "../../common/utils";
 import { Button, IButtonProps } from "../button/buttons";
 import { Dialog, IDialogProps } from "./dialog";
 import { DialogStep, DialogStepId, IDialogStepProps } from "./dialogStep";
 
 type DialogStepElement = React.ReactElement<IDialogStepProps & { children: React.ReactNode }>;
+type AllowedButtonProps = Pick<IButtonProps, "disabled" | "text">;
 
 export interface IMultistepDialogProps extends IDialogProps {
     /**
      * Props for the back button.
      */
-    backButtonProps?: Partial<Pick<IButtonProps, "disabled" | "text">>;
+    backButtonProps?: Partial<AllowedButtonProps> | ((step: DialogStepId) => Partial<AllowedButtonProps>);
 
     /**
      * Props for the button to display on the final step.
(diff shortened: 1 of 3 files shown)


packages/core/test/multistep-dialog/multistepDialogTests.tsx
```diff
@@ -200,6 +200,31 @@ describe("<MultistepDialog>", () => {
...
+        const getNextButtonProps = (stepId: DialogStepId) => {
+            if (stepId === "two") return { disabled: true };
+
+            return {};
+        };
+
+        const dialog = mount(
+            <MultistepDialog nextButtonProps={getNextButtonProps} isOpen={true} usePortal={false}>
+                <DialogStep id="one" title="Step 1" panel={<Panel />} />
+                <DialogStep id="two" title="Step 2" panel={<Panel />} />
```
adidahiya: hmm, on second thought, I think a better API would look like this:

```
<DialogStep id="two" title="Step 2" panel={<Panel />} nextButtonProps={{ disabled: true }} />
```

since we're calling `this.getDialogStepChildren()` and accessing each DialogStep's props while rendering the buttons anyway... we can make the API more straightforward and ergonomic this way

adidahiya (changes requested):

packages/core/test/multistep-dialog/multistepDialogTests.tsx
tanmoyopenroot: Done

packages/core/src/components/dialog/multistepDialog.tsx
```diff
@@ -164,28 +164,34 @@ export class MultistepDialog extends AbstractPureComponent2<IMultistepDialogProp
 
     private renderButtons() {
         const { selectedIndex } = this.state;
+        const steps = this.getDialogStepChildren();
         const buttons = [];
+
         if (this.state.selectedIndex > 0) {
+            const backButtonProps = steps[selectedIndex].props.backButtonProps || this.props.backButtonProps;
```
adidahiya: nit: avoid implicit boolean type coercion

```suggestion
            const backButtonProps = steps[selectedIndex].props.backButtonProps ?? this.props.backButtonProps;
```

packages/core/src/components/dialog/multistepDialog.tsx
```diff
@@ -164,28 +164,34 @@ export class MultistepDialog extends AbstractPureComponent2<IMultistepDialogProp
...
-                    {...this.props.backButtonProps}
+                    {...backButtonProps}
                 />,
             );
         }
 
         if (selectedIndex === this.getDialogStepChildren().length - 1) {
             buttons.push(<Button intent="primary" key="final" text="Submit" {...this.props.finalButtonProps} />);
         } else {
+            const nextButtonProps = steps[selectedIndex].props.nextButtonProps || this.props.nextButtonProps;
```
adidahiya: nit: same here

```suggestion
            const nextButtonProps = steps[selectedIndex].props.nextButtonProps ?? this.props.nextButtonProps;
```

adidahiya (changes requested):
thanks for making those changes

adidahiya (approved):

PR #4612, opened 2021-03-29
Fix unsafe ref usage

### LINKED ISSUE #4611: refHandler not typesafe safe and causes components to break when passed object ref
#### Environment

- __Package version(s)__: 
  - @blueprintjs/core: 3.40.0
- __Operating System__: any
- __Browser name and version__: any

#### Code Sandbox

https://codesandbox.io/s/blueprint-sandbox-forked-vmjrb?file=/src/index.tsx

#### Steps to reproduce

1. Click into numeric input

#### Actual behavior

The entire page crashes with the error `_a.setSelectionRange is not a function` which originates here: https://github.com/palantir/blueprint/blob/6906ffce4482d9910cbe311ed9937a2c84e829b8/packages/core/src/components/forms/numericInput.tsx#L347

#### Expected behavior

The component should not throw and error and it should be possible to pass object refs to components.

#### Possible solution

The core of the problem is the type signature on the utility function [`refHandler`](https://github.com/palantir/blueprint/blob/6906ffce4482d9910cbe311ed9937a2c84e829b8/packages/core/src/common/refs.ts#L78-L91). This signature allows for the following call:

```ts
const myRef = useRef<HTMLElement>(null);
const myParent: {el: HTMLElement | null} = null;

refHandler(myParent, "el", myRef);
myParent.el?.contains(otherEl); // throws at runtime
```

Which breaks since `myParent.el` is actually a `RefObject<HTMLElement>` at runtime and _not_ a `HTMLElement`.

#### Related
- https://github.com/palantir/blueprint/pull/4438
- https://github.com/palantir/blueprint/issues/4382

maclockard:
#### Checklist

- [x] Includes tests
- [x] Update documentation

This changes a couple aspects of ref usage. 
* Makes it so that `refHandler` always creates a ref callback and always assigns an element to the local target.
  * This makes `refHandler` (more) type safe to use
  * This also prevents components from incorrectly accessing the local target element.
* Updates the local ref handler when a new ref callback is passed.
  * Previously, when passing a new ref callback it would only get called once, while the old ref callback would continue getting called.
* Properly set the previously passed ref to `null`, according to https://reactjs.org/docs/refs-and-the-dom.html#caveats-with-callback-refs

Changes to the ref utility functions and component lifecycle methods.

### DIFF
--- a/packages/core/src/common/refs.ts
+++ b/packages/core/src/common/refs.ts
@@ -42,6 +42,11 @@ export function setRef<T extends HTMLElement>(refTarget: IRef<T> | undefined | n
     }
 }
 
+/** @deprecated use mergeRefs() instead */
+export function combineRefs<T extends HTMLElement>(ref1: IRefCallback<T>, ref2: IRefCallback<T>) {
+    return mergeRefs(ref1, ref2);
+}
+
 /**
  * Utility for merging refs into one singular callback ref.
  * If using in a functional component, would recomend using `useMemo` to preserve function identity.
--- a/packages/core/src/common/refs.ts
+++ b/packages/core/src/common/refs.ts
@@ -22,51 +22,44 @@ export interface IRefObject<T extends HTMLElement = HTMLElement> {
 }
 
 export function isRefObject<T extends HTMLElement>(value: IRef<T> | undefined | null): value is IRefObject<T> {
-    return value != null && typeof (value as IRefObject<T>).current !== "undefined";
+    return value != null && typeof value !== "function";
 }
 
-export type IRefCallback<T = HTMLElement> = (ref: T | null) => any;
+export type IRefCallback<T extends HTMLElement = HTMLElement> = (ref: T | null) => any;
 
 export function isRefCallback<T extends HTMLElement>(value: IRef<T> | undefined | null): value is IRefCallback<T> {
     return typeof value === "function";
 }
 
-export function combineRefs<T extends HTMLElement>(ref1: IRefCallback<T>, ref2: IRefCallback<T>) {
-    return mergeRefs(ref1, ref2);
-}
-
-export function mergeRefs<T extends HTMLElement>(...refs: Array<IRef<T> | null>): IRefCallback<T> {
-    return value => {
-        refs.forEach(ref => {
-            if (isRefCallback(ref)) {
-                ref(value);
-            } else if (isRefObject(ref)) {
-                ref.current = value;
-            }
-        });
-    };
-}
-
-export function getRef<T extends HTMLElement>(ref: T | IRefObject<T> | null) {
-    if (ref === null) {
-        return null;
-    }
-
-    return (ref as IRefObject<T>).current ?? (ref as T);
-}
-
 /**
  * Assign the given ref to a target, either a React ref object or a callback which takes the ref as its first argument.
  */
-export function setRef<T extends HTMLElement>(refTarget: IRef<T> | undefined, ref: T | null) {
-    if (refTarget === undefined) {
-        return;
-    }
+export function setRef<T extends HTMLElement>(refTarget: IRef<T> | undefined | null, ref: T | null): void {
     if (isRefObject<T>(refTarget)) {
         refTarget.current = ref;
     } else if (isRefCallback(refTarget)) {
         refTarget(ref);
     }
+}
+
+/**
+ * Utility for merging refs into one singular callback ref.
+ * If using in a functional component, would recomend using `useMemo` to preserve function identity.
+ */
+export function mergeRefs<T extends HTMLElement>(...refs: Array<IRef<T> | null>): IRefCallback<T> {
+    return value => {
+        refs.forEach(ref => {
+            setRef(ref, value);
+        });
+    };
+}
+
+export function getRef<T extends HTMLElement>(ref: T | IRefObject<T> | null): T | null {
+    if (ref === null) {
+        return null;
+    }
+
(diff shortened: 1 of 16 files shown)


packages/core/src/common/refs.ts
```diff
@@ -22,53 +22,46 @@ export interface IRefObject<T extends HTMLElement = HTMLElement> {
 }
 
 export function isRefObject<T extends HTMLElement>(value: IRef<T> | undefined | null): value is IRefObject<T> {
-    return value != null && typeof (value as IRefObject<T>).current !== "undefined";
+    return value != null && typeof value !== "function";
```
maclockard, line 25: it is valid for `RefObject.current` to be `undefined` when using refs for something other than HTML elements. This pattern is much more common when using functional components. While Blueprint doesn't really use FCs yet, and the type constraint here _should_ prevent it, this change helps future proof.

packages/core/src/common/refs.ts
```diff
@@ -22,53 +22,46 @@ export interface IRefObject<T extends HTMLElement = HTMLElement> {
...
 
-export type IRefCallback<T = HTMLElement> = (ref: T | null) => any;
+export type IRefCallback<T extends HTMLElement = HTMLElement> = (ref: T | null) => any;
 
 export function isRefCallback<T extends HTMLElement>(value: IRef<T> | undefined | null): value is IRefCallback<T> {
     return typeof value === "function";
 }
 
-export function combineRefs<T extends HTMLElement>(ref1: IRefCallback<T>, ref2: IRefCallback<T>) {
-    return mergeRefs(ref1, ref2);
```
maclockard: This function was duplicative of `mergeRefs` without changing behavior. If trying to avoid breaks, I can add it back with a `@deprecated` annotation.

packages/core/src/common/refs.ts
```diff
@@ -22,53 +22,46 @@ export interface IRefObject<T extends HTMLElement = HTMLElement> {
...
+/**
+ * Assign the given ref to a target, either a React ref object or a callback which takes the ref as its first argument.
+ */
+export function setRef<T extends HTMLElement>(refTarget: IRef<T> | undefined | null, ref: T | null): void {
+    if (isRefObject<T>(refTarget)) {
```
maclockard, line 38: the type guards here prevent this from being called with `undefined | null`, so I removed the `undefined` check

packages/core/src/common/refs.ts
```diff
@@ -78,25 +71,10 @@ export function setRef<T extends HTMLElement>(refTarget: IRef<T> | undefined, re
 export function refHandler<T extends HTMLElement, K extends string>(
     refTargetParent: { [k in K]: T | null },
     refTargetKey: K,
-): IRefCallback<T>;
-export function refHandler<T extends HTMLElement, K extends string>(
-    refTargetParent: { [k in K]: T | IRefObject<T> | null },
```
maclockard, line 83: This type made it pretty easy to cause runtime errors since it does not require the passed type to include `IRefObject<T>`. The problem here being that `k` here should be contravariant, but since typescript doesn't allow for this it defaults to being covariant.

This kind of typing issue is pretty common with refs and Typescript, and leads to things like the bivariance hack in the base react types.

More detail here: https://github.com/palantir/blueprint/issues/4611

packages/datetime/src/dateInput.tsx
```diff
@@ -212,7 +211,7 @@ export class DateInput extends AbstractPureComponent2<IDateInputProps, IDateInpu
                 if (
                     e.key === "Tab" &&
                     !e.shiftKey &&
-                    this.lastTabbableElement.classList.contains(Classes.DATEPICKER_DAY)
+                    this.lastTabbableElement?.classList.contains(Classes.DATEPICKER_DAY)
```
maclockard, line 214: TBH, I'm not sure why the type checker wasn't catching this null call here. It might be worth doing a deeper pass of all current usage of refs to make sure they are properly null checked.

adidahiya: not all packages use strict null checks, unfortunately. this one has that compiler option turned off at the moment.

maclockard: gotcha, makes sense. I'm so used to having that compiler option turned on, I forgot TS doesn't check that by default haha

packages/core/src/common/refs.ts
adidahiya: yeah, let's avoid a break, please add it back with `@deprecated`

adidahiya (changes requested):
lgtm aside from the breaking change

packages/core/src/common/refs.ts
```diff
@@ -22,53 +22,46 @@ export interface IRefObject<T extends HTMLElement = HTMLElement> {
...
+/**
+ * Assign the given ref to a target, either a React ref object or a callback which takes the ref as its first argument.
+ */
+export function setRef<T extends HTMLElement>(refTarget: IRef<T> | undefined | null, ref: T | null): void {
+    if (isRefObject<T>(refTarget)) {
+        refTarget.current = ref;
+    } else if (isRefCallback(refTarget)) {
+        refTarget(ref);
+    }
 }
```
adidahiya, line 44:
```suggestion

/** @deprecated use mergeRefs() instead */
export function combineRefs<T extends HTMLElement>(ref1: IRefCallback<T>, ref2: IRefCallback<T>) {
    return mergeRefs(ref1, ref2);
}

```

adidahiya (approved):

PR #4618, opened 2021-04-01
[popover2] docs: fix Tooltip2 example

nvizzari:
- `ref2` is not defined, and it is supposed to be `ref` renamed
- `combineRefs` only accepts `IRefCallback` arguments, but `ref1` and `ref2` are `IRef`, so need to use `mergeRefs` instead


adidahiya (approved):
thanks @ntrinquier!

PR #4620, opened 2021-04-06
[core] feat(Drawer): add shouldReturnFocusOnClose prop

adidahiya:

This new prop provides a way to disable Drawer's default behavior of returning focus to the previous active element (introduced in https://github.com/palantir/blueprint/pull/4422)

### DIFF
--- a/packages/core/src/components/drawer/drawer.tsx
+++ b/packages/core/src/components/drawer/drawer.tsx
@@ -63,7 +63,7 @@ export interface IDrawerProps extends IOverlayableProps, IBackdropProps, IProps 
      *
      * @default true
      */
-    returnFocusOnClose?: boolean;
+    shouldReturnFocusOnClose?: boolean;
 
     /**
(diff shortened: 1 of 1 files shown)


packages/core/src/components/drawer/drawer.tsx
```diff
@@ -57,6 +57,14 @@ export interface IDrawerProps extends IOverlayableProps, IBackdropProps, IProps
      */
     position?: Position;
 
+    /**
+     * Whether the application should return focus to the last active element in the
+     * document after this drawer closes.
+     *
+     * @default true
+     */
+    returnFocusOnClose?: boolean;
```
adidahiya: "return" sounds like an action / function. the name should probably be `shouldReturnFocusOnClose`

Nantris: First thing that comes to mind for me, since it uses `document.activeElement`, is `focusPreviouslyActiveElementOnClose` but that's pretty damn wordy.

PR #4623, opened 2021-04-07
[v4] fix: remove usage of findDOMNode

### LINKED ISSUE #3979: Removing findDOMNode

adidahiya:
#### Checklist

- [x] Includes tests
- [x] Update documentation

- Remove usage of `React.findDOMNode` in `<Overlay>` by using the new `nodeRef` prop on `<CSSTransition>` instead (we only expect one overlay child, so just one ref on a single transition child should be sufficient).
- Remove usage of `React.findDOMNode` in `<ResizeSensor>` by ~refactoring its implementation into a function component and~ using the `React.cloneElement` approach described in https://github.com/palantir/blueprint/issues/3979#issue-571047705.
- Remove defensive code in `<Portal>` which guarded against `React.createPortal` not being available (since we don't support React 15 anymore)

No regressions

### DIFF
--- a/packages/core/src/common/utils/reactUtils.ts
+++ b/packages/core/src/common/utils/reactUtils.ts
@@ -105,10 +105,3 @@ export function isElementOfType<P = {}>(
         element.type.displayName === ComponentType.displayName
     );
 }
-
-/**
- * Returns React.createRef if it's available, or a ref-like object if not.
- */
-export function createReactRef<T>() {
-    return typeof React.createRef !== "undefined" ? React.createRef<T>() : { current: null };
-}
--- a/packages/core/src/components/html/html.tsx
+++ b/packages/core/src/components/html/html.tsx
@@ -15,24 +15,23 @@
  */
 
 import classNames from "classnames";
-import React from "react";
+import React, { forwardRef } from "react";
 
-import { ElementRefProps } from "../../common";
 import { BLOCKQUOTE, CODE, CODE_BLOCK, HEADING, LABEL, LIST } from "../../common/classes";
 
 function htmlElement<E extends HTMLElement>(
     tagName: keyof JSX.IntrinsicElements,
     tagClassName: string,
-): React.FunctionComponent<React.HTMLProps<E> & ElementRefProps<E>> {
+): React.FC<React.HTMLAttributes<E> & React.RefAttributes<E>> {
     /* eslint-disable-next-line react/display-name */
-    return props => {
-        const { className, elementRef, ...htmlProps } = props;
+    return forwardRef<E, React.HTMLAttributes<E>>((props, ref) => {
+        const { className, ...htmlProps } = props;
         return React.createElement(tagName, {
             ...htmlProps,
             className: classNames(tagClassName, className),
-            ref: elementRef,
+            ref,
         });
-    };
+    });
 }
 
 // the following components are linted by blueprint-html-components because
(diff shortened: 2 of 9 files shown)


adidahiya:
The refactor of ResizeSensor into a function component didn't work out easily without regressions, so I'm reverting it to a class component. We can try to refactor it to use hooks in a future PR. Here's my attempt, for posterity:

```ts

export const ResizeSensor: React.FC<ResizeSensorProps> = props => {
    const elementRef = useRef<HTMLElement>();
    const onlyChild = React.Children.only(props.children);
    const observer = useMemo(() => new ResizeObserver(entries => props.onResize?.(entries)), [props.onResize]);

    useEffect(() => {
        if (elementRef.current instanceof Element) {
            // N.B. observer callback is invoked immediately when observing new elements
            observer.observe(elementRef.current);
        }

        return () => observer.disconnect();
    }, [observer, elementRef]);

    // effect to observe parents
    useEffect(() => {
        if (props.observeParents && elementRef.current instanceof Element) {
            let parent = elementRef.current.parentElement;
            while (parent != null) {
                observer.observe(parent);
                parent = parent.parentElement;
            }
        }
    }, [observer, elementRef, props.observeParents]);

    return cloneElement(onlyChild, { ref: elementRef });
};
ResizeSensor.displayName = `${DISPLAYNAME_PREFIX}.ResizeSensor`;
ResizeSensor.defaultProps = {
    observeParents: false,
};
```

PR #4630, opened 2021-04-12
[core] feat(MultiStepDialog): improve step styling

kmblake:
#### Checklist

- [ ] Includes tests
- [ ] Update documentation

- Only make currently active step blue.
- Adjust right to left panel ratios to match original mocks.
- Set min-width on right panel to prevent overflow
- Make step title flex: 1 to allow clients to provide a title + right-aligned icon for a step.

#### Screenshot

Before this PR:
image

After this PR:
image

### DIFF
--- a/packages/core/src/components/dialog/_multistep-dialog.scss
+++ b/packages/core/src/components/dialog/_multistep-dialog.scss
@@ -121,6 +121,7 @@ $step-radius: $pt-border-radius * 2 !default;
 
 .#{$ns}-dialog-step-title {
   color: $gray1;
+  flex: 1;
   padding-left: 10px;
 
   .#{$ns}-dark & {
--- a/packages/core/src/components/dialog/_multistep-dialog.scss
+++ b/packages/core/src/components/dialog/_multistep-dialog.scss
@@ -129,8 +129,8 @@ $step-radius: $pt-border-radius * 2 !default;
     color: $pt-dark-text-color-disabled;
   }
 
-  // by default, step title is active only when the step is selected
-  .#{$ns}-active.#{$ns}-active.#{$ns}-dialog-step-viewed & {
+  // step title is active only when the step is selected
+  .#{$ns}-active.#{$ns}-dialog-step-viewed & {
     color: $blue4;
   }
 
--- a/packages/core/src/components/dialog/_multistep-dialog.scss
+++ b/packages/core/src/components/dialog/_multistep-dialog.scss
@@ -57,7 +57,7 @@ $step-radius: $pt-border-radius * 2 !default;
     border-bottom: 1px solid $pt-dark-divider-black;
   }
 
-  &.#{$ns}-active {
+  &.#{$ns}-dialog-step-viewed {
     background-color: $white;
     .#{$ns}-dark & {
       background: $dark-gray4;
@@ -79,7 +79,7 @@ $step-radius: $pt-border-radius * 2 !default;
   }
 
   // by default, steps are inactive until they are visited
-  .#{$ns}-active & {
+  .#{$ns}-dialog-step-viewed & {
     background-color: $white;
     cursor: pointer;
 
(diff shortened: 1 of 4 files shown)


packages/core/src/components/dialog/_multistep-dialog.scss
```diff
@@ -110,13 +110,18 @@ $step-radius: $pt-border-radius * 2 !default;
     background-color: $pt-dark-icon-color-disabled;
   }
 
-  .#{$ns}-active & {
+  &.#{$ns}-active {
     background-color: $blue4;
   }
+
+  &.#{$ns}-dialog-step-viewed {
+    background-color: $gray3;
```
invliD, line 119: Does this need to be different from the not-yet-viewed ones? If so: Might need dark support.

packages/core/src/components/dialog/multistepDialog.tsx
```diff
@@ -123,7 +123,14 @@ export class MultistepDialog extends AbstractPureComponent2<IMultistepDialogProp
         return (
             <div className={classNames(Classes.DIALOG_STEP_CONTAINER, { [Classes.ACTIVE]: hasBeenViewed })} key={index}>
                 <div className={Classes.DIALOG_STEP} onClick={this.handleClickDialogStep(index)}>
-                    <div className={Classes.DIALOG_STEP_ICON}>{stepNumber}</div>
+                    <div
+                        className={classNames(Classes.DIALOG_STEP_ICON, {
+                            [Classes.ACTIVE]: currentlySelected,
```
invliD: The container div above is already `Classes.ACTIVE`. That should probably change to only be applied to the current step, and then you can use CSS to target the icon based on that and the new viewed class if necessary (see other comment).

packages/core/src/components/dialog/_multistep-dialog.scss
kmblake: Yes, the intent is that previously viewed steps are a darker gray because, unlike not-yet-viewed steps, you can click on them to navigate to that step. I think the gray3 looks fine in dark mode though:
image

kmblake:
Also adjusted step colors slightly per design feedback:

> "Seen" step title is in the Default text colour
Dark mode matches so "Seen" step title is also in the Default dark mode colour (#F5F8FA, 100%)
I had "Unseen" step title as Muted text before, but I think it makes more sense to use Disabled Default text so it translates to dark mode easier. So for light mode (#5C7080 60%;) and dark mode (#A7B6C2 60%;)

packages/core/src/components/dialog/_multistep-dialog.scss
```diff
@@ -110,21 +111,34 @@ $step-radius: $pt-border-radius * 2 !default;
...
+  flex: 1;
   padding-left: 10px;
 
   .#{$ns}-dark & {
     color: $pt-dark-text-color-disabled;
   }
 
   // by default, step title is active only when the step is selected
-  &.#{$ns}-active {
+  .#{$ns}-active.#{$ns}-active.#{$ns}-dialog-step-viewed & {
```
adidahiya: duplicate class name here. should be:

```suggestion
  .#{$ns}-active.#{$ns}-dialog-step-viewed & {
```

packages/core/src/components/dialog/_multistep-dialog.scss
```diff
@@ -110,21 +111,34 @@ $step-radius: $pt-border-radius * 2 !default;
...
-  color: $gray1;
+  color: $pt-text-color-disabled;
+  flex: 1;
   padding-left: 10px;
 
   .#{$ns}-dark & {
     color: $pt-dark-text-color-disabled;
   }
 
   // by default, step title is active only when the step is selected
```
adidahiya: not sure if "by default" has any meaning here...

```suggestion
  // step title is active only when the step is selected
```

adidahiya (changes requested):
looks good aside from these minor comments

adidahiya (approved):

PR #4632, opened 2021-04-12
alwaysRenderOverflow prop for overflowList

### LINKED ISSUE #4631: Overflow List always render overflow
#### Environment

- __Package version(s)__: 

#### Feature request

If there is no overflow items then the overflow renderer is never run and returns null. We should allow the option to render an overflow item with no passed in items. Main usage will be for a drop down, if the items change then the popover will be re-rendered meaning that the drop down closes. We'd like it to stay open so would still render the menu but with a display none. 

#### Examples
```
<OverflowList
  alwaysRenderOverflow={true}
  overflowRenderer:{(items: T[]) => <Tag>+ {items.length} items</Tag>
/>
```

rhysbrettbowen:
#### Checklist

- [ ] Includes tests
- [X] Update documentation

Adds a prop to always render the overflow for overflow list


adidahiya:
I think there was an issue with your repo config, Circle wasn't reporting any status for any CI jobs -- you may have inadvertently enabled Circle on your fork (see [contributing docs](https://github.com/palantir/blueprint/blob/develop/CONTRIBUTING.md#developing)), which prevents palantir's CI instance from running the checks. I've pushed your branch to this repo, let's switch to this: https://github.com/palantir/blueprint/pull/4648

PR #4633, opened 2021-04-12
className prop moved to outer component element

andreash:
#### Fixes #0000

The Drawer component maps the className prop to an inner `<div>.` It seems to be common - and realized by all other blueprintjs components - that the most outer DOM element receives the className prop in its class name list.

#### Checklist

- [ ] Includes tests
- [ ] Update documentation

Move the className prop to the class name list of the outer `<Overlay>` component.

### DIFF
--- a/packages/core/src/components/drawer/drawer.tsx
+++ b/packages/core/src/components/drawer/drawer.tsx
@@ -132,14 +132,18 @@ export class Drawer extends AbstractPureComponent2<IDrawerProps> {
         const { size, style, position, vertical } = this.props;
         const realPosition = position ? getPositionIgnoreAngles(position) : undefined;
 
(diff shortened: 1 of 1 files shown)


palantirtech:
Thanks for your interest in [palantir/blueprint](https://github.com/palantir/blueprint), @andreash! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

adidahiya:
I'm not sure what problem this actually solves, and it is a breaking change. Also the code doesn't pass lint checks. If you can demonstrate a problem solved by this PR and push code that passes CI checks, I'd be happy to re-open the PR.

PR #4648, opened 2021-04-16
[core] feat(OverflowList): alwaysRenderOverflow prop

### LINKED ISSUE #4631: Overflow List always render overflow

adidahiya:
#### Fixes #4631

Supercedes https://github.com/palantir/blueprint/pull/4632

#### Checklist

- [ ] Includes tests
- [x] Update documentation

Add `alwaysRenderOverflow` prop, which forces the overflow renderer to always be called if true.

This feature is useful in situations where, for example, the overflow dropdown popover is open and the list items change. In this case you may not want the popover to close.

Should this be the default behavior of OverflowList (does it even need to be guarded behind a prop)?

#### Screenshot

TODO


adidahiya:
Going to leave it behind the prop for now for backcompat

PR #4663, opened 2021-04-21
[v4] [core] fix(Portal): use new React context API

adidahiya:

- Remove usage of [legacy context API](https://reactjs.org/docs/legacy-context.html), switch to [stable context API](https://reactjs.org/docs/context.html)
- Expose new `<PortalProvider>` component which supports the `portalClassName` prop (previously, this was `blueprintPortalClassName` on `this.getChildContext()`)

No regressions.

Note that I considered moving the portal container element customization API into `<PortalProvider>`, but this turned out to be needlessly complicated (at least for now) and out of scope for this PR.

### DIFF
--- a/packages/core/src/components/portal/portal.tsx
+++ b/packages/core/src/components/portal/portal.tsx
@@ -19,7 +19,7 @@ import ReactDOM from "react-dom";
 
 import * as Classes from "../../common/classes";
 import { DISPLAYNAME_PREFIX, Props } from "../../common/props";
-import { PortalContext, PortalContextState } from "../../context";
+import { PortalContext, PortalContextState } from "../../context/portal/portalProvider";
 
 export interface PortalProps extends Props {
     /**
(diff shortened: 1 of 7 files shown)


packages/core/src/components/portal/portal.tsx
```diff
@@ -106,18 +89,14 @@ export class Portal extends React.Component<PortalProps, PortalState> {
...
 
-    private createContainerElement() {
+    private createPortalElement() {
         const container = document.createElement("div");
         container.classList.add(Classes.PORTAL);
         maybeAddClass(container.classList, this.props.className);
-        if (this.context != null) {
-            maybeAddClass(container.classList, this.context.blueprintPortalClassName);
-        }
+        maybeAddClass(container.classList, (this.context as PortalContextState)?.portalClassName);
```
adidahiya, line 99: This is kind of awkward, but there's no better way around it with the new context API when used in class components. I guess React really wants you to use function components. We can remove this type cast when we refactor this component to be a function.

PR #4681, opened 2021-04-27
[v4] feat: convert icons to FCs, wrap with forwardRef

adidahiya:

- Convert `<Icon>` and auto-generated icon components in packages/icons/src/generated into function components
  - :warning: breaking change: `Icon` no longer has static constants on its public API; use `IconSize` from `@blueprintjs/icons` (or `@blueprintjs/core`, where it is re-exported) instead.
- Wrap these components with `React.forwardRef()` so that they can be more easily used as children of `<ResizeSensor>`

- No regressions
- ~Is removal of `Icon.SIZE_STANDARD` and `Icon.SIZE_LARGE` constants ok?~
  - Deprecation done on develop with #4688


adidahiya:
- [x] Need to update based on changes committed in https://github.com/palantir/blueprint/pull/4688

PR #4683, opened 2021-04-28
Add @blueprintjs/stylelint-plugin package and no-prefix-literal rule

p-szm:
#### Checklist

- [x] Includes tests
- [x] Update documentation

This PR adds a stylelint rule that bans the usage of `.bp3`/`.bp4` prefixes in favor of using the `bp-ns` variables. The rule comes with an autofixer that automatically converts the literal prefixes to the `bp-ns` variable and also inserts the Blueprint variables import.

#### Screenshot

bp-stylelint


packages/stylelint-plugin/src/rules/no-prefix-literal.ts
```diff
@@ -0,0 +1,147 @@
...
+import { checkImportExists } from "../utils/checkImportExists";
+import { insertImport } from "../utils/insertImport";
+
+const ruleName = "@blueprintjs/no-prefix-literal";
+
+const messages = stylelint.utils.ruleMessages(ruleName, {
+    expected: (unfixed: string, fixed: string) => `Use the \`${fixed}\` variable instead of the \`${unfixed}\` literal`,
+});
+
+const bannedPrefixes = ["bp1", "bp2", "bp3", "bp4"];
```
styu: i would just ban bp, bp3, and bp4. the first 2 don't exist afaik

packages/stylelint-plugin/src/rules/no-prefix-literal.ts
```diff
@@ -0,0 +1,147 @@
...
+                    // Check that the keys and their values are correct
+                    const allowedKeys = new Set<string>(Object.values(CssSyntax).filter(v => v !== CssSyntax.OTHER));
+                    return Object.keys(obj).every(key => allowedKeys.has(key) && typeof (obj as any)[key] === "string");
+                },
+            },
+        },
+    );
+
+    if (!validOptions) {
+        return;
```
styu, line 72: if the options are invalid, does this mean it will silently fail?

packages/stylelint-plugin/src/rules/no-prefix-literal.ts
```diff
@@ -0,0 +1,147 @@
...
+    }
+
+    const disableFix = options?.disableFix ?? false;
+
+    const cssSyntax = getCssSyntax(root.source?.input.file || "");
+    if (cssSyntax === CssSyntax.OTHER) {
+        return;
+    }
+
+    let hasBpVariablesImport: boolean | undefined; // undefined means not checked yet
```
styu, line 82: I'm assuming this code runs per-file? (asking mostly since I've never written a stylelint rule before :) )

packages/stylelint-plugin/src/rules/no-prefix-literal.ts
```diff
@@ -0,0 +1,147 @@
...
+    OTHER = "other",
+}
+
+const BpPrefixVariableMap: Record<Exclude<CssSyntax, CssSyntax.OTHER>, string> = {
+    [CssSyntax.SASS]: "#{$ns}",
+    [CssSyntax.LESS]: "@{ns}",
+};
+
+const BpVariableImportMap: Record<Exclude<CssSyntax, CssSyntax.OTHER>, string> = {
+    [CssSyntax.SASS]: "~@blueprntjs/core/lib/scss/variables",
```
styu:
```suggestion
    [CssSyntax.SASS]: "~@blueprintjs/core/lib/scss/variables",
```

packages/stylelint-plugin/src/utils/checkImportExists.ts
```diff
@@ -0,0 +1,41 @@
...
+limitations under the License.*/
+
+import type { Root } from "postcss";
+
+/**
+ * Returns true if the given import exists in the file, otherwise returns false.
+ */
+export function checkImportExists(root: Root, importPath: string): boolean {
+    let hasBpVarsImport = false;
+    root.walkAtRules(/^import$/i, atRule => {
```
styu, line 24: does the AST at this point have stripped off the `@` from `@import "~@blueprintjs/core/lib/scss/variables";` ?

packages/stylelint-plugin/test/no-prefix-literal.test.js
```diff
@@ -0,0 +1,57 @@
...
+
+it("Warns when .bp3 is present", async () => {
+    const result = await stylelint.lint({
+        files: "test/fixtures/contains-bp3.scss",
+        config,
+    });
+    expect(result.errored).to.be.true;
+});
+
+it("Warns when nested .bp3 is present", async () => {
```
styu:
```suggestion
it("Warns when nested .bp3 is present even when not first selector", async () => {
```

packages/stylelint-plugin/test/no-prefix-literal.test.js
```diff
@@ -0,0 +1,57 @@
...
+
+const config = {
+    plugins: ["@blueprintjs/stylelint-plugin"],
+    rules: {
+        "@blueprintjs/no-prefix-literal": true,
+        "at-rule-semicolon-newline-after": "always",
+    },
+};
+
+it("Warns when .bp3 is present", async () => {
```
styu: is there a way to test the autofix portion?

styu: could we also test css module syntax? e.g.

```
.my-class {
  :global(.bp3-card) {
  }
}
```

should trigger as well

packages/stylelint-plugin/src/rules/no-prefix-literal.ts
p-szm: 👍

packages/stylelint-plugin/src/rules/no-prefix-literal.ts
p-szm: VS code will display an error in the bottom right corner. This is also the pattern that all stylelint plugins that I found use

packages/stylelint-plugin/src/rules/no-prefix-literal.ts
p-szm: Yes. That is my understanding

packages/stylelint-plugin/src/rules/no-prefix-literal.ts
p-szm: Great catch. Thank you

packages/stylelint-plugin/src/utils/checkImportExists.ts
p-szm: Yes

packages/stylelint-plugin/test/no-prefix-literal.test.js
p-szm: I spent an hour trying to figure out how to test the autofixer and I couldn't figure it out. The problem is that when `fix: true` is passed to stylelint then it modifies the original file which we don't want. If someone finds a working example online I'd love to test it, but otherwise I could try adding tests for some of the functions that I'm using to perform autofixing (like inserting imports and etc)

styu: here's one example i found: https://github.com/hudochenkov/stylelint-order/blob/master/rules/properties-alphabetical-order/tests/index.js

i care more about catching the css module syntax than the fixer testing if that helps (unit testing the individual functions sound ok to me too)

styu (approved):

packages/stylelint-plugin/test/no-prefix-literal.test.js
p-szm: This uses jest which blueprint unfortunately doesn't use. I added some tests for the util functions that we use to do autofixing

packages/stylelint-plugin/src/index.ts
```diff
@@ -0,0 +1,18 @@
+/* Copyright 2020 Palantir Technologies, Inc. All rights reserved.
```
adidahiya: nit: formatting of block comment is slightly off (is the autofixer broken?)

packages/stylelint-plugin/src/index.ts
```diff
@@ -0,0 +1,18 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.*/
+
+import noPrefixLiteral from "./rules/no-prefix-literal";
+
+// eslint-disable-next-line import/no-default-export
```
adidahiya: we can just turn this rule off in packages/stylelint-plugin/.eslintrc.json

packages/stylelint-plugin/test/checkImportExists.test.js
```diff
@@ -0,0 +1,70 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.*/
+
+// eslint-disable-next-line import/no-extraneous-dependencies
+const { expect } = require("chai");
+const postcss = require("postcss");
+
+const { checkImportExists } = require("../lib/utils/checkImportExists");
+
+it("Returns false if no imports exist", () => {
```
adidahiya: should be in a `describe()` block... I'm surprised this works without one

packages/stylelint-plugin/test/no-prefix-literal.test.js
adidahiya: I'm open to adding Jest in a future PR for certain BP packages if it helps us leverage more things in the ecosystem

adidahiya (commented):
overall looks good! After seeing this rule fixer work in practice, I think we should introduce a more verbose variable which consumers import... `$ns` feels a little to terse to be littering all over BP consumer code via the autofixer. it could be a simple alias in core/src/common/_variables.scss:

```scss
$ns: bp3 !default;
// alias for BP users outside this repo
$bp-ns: $ns;
```

styu:
@adidahiya i love the idea of introducing `$bp-ns` or something like that instead of requiring `$ns`, since the latter is fairly generic in retrospect

packages/stylelint-plugin/src/index.ts
p-szm: I fixed this manually, but actually it seems like this is also a valid format and the autofixer always does it this way on my computer. Could you test what happens on your machine?

p-szm: It looks like the plugin only inserts starts at the beginning and end https://github.com/Stuk/eslint-plugin-header/blob/66f5269c8a7e3282afd02b914b502f7bb95fe702/lib/rules/header.js#L46

Do you have a template for ts/js files configured in VS code by any chance?

adidahiya (approved):

packages/stylelint-plugin/README.md
```diff
@@ -0,0 +1,69 @@
...
+}
+```
+
+```diff
+-.bp3-button > div {
+-    border: 1px solid black;
+-}
++ @import "~@blueprntjs/core/lib/scss/variables";
++
++.#{$ns}-button > div {
```
styu:
```suggestion
+.#{$bp-ns}-button > div {
```

packages/stylelint-plugin/README.md
```diff
@@ -0,0 +1,69 @@
...
+        "@blueprintjs/no-prefix-literal": true
+    }
+}
+```
+
+## Rules
+
+### `@blueprintjs/no-prefix-literal` (autofixable)
+
+Enforce usage of the `ns` constant over namespaced string literals.
```
styu:
```suggestion
Enforce usage of the `bp-ns` constant over namespaced string literals.
```

packages/stylelint-plugin/README.md
```diff
@@ -0,0 +1,69 @@
...
+
+The `@blueprintjs` package exports a `ns` CSS variable which contains the prefix for the current version of Blueprint (`bp3` for Blueprint 3, `bp4` for Blueprint 4, and etc). Using the variable instead of hardcoding the prefix means that your code will still work when new major version of Blueprint is released.
```
styu:
```suggestion
The `@blueprintjs` package exports a `bp-ns` CSS variable which contains the prefix for the current version of Blueprint (`bp3` for Blueprint 3, `bp4` for Blueprint 4, and etc). Using the variable instead of hardcoding the prefix means that your code will still work when new major version of Blueprint is released.
```

packages/stylelint-plugin/test/fixtures/does-not-contain-bp3.module.scss
```diff
@@ -0,0 +1,5 @@
+.some-class {
+  :global(.#{$ns})-button {
```
styu:
```suggestion
  :global(.#{$bp-ns})-button {
```

packages/stylelint-plugin/test/fixtures/does-not-contain-bp3.scss
```diff
@@ -0,0 +1,5 @@
+.some-class {
+  .#{$ns}-button {
```
styu:
```suggestion
  .#{$bp-ns}-button {
```

adidahiya (approved):

PR #4687, opened 2021-04-29
fix: make refs compatible with React refs

kuoruan:

This change uses React's refs definition, should fix following type error:

```
Type 'Ref<HTMLInputElement>' is not assignable to type 'IRef<HTMLInputElement> | undefined'.
  Type 'null' is not assignable to type 'IRef<HTMLInputElement> | undefined'.
```

This error shows when I try to pass a `React.Ref` to blueprint's `inputRef`

React's Ref contains `null`, so we should add it too.

### DIFF
--- a/packages/core/src/common/refs.ts
+++ b/packages/core/src/common/refs.ts
@@ -14,27 +14,30 @@
  * limitations under the License.
  */
 
-export type IRef<T extends HTMLElement = HTMLElement> = IRefObject<T> | IRefCallback<T>;
+import * as React from "react";
 
-// compatible with React.Ref type in @types/react@^16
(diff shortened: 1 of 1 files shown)


palantirtech:
Thanks for your interest in [palantir/blueprint](https://github.com/palantir/blueprint), @kuoruan! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

adidahiya:
@kuoruan can you file an issue and create a sandbox demonstrating the problem first? in general I think you can work around the issues you're having by using `React.useRef(null)`. Let's discuss the problem and potential solution in an issue first before opening this PR.

PR #4688, opened 2021-04-29
[core] deprecate Icon, Drawer, Spinner size constant static members

adidahiya:

In conjunction with upcoming changes in v4.0 where class components are being refactored into function components (like https://github.com/palantir/blueprint/pull/4681), we must remove support for static members and provide those constants as new symbols in the public API.

Deprecated:

- `Icon.SIZE_STANDARD`
- `Icon.SIZE_LARGE`
- `Drawer.SIZE_SMALL`
- `Drawer.SIZE_STANDARD`
- `Drawer.SIZE_LARGE`
- `Spinner.SIZE_SMALL`
- `Spinner.SIZE_STANDARD`
- `Spinner.SIZE_LARGE`

Added: 

- `IconSize`
- `DrawerSize`
- `SpinnerSize`

Not addressed in this PR:

- `OverflowList.ofType()`, `Select.ofType()`, etc...

### DIFF
--- a/packages/core/src/components/button/abstractButton.tsx
+++ b/packages/core/src/components/button/abstractButton.tsx
@@ -27,7 +27,7 @@ import {
     MaybeElement,
     Utils,
 } from "../../common";
-import { Icon, IconName } from "../icon/icon";
+import { Icon, IconName, IconSize } from "../icon/icon";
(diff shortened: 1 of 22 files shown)


adidahiya:
Need to add spinner size constants

PR #4693, opened 2021-05-01
fix(setState usage): doesn't clobber existing keys

### LINKED ISSUE #4629: Bug in documentation example
https://github.com/palantir/blueprint/blob/develop/packages/docs-app/src/examples/table-examples/tableEditableExample.tsx#L128

When Lines 128 and 134 run, they delete the other fields in this state object.

#### Steps to reproduce
Run your documentation code and try to edit a cell in the browser.

#### Actual behavior
Error is thrown in `renderColumnHeader` since `state.columnNames[columnIndex]` does not exist once either `setSparseState` or `setArrayState` is ran.

#### Expected behavior
No error.

#### Possible solution
I resolved this by first ensuring previous state is destructured into the new state object, before setting the key & value of the field being replaced/updated like so, in both `setSparseState` and `setArrayState`:

`setState({ ...state, [stateKey]: values })`

(The docs just have:   `this.setState({ [key]: values });` on lines 128 and 134 which wipes any field not being updated from state.)

caffeinateoften:
When setting state from `setArrayState` and `setSparseState`, set existing state values first, and then overwrite the intended field with `stateKey`, without clobbering over all existing state values that are not `stateKey`.

#### Checklist

- [x] Includes tests
- [x] Update documentation (well technically this _is_ the update to the documentation (code example))

Fixes two bugs in EditableTable component examples in Blueprintjs docs.

Do doc examples have associated test cases? Where are they? (My bad if I'm being blind).

#### Screenshot
n/a

### DIFF
--- a/packages/docs-app/src/examples/table-examples/tableEditableExample.tsx
+++ b/packages/docs-app/src/examples/table-examples/tableEditableExample.tsx
@@ -125,12 +125,12 @@ export class TableEditableExample extends React.PureComponent<IExampleProps, ITa
     private setArrayState<T>(key: string, index: number, value: T) {
         const values = (this.state as any)[key].slice() as T[];
         values[index] = value;
-        this.setState({ [key]: values });
+        this.setState({ ...this.state, [key]: values });
     }
 
     private setSparseState<T>(stateKey: string, dataKey: string, value: T) {
         const stateData = (this.state as any)[stateKey] as { [key: string]: T };
         const values = { ...stateData, [dataKey]: value };
-        this.setState({ [stateKey]: values });
+        this.setState({ ...this.state, [stateKey]: values });
     }
 }


palantirtech:
Thanks for your interest in [palantir/blueprint](https://github.com/palantir/blueprint), @caffeinateoften! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

caffeinateoften:
For https://github.com/palantir/blueprint/issues/4629

caffeinateoften:
> Thanks for your interest in [palantir/blueprint](https://github.com/palantir/blueprint), @caffeinateoften! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

I signed this. Not sure if something should signify that on here or not, not sure why merging is blocked.

caffeinateoften:
@Adidahiya EDIT: I approved policy bot

caffeinateoften:
uhh Looks like Palantir's policy-bot is now asking for permission to act on behalf of my github account.
I'm not willing to permit that. Is there something else that can be done?

adidahiya:
This is already how `this.setState()` works in a React component. It only updates the specified key(s) and does not alter/remove existing state keys.

> Fixes two bugs in EditableTable component examples in Blueprintjs docs.

What bugs?

caffeinateoften:
Hm, is it because I used functional components and the behavior of setState
changes if using useState hooks?

The behavior I saw didn’t update only the specified keys, but if I recall
correctly I used `useState` And kept everything else the same.

I see now in Reactjs docs that `this.setState` actually performs a shallow
merge and erroneously assumed a `setState` call from `useState` would
behave the same.

Thanks for clarifying- this is not a bug!

PR #4699, opened 2021-05-06
[stylelint-plugin] fix: handle less import references

p-szm:
This fixes an edge case with less references that I found.


adidahiya (approved):

p-szm:
👉 poke policy bot

p-szm:
@adidahiya would you mind merging this for me? Policy bot is stuck

PR #4717, opened 2021-05-12
[popover2] feat(ContextMenu2): support ref prop

adidahiya:

Use `React.forwardRef` to forward refs to the context menu container element


packages/popover2/src/contextMenu2.tsx
```diff
@@ -227,22 +229,22 @@ export const ContextMenu2: React.FC<ContextMenu2Props> = ({
             contentProps,
             onContextMenu: handleContextMenu,
             popover: maybePopover,
-            ref: containerRef,
+            ref: mergeRefs(containerRef, userRef),
```
adidahiya: we probably don't need this. the user has the ref already so we don't need to send it back to them in the child render callback

adidahiya:
```suggestion
            ref: containerRef,
```

PR #4723, opened 2021-05-17
[core] fix(EditableText): optimize updateInputDimensions

SatishGandham:
#### Checklist

- [ ] Includes tests
- [ ] Update documentation

`componentDidUpdate` of Editable text calls `updateInputDimensions` on every rerender, this triggers the layout calculations.
This PR addresses that issue by only calling `updateInputDimensions`  only when the props it depends on change

ref: https://github.com/palantir/blueprint/pull/4639#issuecomment-820503242

Confirm if calling `updateInputDimensions` conditionally is a safe change

### DIFF
--- a/packages/core/src/components/editable-text/editableText.tsx
+++ b/packages/core/src/components/editable-text/editableText.tsx
@@ -283,7 +283,9 @@ export class EditableText extends AbstractPureComponent2<EditableTextProps, IEdi
             this.props.maxLines !== prevProps.maxLines ||
             this.props.minLines !== prevProps.minLines ||
             this.props.minWidth !== prevProps.minWidth ||
-            this.props.multiline !== prevProps.multiline
+            this.props.multiline !== prevProps.multiline ||
+            this.state.value !== prevState.value ||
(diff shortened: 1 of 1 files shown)


adidahiya:
@SatishGandham thanks for taking my suggestion and running with it. Looks like it is an incomplete implementation and will need to be adjusted / tweaked to get the right behavior. There are a few problems with this approach right now.

(1) The input is not expanding its height past its minLines to accommodate longer text:

image

(2) When switching to `alwaysRenderInput={false}`, the component collapses completely:

adidahiya (changes requested):
see my feedback above

SatishGandham:
@adidahiya 
Is `alwaysRenderInput` expected to change during components lifetime?

SatishGandham:
@adidahiya, Made the necessary changes.
We are now watching `state.value` and `props.alwaysRenderInput`

adidahiya (approved):

adidahiya:
thanks @SatishGandham!

iamakulov:
Thank you @SatishGandham 🙌 Never got time to finish the implementation in #4639 myself.

PR #4729, opened 2021-05-20
[core] fix(Portal): handle more className edge cases

### LINKED ISSUE #4519: portalClassName breaks with empty string
#### Environment

- __Package version(s)__:  Most recent
- __Operating System__:  Mac
- __Browser name and version__:  Google beta

#### Code Sandbox

``` js
// Some portal component (Dialog, Popover, etc.):
...
portalClassName={theme}
...
// Update theme somewhere in the app:
setSettings({
    ...settings,
    theme: theme === '' ? 'bp3-dark' : '',
});
```

#### Steps to reproduce

Shown above, global state for theme with two values ('') and ('bp3-dark')

#### Actual behavior

``` js
// Line 112 - 117 in blueprint/packages/core/src/components/portal/portal.tsx

        if (this.portalElement != null && prevProps.className !== this.props.className) {
            if (prevProps.className !== undefined) {
                this.portalElement.classList.remove(prevProps.className);
            }
            maybeAddClass(this.portalElement.classList, this.props.className);
        }

```

Here, the app breaks at line 114. Why? Because prevProps.className === '' rather than undefined and '' classList.remove breaks on ''.

Proof:

Screenshot 2021-02-05 at 12 03 42 AM

#### Expected behavior

Either it handles the empty string or a better error message... something like you cannot input an empty string into the portalClassName (especially when you are utilising a button that switches the theme). I fixed it by using 'bp3-dark' and 'light' but this is not ideal as i'm adding a useless class.

#### Possible solution

Mentioned above. I'd do it myself but the build process is quite time-consuming. Just want to put this out there as I wasted quite a bit of time confused about this and the solution is very simple.

Thanks, jerry

jkillian:
#### Checklist

- [x] Includes tests
- [ ] Update documentation

Updates how `Portal` removes classes so that it will not crash when 1) an empty string was previously provided or 2) when removing multiple classes at once.

Any logic errors in removing the class names.

### DIFF
--- a/packages/core/src/components/portal/portal.tsx
+++ b/packages/core/src/components/portal/portal.tsx
@@ -111,9 +111,7 @@ export class Portal extends React.Component<IPortalProps, IPortalState> {
     public componentDidUpdate(prevProps: IPortalProps) {
         // update className prop on portal DOM element
         if (this.portalElement != null && prevProps.className !== this.props.className) {
-            if (prevProps.className !== undefined) {
-                this.portalElement.classList.remove(prevProps.className);
-            }
+            maybeRemoveClass(this.portalElement.classList, prevProps.className);
             maybeAddClass(this.portalElement.classList, this.props.className);
         }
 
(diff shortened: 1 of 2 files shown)


adidahiya (approved):
thanks @jkillian!

jkillian:
👋 😄

PR #4730, opened 2021-05-20
[core] feat(MultiStepDialog): Allow skipping to a specific step in the dialog

styu:
#### Checklist

- [ ] Includes tests
- [x] Update documentation

- Allows for providing an optional `initialStepIndex` prop which will start the dialog off at the specified step index (0-indexed)
- If the index < 0, it will start from the first step
- If the index exceeds the number of steps, it will start from the last step

Is there more validation I should be doing?

#### Screenshot

dialog

### DIFF
--- a/packages/core/src/components/dialog/multistepDialog.tsx
+++ b/packages/core/src/components/dialog/multistepDialog.tsx
@@ -232,7 +232,7 @@ export class MultistepDialog extends AbstractPureComponent2<MultistepDialogProps
     }
 
     private getInitialIndexFromProps(props: MultistepDialogProps) {
-        if (props.initialStepIndex) {
+        if (props.initialStepIndex !== undefined) {
             const boundedInitialIndex = Math.max(
                 0,
                 Math.min(props.initialStepIndex, this.getDialogStepChildren(props).length - 1),
--- a/packages/core/src/components/dialog/multistepDialog.tsx
+++ b/packages/core/src/components/dialog/multistepDialog.tsx
@@ -57,6 +57,8 @@ export interface IMultistepDialogProps extends DialogProps {
     /**
      * Whether to reset the dialog state to its initial state on close.
      * By default, closing the dialog will reset its state.
+     *
+     * @default true
      */
     resetOnClose?: boolean;
 
@@ -77,11 +79,6 @@ const PADDING_BOTTOM = 0;
 
 const MIN_WIDTH = 800;
 
-const INITIAL_STATE = {
-    lastViewedIndex: 0,
-    selectedIndex: 0,
-};
-
 @polyfill
 export class MultistepDialog extends AbstractPureComponent2<MultistepDialogProps, IMultistepDialogState> {
     public static displayName = `${DISPLAYNAME_PREFIX}.MultistepDialog`;
@@ -89,15 +86,10 @@ export class MultistepDialog extends AbstractPureComponent2<MultistepDialogProps
     public static defaultProps: Partial<MultistepDialogProps> = {
         canOutsideClickClose: true,
         isOpen: false,
+        resetOnClose: true,
     };
 
-    public state: IMultistepDialogState;
-
-    constructor(props: MultistepDialogProps) {
-        super(props);
-
-        this.state = this.getInitialIndexFromProps(props);
-    }
+    public state: IMultistepDialogState = this.getInitialIndexFromProps(this.props);
 
     public render() {
         return (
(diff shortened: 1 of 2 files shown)


packages/core/src/components/dialog/multistepDialog.tsx
```diff
@@ -84,7 +91,13 @@ export class MultistepDialog extends AbstractPureComponent2<MultistepDialogProps
         isOpen: false,
     };
 
-    public state: IMultistepDialogState = INITIAL_STATE;
+    public state: IMultistepDialogState;
+
+    constructor(props: MultistepDialogProps) {
+        super(props);
+
+        this.state = this.getInitialIndexFromProps(props);
```
adidahiya: I think this can actually stay in the class member initializer, no constructor necessary:

```ts
public state: IMultistepDialogState = this.getInitialIndexFromProps(this.props);
```

packages/core/src/components/dialog/multistepDialog.tsx
```diff
@@ -225,6 +238,21 @@ export class MultistepDialog extends AbstractPureComponent2<MultistepDialogProps
     private getDialogStepChildren(props: MultistepDialogProps & { children?: React.ReactNode } = this.props) {
         return React.Children.toArray(props.children).filter(isDialogStepElement);
     }
+
+    private getInitialIndexFromProps(props: MultistepDialogProps) {
+        if (props.initialStepIndex) {
```
adidahiya: nit: `props.initialStepIndex !== undefined`

packages/core/src/components/dialog/multistepDialog.tsx
```diff
@@ -59,6 +59,13 @@ export interface IMultistepDialogProps extends DialogProps {
      * By default, closing the dialog will reset its state.
```
adidahiya, line 59:
```suggestion
     * By default, closing the dialog will reset its state.
     *
     * @default true
```

packages/core/src/components/dialog/multistepDialog.tsx
```diff
@@ -225,6 +238,21 @@ export class MultistepDialog extends AbstractPureComponent2<MultistepDialogProps
...
+            const boundedInitialIndex = Math.max(
+                0,
+                Math.min(props.initialStepIndex, this.getDialogStepChildren(props).length - 1),
+            );
+            return {
+                lastViewedIndex: boundedInitialIndex,
+                selectedIndex: boundedInitialIndex,
+            };
+        } else {
+            return INITIAL_STATE;
```
adidahiya: can we please inline this value here and get rid of the variable? it doesn't really make sense as this variable name anymore

packages/core/src/components/dialog/multistepDialog.tsx
```diff
@@ -84,7 +91,13 @@ export class MultistepDialog extends AbstractPureComponent2<MultistepDialogProps
         isOpen: false,
```
adidahiya, line 88:
```suggestion
        isOpen: false,
        resetOnClose: true,
```

packages/core/src/components/dialog/multistepDialog.tsx
```diff
@@ -103,7 +116,7 @@ export class MultistepDialog extends AbstractPureComponent2<MultistepDialogProps
             !prevProps.isOpen &&
```
adidahiya, line 108: with the changes I suggested above, we can get rid of the `prevProps.resetOnClose === undefined` check. I think that will make this easier to read (it took me a second to figure out what's going on)...

in addition, I think we should check if `props.initialStepIndex` changed, and reset the state in that case too...

```ts
if (
    (prevProps.resetOnClose && !prevProps.isOpen && this.props.isOpen) ||
    (prevProps.initialStepIndex !== this.props.initialStepIndex)
) {
```

adidahiya (approved):
works fine in the PR preview

adidahiya (changes requested):
sorry, I meant to request changes

packages/core/src/components/dialog/multistepDialog.tsx
styu: hm... it feels weird to change the current index if the `initialStepIndex` changes while it's still open?

styu: what about
```
if (
            (prevProps.resetOnClose || prevProps.initialStepIndex !== this.props.initialStepIndex) &&
            !prevProps.isOpen &&
            this.props.isOpen
        ) {
```

adidahiya (approved):

PR #4735, opened 2021-05-26
[popover2] close popovers/tooltips on ContextMenu2 opening

### LINKED ISSUE #4719: ContextMenu2 does not dismiss tooltip on open
#### Environment

- __Package version(s)__: popover2 0.8.0, core 3.44.2
- __Operating System__: macOS

#### Code Sandbox

https://codesandbox.io/s/blueprint-sandbox-forked-w5nv6?file=/src/CoreExample.tsx

#### Steps to reproduce

Attach a ContextMenu2 to the same target as a Tooltip or Tooltip2, as in the above sandbox.

Hover over and right click on the various tags in the sandbox.

#### Actual behavior

The ones using ContextMenu2 do not dismiss their tooltips when opened

image

#### Expected behavior

They _should_ dismiss their tooltips when opened, like the tags using the legacy `@ContextMenuTarget` API:

image

adidahiya:
#### Checklist

- [x] Includes tests
- [x] Update documentation

Add special event handling to Popover2 so that parent and child popovers/tooltips of a ContextMenu2 will get dismissed when the custom context menu is opened.

No regressions

### DIFF
--- a/packages/popover2/src/contextMenu2.tsx
+++ b/packages/popover2/src/contextMenu2.tsx
@@ -174,6 +174,7 @@ export const ContextMenu2: React.FC<ContextMenu2Props> = React.forwardRef<any, C
         menu === undefined ? undefined : (
             <Popover2
                 {...popoverProps}
+                autoFocus={false}
                 content={
                     // this prevents right-clicking inside our context menu
                     <div onContextMenu={cancelContextMenu}>{menu}</div>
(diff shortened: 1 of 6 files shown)


packages/popover2/src/contextMenu2.tsx
```diff
@@ -174,6 +174,7 @@ export const ContextMenu2: React.FC<ContextMenu2Props> = React.forwardRef<any, C
         menu === undefined ? undefined : (
             <Popover2
                 {...popoverProps}
+                autoFocus={false}
```
adidahiya, line 177: this is the crux of the bugfix, but it caused a regression in https://github.com/palantir/blueprint/issues/4742

PR #4734, opened 2021-05-26
Add dialog role and tests

jabsatz:
#### Checklist

- [x] Includes tests

- [ ] Update documentation

#### Changes proposed in this pull request:

Add dialog role to the Dialog component.

#### Reviewers should focus on:

Correctness of the test, since the change itself pretty simple.

PR #4739, opened 2021-05-27
Add dialog role and tests

### LINKED ISSUE #4528: Dialogs should have proper a11y markup
#### Feature request

Dialogs should have `role="dialog"`, `aria-labeledby`, and optionally `aria-describedby` attributes for accessibility. At the very least there should be docs about how to configure this markup properly.

#### References

- https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/dialog_role

jabsatz:
Recreating pull https://github.com/palantir/blueprint/pull/4734 so it correctly builds in CircleCI

#### Checklist

- [x] Includes tests

- [ ] Update documentation

Add dialog role to the Dialog component.

Correctness of the test, since the change itself pretty simple.

### DIFF
--- a/packages/core/src/components/dialog/dialog.tsx
+++ b/packages/core/src/components/dialog/dialog.tsx
@@ -90,7 +90,7 @@ export class Dialog extends AbstractPureComponent2<DialogProps> {
     public render() {
         return (
             <Overlay {...this.props} className={Classes.OVERLAY_SCROLL_CONTAINER} hasBackdrop={true}>
-                <div className={Classes.DIALOG_CONTAINER}>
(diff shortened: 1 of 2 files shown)


jabsatz:
Looks like some tests are effectively breaking with this change. I wasn't able to test the project properly yesterday because of some environment issues on my home PC. I was hoping since the change was small they would pass directly in CircleCI, but unfortunately I didn't have such luck. 
I will try again later today from another computer and see if I can build it and test it correctly before pushing another fix.

adidahiya:
Closing for now, happy to re-open if you push a new commit

PR #4740, opened 2021-05-27
[popover2] fix(ContextMenu2): simpler target positioning logic

adidahiya:
#### Checklist

- [x] Includes tests
- [x] Update documentation

- Wrap ContextMenu2's generated popover target in a `<Portal>` to make its implementation align more closely with the legacy ContextMenuTarget API.
	- This obviates the need for attaching our own container ref and querying its containing positioning block with `Classes.FIXED_POSITIONING_CONTAINING_BLOCK`, making the API simpler for users in many use cases
	- This fixes a bug where the offset calculation was wrong in cases where the context menu child scrolled around in its container and/or the developer was not able to attach `Classes.FIXED_POSITIONING_CONTAINING_BLOCK` to the most relevant DOM element (for example, when using third-party libraries)
- Deprecate `Classes.FIXED_POSITIONING_CONTAINING_BLOCK`
- ❌ BREAKING CHANGE: Removed `ref` from `ContextMenu2ChildrenProps`

No regressions in ContextMenu2 and Tree examples in docs

#### Screenshot

image

image

### DIFF
--- a/packages/core/src/common/classes.ts
+++ b/packages/core/src/common/classes.ts
@@ -78,7 +78,10 @@ export const LIST_UNSTYLED = `${NS}-list-unstyled`;
 export const RTL = `${NS}-rtl`;
 
 // layout utilities
-/** @see https://developer.mozilla.org/en-US/docs/Web/CSS/Containing_block#identifying_the_containing_block */
+/**
+ * @see https://developer.mozilla.org/en-US/docs/Web/CSS/Containing_block#identifying_the_containing_block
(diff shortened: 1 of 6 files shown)


adidahiya:
~The semantics of `targetOffset` have changed, but now they actually match what ContextMenuTarget (the "v1" API) was doing:~

edit: nvm, this was always the case with ContextMenu2

image

adidahiya:
there is a bug when scrolled, we should actually move back to using clientX/Y:

image

PR #4744, opened 2021-06-01
[popover2] fix(ContextMenu2): Tooltip2 and autofocus interactions

### LINKED ISSUE #4742: ContextMenu2 does not close on escape keypress
#### Environment

- __Package version(s)__: popover2 v0.10.0
- __Operating System__: macOS
- __Browser name and version__: Firefox / Chrome

#### Code Sandbox

https://codesandbox.io/s/blueprint-sandbox-forked-epowp?file=/src/index.tsx

#### Steps to reproduce

1. Right click on button to open context menu
1. Press ESC key

#### Actual behavior

Context menu stays open

#### Expected behavior

Context menu should close, as it does with the legacy ContextMenuTarget API

#### Possible solution

[This line change](https://github.com/palantir/blueprint/pull/4735/files#diff-f3fe6a0c83840dc0f252cf79338758d8fe174cb4921f1fc865f3a00bc05129b9R177) fixed #4719, but it caused context menu popovers to no longer autofocus, so ESC keypresses no longer close the overlay unless the user explicitly clicks into the menu to move focus inside it. We'll need to figure out a different way to fix #4719.

adidahiya:
#### Checklist

- [x] Includes tests
- [x] Update documentation

An alternative (much more complicated) fix for #4719, which avoids the regression described in #4742.

I was not able to implement #4719 by simply using mouse events when `autoFocus={true}` on the ContextMenu2 Popover2. For some reason the React event system seems to propagate mouse enter events to the target after the ContextMenu2 is opened, and that re-opens the tooltip on the shared target.

So I decided to try this alternative approach where ContextMenu2 and Tooltip2 communicate with each other via React context.

In the `<ContextMenu2><Tooltip2>...</Tooltip2></ContextMenu2>` case, the parent ContextMenu2 creates a Tooltip2Context with `forceDisable: true`, and the child Tooltip2 reacts to that by consuming the context state.

In the `<Tooltip2><ContextMenu2>...</ContextMenu2></Tooltip2>` case, the child ContextMenu2 dispatches actions to the parent Tooltip2 context, telling it to force disable itself.

Is this React context approach overkill?

### DIFF
--- a/packages/popover2/src/contextMenu2.tsx
+++ b/packages/popover2/src/contextMenu2.tsx
@@ -132,7 +132,8 @@ export const ContextMenu2: React.FC<ContextMenu2Props> = React.forwardRef<any, C
 
     // ancestor Tooltip2Context state doesn't affect us since we don't care about parent ContextMenu2s, we only want to
     // force disable parent Tooltip2s in certain cases through dispatching actions
-    const [, popover2Dispatch] = React.useContext(Tooltip2Context);
+    // N.B. any calls to this dispatch function will be no-ops if there is no Tooltip2Provider ancestor of this component
+    const [, tooltipCtxDispatch] = React.useContext(Tooltip2Context);
     // click target offset relative to the viewport (e.clientX/clientY), since the target will be rendered in a Portal
     const [targetOffset, setTargetOffset] = React.useState<Offset | undefined>(undefined);
     // hold a reference to the click mouse event to pass to content/child render functions
@@ -145,7 +146,7 @@ export const ContextMenu2: React.FC<ContextMenu2Props> = React.forwardRef<any, C
     // for this component (that will lead to unpredictable behavior).
     React.useEffect(() => {
         setIsOpen(false);
-        popover2Dispatch({ type: "RESET_DISABLED_STATE" });
+        tooltipCtxDispatch({ type: "RESET_DISABLED_STATE" });
     }, [disabled]);
 
     const cancelContextMenu = React.useCallback((e: React.SyntheticEvent<HTMLDivElement>) => e.preventDefault(), []);
@@ -154,7 +155,7 @@ export const ContextMenu2: React.FC<ContextMenu2Props> = React.forwardRef<any, C
         if (!nextOpenState) {
             setIsOpen(false);
             setMouseEvent(undefined);
-            popover2Dispatch({ type: "RESET_DISABLED_STATE" });
+            tooltipCtxDispatch({ type: "RESET_DISABLED_STATE" });
         }
     }, []);
 
@@ -191,6 +192,7 @@ export const ContextMenu2: React.FC<ContextMenu2Props> = React.forwardRef<any, C
                 // when offset changes, to force recomputing position.
                 key={getPopoverKey(targetOffset)}
                 hasBackdrop={true}
+                backdropProps={{ className: Classes.CONTEXT_MENU2_BACKDROP }}
                 isOpen={isOpen}
                 minimal={true}
                 onInteraction={handlePopoverInteraction}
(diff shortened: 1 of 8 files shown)


adidahiya:
I probably need to write some more unit tests testing various nesting scenarios, making sure that multiple contexts don't interact with each other unexpectedly.

packages/popover2/src/tooltip2.tsx
```diff
@@ -87,7 +88,24 @@ export class Tooltip2<T> extends React.PureComponent<Tooltip2Props<T>> {
     private popover: Popover2<T> | null = null;
 
     public render() {
-        const { children, intent, popoverClassName, ...restProps } = this.props;
+        // if we have an ancestor Tooltip2Context, we should take its state into account in this render path,
+        // it was likely created by a parent ContextMenu2
+        return (
+            <Tooltip2Context.Consumer>
+                {([state]) => <Tooltip2Provider initialState={state}>{this.renderPopover}</Tooltip2Provider>}
```
styu: what happens if you have nested tooltips? (acknowledging that it's not the best UX, but not ruling out that its possibility 😬 )

styu (commented):
ooc why is it that we need to do this for context menus but not popovers? e.g. in the scenario where you nest `<Tooltip2><Popover2></Popover2></Tooltip2>`

packages/popover2/src/contextMenu2.tsx
```diff
@@ -129,6 +130,9 @@ export const ContextMenu2: React.FC<ContextMenu2Props> = React.forwardRef<any, C
         ...restProps
     } = props;
 
+    // ancestor Tooltip2Context state doesn't affect us since we don't care about parent ContextMenu2s, we only want to
+    // force disable parent Tooltip2s in certain cases through dispatching actions
+    const [, popover2Dispatch] = React.useContext(Tooltip2Context);
```
adidahiya:
```suggestion
    const [, tooltip2Dispatch] = React.useContext(Tooltip2Context);
```

should be renamed here and below...

packages/popover2/src/tooltip2.tsx
adidahiya: I don't think it would be a possibility in terms of tooltips nested inside tooltip _content_, but nesting of tooltip targets could be possible (just like nesting of ContextMenu2 targets, as demonstrated in the ContextMenu2 docs example). I'll add a test for it

styu: oops yes, i meant targets, thanks for clarifying!

adidahiya:
@styu 

> ooc why is it that we need to do this for context menus but not popovers? e.g. in the scenario where you nest `<Tooltip2><Popover2></Popover2></Tooltip2>`

so there is actually some brittle code in Popover/Popover2 which force disables direct Tooltip/Tooltip2 children here: https://github.com/palantir/blueprint/blob/b2f3862a51ea9147532461412a5fe9da919fb3cd/packages/popover2/src/popover2.tsx#L352

the above code breaks when there is even one level of indirection (consider a reusable component which renders a Tooltip around something). In places where I'm upgrading to ContextMenu2 I knew that I would need a more robust solution than this existing Popover > Tooltip pattern, so I opted for the React context approach. I may need to consider doing the same thing for Popover2 <> Tooltip2 interactions...

adidahiya:
I've added tests for some nesting scenarios, and verified these scenarios work in docs-app:





I didn't change the Popover2 > Tooltip2 disable behavior in this PR because I didn't want to introduce scope creep here. Might revisit that in a future PR.

styu (approved):

packages/popover2/src/tooltip2Context.tsx
```diff
@@ -0,0 +1,67 @@
...
+};
+
+interface Tooltip2ProviderProps {
+    children: React.ReactNode | ((ctxState: Tooltip2ContextState) => React.ReactNode);
+    forceDisable?: boolean;
+}
+
+export const Tooltip2Provider = ({ children, forceDisable }: Tooltip2ProviderProps) => {
+    const [state, dispatch] = React.useReducer(tooltip2Reducer, {});
+    // // if we have a parent context controlling our state, just use its value and don't use the local state
```
adidahiya: remove commented code

adidahiya:
Editor's note: this feature was straightforward enough to implement, but notoriously difficult to test. I think I would have more success / an easier time with [React Testing Library](https://testing-library.com/docs/react-testing-library/intro/). Blueprint's testing infra is due for a refresh :/

PR #4756, opened 2021-06-09
[popover2] fix(ContextMenu2): detect dark theme correctly

### LINKED ISSUE #4743: ContextMenu2 does not detect dark theme target correctly
#### Environment

popover2 v0.10.0

#### Actual behavior

image

#### Expected behavior

the popover should use dark theme

adidahiya:
#### Checklist

- [x] Includes tests
- [x] Update documentation

- Fix ContextMenu2 dark theme detection by restoring an inline `<div>` rendered as the popover target, so that we can attach a ref to it and properly query for dark theme. Before this change, we were checking for dark theme on the "virtual" target we rendered inside a `<Portal>`, which will almost always be an invalid check since applications rarely add `.bp3-dark` to the `<body>` element.
- 🔥 BREAKING CHANGE: rename `Classes.CONTEXT_MENU2_POPOVER2_TARGET` to `Classes.CONTEXT_MENU2_VIRTUAL_TARGET`
- 🔥 BREAKING CHANGE: add `ref` property _back_ to `ContextMenu2ChildrenProps` after it was removed in #4740 (sorry for the API whiplash). If upgrading from popover2 `> 0.8.0 < 0.11.0` and you are using the advanced children render function API, you'll need to attach this ref to whatever you return from the children render function.

N/A

### DIFF
--- a/packages/docs-app/src/examples/core-examples/contextMenu2Example.tsx
+++ b/packages/docs-app/src/examples/core-examples/contextMenu2Example.tsx
@@ -76,6 +76,7 @@ const GraphNode: React.FC = () => {
                     "docs-context-menu-open": props.contentProps.isOpen,
                 })}
                 onContextMenu={props.onContextMenu}
+                ref={props.ref}
             >
                 {props.popover}
             </div>
--- a/packages/popover2/src/_context-menu2.scss
+++ b/packages/popover2/src/_context-menu2.scss
@@ -7,6 +7,6 @@
   display: block;
 }
 
-.#{$ns}-context-menu2-popover2-target {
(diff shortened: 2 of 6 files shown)


adidahiya:
This solution doesn't work because the popper ref needs to be attached to the root node returned from `renderTarget`, it can't be on a descendant "virtual element". Need to re-introduce the [ref which was removed](https://github.com/palantir/blueprint/commit/a649dedfbcd7dd5e61218571ef3b5bd74f33db87#diff-f3fe6a0c83840dc0f252cf79338758d8fe174cb4921f1fc865f3a00bc05129b9L74) in #4740.

PR #4767, opened 2021-06-16
[core] feat(HotkeysProvider): support advanced nesting use cases

adidahiya:
#### Checklist

- [ ] Includes tests
- [x] Update documentation

- Add `value` prop to `<HotkeysProvider>`. If supplied, the component will re-use an existing context instance.
- Export `HotkeysContextInstance` type
- Fix HotkeysTarget2 example (piano keys were not working because the event handlers were not bound correctly in the example code)


packages/core/src/context/hotkeys/hotkeys-provider.md
```diff
@@ -43,6 +43,74 @@ ReactDOM.render(
 );
 ```
 
+@## Advanced usage
+
+HotkeysProvider should not be nested, except in special cases. If you have a rendering boundary within your application
+through which React context is not preserved (for example, a plugin system which uses `ReactDOM.render()`) and you wish
+to use hotkeys in a descendant part of the tree below such a boundary, you may render a descendant provider and hydrate it
```
adidahiya: not sure if "hydrate" is the right word choice. might be too overloaded.

packages/core/src/context/hotkeys/hotkeys-provider.md
```diff
@@ -43,6 +43,74 @@ ReactDOM.render(
...
+with the root context instance. Without this context plumbing, you would end up with two "global" hotkeys dialogs in an
```
adidahiya: paragraph structure is slightly awkward. this presents a potential problem after the solution. maybe we should just clearly state the intended outcome, which is a single global hotkeys dialog no matter how many HotkeysProviders are rendered in the application.

packages/core/src/context/hotkeys/hotkeys-provider.md
```diff
@@ -43,6 +43,84 @@ ReactDOM.render(
 );
 ```
 
+@## Advanced usage
+
+HotkeysProvider should not be nested, except in special cases. If you have a rendering boundary within your application
+through which React context is not preserved (for example, a plugin system which uses `ReactDOM.render()`) and you wish
+to use hotkeys in a descendant part of the tree below such a boundary, you may render a descendant provider and initialize
+it with the root context instance. This ensure that there will only be one "global" hotkeys dialogs in an application
+taht has multiple HotkeysProviders.
```
adidahiya:
```suggestion
it with the root context instance. This ensures that there will only be one "global" hotkeys dialogs in an application
which has multiple HotkeysProviders.
```

packages/core/src/context/hotkeys/hotkeysProvider.tsx
```diff
@@ -31,14 +31,23 @@ type HotkeysAction =
...
+// N.B. we can remove this optional call guard once Blueprint depends on React 16
+/**
+ * A React context used to register and deregister hotkeys as components are mounted and unmounted in an application.
+ * Users should take care to make sure that only _one_ of these is instantiated and used within an application, especially
+ * if using global hotkeys.
+ *
+ * You will likely not be using this HotkeysContext directly, except in cases where you need to get a direct handle on an
+ * exisitng context instance for advanced use cases involving nested HotkeysProviders.
+ *
+ * For more information, see the HotkeysProvider documentation.
```
adidahiya:
```suggestion
 * For more information, see the [HotkeysProvider documentation](https://blueprintjs.com/docs/#core/context/hotkeys-provider).
```

PR #4768, opened 2021-06-17
[core] fix(InputGroup): adjust classes for leftElement span

### LINKED ISSUE #4535: leftElement inside InputGroup inside ControlGroup is invisible
#### Environment

- __Package version(s)__:  "@blueprintjs/core": "^3.39.0"
- __Operating System__: macOS Big Sur 11.1
- __Browser name and version__: Brave Browser v1.20.103, Firefox: 85.0.2

#### Code Sandbox

Link to a minimal repro (fork [this code sandbox](https://codesandbox.io/s/blueprint-sandbox-et9xy)): https://codesandbox.io/s/blueprint-sandbox-forked-3cr0u

#### Steps to reproduce

1. Create a `ControlGroup` 
2. Create a `InputGroup` with `leftElement`
3. Insert `<Icon icon="tag" />` (Any icon is reproducible)

OR  - 
Just click the forked codesandbox

#### Actual behavior

Following the example from [here](https://github.com/palantir/blueprint/blob/develop/packages/docs-app/src/examples/core-examples/inputGroupExample.tsx).
Tag icon should be shown on left element but instead invisible

#### Expected behavior

Icon on left element should be visible. 

#### Possible solution

~~`z-index` needs to be greater than or equal to 3 in order for it to show.~~
When input is in focus state, it has a `z-index` at 14, and therefore icon's `z-index` needs to be greater than or equal to 15 in order for it to show.

image

image

ejose19:
#### Checklist

- [ ] Includes tests
- [ ] Update documentation

Reuse existing logic provided for `Classes.INPUT_ACTION`, which correctly handles `z-index` for `rightElement` when inside a `ControlGroup` (due to higher specificity), without affecting `Classes.INPUT_LEFT_CONTAINER` when not inside a `ControlGroup` element.

### DIFF
--- a/packages/core/src/components/forms/inputGroup.tsx
+++ b/packages/core/src/components/forms/inputGroup.tsx
@@ -264,7 +264,10 @@ export class InputGroup extends AbstractPureComponent2<InputGroupProps2, IInputG
 
         if (leftElement != null) {
             return (
-                <span className={Classes.INPUT_LEFT_CONTAINER} ref={this.refHandlers.leftElement}>
+                <span
+                    className={classNames(Classes.INPUT_LEFT_CONTAINER, Classes.INPUT_ACTION)}
+                    ref={this.refHandlers.leftElement}
+                >
                     {leftElement}
                 </span>
             );


adidahiya:
This isn't really the right fix, since it creates ambiguity around "input action". I know that's probably not the best name for what it is ("something on the right side of the input"), but it's what we have now. https://github.com/palantir/blueprint/pull/4771 is the correct fix.

PR #4790, opened 2021-07-01
[table] feat: add Table2 and EditableCell2, use new hotkeys API

### LINKED ISSUE #4789: HotkeysProvider does not work properly when Table is rendered on the same page
#### Environment

- __Package version(s)__: core 3.46.0, table 3.8.34
- __Operating System__: macOS

#### Steps to reproduce

1. Use HotkeysProvider, register some global hotkeys
1. Render a Table on the same page
1. Open the hotkeys dialog with "?"

#### Actual behavior

You get two dialogs open, one of them is empty (the one triggered by the old `@HotkeysTarget` API, used in `<Table>`).

You can even repro this in the documentation "Table features" page

image

#### Expected behavior

The two hotkeys APIs should not interfere

#### Possible solution

Converting Table to HotkeysTarget2 seems to work (despite some duplicate keys displayed), but we can’t introduce a breaking change to Table right now (requiring HotkeysProvider in place)... so we can either
- create a new Table2 component which uses HotkeysTarget2, or
- edit HotkeysEvents to not display the dialog if there are no hotkeys to show (this is a bit of a band-aid)

Since we’re going to need to migrate people to a new version of Table which uses new hotkeys anyway, we should probably make Table2...

adidahiya:
#### Checklist

- [x] Includes tests
- [x] Update documentation

- Abstract out hotkeys interaction logic from Table into a plain class (not a React component) called `TableHotkeys`
	- We instantiate this class in the Table constructor, and keep it up to date with Table props, state, and its Grid as necessary
- Create a new component, `<Table2>`, which copies much of its implementation from Table, but has the enhancement of using `<HotkeysTarget2>` instead of `@HotkeysTarget`
	- This new component requires that you configure your application with a `<HotkeysProvider>` somewhere.
- Create a new compoennt, `<EditableCell2>`, which copies much of its implementation from EditableCell, but has the enhancement of using `<HotkeysTarget2>` instead of `@HotkeysTarget`
- Use Table2 and EditableCell2 in all the docs-app examples

Note that #4789 contains some discussion of the tradeoffs of alternative solutions to the problem.

- All the docs examples should work with no regressions


packages/table/test/tableTests.tsx
```diff
@@ -1493,7 +1493,9 @@ describe("<Table>", function (this) {
         }
     });
 
-    describe("Autoscrolling when rows/columns decrease in count or size", () => {
+    // HACKHACK: these tests were not running their assertions correctly for a while, and when that
```
adidahiya, line 1496: note: these tests should get fixed in a future PR

PR #4799, opened 2021-07-05
[v4] [core] feat(PanelStack): remove prop injection

invliD:
The prop injection only really existed because the old (pre-v4) PanelStack API required users to pass in a component. That component would be injected with the PanelProps and the custom additional props. Now that we've switched to a render function-based API, there is no more reason to do this. You can just refactor `openPanel({ props: { someProp: "value" }, renderPanel: PanelComponent)` to `openPanel({ renderPanel: panelProps => <PanelComponent {...panelProps} someProp="value" /> })`.

This is a breaking change, which is why I'm not proposing it to PanelStack2 on develop. I believe breaks are still ok on the v4 branch, since 4.0.0 has not been released.

### DIFF
--- a/packages/core/src/components/panel-stack/panel-stack.md
+++ b/packages/core/src/components/panel-stack/panel-stack.md
@@ -18,15 +18,15 @@ React trees mounted, change the `renderActivePanelOnly` prop.
 
 @## Panels
 
-Panels are supplied as `Panel<T>` objects, where `renderPanel` and `props` are
-used to render the panel element and `title` will appear in the header and back button.
-This breakdown allows the component to avoid cloning elements.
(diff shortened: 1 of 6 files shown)


adidahiya:
Wow so clean! In general, this sounds like a good idea. Can you please update `panel-stack.md` as well?
This means we'll need to add PanelStack3 to `

adidahiya:
Wow so clean! In general, this sounds like a good idea. Can you please update `panel-stack.md` as well?

This change means we'll need to add PanelStack3 to develop for a smooth upgrade path... I can look into that after this PR merges.

btw, we're using `next` instead of `v4` branch now (it allows us to release alpha/beta versions).

adidahiya (approved):

PR #4800, opened 2021-07-05
[core] fix(PanelStack2): animation direction in controlled mode

### LINKED ISSUE #4686: PanelStack controlled mode transitions
#### Environment

- __Package version(s)__: 3.43.0
- __Browser and OS versions__: Google Chrome, macOS Big Sur 11.3

#### Question

In controlled mode, the panel stack doesnt seem like its animating correctly. It currently on has the push animation (pushing a new panel onto the stack); however if I remove a panel from the stack array, it "pushes" the older panel on rather than animating it in from the lefthand side.

invliD:
#### Fixes #4686

I have a cherry-pick onto v4's renamed PanelStack ready locally (based on #4799, so not pushing just yet).

### DIFF
--- a/packages/core/src/components/panel-stack2/panelStack2.tsx
+++ b/packages/core/src/components/panel-stack2/panelStack2.tsx
@@ -15,8 +15,7 @@
  */
 
 import classNames from "classnames";
-import * as React from "react";
-import { useEffect, useMemo, useRef } from "react";
+import React, { useEffect, useMemo, useRef } from "react";
 import { CSSTransition, TransitionGroup } from "react-transition-group";
 
 import { Classes, DISPLAYNAME_PREFIX, Props } from "../../common";
--- a/packages/core/src/components/panel-stack2/panelStack2.tsx
+++ b/packages/core/src/components/panel-stack2/panelStack2.tsx
@@ -15,7 +15,7 @@
  */
 
 import classNames from "classnames";
-import React, { useEffect, useMemo, useRef } from "react";
+import * as React from "react";
 import { CSSTransition, TransitionGroup } from "react-transition-group";
 
 import { Classes, DISPLAYNAME_PREFIX, Props } from "../../common";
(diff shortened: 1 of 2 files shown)


packages/core/src/components/panel-stack2/panelStack2.tsx
```diff
@@ -15,7 +15,7 @@
  */
 
 import classNames from "classnames";
-import * as React from "react";
+import React, { useEffect, useMemo, useRef } from "react";
```
adidahiya: please revert this line, we need to continue using the namespace import, because otherwise it's a breaking change to force consumers to use esModuleInterop + allowSyntheticDefaultImports

adidahiya (changes requested):
the gist of the change seems reasonable, but it causes issues when `renderActivePanelOnly={false}`, the page crashes:

invliD:
@adidahiya That crash is actually not caused by this PR. I've found two issues that would cause a crash due to inconsistent number of hooks called:
1. Hooks are called after this early return:
    https://github.com/palantir/blueprint/blob/55f89fa7aa79a0bfe95120a8295cfb5b7d1b12a8/packages/core/src/components/panel-stack2/panelStack2.tsx#L95-L99

1. The `renderPanel` function is _called_, not added as a JSX.Element. This is the issue the example is running into when using controlled mode (in fact just the 2-line change switching the example to controlled mode reproduces the issue). If a component uses hooks, it _must_ be created as a JSX.Element, otherwise the hook calls are attributed to the parent component. In that case, unless the current panel or all panels _together_ (depending on renderActivePanelOnly) call the same number of hooks, we'll run into that crash.
    https://github.com/palantir/blueprint/blob/55f89fa7aa79a0bfe95120a8295cfb5b7d1b12a8/packages/core/src/components/panel-stack2/panelView2.tsx#L89-L93
    I believe the three options we have are either (1) pretend renderPanel is a component and call createElement for it, (2) wrap each renderPanel call in a simple FC that essentially takes ownership of potential hook calls, or (3) break typings of renderPanel to disallow passing of an FC itself, always requiring a function. Not sure that one's possible.

Either way, given that the issue already exists, let's fix that in a follow-up.

adidahiya (approved):

adidahiya:
Filed https://github.com/palantir/blueprint/issues/4807 for follow-up bug

PR #4808, opened 2021-07-15
[core] fix(PanelStack2): more stable controlled mode

### LINKED ISSUE #4807: PanelStack2 crashes due to inconsistent number of hooks called
#### Environment

- __Package version(s)__: core 3.46.0

#### Code Sandbox

Link to a minimal repro: https://github.com/palantir/blueprint/pull/4800#pullrequestreview-703348256

#### Steps to reproduce

1. Render a controlled PanelStack2 with `renderActivePanelOnly={false}`

#### Actual behavior

Component crashes with React error

#### Expected behavior

No crash

#### Possible solution

From @invliD in [comment](https://github.com/palantir/blueprint/pull/4800#issuecomment-878204362):

> 1. Hooks are called after this early return:
>   https://github.com/palantir/blueprint/blob/55f89fa7aa79a0bfe95120a8295cfb5b7d1b12a8/packages/core/src/components/panel-stack2/panelStack2.tsx#L95-L99
> 
> 1. The `renderPanel` function is _called_, not added as a JSX.Element. This is the issue the example is running into when using controlled mode (in fact just the 2-line change switching the example to controlled mode reproduces the issue). If a component uses hooks, it _must_ be created as a JSX.Element, otherwise the hook calls are attributed to the parent component. In that case, unless the current panel or all panels _together_ (depending on renderActivePanelOnly) call the same number of hooks, we'll run into that crash.
>  https://github.com/palantir/blueprint/blob/55f89fa7aa79a0bfe95120a8295cfb5b7d1b12a8/packages/core/src/components/panel-stack2/panelView2.tsx#L89-L93
> I believe the three options we have are either (1) pretend renderPanel is a component and call createElement for it, (2) wrap each renderPanel call in a simple FC that essentially takes ownership of potential hook calls, or (3) break typings of renderPanel to disallow passing of an FC itself, always requiring a function. Not sure that one's possible.

adidahiya:
#### Checklist

- [x] Includes tests
- [x] Update documentation

- Avoid React "inconsistent number of hook called" errors with two changes to the PanelStack2 implementation:
  - Move PanelStack2's early `return` in its render path to _after_ all hooks are called
  - Wrap each `renderPanel()` call in a simple functional component wrapper to encapsulate the lifecycle of any hooks potentially called inside those functions
- Fix the PanelStack2 docs example state management to append and remove panels from the _end_ of the stack array, rather than the beginning
  - The test suite and docs example were inconsistent about this. I believe the test suite had the correct behavior, and I've updated the docs to point this out for controlled mode users.

- Any performance implications of the new `<PanelWrapper />`

### DIFF
--- a/packages/core/src/components/panel-stack2/panelView2.tsx
+++ b/packages/core/src/components/panel-stack2/panelView2.tsx
@@ -76,12 +76,15 @@ export const PanelView2: PanelView2Component = <T extends Panel<object>>(props: 
     // instantiated with a type unrelated to our generic constraint `T` here. We know
     // we're sending the right values here though, and it makes the consumer API for this
     // component type safe, so it's ok to do this...
-    const PanelWrapper: React.FunctionComponent = () =>
-        props.panel.renderPanel({
-            closePanel: handleClose,
-            openPanel: props.onOpen,
-            ...props.panel.props,
-        } as PanelProps<T>);
+    const PanelWrapper: React.FunctionComponent = React.useMemo(
+        () => () =>
+            props.panel.renderPanel({
+                closePanel: handleClose,
+                openPanel: props.onOpen,
+                ...props.panel.props,
+            } as PanelProps<T>),
+        [props.panel.renderPanel, handleClose, props.onOpen, props.panel.props],
+    );
 
     return (
         <div className={Classes.PANEL_STACK2_VIEW}>
--- a/packages/core/src/components/panel-stack2/panelView2.tsx
+++ b/packages/core/src/components/panel-stack2/panelView2.tsx
@@ -68,6 +68,21 @@ export const PanelView2: PanelView2Component = <T extends Panel<object>>(props: 
             />
         );
 
+    // `props.panel.renderPanel` is simply a function that returns a JSX.Element. It may be an FC which
+    // uses hooks. In order to avoid React errors due to inconsistent hook calls, we must encapsulate
+    // those hooks with their own lifecycle through a very simple wrapper component.
+
+    // N.B. A type cast is required because of error TS2345, where technically `panel.props` could be
+    // instantiated with a type unrelated to our generic constraint `T` here. We know
+    // we're sending the right values here though, and it makes the consumer API for this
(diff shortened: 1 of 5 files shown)


packages/core/src/components/panel-stack2/panelView2.tsx
```diff
@@ -68,6 +68,21 @@ export const PanelView2: PanelView2Component = <T extends Panel<object>>(props:
...
 
+    // `props.panel.renderPanel` is simply a function that returns a JSX.Element. It may be an FC which
+    // uses hooks. In order to avoid React errors due to inconsistent hook calls, we must encapsulate
+    // those hooks with their own lifecycle through a very simple wrapper component.
+
+    // N.B. A type cast is required because of error TS2345, where technically `panel.props` could be
+    // instantiated with a type unrelated to our generic constraint `T` here. We know
+    // we're sending the right values here though, and it makes the consumer API for this
+    // component type safe, so it's ok to do this...
+    const PanelWrapper: React.FunctionComponent = () =>
```
invliD: You'll probably want to useMemo this so you don't end up with a different component on every render.

packages/core/test/panel-stack2/panelStack2Tests.tsx
```diff
@@ -68,199 +68,201 @@ describe("<PanelStack2>", () => {
...
-
-        const firstPanelHeader = panelStackWrapper.findClass(Classes.HEADING);
-        assert.exists(firstPanelHeader);
-        assert.equal(firstPanelHeader.at(0).text(), "Test Title");
-    });
+        it("can render a panel stack with multiple initial panels and close one", () => {
+            let stack: Array<Panel<TestPanelInfo>> = [initialPanel, { renderPanel: TestPanel, title: "New Panel 1" }];
+            panelStackWrapper = renderPanelStack({
+                onClose: () => {
+                    stack = stack.slice(0, stack.length - 1);
```
invliD:
```suggestion
                    stack = stack.slice(0, -1);
```

PR #4816, opened 2021-07-22
[popover2] fix(ContextMenu2): remove obsolete target CSS

invliD:
The `context-menu` class was previously applied to the context menu itself. In `ContextMenu2`, it's applied to the _target_. Since it's not using a direct child operator, _any_ descendant popover target inside a context menu target will end up getting `display: block`.

Given that this was previously not set on the target, and there doesn't appear to ever _be_ a `popover-target` for the context menu, we may be able to just remove this entirely.

### DIFF
--- a/packages/popover2/src/_context-menu2.scss
+++ b/packages/popover2/src/_context-menu2.scss
@@ -3,7 +3,7 @@
 
 @import "./common";
 
-.#{$ns}-context-menu2 .#{$ns}-popover2-target {
+.#{$ns}-context-menu2 > .#{$ns}-popover2-target {
   display: block;
 }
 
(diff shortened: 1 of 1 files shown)


adidahiya (changes requested):
I think you're right, we can remove this `display: block` style entirely. It's not doing anything useful anymore. I believe it was added to allow the `FIXED_POSITIONING_CONTAINING_BLOCK` layout to work in https://github.com/palantir/blueprint/commit/0064ea4793beed967d842f32a1cfcac32dda4926, but it was never removed as it should have been in https://github.com/palantir/blueprint/pull/4740.

adidahiya (approved):

PR #4824, opened 2021-07-27
[icons] feat: add stadium-geometry + area-of-interest, edit send-message

candyyyji:
#### Checklist

- [ ] Includes tests
- [ ] Update documentation

Added new icon for:
- area of interest
- stadium

Edited icon for:
- send message

Everything renders ok and right files were changed

#### Screenshot

Screen Shot 2021-07-27 at 11 43 31 AM
Screen Shot 2021-07-27 at 11 43 39 AM


adidahiya:
@candyyyji the send-message icon doesn't look any different? am I missing something?

[current live docs](https://blueprintjs.com/docs/#icons):

image


[this PR preview](https://45159-71939872-gh.circle-artifacts.com/0/packages/docs-app/dist/index.html#icons):

image


but the new ones look fine:

image

image

btw are we sure this minimal representation of a stadium will be clear enough? IMO it looks like a generic oval, something to do with shapes and drawing rather than a location. have we considered a less minimal representation of a stadium?

candyyyji:
allen asked me to move two vertices by half pixel so the corners snapped to whole pixels... you shouldn't be able to see any difference... 😬

adidahiya:
cool. I'm still curious about my question from above:

> have we considered a less minimal representation of a stadium?

candyyyji:
oh sorry i missed that question! 
it's for "holding pattern" in our application https://en.wikipedia.org/wiki/Holding_(aeronautics), but holding pattern isn't generally applicable elsewhere so i found that the mathematical word for this shape, which is "stadium" https://mathworld.wolfram.com/Stadium.html

candyyyji:
we could change the name of this icon though! in case in future we need an actual stadium?

adidahiya:
Discussed offline, renamed to "stadium-geometry"

PR #4832, opened 2021-08-02
[core,popover2] Fix heading colors in tooltips in dark mode

### LINKED ISSUE #3773: Headings in tooltips in dark mode are unreadable
#### Environment

- __Package version(s)__: 3.17.1
- __Browser and OS versions__: MacOS 10.14.6, Chrome 77

If possible, link to a minimal repro: https://codesandbox.io/embed/blueprint-sandbox-b69x5

#### Steps to reproduce

1. Add Classes.DARK to a div.
1. Add a Tooltip as a child.
1. Add a heading to the tooltip contents.

#### Actual behavior

Screen Shot 2019-10-09 at 10 53 23 AM

#### Expected behavior

Screen Shot 2019-10-09 at 10 55 44 AM

#### Possible solution

- Dark background for tooltips
- Special case heading colors inside tooltips

tomshen:
#### Checklist

- [ ] Includes tests N/A
- [ ] Update documentation N/A

- Uses text color to set heading color explicitly to override default header colors in dark mode tooltips
- Fixed in both soon-to-be-deprecated Tooltip and Tooltip2

- Does this lead to any unexpected styling issues, since this changes the popover appearance mixin?
- Should I add a "test" example for this case? i.e. adding an example to the tooltip docs that uses an header

Here is a Tooltip2 repro: https://codesandbox.io/s/blueprint-sandbox-forked-u378j

#### Screenshot

Before:

Screen Shot 2021-08-02 at 3 24 45 PM

After:

Screen Shot 2021-08-02 at 3 24 31 PM

### DIFF
--- a/packages/core/src/components/popover/_common.scss
+++ b/packages/core/src/components/popover/_common.scss
@@ -139,6 +139,10 @@ $dark-popover-background-color: $dark-gray4 !default;
 
   .#{$ns}-popover-content {
     background: $background-color;
+  }
+
+  .#{$ns}-popover-content,
+  .#{$ns}-heading {
     color: $text-color;
   }
 
--- a/packages/core/src/components/popover/_common.scss
+++ b/packages/core/src/components/popover/_common.scss
@@ -141,6 +141,9 @@ $dark-popover-background-color: $dark-gray4 !default;
     background: $background-color;
   }
 
+  // Some popovers (like tooltips) invert their foreground/background colors relative
+  // to what we expect for the theme. In those cases, we need to override global typography
+  // styles to get the right colors.
   .#{$ns}-popover-content,
   .#{$ns}-heading {
     color: $text-color;
--- a/packages/popover2/src/_common.scss
+++ b/packages/popover2/src/_common.scss
@@ -58,6 +58,9 @@ $pt-dark-tooltip2-box-shadow: $pt-dark-popover-box-shadow !default;
     background: $background-color;
   }
 
+  // Some popovers (like tooltips) invert their foreground/background colors relative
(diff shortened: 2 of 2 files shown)


packages/core/src/components/popover/_common.scss
```diff
@@ -139,6 +139,10 @@ $dark-popover-background-color: $dark-gray4 !default;
 
   .#{$ns}-popover-content {
     background: $background-color;
+  }
+
+  .#{$ns}-popover-content,
```
adidahiya, line 147:
```suggestion
  // Some popovers (like tooltips) invert their foreground/background colors relative
  // to what we expect for the theme. In those cases, we need to override global typography
  // styles to get the right colors.
  .#{$ns}-popover-content,
```

packages/popover2/src/_common.scss
```diff
@@ -56,6 +56,10 @@ $pt-dark-tooltip2-box-shadow: $pt-dark-popover-box-shadow !default;
 
   .#{$ns}-popover2-content {
     background: $background-color;
+  }
+
+  .#{$ns}-popover2-content,
```
adidahiya, line 64:
```suggestion
  // Some popovers (like tooltips) invert their foreground/background colors relative
  // to what we expect for the theme. In those cases, we need to override global typography
  // styles to get the right colors.
  .#{$ns}-popover2-content,
```

adidahiya (changes requested):
seems fine to add here. the alternative would be to put this style in `_tooltip.scss` / `_tooltip2.scss`, but I think it makes sense to leave it here as well.

just added some code comments to make the code more clear.

adidahiya (approved):

PR #4835, opened 2021-08-04
[core] feat(Icon): add aria-hidden for unlabeled icons

### LINKED ISSUE #4793: Callout icons should have aria-hidden="true" to indicate that the SVG is decorative in nature
Images that are purely decorative in nature and/or have adjacent text that are explanatory already can be ignored by assistive technologies: https://www.w3.org/WAI/tutorials/images/decorative/

For callouts, the icons next to the title (which is the informative part of the component) are more than likely purely decorative, so I'd imagine it would be safe to mark them as `aria-hidden="true"` by default

#### Environment

- __Package version(s)__: 3.46.0
- __Operating System__: MacOS

#### Screenshot

image

evansjohnson:
#### Checklist

- [x] Includes tests
- [x] Update documentation

set `aria-hidden="true"` on `Icon`s that don't have a `title` defined.

this is a change in documented behavior though I think it may be reasonable to view the previous behavior as undesired for accessibility and so okay to change within a major version. previously `desc` was set to the icon's name which wasn't formatted. perhaps screen readers handle this but I wonder if `less-than-or-equal-to` is read as `lessdashthandashordashequaldashto` (any way to check this?). regardless of whether that is handled, I think it's a better default to not announce every icon.

it could be considered to release this as part of v4 to include guidance in the upgrade notes that standalone icons without other text should now provide the `title` prop to be labeled appropriately

opinion on if this is an appropriate thing to do. I started off setting `aria-hidden` for decorative icons but realized it was almost all of them

#### Screenshot
Screen Shot 2021-08-04 at 1 56 41 PM

### DIFF
--- a/packages/core/src/components/breadcrumbs/breadcrumb.tsx
+++ b/packages/core/src/components/breadcrumbs/breadcrumb.tsx
@@ -27,6 +27,12 @@ export type BreadcrumbProps = IBreadcrumbProps;
 export interface IBreadcrumbProps extends ActionProps, LinkProps {
     /** Whether this breadcrumb is the current breadcrumb. */
     current?: boolean;
+
+    /**
+     * Pass though value to icon's title attribute. Should be used for breadcrumbs without
+     * text or children defined
+     */
+    iconTitle?: string;
 }
 
 export const Breadcrumb: React.FunctionComponent<BreadcrumbProps> = breadcrumbProps => {
@@ -39,7 +45,10 @@ export const Breadcrumb: React.FunctionComponent<BreadcrumbProps> = breadcrumbPr
         breadcrumbProps.className,
     );
 
-    const icon = breadcrumbProps.icon != null ? <Icon icon={breadcrumbProps.icon} /> : undefined;
+    const icon =
+        breadcrumbProps.icon != null ? (
+            <Icon aria-label={breadcrumbProps.iconTitle} icon={breadcrumbProps.icon} />
+        ) : undefined;
 
     if (breadcrumbProps.href == null && breadcrumbProps.onClick == null) {
         return (
--- a/packages/core/src/components/breadcrumbs/breadcrumb.tsx
+++ b/packages/core/src/components/breadcrumbs/breadcrumb.tsx
@@ -29,8 +29,8 @@ export interface IBreadcrumbProps extends ActionProps, LinkProps {
     current?: boolean;
 
     /**
-     * Pass though value to icon's title attribute. Should be used for breadcrumbs without
-     * text or children defined
+     * Pass through value to icon's title attribute. Should be used for breadcrumbs without
+     * text or children defined.
      */
     iconTitle?: string;
 }
--- a/packages/core/src/components/breadcrumbs/breadcrumb.tsx
+++ b/packages/core/src/components/breadcrumbs/breadcrumb.tsx
@@ -47,7 +47,7 @@ export const Breadcrumb: React.FunctionComponent<BreadcrumbProps> = breadcrumbPr
 
     const icon =
         breadcrumbProps.icon != null ? (
-            <Icon aria-label={breadcrumbProps.iconTitle} icon={breadcrumbProps.icon} />
+            <Icon title={breadcrumbProps.iconTitle} icon={breadcrumbProps.icon} />
         ) : undefined;
 
     if (breadcrumbProps.href == null && breadcrumbProps.onClick == null) {
--- a/packages/core/src/components/button/abstractButton.tsx
+++ b/packages/core/src/components/button/abstractButton.tsx
@@ -91,8 +91,7 @@ export interface IButtonProps<E extends HTMLButtonElement | HTMLAnchorElement = 
 
     /**
      * Provide this to give a text label to icon only buttons. You likely don't need this prop
-     * if there is other text in the button already. There is not currently a way to provide
-     * an iconTitle to `rightIcon` - please open a feature request if you find a need for this.
+     * if there is other text in the button already.
      */
     iconTitle?: string;
 }
--- a/packages/core/src/components/button/abstractButton.tsx
+++ b/packages/core/src/components/button/abstractButton.tsx
@@ -88,6 +88,13 @@ export interface IButtonProps<E extends HTMLButtonElement | HTMLAnchorElement = 
      * @default "button"
      */
     type?: "submit" | "reset" | "button";
+
+    /**
+     * Provide this to give a text label to icon only buttons. You likely don't need this prop
+     * if there is other text in the button already. There is not currently a way to provide
+     * an iconTitle to `rightIcon` - please open a feature request if you find a need for this.
+     */
+    iconTitle?: string;
 }
 
 /** @deprecated use AnchorButtonProps */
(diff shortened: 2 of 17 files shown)


palantirtech:
Thanks for your interest in [palantir/blueprint](https://github.com/palantir/blueprint), @evansjohnson! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

packages/core/src/components/non-ideal-state/nonIdealState.tsx
```diff
@@ -49,6 +49,12 @@ export interface INonIdealStateProps extends Props {
 
     /** The title of the non-ideal state. */
     title?: React.ReactNode;
+
+    /**
+     * Pass though value to icon's title attribute. Should be used for breadcrumbs without
+     * text or children defined
+     */
+     iconTitle?: string;
```
adidahiya: I would argue that NonIdealState icons are purely decorative. So there's no need for an icon title.

packages/core/src/components/menu/menuItem.tsx
```diff
@@ -170,7 +170,13 @@ export class MenuItem extends AbstractPureComponent2<MenuItemProps & React.Ancho
                 {text}
             </Text>,
             this.maybeRenderLabel(labelElement),
-            hasSubmenu ? <Icon icon="caret-right" /> : undefined,
+            hasSubmenu ? (
+                <Icon
+                    // TODO(accessibility): uncomment below when this component is accessible
```
adidahiya: what's this about?

packages/core/src/components/icon/icon.tsx
```diff
@@ -139,13 +141,18 @@ export class Icon extends AbstractPureComponent2<IconProps & Omit<React.HTMLAttr
         const classes = classNames(Classes.ICON, Classes.iconClass(icon), Classes.intentClass(intent), className);
         const viewBox = `0 0 ${pixelGridSize} ${pixelGridSize}`;
 
+        const props = {
+            ...htmlprops,
+            className: classes,
+            title: htmlTitle,
+        };
+        if (!title) {
+            props["aria-hidden"] = true;
```
adidahiya: I think you can leave the code mostly as-is... just add a key-value pair at L148 `"aria-hidden": title == null ? true : undefined,`

packages/core/src/components/breadcrumbs/breadcrumb.tsx
```diff
@@ -27,6 +27,12 @@ export type BreadcrumbProps = IBreadcrumbProps;
 export interface IBreadcrumbProps extends ActionProps, LinkProps {
     /** Whether this breadcrumb is the current breadcrumb. */
     current?: boolean;
+
+    /**
+     * Pass though value to icon's title attribute. Should be used for breadcrumbs without
+     * text or children defined
```
adidahiya:
```suggestion
     * Pass through value to icon's title attribute. Should be used for breadcrumbs without
     * text or children defined.
```

packages/core/src/components/tree/treeNode.tsx
```diff
@@ -158,7 +158,15 @@ export class TreeNode<T = {}> extends React.Component<ITreeNodeProps<T>> {
                 isExpanded ? Classes.TREE_NODE_CARET_OPEN : Classes.TREE_NODE_CARET_CLOSED,
             );
             const onClick = disabled === true ? undefined : this.handleCaretClick;
-            return <Icon className={caretClasses} onClick={onClick} icon={"chevron-right"} />;
+            return (
+                <Icon
+                    // TODO(accessibility): add below when component is accessible
```
adidahiya: what's this about?

packages/table/src/headers/header.tsx
```diff
@@ -390,7 +390,8 @@ export class Header extends React.Component<IInternalHeaderProps, IHeaderState>
     private maybeRenderReorderHandle(index: number) {
         return !this.isReorderHandleEnabled()
             ? undefined
-            : this.wrapInDragReorderable(
+            : // TODO(accessibility): add aria-label when this drag handle is accessible
```
adidahiya: this seems unrelated to the overall PR

adidahiya (changes requested):
this seems like the right direction w/r/t icon aria attributes. there are some more cases we need to consider:

- icon-only buttons without text/children (need a way to send `title` to the rendered `<Icon>`)
- "double-caret-vertical" icon inside `<HTMLSelect>` needs a title
- `<KeyCombo>` icons probably need titles
- `<TimePicker>` arrow icons probably need titles

packages/core/src/components/non-ideal-state/nonIdealState.tsx
evansjohnson: yup - nice catch!

packages/core/src/components/menu/menuItem.tsx
evansjohnson: will remove these from this PR

was thinking we don't want to add a `title` if the component isn't actually accessible but didn't think about `title`s being somewhat useful regardless

evansjohnson:
did a pass at the extra cases you mentioned, I did search through every `<Icon` but went quick so missed at least those and there could be more. would hope to be able to merge with this incremental progress knowing there may be some caught over time.

only one I didn't add for now:
- "double-caret-vertical" icon inside <HTMLSelect> - I imagine a `<select>` component is already semantic usage that we can expect a user to understand and the icon we add is decorative but happy to add "Open options" or another title if you could make a decision there?

packages/core/src/components/button/abstractButton.tsx
```diff
@@ -88,6 +88,13 @@ export interface IButtonProps<E extends HTMLButtonElement | HTMLAnchorElement =
      * @default "button"
      */
     type?: "submit" | "reset" | "button";
+
+    /**
+     * Provide this to give a text label to icon only buttons. You likely don't need this prop
+     * if there is other text in the button already. There is not currently a way to provide
+     * an iconTitle to `rightIcon` - please open a feature request if you find a need for this.
```
adidahiya: IMO this kind of thing doesn't belong in prop documentation, it can be discussed in the github repo.

```suggestion
     * if there is other text in the button already.
```

Also, why would you have a `rightIcon` if you don't have any text?

packages/core/test/icon/iconTests.tsx
```diff
@@ -77,17 +73,15 @@ describe("<Icon>", () => {
...
 
-    /** Asserts that rendered icon has given className. */
-    function assertIcon(icon: React.ReactElement<IIconProps>, iconName: IconName) {
-        const wrapper = shallow(icon);
-        assert.strictEqual(wrapper.text(), iconName);
-        assert.isAbove(wrapper.find("path").length, 0, "should find at least one path element");
-    }
+    it("applies aria-hidden=true if title is not defined", () => {
+        const icon = shallow(<Icon icon="airplane" />);
+        assert.isTrue(icon.find(`.${Classes.ICON}`).prop("aria-hidden"));
```
adidahiya: you may need a `.hostNodes()` in here after the `.find()`

adidahiya (changes requested):

adidahiya:
Regarding `<HTMLSelect>`... I guess I could go either way but slightly lean towards adding the title for now (something like "Open dropdown")

adidahiya (approved):

adidahiya:
I am reconsidering one of the suggestions I made earlier:

> icon-only buttons without text/children (need a way to send `title` to the rendered `<Icon>`)

I believe this turns out to be superfluous. icon-only buttons without text/children should have their a11y labels set with `aria-label` and/or `title`, not by labeling the icon. I will make a PR to remove the `iconTitle` prop.

PR #4838, opened 2021-08-05
[core] feat(Dialog): accessibility aria attributes

### LINKED ISSUE #4528: Dialogs should have proper a11y markup

evansjohnson:
#### Checklist

- [x] Includes tests
- [x] Update documentation

add `aria-labelledby` and `aria-describedby` props to `Dialog` with `aria-labelledby` defaulting to `title` element

#### Screenshot
Screen Shot 2021-08-05 at 8 58 24 AM
Screen Shot 2021-08-05 at 8 58 49 AM


packages/core/src/components/dialog/dialog.tsx
```diff
@@ -17,6 +17,7 @@
 import classNames from "classnames";
 import * as React from "react";
 import { polyfill } from "react-lifecycles-compat";
+import { uniq } from "lodash-es";
```
adidahiya: We don't depend on lodash in core. I don't really want to add that dependency now, so we'll need to come up with our own "unique ID" utility function. it can live in packages/core/src/common/utils/jsUtils.ts

Also, did you mean to use `uniqueId` instead of `uniq` here?

packages/core/test/dialog/dialogTests.tsx
```diff
@@ -64,6 +64,62 @@ describe("<Dialog>", () => {
         document.body.removeChild(container);
     });
 
+    describe("ally support", () => {
```
adidahiya: wrong spelling of "a11y", and actually I'd rather we just spell out the word here

```suggestion
    describe("accessibility support", () => {
```

packages/core/src/components/dialog/dialog.tsx
```diff
@@ -76,6 +77,17 @@ export interface IDialogProps extends OverlayableProps, IBackdropProps, Props {
      * name here will require defining new CSS transition properties.
      */
     transitionName?: string;
+
+    /**
+     * id of element that contains title or label text for this dialog
+     * by default the element displaying the `title` prop will be used
```
adidahiya:
```suggestion
     * ID of the element that contains title or label text for this dialog.
     *
     * By default, if the `title` prop is supplied, this component will generate
     * a unique ID for the `<H4>` title element and use that ID here.
```

packages/core/src/components/dialog/dialog.tsx
```diff
@@ -76,6 +77,17 @@ export interface IDialogProps extends OverlayableProps, IBackdropProps, Props {
...
+     */
+    "aria-labelledby"?: string;
+
+    /**
+     * id of element that contains description text for this dialog
```
adidahiya:
```suggestion
     * ID of an element that contains description text inside this dialog.
```

adidahiya (changes requested):
let's update `dialogExample.tsx` as well... I think we just need to add an `aria-describedby` prop and point it to an ID we add to the `Classes.DIALOG_BODY` element

adidahiya:
sorry about the test flakes in the table package test suite... I will look into that soon

packages/core/src/common/utils/jsUtils.ts
```diff
@@ -70,3 +70,10 @@ export function countDecimalPlaces(num: number) {
     }
     return p;
 }
+
+const uniqueCountForPrefix = new Map<string, number>();
+export function uniqueId(prefix: string) {
+    const curCount = uniqueCountForPrefix.get(prefix) || 0;
+    uniqueCountForPrefix.set(prefix, curCount + 1);
+    return `///${prefix}-${curCount}///`;
```
adidahiya: why the slashes? IMO they don't signify much, and could be misleading, suggesting some kind of regex

packages/core/test/dialog/dialogTests.tsx
```diff
@@ -64,6 +64,69 @@ describe("<Dialog>", () => {
         document.body.removeChild(container);
     });
 
+    describe("accessibility support", () => {
+        const mountDefaults = (className: string) => {
+            const container = document.createElement("div");
+            document.body.appendChild(container);
```
adidahiya: I think these could be in a `beforeEach` block. and `container` can live as a `let container: HTMLElement;` outside that block.

packages/core/test/dialog/dialogTests.tsx
```diff
@@ -64,6 +64,69 @@ describe("<Dialog>", () => {
...
+                    isOpen={true}
+                    portalContainer={container}
+                    aria-labelledby="dialog-title"
+                    aria-describedby="dialog-description"
+                >
+                    {createDialogContents()}
+                </Dialog>,
+            );
+
+            return () => document.body.removeChild(container);
```
adidahiya: similar feedback here... I think it would be more idiomatic to put this in a `afterEach` block...


```
afterEach(() => {
  if (container != null) {
    document.body.removeChild(container);
  }
});
```

packages/core/test/dialog/dialogTests.tsx
```diff
@@ -64,6 +64,69 @@ describe("<Dialog>", () => {
...
+        };
+
+        it("renders with role={dialog}", () => {
+            const cleanUp = mountDefaults("check-role");
+            const dialogElement = document.querySelector(`.check-role`);
+            assert.equal(dialogElement!.getAttribute("role"), "dialog", "missing dialog role!!");
+            cleanUp();
```
adidahiya: if you switch to `afterEach`, you won't need this function call in every test

adidahiya (changes requested):

adidahiya (approved):

Issue #4810, opened 2021-07-20
Clickable menu items should specify tabindex="0" for keyboard accessible controls


adidahiya:
#### Environment

- __Package version(s)__: core 3.47.0

#### Code Sandbox

Link to a minimal repro: https://codesandbox.io/s/blueprint-sandbox-forked-glq4i?file=/src/CoreExample.tsx

#### Steps to reproduce

1. Click inside "Core sandbox example" (this step is optional, but gets you quicker to focus on the first menu item)
1. Press TAB key to focus first menu item
1. Press TAB key again

#### Actual behavior

Focus goes to third menu item, which has an `href` attribute

#### Expected behavior

Focus goes to second menu item (and focus ring indicator is displayed)

#### Possible solution

We should apply `tabIndex={0}` by default to all MenuItems, even if they don't have an `href` (which would make them focusable by default). The reasoning for this is that we make MenuItems look interactive regardless of whether they are actually hooked up to _do_ something in an application (with `:hover` and `:active` styles), so for accessible consistency we should also make them focusable via the keyboard.

Note that this issue is a subset of https://github.com/palantir/blueprint/issues/3079, but is more narrow and easily actionable.

Issue #4782, opened 2021-06-25
Tag onClick does not fire with spacebar or enter press when focused


evansjohnson:
#### Code Sandbox
https://codesandbox.io/s/palantir-blueprint-forked-s0reo

#### Steps to reproduce
1. click tag, observe onClick handler works
2. use tab to navigate to tag, onClick does not fire with spacebar or enter

`Button` included to show expected behavior

#### Actual behavior
`onClick` of `Tag` does not fire with spacebar or enter press when focused

#### Expected behavior
spacebar/enter should trigger `onClick` of `Tag`

#### Possible solution
if `Tag` has `onClick` defined, render as a button but with the `Tag` styles

adidahiya:
I'm not sure this is expected behavior for tags. If you really want something so interactive that it can be focussed and clicked via the keyboard, you probably want a minimal or outlined button instead. I don't think your requested feature exists in other design toolkits like Bootstrap or Ant (tags are not even focussable by default in Ant).

evansjohnson:
> I'm not sure this is expected behavior for tags. If you really want something so interactive that it can be focussed and clicked via the keyboard, you probably want a minimal or outlined button instead. I don't think your requested feature exists in other design toolkits like Bootstrap or Ant (tags are not even focussable by default in Ant).

"so interactive that ..." seems to indicate levels of interactivity, I think it's a binary thing. Either an element is interactive and it should be accessible by keyboard users, or it's not interactive and doesn't need to receive focus/be triggered by keyboard.

I'd support `Tag`s not accepting the `onClick` prop/no interactive style but that sounds like a large breaking change, but don't think that's a reason for `Tag`s to stay in a state of partial support for interactivity.

I don't need an immediate solution but thought raising this would be helpful to track accessibility of Blueprint components, particularly if the issue sees other activity. Perhaps a "Domain: accessibility" label is appropriate here.

PR #4839, opened 2021-08-05
Add initial AccessibleButton component

evansjohnson:
#### Checklist

- [ ] Includes tests
- [x] Update documentation

Will add tests if this gets general go ahead nod.

Add an `AccessibleButton` utility component to abstract away making a non `<button>` component properly clickable

Would come in handy for https://github.com/palantir/blueprint/issues/3079, https://github.com/palantir/blueprint/issues/4810 (`tabIndex={0}` won't really help if can't be triggered by keyboard while focused), https://github.com/palantir/blueprint/issues/4782, and other tickets requiring keyboard support

Is this a component Blueprint wants to support?

Is it implemented correctly?
- Also considered if it should be a `<button>` component that removes all styling, though perhaps less flexible for consumers

`AccessibleButtonProps` doesn't extend anything right now, had trouble finding something that worked everything seemed to cause some issue. Will fix up if this gets a general go ahead nod.

#### Screencap
AccessibleButtonDemo

### DIFF
--- a/packages/core/src/components/_index.scss
+++ b/packages/core/src/components/_index.scss
@@ -3,6 +3,7 @@
 
 @import "alert/alert";
 @import "breadcrumbs/breadcrumbs";
+@import "button/accessible-button";
 @import "button/button";
 @import "button/button-group";
 @import "callout/callout";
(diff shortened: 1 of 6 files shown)


adidahiya:
we added this to an internal component library; will reconsider adding it to Blueprint after it's had some time to bake there

PR #4843, opened 2021-08-05
[select] feat(Select): new prop "fill"

### LINKED ISSUE #4721: [4.0.0-alpha.0] `fill` doesn't work on MultiSelect, Suggest, Select
__Package version(s)__: 4.0.0-alpha.0
This can be reproduced in the docs page: https://blueprintjs.com/docs/versions/4/#select/multi-select

It's possible that `fill` might also not work in `Popover` and `InputGroup` although I haven't tested (see https://github.com/palantir/blueprint/pull/3636).

I upgraded to 4.0.0-alpha.0 in hopes that `fill` was implemented for `Select` but apparently it's not (see https://github.com/palantir/blueprint/issues/2956).

PS: Blueprint is amazing. Thank you, to all the developers, maintainers and contributors!

millasml:
#### Checklist

- [x] Update documentation

Fill prop that makes the select component fill component width. 

#### Screenshot

image


packages/select/src/components/select/select.tsx
```diff
@@ -39,6 +39,12 @@ import { IQueryListRendererProps, QueryList } from "../query-list/queryList";
 export type SelectProps<T> = ISelectProps<T>;
 /** @deprecated use SelectProps */
 export interface ISelectProps<T> extends IListItemsProps<T> {
+    /**
+     * Whether the component should take up the full width of its container.
+     * This overrides `popoverProps.fill` and `tagInputProps.fill`.
```
adidahiya: there is no `tagInputProps` for this component.

```suggestion
     * This overrides `popoverProps.fill`.
```

packages/select/src/components/select/select.tsx
```diff
@@ -129,7 +135,18 @@ export class Select<T> extends AbstractPureComponent2<SelectProps<T>, ISelectSta
...
+        const { fill, filterable = true, disabled = false, inputProps = {}, popoverProps = {} } = this.props;
+
+        const childProps: { fill?: boolean } = {};
+
+        if (fill) {
+            popoverProps.fill = true;
+            childProps.fill = true;
+        }
+
+        const children = React.Children.map(this.props.children, (child: React.ReactNode) => {
```
adidahiya: I don't love this approach and would prefer not to clone children if we can avoid it. We can't guarantee that the children will support the `fill` prop, and I'd rather not send along extraneous HTML attributes which might get rendered out to the DOM (and produce a React warning). 

If we stick to just rendering `{this.props.children}`, then `<Select>` users will have to add `fill` to children themselves (or style it some other way), but I think that's ok. We should make a note of that in the fill prop documentation.

adidahiya (changes requested):
Overall looks good, just a few comments I'd like addressed

adidahiya:
the docs example doesn't work anymore, it needs to be updated so that the button uses `fill={true}` when this switch is toggled:

adidahiya (approved):
works now

PR #4845, opened 2021-08-06
[datetime] fix(TimePicker): reset input fields when value={null}

### LINKED ISSUE #4775: Pressing "Clear" on DatePicker with TimePicker doesn't clear time inputs
#### Environment

- __Package version(s)__: 
  - "@blueprintjs/core": "3.31.0"
  - "@blueprintjs/datetime": "^3.23.4"
- __Operating System__: Mac OS 10.15.7
- __Browser name and version__: Chrome 91.0.4472.114

#### Code Sandbox

Link to a minimal repro: https://codesandbox.io/embed/timepicker-not-clearing-u08v6?fontsize=14&hidenavigation=1&theme=dark

Also here: 

https://user-images.githubusercontent.com/4326691/123145098-77e2cf80-d42a-11eb-8980-610593e112b2.mov

#### Steps to reproduce

1. Click Today to set date to now
2. Click "Clear".

#### Actual behavior

Clear does not set the timepicker fields back to 00s.
**Note: this _does_ clear the actual underlying component value, just not the input fields for time**

#### Expected behavior

Clear sets the timepicker fields back to 00s.

#### Possible solution

Not sure of the inner workings of how datepicker talks to timepicker 🤔

dummycode:
#### Checklist

- [x] Includes tests
- [x] Update documentation

Updating `TimePickers` value to null should behave the same way as when it was constructed: get min time or default value. 

Whether this breaks any existing functionality.

#### Screenshot

output

### DIFF
--- a/packages/datetime/src/dateTimePicker.tsx
+++ b/packages/datetime/src/dateTimePicker.tsx
@@ -100,7 +100,11 @@ export class DateTimePicker extends AbstractPureComponent2<IDateTimePickerProps,
                     onChange={this.handleDateChange}
                     value={value}
                 />
-                <TimePicker {...this.props.timePickerProps} onChange={this.handleTimeChange} value={value} />
+                <TimePicker
+                    {...this.props.timePickerProps}
+                    onChange={this.handleTimeChange}
+                    value={this.state.timeValue}
+                />
             </div>
         );
     }
--- a/packages/datetime/src/timePicker.tsx
+++ b/packages/datetime/src/timePicker.tsx
@@ -169,9 +169,7 @@ export class TimePicker extends React.Component<TimePickerProps, ITimePickerStat
     public constructor(props?: TimePickerProps, context?: any) {
         super(props, context);
 
-        const value = this.getInitialValue();
-
-        this.state = this.getFullStateFromValue(value, props.useAmPm);
+        this.state = this.getFullStateFromValue(this.getInitialValue(), props.useAmPm);
     }
 
     public render() {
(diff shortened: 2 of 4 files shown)


palantirtech:
Thanks for your interest in [palantir/blueprint](https://github.com/palantir/blueprint), @dummycode! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

packages/docs-app/src/examples/datetime-examples/dateTimePickerExample.tsx
```diff
@@ -16,23 +16,34 @@
 
 import * as React from "react";
 
-import { Classes } from "@blueprintjs/core";
+import { Classes, H5, Switch } from "@blueprintjs/core";
```
adidahiya: changes to this file are unrelated to the issue, please revert them. also this is a deprecated component, people shouldn't be using this docs page anyway.

packages/datetime/src/dateTimePicker.tsx
```diff
@@ -100,7 +100,11 @@ export class DateTimePicker extends AbstractPureComponent2<IDateTimePickerProps,
                     onChange={this.handleDateChange}
                     value={value}
                 />
-                <TimePicker {...this.props.timePickerProps} onChange={this.handleTimeChange} value={value} />
+                <TimePicker
```
adidahiya, line 103: this component is deprecated, please revert these changes. users should be using `<DatePicker>` instead.

packages/datetime/src/timePicker.tsx
```diff
@@ -222,9 +215,12 @@ export class TimePicker extends React.Component<TimePickerProps, ITimePickerStat
         const didMaxTimeChange = prevProps.maxTime !== this.props.maxTime;
         const didBoundsChange = didMinTimeChange || didMaxTimeChange;
         const didPropValueChange = prevProps.value !== this.props.value;
-        const shouldStateUpdate = didMinTimeChange || didMaxTimeChange || didBoundsChange || didPropValueChange;
```
adidahiya, line 225: good catch... this expression was redundant

adidahiya (changes requested):
nice, simple fix. I'll make sure to note this change in behavior in the release notes.

packages/datetime/src/dateTimePicker.tsx
dummycode: Without this change, the test `changing the time before selecting a date works as expected` fails because clearing the date also clears the time because the `TimePicker` component now responds to being passed `null`, whereas before it was a no-op. Should I still revert this change?

packages/docs-app/src/examples/datetime-examples/dateTimePickerExample.tsx
dummycode: Reverted.

packages/datetime/src/dateTimePicker.tsx
adidahiya: Ok, fair enough, we can include this change in the PR.

adidahiya (approved):
looks good in [docs preview](https://45857-71939872-gh.circle-artifacts.com/0/packages/docs-app/dist/index.html#datetime/datepicker)

PR #4852, opened 2021-08-11
[stylelint-plugin] Fix no-prefix-literal test

p-szm:
https://github.com/palantir/blueprint/pull/4715 made the `bp` prefix not banned, but I forgot to update this test.

Also, looks like these tests are not being run in CI for some reason and they probably should be.


adidahiya (approved):

adidahiya:
I'll look into why the tests are not running in CI...

PR #4853, opened 2021-08-11
[stylelint-plugin] feat: add no-color-literal rule

p-szm:
This PR adds a new stylelint rule that encourages using color variables (`$blue3` and etc) instead of color literals (`#137CBD` and etc).


packages/stylelint-plugin/test/hexColor.test.js
```diff
@@ -0,0 +1,49 @@
+/* Copyright 2020 Palantir Technologies, Inc. All rights reserved.
```
adidahiya, line 1: minor nit: these copyright headers don't have the right comment syntax. I can fix in a follow-up.

adidahiya (approved):

PR #4858, opened 2021-08-16
feat: new package "@blueprintjs/colors"

p-szm:


adidahiya:
I fixed `dist:variables` by just reaching up outside of packages/core into packages/colors in the script CLI arguments. Should be ok. In a future PR we should consider adding a separate docs section for the colors package, but I'm fine with shipping this for now as we're continuing to export colors through the core package.

Note: the table test suite failures are unrelated flakes, going to investigate those more this week...

adidahiya (approved):

Issue #3941, opened 2020-01-24
Blank table when resizing columns


comfroels:
#### Environment

- __Package version(s)__: 3.8.3 (table)
- __Browser and OS versions__: Chrome 79.0.3945.117

If possible, link to a minimal repro (fork [this code sandbox](https://codesandbox.io/s/blueprint-sandbox-et9xy)): You can repro on one of your examples: https://blueprintjs.com/docs/#table/features.freezing

#### Steps to reproduce

1. Resize any of the columns in the above example

#### Actual behavior

Table goes white

#### Expected behavior

Table renders cell data

#### Possible solution

Not quite sure why this is happening yet, still trying to dig. Other examples on your site are rendering correctly after the resize, so I'll have to try and figure out what's different

adidahiya:
I'm not able to repro this in Chrome 79.0.3945.130 or Firefox 72.0.2 on macOS Mojave... do you have a screenshot or gif/video of the problem?

waltonseymour:
I believe I'm seeing the same issue on Chrome 81.0.4044.138. It seems to be triggered by resizing columns and can be recovered by scrolling.

![bp-table-bug](https://user-images.githubusercontent.com/7594493/81597025-04e7d800-9393-11ea-81a3-4bab87ca838e.gif)

strickland84:
![table_bug](https://user-images.githubusercontent.com/2702496/82214770-3a9a3d00-990e-11ea-9156-cdc1701c470f.gif)

I was able to replicate this issue on our project in both Chrome and Firefox so I don't think it's browser specific. I have tried adjusting the `Table` props to see if I could determine a trigger for this but I could not. As mentioned above it occurs when re-sizing the columns and can be reverted by scrolling.

strickland84:
Also, as per my example and the one from @waltonseymour I notice that the first 4 columns remain rendered. Perhaps this is coincidence. 🤷

ae0n:
I have the same issue with resizing. I'm able to repeat this bug in chrome 83.0.4103 and safari 13.1.1. Also only 4 columns are visible after resize and others are gone until I scroll the page.

adidahiya:
It sounds like this is connected to [all the other lifecycle bugs in `@blueprintjs/table` 3.8.x](https://github.com/palantir/blueprint/milestone/41). For now I would recommend downgrading to 3.7

ryanrabello:
[...]

PR #4884, opened 2021-08-30
[table] fix issues with blank rows/columns on mount/resize

### LINKED ISSUE #3779: [Table] only 4 columns render on table mounted if table horizontal overflow

adidahiya:
#### Checklist

- [ ] Includes tests
- [x] Update documentation

- Update the "basic usage" table example to repro the bug described in #3941 
- Adjust some viewport rect code to be more defensive about and provide defaults (top = 0, left = 0) for when the scroll container element ref is null, or doesn't have client coordinates. This fixes the edge cases where the visible grid would be incorrectly calculated as the top left 4x4 cells, and only those cells would render

N/A

### DIFF
--- a/packages/docs-app/src/examples/table-examples/tableDollarExample.tsx
+++ b/packages/docs-app/src/examples/table-examples/tableDollarExample.tsx
@@ -19,13 +19,20 @@ import * as React from "react";
 import { Example, IExampleProps } from "@blueprintjs/docs-theme";
 import { Cell, Column, Table2 } from "@blueprintjs/table";
 
+// this will obviously get outdated, it's valid only as of August 2021
+const USD_TO_EURO_CONVERSION = 0.85;
+
 export class TableDollarExample extends React.PureComponent<IExampleProps> {
     public render() {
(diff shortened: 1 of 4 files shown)


adidahiya:
This is a relatively safe change and it passes the test suite, so I'm going to push it through to verify that it fixes the related bugs in practice.

PR #4894, opened 2021-09-09
[core] fix(Overlay): use focus traps to fix enforceFocus

### LINKED ISSUE #3445: Dialog and Popover enforceFocus has no affect
#### Environment

- __Package version(s)__: 13.14.x
- __Browser and OS versions__: Windows 10, Chrome  73.0.3683.86 (Official Build) (64-bit)

#### Steps to reproduce

I've created both a popover and a dialog with a variety of controls including check boxes and input controls. I set enforceFocus to true, autoFocus to true. Tabbing between items in the dialog  or the popover does not keep the focus within the dialog/popover. 

For the dialog, you can verify this behavior by doing the following:
1. Go to BluepringJS dialog page - all props are on by default
2. Click ShowDialog
3. Tab through all of the controls

#### Actual behavior
Tabbing past 'Visit the Foundry website' button puts you into the chrome address bar instead of back to the first focus-able item in the dialog.

As an aside, the 'Visit the Foundry website' button has no focus rect (hard to see that it is the focused element)

#### Expected behavior
Tabbing past the `Visit the Foundry website' button sets focus to the first focus-able item in the dialog or popover.

michael-yx-wu:
#### Checklist

- [x] Includes tests
- [x] Update documentation (n/a)

When enforceFocus is enabled, surround Overlay content with dummy div
elements with focus event listeners that keep focus inside the Overlay.

I want to flag that I've removed the `tabindex` attribute from the backdrop to prevent it from being keyboard focusable.

This change also fixes an issue where multiple open overlays will compete with each for focus. As an example, go to https://blueprintjs.com/docs/#core/components/overlay and open the overlay and then hit shift+s to open the omnibar search (also powered by Overlay). The search input will not get autofocused. Repeat the same workflow on https://46971-71939872-gh.circle-artifacts.com/0/packages/docs-app/dist/index.html#core/components/overlay and note that the omnibar search does get focus correctly.

##### Before

When using a portal, the portal is the last element on the page so tabbing will eventually cause focus to leave the document and none of the existing event handlers run before that happens. Note that shift+tabbing is not closing the overlay, the gif is just looping.

##### After

Tab/shift+tab will never allow focus to leave the overlay if `enforceFocus`.

### DIFF
--- a/packages/core/src/components/overlay/overlay.tsx
+++ b/packages/core/src/components/overlay/overlay.tsx
@@ -435,6 +435,8 @@ export class Overlay extends AbstractPureComponent2<OverlayProps, IOverlayState>
     }
 
     private handleStartFocusTrapElementFocusIn = (e: FocusEvent) => {
+        e.preventDefault();
+        e.stopImmediatePropagation();
         if (
             e.relatedTarget != null &&
             this.containerElement!.contains(e.relatedTarget as Element) &&
@@ -447,6 +449,8 @@ export class Overlay extends AbstractPureComponent2<OverlayProps, IOverlayState>
     };
 
     private handleEndFocusTrapElementFocusIn = (e: FocusEvent) => {
+        e.preventDefault();
+        e.stopImmediatePropagation();
         if (
             e.relatedTarget != null &&
             this.containerElement!.contains(e.relatedTarget as Element) &&
@@ -462,6 +466,7 @@ export class Overlay extends AbstractPureComponent2<OverlayProps, IOverlayState>
         const focusableElements: HTMLElement[] =
             this.containerElement !== null
                 ? Array.from(
+                      // Order may not be correct if children elements use tabindex values > 0
                       this.containerElement.querySelectorAll(
                           [
                               "a[href]",
@@ -485,7 +490,6 @@ export class Overlay extends AbstractPureComponent2<OverlayProps, IOverlayState>
     }
 
     private overlayWillClose() {
-        document.removeEventListener("focus", this.handleDocumentFocus, /* useCapture */ true);
         document.removeEventListener("mousedown", this.handleDocumentClick);
         this.startFocusTrapElement?.removeEventListener("focusin", this.handleStartFocusTrapElementFocusIn);
         this.endFocusTrapElement?.removeEventListener("focusin", this.handleEndFocusTrapElementFocusIn);
@@ -497,7 +501,7 @@ export class Overlay extends AbstractPureComponent2<OverlayProps, IOverlayState>
             if (openStack.length > 0) {
                 const lastOpenedOverlay = Overlay.getLastOpened();
                 if (lastOpenedOverlay.props.enforceFocus) {
-                    document.addEventListener("focus", lastOpenedOverlay.handleDocumentFocus, /* useCapture */ true);
+                    lastOpenedOverlay.bringFocusInsideOverlay();
                 }
             }
 
@@ -508,20 +512,13 @@ export class Overlay extends AbstractPureComponent2<OverlayProps, IOverlayState>
     }
 
     private overlayWillOpen() {
-        const { openStack } = Overlay;
-        if (openStack.length > 0) {
-            document.removeEventListener("focus", Overlay.getLastOpened().handleDocumentFocus, /* useCapture */ true);
-        }
-        openStack.push(this);
+        Overlay.openStack.push(this);
 
         if (this.props.autoFocus) {
             this.bringFocusInsideOverlay();
         }
-        if (this.props.enforceFocus) {
-            document.addEventListener("focus", this.handleDocumentFocus, /* useCapture */ true);
-        }
 
-        if (this.props.canOutsideClickClose && !this.props.hasBackdrop) {
+        if (this.props.canOutsideClickClose || this.props.enforceFocus) {
             document.addEventListener("mousedown", this.handleDocumentClick);
         }
 
@@ -537,15 +534,14 @@ export class Overlay extends AbstractPureComponent2<OverlayProps, IOverlayState>
             onClose?.(e);
         }
         if (enforceFocus) {
-            // make sure document.activeElement is updated before bringing the focus back
             this.bringFocusInsideOverlay();
         }
         backdropProps?.onMouseDown?.(e);
     };
 
     private handleDocumentClick = (e: MouseEvent) => {
-        const { canOutsideClickClose, isOpen, onClose } = this.props;
-        // get the actually target even if we are in an open mode Shadow DOM
+        const { canOutsideClickClose, enforceFocus, isOpen, onClose } = this.props;
+        // get the actual target even if we are in an open mode Shadow DOM
         const eventTarget = (e.composed ? e.composedPath()[0] : e.target) as HTMLElement;
 
         const stackIndex = Overlay.openStack.indexOf(this);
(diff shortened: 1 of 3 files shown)


packages/core/src/components/overlay/overlay.tsx
```diff
@@ -475,25 +553,14 @@ export class Overlay extends AbstractPureComponent2<OverlayProps, IOverlayState>
...
-        if (
-            this.props.enforceFocus &&
-            this.containerElement != null &&
-            eventTarget instanceof Node &&
-            !this.containerElement.contains(eventTarget as HTMLElement)
-        ) {
-            // prevent default focus behavior (sometimes auto-scrolls the page)
-            e.preventDefault();
-            e.stopImmediatePropagation();
-            this.bringFocusInsideOverlay();
```
michael-yx-wu, line 496: when `enforceFocus=true`, focus will start on the overlay and will only leave if another overlay with autoFocus or enforceFocus is opened. because we take care to focus the last opened overlay when closing the current overlay and we call `bringFocusInsideOverlay` in the mousedown event handler, we don't need this focus event handler anymore

packages/core/src/components/overlay/overlay.tsx
```diff
@@ -415,7 +501,7 @@ export class Overlay extends AbstractPureComponent2<OverlayProps, IOverlayState>
             if (openStack.length > 0) {
                 const lastOpenedOverlay = Overlay.getLastOpened();
                 if (lastOpenedOverlay.props.enforceFocus) {
-                    document.addEventListener("focus", lastOpenedOverlay.handleDocumentFocus, /* useCapture */ true);
+                    lastOpenedOverlay.bringFocusInsideOverlay();
```
michael-yx-wu, line 522: this handler was a roundabout way to bring focus back to the last opened overlay. we can just call `bringFocusInsideOverlay` directly.

packages/core/test/multistep-dialog/multistepDialogTests.tsx
```diff
@@ -15,18 +15,47 @@
...
+
+    /**
+     * Mount the `content` into `testsContainerElement` and assign to local `wrapper` variable.
+     * Use this method in this suite instead of Enzyme's `mount` method.
+     */
+    function mountWrapper(content: JSX.Element) {
+        wrapper = mount(content, { attachTo: testsContainerElement });
+        isMounted = true;
+        return wrapper;
+    }
```
michael-yx-wu: these tests were causing Overlay tests to fail. see https://app.circleci.com/pipelines/github/palantir/blueprint/1539/workflows/f959324a-857b-4131-a89e-f3c315416103/jobs/46941: 

> \<Overlay\> Focus management does not focus overlay when closed - <Overlay> Focus management
> `Error: Uncaught AssertionError: expected <button></button> to equal <div class="bp3-dialog-container bp3-overlay-content bp3-overlay-appear-done bp3-overlay-enter-done" tabindex="0"><div class="bp3-dialog" role="dialog" style="min-width: 800px; padding-bottom: 0px;"><div class="bp3-multistep-dialog-panels"><div class="bp3-multistep-dialog-left-panel"></div></div></div></div>`

Note the **`bp3-multistep-dialog-left-panel`** class. The error suggests that something is causing these tests to affect the overlay tests but I'm not sure why...

Making this suite use a mountWrapper like the overlayTests seems to fix the issue.

michael-yx-wu: on closer inspection, I suspect the reason is due to some of these tests previously mounting but _not_ unmounting. For example: https://github.com/palantir/blueprint/blob/88a158b38a96350109650a7fe821a5091a25b3ed/packages/core/test/multistep-dialog/multistepDialogTests.tsx#L169-L180

Since this suite is alphabetically before overlayTests and because we've now fixed the `enforceFocus` behavior (which is enabled for both Dialog and MultistepDialog), these unmounted overlays are keeping focus. I'll revert this change and just unmount properly for a smaller diff, but lmk if you actually prefer this wrapper instead.

packages/core/test/overlay/overlayTests.tsx
```diff
@@ -60,6 +60,10 @@ describe("<Overlay>", () => {
         }
     });
 
+    after(() => {
+        document.documentElement.removeChild(testsContainerElement);
+    });
```
michael-yx-wu, line 65: not really important, but seems good to clean up?

packages/core/src/components/overlay/overlay.tsx
```diff
@@ -319,11 +335,11 @@ export class Overlay extends AbstractPureComponent2<OverlayProps, IOverlayState>
             if (isFocusOutsideModal) {
                 // element marked autofocus has higher priority than the other clowns
                 const autofocusElement = this.containerElement.querySelector("[autofocus]") as HTMLElement;
```
michael-yx-wu, line 339: I noticed that `InputGroup`'s `autoFocus` prop is passed to the `input` element as is. React doesn't actually render the autofocus attribute to the DOM (see https://github.com/facebook/react/issues/11851#issuecomment-351672131), so this querySelector is unlikely to work unless the consumer explicitly sets the DOM `autofocus` (no caps) attribute instead. Not sure if we want to do anything about this though, since React will warn if you set the DOM attribute directly

packages/core/src/components/overlay/overlay.tsx
```diff
@@ -229,10 +229,22 @@ export class Overlay extends AbstractPureComponent2<OverlayProps, IOverlayState>
     // an HTMLElement that contains the backdrop and any children, to query for focus target
     public containerElement: HTMLElement | null = null;
 
+    private startFocusTrapElement: HTMLDivElement | null = null;
```
adidahiya, line 233: would be nice to leave a brief `// comment` here about this member's purpose. same for `endFocusTrapElement`.

packages/core/src/components/overlay/overlay.tsx
```diff
@@ -404,9 +419,80 @@ export class Overlay extends AbstractPureComponent2<OverlayProps, IOverlayState>
...
+            this.containerElement!.contains(e.relatedTarget as Element) &&
+            e.relatedTarget !== this.startFocusTrapElement
+        ) {
+            this.startFocusTrapElement?.focus();
+        } else {
+            this.keyboardFocusableElements().pop()?.focus();
+        }
+    };
+
+    private keyboardFocusableElements() {
```
adidahiya: nit:

```suggestion
    private getKeyboardFocusableElements() {
```

packages/core/src/components/overlay/overlay.tsx
```diff
@@ -404,9 +419,80 @@ export class Overlay extends AbstractPureComponent2<OverlayProps, IOverlayState>
...
+        const focusableElements: HTMLElement[] =
+            this.containerElement !== null
+                ? Array.from(
+                      // Order may not be correct if children elements use tabindex values > 0
+                      this.containerElement.querySelectorAll(
+                          [
```
adidahiya, line 489: I'm curious how you decided on this list? Is there some documentation we can link to? It would be nice to leave some more comments about our reasoning for this code

adidahiya (commented):
overall looks pretty good!

packages/core/test/multistep-dialog/multistepDialogTests.tsx
adidahiya: unmounting properly sounds like a better change to do in this PR :+1:

packages/core/src/components/overlay/overlay.tsx
michael-yx-wu: I cobbled this together from jquery-ui's [`focusable` selector](https://github.com/jquery/jquery-ui/blob/70dae67b73dfea9126f126f516fe8286f1e73417/ui/focusable.js#L45) and this [SO answer](https://stackoverflow.com/a/1600194/2213364). I will add a comment

michael-yx-wu:
I found a bug where this doesn't properly enforce focus if focus is coming back to the document _from_ another part of the browser (e.g. the URL or bookmarks bar). I will restore the focus event handler logic, but I will listen instead for `focusin` instead of `focus` since the latter isn't supposed to bubble

update: as suggested, I'll address this in a flup

adidahiya (approved):
nice :+1:

packages/core/src/components/overlay/overlay.tsx
adidahiya: we should do something about this in a FLUP. not a high priority though

michael-yx-wu: Filed https://github.com/palantir/blueprint/issues/4908

ro-savage:
@michael-yx-wu - The appears to be a bug or unintentional behaviour with this change on dialogs with scrolls.

See: #4920

PR #4898, opened 2021-09-12
[popover2] feat(Popover2): allow enter key to open popover

styu:
#### Checklist

- [ ] Includes tests
- [x] Update documentation

Currently if the target is focused and the popover is meant to open on click, then you cannot open the popover via keyboard-only navigation.

This change makes it at least so that pressing enter while the target is focused the popover will open. Note that the popover currently always steals focus away from the target, so that you cannot press enter again in the same way that you could click the target again to close the popover. I'm not sure what was intended behavior (stealing focus) or not, so I could use advice as to how would we allow focus to remain on the target such that 2 enter key presses opens and then closes the popover

Any other gotchas with adding this event listener?

#### Screenshot

focus

### DIFF
--- a/packages/popover2/src/popover2.tsx
+++ b/packages/popover2/src/popover2.tsx
@@ -24,6 +24,7 @@ import {
     Classes as CoreClasses,
     DISPLAYNAME_PREFIX,
     HTMLDivProps,
+    Keys,
     refHandler,
     mergeRefs,
     Overlay,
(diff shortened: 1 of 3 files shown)


packages/popover2/src/popover2.tsx
```diff
@@ -308,6 +308,9 @@ export class Popover2<T> extends AbstractPureComponent2<Popover2Props<T>, IPopov
             : {
                   // CLICK needs only one handler
                   onClick: this.handleTargetClick,
+                  // For keyboard accessibility, we trigger the same behavior as a click event upon pressing enter
+                  onKeyDown: (event: React.KeyboardEvent<HTMLElement>) =>
+                      event.key === "Enter" && this.handleTargetClick(event),
```
adidahiya: We use a util function called `Keys.isKeyboardClick` in other places to enable this kind of behavior. I suggest we do the same here, and make the popover open on either ENTER or SPACE. https://github.com/palantir/blueprint/search?q=Keys.isKeyboardClick

adidahiya (changes requested):
in general this sounds like a good idea. let's make it work for the SPACE key too

adidahiya (approved):

PR #4899, opened 2021-09-13
[core,datetime,table] feat: Add semantic color variables

### LINKED ISSUE #4127: Proposal: Semantic Color Variables For Easy Extensibility
#### Feature request
Colors variables in the SCSS should be semantic base on Intent, rather than a specific color. This would make colors much more easily overridden and extended. 
#### Examples

[This is touched on this issue about overriding about button's styling.](https://github.com/palantir/blueprint/issues/3673)

When the $Primary intent color is overridden in Scss, the hover states remain Blue because they are not related to the primary intent. 
This
```scss
// "intent": (default, hover, active colors)
$button-intents: (
  "primary": ($pt-intent-primary, $blue2, $blue1),
  "success": ($pt-intent-success, $green2, $green1),
  "warning": ($pt-intent-warning, $orange2, $orange1),
  "danger": ($pt-intent-danger, $red2, $red1)
) !default;
````
Could be changed to 
```scss
// "intent": (default, hover, active colors)
$button-intents: (
  "primary": ($pt-intent-primary, $pt-intent-primary2, $pt-intent-primary1),
  "success": ($pt-intent-success, $pt-intent-success2, $pt-intent-success1),
  "warning": ($pt-intent-success, $pt-intent-success2, $pt-intent-success1),
  "danger": ($pt-intent-danger, $pt-intent-danger2, $pt-intent-danger1)
) !default;
````

The new semantic color variables would have to be added to the color variables, assigned to default base colors and then changed throughout the SCSS.

Currently, to override a blueprint default color reliably you would have to reassign a Color to a different one, or go through all of the styling a change the color variables. Neither are intuitive or maintainable. This change would make it very easy to extend or override the default Blueprint Colors with some variable changes. 

There may be implementation details I'm not thinking of that would prevent this from going in, but I would love to open a PR if you guys think it is a good idea!

adamboros:
#### Checklist

- [ ] Includes tests
- [ ] Update documentation

Changing theme colors in Blueprint only works if you overwrite the colors itself. For example (as mentioned in #4127), for the primary-intent this means that users need to overwrite `$blue2` and `$blue1` to change the color of hover/active states in buttons. This affects success, warning and danger intents as well. This PR replaces the 'hardcoded' color variables with color-aliases. So everyone can theme as they like, without changing the meaning of `$blue`.

To avoid multiple dark-intent-text color-aliases, in `table/src/common/_variables.scss` this PR levels up colors from level 4 to 5. In my tests level 5 gave better contrast ratios in dark table cells and it was already used in `core/src/common/_mixins.scss` >> `$pt-dark-intent-text-colors`.

It's a non-breaking change, the same colors are referenced as before, but with one additional step (except for dark tables as detailed above where colors actually changed).

This change does not modify the default blueprint setup, only gives a cleaner option to overwrite intent colors.
Contrast ratio for texts in dark tables might worth a look, as the color got a bit brighter.

### DIFF
--- a/packages/core/src/common/_color-aliases.scss
+++ b/packages/core/src/common/_color-aliases.scss
@@ -2,9 +2,24 @@
 // Licensed under the Apache License, Version 2.0.
 
 $pt-intent-primary: $blue3 !default;
+$pt-intent-primary2: $blue2 !default;
+$pt-intent-primary1: $blue1 !default;
+$pt-dark-intent-primary: $blue5 !default;
+
 $pt-intent-success: $green3 !default;
+$pt-intent-success2: $green2 !default;
+$pt-intent-success1: $green1 !default;
+$pt-dark-intent-success: $green5 !default;
+
 $pt-intent-warning: $orange3 !default;
+$pt-intent-warning2: $orange2 !default;
+$pt-intent-warning1: $orange1 !default;
+$pt-dark-intent-warning: $orange5 !default;
+
 $pt-intent-danger: $red3 !default;
+$pt-intent-danger2: $red2 !default;
+$pt-intent-danger1: $red1 !default;
+$pt-dark-intent-danger: $red5 !default;
 
 $pt-app-background-color: $light-gray5 !default;
 $pt-dark-app-background-color: $dark-gray3 !default;
--- a/packages/core/src/common/_mixins.scss
+++ b/packages/core/src/common/_mixins.scss
@@ -12,17 +12,17 @@ $pt-intent-colors: (
 ) !default;
 
 $pt-intent-text-colors: (
-  "primary": $blue2,
-  "success": $green2,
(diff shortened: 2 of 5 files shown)


palantirtech:
Thanks for your interest in [palantir/blueprint](https://github.com/palantir/blueprint), @adamboros! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

adidahiya:
Hey @adamboros, thanks for the contribution. We're currently in the process of modernizing our colors to achieve better contrast; this change will be released soon with Blueprint 5.0 (coming shortly after 4.0). Because of this work, I'd rather not make changes to the color variables at this time. We can revisit in a few months after those changes go through.

PR #4904, opened 2021-09-14
[core] feat(Overlay): add shouldReturnFocusOnClose prop

adidahiya:
#### Checklist

- [ ] Includes tests
- [x] Update documentation

Building on #4422 and #4620, this PR moves the `shouldReturnFocusOnClose` functionality to the `Overlay` component, thereby enabling it by default for Popover/Popover2, Tooltip/Tooltip2, and Dialogs. This is desirable for keyboard accessibility. Drawers still retain this behavior since they extend Overlay / OverlayableProps.

### DIFF
--- a/packages/core/src/components/drawer/drawer.tsx
+++ b/packages/core/src/components/drawer/drawer.tsx
@@ -67,14 +67,6 @@ export interface IDrawerProps extends OverlayableProps, IBackdropProps, Props {
     position?: Position;
 
     /**
-     * Whether the application should return focus to the last active element in the
-     * document after this drawer closes.
-     *
-     * @default true
-     */
-    shouldReturnFocusOnClose?: boolean;
-
-    /**
      * CSS size of the drawer. This sets `width` if `vertical={false}` (default)
      * and `height` otherwise.
      *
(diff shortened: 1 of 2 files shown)


packages/core/src/components/drawer/drawer.tsx
```diff
@@ -66,14 +66,6 @@ export interface IDrawerProps extends OverlayableProps, IBackdropProps, Props {
      */
     position?: Position;
 
-    /**
-     * Whether the application should return focus to the last active element in the
-     * document after this drawer closes.
-     *
-     * @default true
-     */
-    shouldReturnFocusOnClose?: boolean;
```
styu, line 75: is this not a break?

styu (commented):
code looks + previews great! just a question on moving the prop

packages/core/src/components/drawer/drawer.tsx
michael-yx-wu: I think it's more like a QOL improvement for anyone that relies on keyboard navigation, since the focus would previously return to `body`. If you were manually setting focus in `onClose`, that still runs at the end of this change so your custom focus behavior will still work.

adidahiya: it's not a break, `DrawerProps` extends from `OverlayableProps`!

styu: ah, i missed that! 👍

PR #4906, opened 2021-09-14
[colors] feat: add modern colors in preparation for v5.x

johnly13:
#### Checklist

- [ ] Includes tests
- [ ] Update documentation

* Adds new colors that have more accessible contrast ratios within Blueprint components.
* Once this is released and `@blueprintjs/colors@5.x` is published, I'll make follow-up PRs to:
  * **Add these new colors into Blueprint core** by upgrading the `@blueprintjs/colors` dependency and creating a new `blueprint-modern-colors.scss`, which contains the line `@import "@blueprintjs/colors/lib/css/colors.css`
  * **Create new less/scss variable files** in `@blueprintjs/core` that has the line `@import "@blueprintjs/colors/lib/css/colors.css`, along with the rest of the blueprint variables (will require an update to `generate-css-variables.js` to accept an output file name)
  * **Add remaining color-related style changes** to `packages/colors/src/colors.scss`


packages/colors/package.json
```diff
@@ -3,9 +3,16 @@
     "version": "3.0.0",
     "description": "Blueprint color definitions",
     "main": "lib/index.js",
+    "style": "lib/css/colors.css",
```
adidahiya: this doesn't produce any CSS, so we don't need it

packages/colors/package.json
```diff
@@ -3,9 +3,16 @@
...
+    "sideEffects": [
```
adidahiya: same here, no side effects if there's no CSS

packages/colors/package.json
```diff
@@ -3,9 +3,16 @@
...
+        "**/*.css"
+    ],
     "scripts": {
         "clean": "rm -rf lib/*",
-        "compile": "tsc -p src/",
+        "compile": "run-p \"compile:*\"",
+        "compile:esm": "tsc -p src/",
+        "compile:css": "sass-compile ./src",
```
adidahiya, line 10: we _do_ still want this so that we can validate that we're using valid Sass syntax.

packages/colors/package.json
```diff
@@ -3,9 +3,16 @@
...
+        "dev": "run-p \"compile:esm -- --watch\" \"compile:css -- --watch\"",
         "lint": "run-p lint:es",
```
adidahiya: please add stylelint and lint the Sass source. currently I noticed some whitespace lint errors.

packages/core/package.json
```diff
@@ -46,7 +46,7 @@
         "verify": "npm-run-all compile -p dist test lint"
     },
     "dependencies": {
-        "@blueprintjs/colors": "^3.0.0",
+        "@blueprintjs/colors": ">3.0.0",
```
adidahiya:
```suggestion
        "@blueprintjs/colors": ">=3.0.0",
```

packages/stylelint-plugin/package.json
```diff
@@ -12,7 +12,7 @@
         "test": "mocha test/index.js"
     },
     "dependencies": {
-        "@blueprintjs/colors": "^3.0.0",
+        "@blueprintjs/colors": ">3.0.0",
```
adidahiya:
```suggestion
        "@blueprintjs/colors": ">=3.0.0",
```

adidahiya (changes requested):
let's add a `legacyColors.ts` file with the old colors...

```ts
// legacyColors.ts

export const LegacyColors = {
  // ...
};
```

and then core should re-export these colors to avoid a breaking change:

```ts
// core/src/common/index.ts

import { LegacyColors } from "@blueprintjs/colors";

export Colors = LegacyColors;
```

consumers who wish to use the new colors can just switch to the new package:

```ts
import { Colors } from "@blueprintjs/colors";
```

packages/colors/package.json
johnly13: ah i see, i noticed that there was a `lib/css/colors.css` that was created but is empty except for the source mapping. wasn't sure if that required adding this or not

adidahiya:
not sure why the CI workflow wasn't getting triggered before... but after pushing my commit merging develop it looks like it's running

adidahiya:
I added `LegacyColors`. Now there are a couple remaining issues with this PR, hoping you can address them @johnly13:

### Cascading variable declarations

We are moving to an API where consumers wishing to use the new colors in 3.x (and the eventual stable API in 5.0) are recommended to import multiple Sass files if they use both colors and other BP variables (like grid size):

```scss
@import "~@blueprintjs/colors/lib/scss/colors";
@import "~@blueprintjs/core/lib/scss/variables";
```

The problem with this is that `@blueprintjs/core/lib/scss/variables.scss` doesn't include [the `! default` syntax](https://sass-lang.com/documentation/variables#default-values) in its variable declarations (the `generate-css-variables` script strips them out), so the color values here will override the new modernized colors. And we can't import `@blueprintjs/colors/lib/scss/colors.scss` _after_ `variables.scss` because the latter includes color aliases which depend on the color hex values. (Note: we should write a stylelint rule which enforces import ordering to avoid this error, as well)

__I recommend changing `generate-css-variables.js` to allow _retaining_ the `! default` syntax through a CLI option so that this API with multiple Sass imports can work__.

### Color variables file name

The packages/colors `dist:variables` script produces a `lib/scss/variables.scss` file, when we want a `lib/scss/colors.scss` file.

__I recommend changing `generate-css-variables.js` to allow specifying output filename with a `--outputFilename` CLI option__.

adidahiya (approved):

adidahiya:
the behavior of the stylelint rule to avoid color literals has changed now. I think we should make it use `LegacyColors` to retain the old behavior. we can think about how to incorporate new `Colors` in 5.0

adidahiya (approved):

ptgamr:
Upgraded `@blueprintjs/core` yesterday and new colors look great! 

However, I noticed the `popover2` package is still using the old colors, is there plan to update this yet?

adidahiya:
@ptgamr use popover2 v1.0.0-beta.x, the new colors are there

PR #4907, opened 2021-09-14
[core] fix(Overlay): capture focus when it returns to document

michael-yx-wu:
Follow up to https://github.com/palantir/blueprint/issues/4894#issuecomment-919384799

#### Screenshot

Before -- focus would not return to overlay if coming back to the document from somewhere outside (e.g. url, native browser controls)

After -- focus immediately returns to overlay

### DIFF
--- a/packages/core/src/components/overlay/overlay.tsx
+++ b/packages/core/src/components/overlay/overlay.tsx
@@ -249,7 +249,7 @@ export class Overlay extends AbstractPureComponent2<OverlayProps, IOverlayState>
     private refHandlers = {
         // HACKHACK: see https://github.com/palantir/blueprint/issues/3979
         /* eslint-disable-next-line react/no-find-dom-node */
-        container: (ref: TransitionGroup) => (this.containerElement = findDOMNode(ref) as HTMLElement),
+        container: (ref: TransitionGroup | null) => (this.containerElement = findDOMNode(ref) as HTMLElement),
         firstFocusable: (ref: HTMLDivElement | null) => {
             this.startFocusTrapElement = ref;
             ref?.addEventListener("focusin", this.handleStartFocusTrapElementFocusIn);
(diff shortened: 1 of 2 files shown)


packages/core/src/components/overlay/overlay.tsx
```diff
@@ -359,13 +359,11 @@ export class Overlay extends AbstractPureComponent2<OverlayProps, IOverlayState>
...
 
         // add a special class to each child element that will automatically set the appropriate
-        // CSS position mode under the hood. also, make the container focusable so we can
-        // trap focus inside it (via `enforceFocus`).
+        // CSS position mode under the hood.
         const decoratedChild =
             typeof child === "object" ? (
                 React.cloneElement(child as React.ReactElement, {
                     className: classNames((child as React.ReactElement).props.className, Classes.OVERLAY_CONTENT),
-                    tabIndex: this.props.enforceFocus || this.props.autoFocus ? 0 : undefined,
```
michael-yx-wu, line 380: removing this fixes an issue w/ the omnibar where the container _and_ the input are both focusable, even though you probably don't want the container to be focusable in that situation. now that we have focus traps, this doesn't seem necessary. if consumers want non-focusable elements to be focusable, they will need to take care to set tabindex themselves. not sure if this constitutes a break

adidahiya (changes requested):
code is a little convoluted, but seems ok...

there is a bug now with dialog focus... seems a little too aggressive. you can't focus the primary button or the close button via the TAB key:

michael-yx-wu:
I think this is a result of the infinite focus bug due to `shouldReturnFocusOnClose` that was fixed in 49a2a4622f85518280a50d71d7a3809cd6f7ed34. Merged in develop and it works properly now:

adidahiya (approved):

PR #4915, opened 2021-09-16
Add core component changes for modern v5.x colors

johnly13:

* [x] Add style/color changes for core components using new v5.x colors
* [X] Write simple test app
* [x] Verify last remaining components
  * [x] Checkbox
  * [x] Radio
  * [ ] Slider
  * [ ] Switch
  * [ ] Tooltip

#### Not done in this PR (punting for future PRs)

* Add styling updates for other component packages (datetime, select, table)
* Clean up `packages/core/src/blueprint-modern.scss` - break Sass code out into individual files based on component


packages/core/src/blueprint-modern.scss
```diff
@@ -0,0 +1,548 @@
...
+        @include pt-button-minimal-intent-modern(
+          map-get($button-active-minimal-intent-text-modern, $intent),
+        );
+      }
+    }
+  }
+
+  .#{$ns}-dark & {
+    // TODO: Fix this maybe
+    &:not([class*="#{$ns}-intent-"]) &.#{$ns}-minimal {
```
johnly13: not sure this is actually valid

packages/core/src/blueprint-modern.scss
```diff
@@ -0,0 +1,548 @@
...
+
+// Import files in the same order that they are documented in the docs
+@import "common/variables-modern";
+@import "~@blueprintjs/colors/lib/css/colors.css";
+
+@import "reset";
+@import "typography";
+@import "accessibility/focus-states";
+
+// Button variables and overrides
```
johnly13: Some of these variables override the existing defaults, e.g. `$button-intents` so these need to be set before importing components. Others are just new variables that I needed to make e.g. `$button-active-minimal-intent-text-modern`, which don't necessarily need to be set before importing but I figured I'd keep all variables in one place

johnly13:
I thought about if these changes should live in `@bluieprintjs/colors/lib/css/colors.css` instead, but I don't think that'll work since the colors would need to be imported before the components in `packages/core`, meaning that the style changes would be overwritten when importing the components. 

I could instead set the variables I need to set in `@blueprintjs/colors/lib/css/colors.css` and then have a separate file (e.g. `$blueprintjs/colors/lib/css/modern-colors-components.css`) with all of the component changes, which I can import after importing the components in `@blueprintjs/core`.

packages/core/src/common/_variables-modern.scss
```diff
@@ -0,0 +1,2 @@
+@import "~blueprintjs/colors/lib/css/colors.css";
```
adidahiya: I think you meant to import the `lib/scss/colors` file

packages/core/src/blueprint-modern.scss
adidahiya: seems like it's used in other places in the code base, and it works fine in the generated CSS (see line 480, for example, [here](https://unpkg.com/browse/@blueprintjs/core@3.50.2/lib/css/blueprint.css))

packages/core/src/blueprint-modern.scss
```diff
@@ -0,0 +1,692 @@
...
+
+      &:active td {
+        background-color: rgba($gray3, 0.4);
+      }
+    }
+  }
+}
+
+// Contrast for icons
+.#{$ns}-icon:not([class*="#{$ns}-intent-"]) {
```
johnly13: By default the non-intent colors use `$pt-text-color` instead of `$pt-icon-color` (i.e. `$pt-text-color-muted`) and likewise for dark mode, because they were being imported from `body`:
Screen Shot 2021-09-21 at 15 33 25

I've overwritten the color here so that we don't just default to the default text colors/to match the colors that are in the proposed designs, but I'm not sure if this will break anything down the line. I.e., if we're breaking icon color inheritance somewhere else

packages/colors/package.json
```diff
@@ -8,9 +8,8 @@
         "compile": "run-p \"compile:*\"",
         "compile:esm": "tsc -p src/",
         "compile:css": "sass-compile ./src",
+        "compile:colors-css": "generate-css-variables --outputFileName colors _colors.scss",
```
johnly13: This has to happen in the compile step, since compiling `packages/core` depends on `@blueprintjs/colors/lib/scss/colors` existing

packages/demo-app/package.json
```diff
@@ -0,0 +1,68 @@
...
+        "@blueprintjs/docs-data": "^3.50.1",
+        "@blueprintjs/docs-theme": "^3.10.11",
+        "@blueprintjs/icons": "^3.30.1",
+        "@blueprintjs/popover2": "^0.12.1",
+        "@blueprintjs/select": "^3.18.3",
+        "@blueprintjs/table": "^3.9.5",
+        "@blueprintjs/test-commons": "^0.10.13",
+        "@blueprintjs/timezone": "^3.9.11",
+        "@documentalist/client": "~3.0.0",
+        "chroma-js": "^2.1.0",
```
adidahiya: we don't need this

packages/demo-app/package.json
```diff
@@ -0,0 +1,68 @@
...
+        "lint:scss": "sass-lint",
+        "lint:es": "es-lint",
+        "lint-fix": "es-lint --fix && sass-lint --fix",
+        "test": "exit 0",
+        "verify": "run-p dist lint"
+    },
+    "dependencies": {
+        "@blueprintjs/core": "^3.50.1",
+        "@blueprintjs/datetime": "^3.23.11",
+        "@blueprintjs/docs-data": "^3.50.1",
```
adidahiya: don't need this

packages/demo-app/package.json
```diff
@@ -0,0 +1,68 @@
...
+        "@blueprintjs/docs-theme": "^3.10.11",
```
adidahiya: I don't think we need this

packages/demo-app/package.json
```diff
@@ -0,0 +1,68 @@
...
+        "@blueprintjs/icons": "^3.30.1",
+        "@blueprintjs/popover2": "^0.12.1",
+        "@blueprintjs/select": "^3.18.3",
+        "@blueprintjs/table": "^3.9.5",
+        "@blueprintjs/test-commons": "^0.10.13",
```
adidahiya: don't need this

packages/demo-app/package.json
```diff
@@ -0,0 +1,68 @@
...
+        "@blueprintjs/timezone": "^3.9.11",
+        "@documentalist/client": "~3.0.0",
```
adidahiya: probably don't need this

packages/demo-app/package.json
```diff
@@ -0,0 +1,68 @@
...
+        "chroma-js": "^2.1.0",
+        "classnames": "^2.2",
+        "core-js": "^3.8.1",
+        "dom4": "^2.1.5",
+        "downloadjs": "^1.4.7",
```
adidahiya: don't need this

packages/demo-app/package.json
```diff
@@ -0,0 +1,68 @@
...
+        "lodash-es": "^4.17.15",
+        "moment": "^2.29.1",
+        "normalize.css": "^8.0.1",
+        "popper.js": "^1.16.1",
```
adidahiya: core imports this for you, so you shouldn't need this

packages/demo-app/package.json
```diff
@@ -0,0 +1,68 @@
...
+        "promise": "^8.1.0",
```
adidahiya: let's remove this; IE11 is going away

packages/demo-app/package.json
```diff
@@ -0,0 +1,68 @@
...
+        "react": "^16.14.0",
+        "react-dom": "^16.14.0",
+        "react-transition-group": "^4.4.1",
+        "tslib": "~1.13.0",
+        "whatwg-fetch": "^3.4.1"
```
adidahiya: same here, let's remove this since IE11 is going away

packages/demo-app/package.json
```diff
@@ -0,0 +1,68 @@
...
+    },
+    "devDependencies": {
+        "@blueprintjs/node-build-scripts": "^1.6.0",
+        "@blueprintjs/webpack-build-scripts": "^2.0.2",
+        "@types/chroma-js": "^2.1.2",
```
adidahiya: remove

packages/demo-app/package.json
```diff
@@ -0,0 +1,68 @@
...
+        "@types/downloadjs": "^1.4.2",
```
adidahiya: remove

packages/demo-app/polyfill.js
```diff
@@ -0,0 +1,52 @@
...
+ * limitations under the License.
+ */
+
+/**
+ * Require the minimal set of ES2015+ polyfills from `core-js`, `promise`, and `whatwg-fetch` to run in IE11+.
+ * Adapted from the "react-app-polyfill" package.
+ * See "NPM Installation" section of docs homepage for more information.
+ */
+
+if (typeof Promise === "undefined") {
```
adidahiya: let's remove this and the `whatwg-fetch` part, since IE11 is going away

packages/demo-app/src/examples/BreadcrumbExample.tsx
```diff
@@ -0,0 +1,36 @@
...
+import { ExampleCard } from "./ExampleCard";
+
+const ITEMS: BreadcrumbProps[] = [
+    { icon: "folder-close", text: "All files" },
+    { icon: "folder-close", text: "Unselectable" },
+    { href: "#", icon: "folder-close", text: "Link" },
+    { icon: "document", text: "Selected", current: true },
+];
+
+export class BreadcrumbExample extends React.PureComponent {
```
adidahiya, line 30: these can all be function components, but since you've already written them as classes I wouldn't worry too much about converting them. new components you write can be FCs, for simplicity...

packages/demo-app/src/examples/ButtonExample.tsx
```diff
@@ -0,0 +1,46 @@
...
+                    ))}
+                </ExampleCard>
+                <ExampleCard width={WIDTH}>
+                    {Object.values(Intent).map(intent => (
+                        <Button
+                            minimal={true}
+                            key={`${intent}-button`}
+                            intent={intent as Intent}
+                            text="Button"
+                            icon={IconNames.ADD}
```
adidahiya, line 62: note: you don't actually have to import `IconNames`, you can just use the string literals, like `"add"` (they get type checked). but it's not a big deal, no need to change what you have here

packages/demo-app/src/examples/CheckboxExample.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+import { Checkbox, Radio } from "@blueprintjs/core";
+
+import { ExampleCard } from "./ExampleCard";
+
+export class CheckboxRadioExample extends React.PureComponent {
+    public render() {
+        return (
+            <div className="example-row">
+                <ExampleCard width={200}>
```
adidahiya: nit: this looks a little awkward center aligned; can we make it left-aligned?

image

packages/demo-app/src/examples/Examples.tsx
```diff
@@ -0,0 +1,81 @@
...
+import { SwitchExample } from "./SwitchExample";
+import { TabsExample } from "./TabsExample";
+import { TagExample } from "./TagExample";
+import { TextExample } from "./TextExample";
+import { ToastExample } from "./ToastExample";
+import { TooltipExample } from "./TooltipExample";
+import { TreeExample } from "./TreeExample";
+
+export class Examples extends React.PureComponent {
+    private renderExamples(isDarkMode: boolean) {
```
adidahiya: nit: I would change the API to just take a class name string:

```
private renderExamples(classes: string) {


this.renderExamples(Classes.DARK)
```

packages/demo-app/src/examples/IconExample.tsx
```diff
@@ -0,0 +1,34 @@
...
+import { IconNames } from "@blueprintjs/icons";
+
+import { ExampleCard } from "./ExampleCard";
+
+export class IconExample extends React.PureComponent {
+    public render() {
+        return (
+            <div className="example-row">
+                <ExampleCard horizontal={true}>
+                    {Object.values(Intent).map(intent => (
```
adidahiya, line 27: I think this section needs a label of some kind... it took me a second to realize which component was being demo'd

image

packages/demo-app/src/examples/SliderExample.tsx
```diff
@@ -0,0 +1,59 @@
...
+        return (
+            <ExampleCard>
+                <Slider
+                    min={0}
+                    max={10}
+                    stepSize={0.1}
+                    labelStepSize={10}
+                    onChange={this.getChangeHandler("value1")}
+                    value={this.state.value1}
+                    vertical={true}
```
adidahiya: I think just the regular slider is sufficient... we don't really need the vertical one since it looks basically the same

packages/demo-app/src/examples/TextExample.tsx
```diff
@@ -0,0 +1,42 @@
...
+import { Classes, Intent, Text } from "@blueprintjs/core";
+
+import { ExampleCard } from "./ExampleCard";
+
+const WIDTH = 200;
+export class TextExample extends React.PureComponent {
+    public render() {
+        return (
+            <div className="example-row">
+                <ExampleCard width={WIDTH}>
```
adidahiya: I think this example also needs a heading... maybe we should add section heading to all the ExampleCards?

packages/demo-app/src/styles/_examples.scss
```diff
@@ -0,0 +1,141 @@
+@import "~@blueprintjs/core/lib/css/blueprint-modern.css";
```
adidahiya: I don't think you need this import; you're already importing this stylesheet in `index.scss`

packages/node-build-scripts/generate-css-variables.js
```diff
@@ -52,7 +52,7 @@ if (!args["retainDefault"]) {
 variablesScss = COPYRIGHT_HEADER + "\n" + variablesScss;
 
 if (!fs.existsSync(`${DEST_DIR}/scss`)) {
-    fs.mkdirSync(`${DEST_DIR}/scss`);
+    fs.mkdirSync(`${DEST_DIR}/scss`, { recursive: true });
```
adidahiya, line 55: interesting; do you remember why you made this change?

scripts/preview.js
```diff
@@ -9,6 +9,7 @@ const ARTIFACTS = {
     "packages/docs-app/dist/index.html": "documentation",
     "packages/landing-app/dist/index.html": "landing",
     "packages/table-dev-app/dist/index.html": "table",
+    "packages/demo-app/dist/index.html": "demo"
```
adidahiya:
```suggestion
    "packages/demo-app/dist/index.html": "modern colors demo"
```

adidahiya (changes requested):

adidahiya:
I will do another review pass for `blueprint-modern.scss` but this is looking nice so far 😎

packages/node-build-scripts/generate-css-variables.js
johnly13: This was to address the circle build was failing during `yarn compile`. i think because this has to run in the compile stage instead of the `dist` stage like it does for `packages/core`, it must have been running into this line before `compile:css` is able to create the `lib` directory

packages/demo-app/src/examples/TextExample.tsx
johnly13: agreed - will get this added

packages/demo-app/package.json
adidahiya: still don't need this

packages/demo-app/src/styles/_examples.scss
```diff
@@ -0,0 +1,146 @@
...
+  display: flex;
+  flex-direction: column;
+
+  > :not(:last-child) {
+    margin-bottom: 10px;
+  }
+}
+
+.html-table-example {
+  .#{$ns}-intent-primary {
```
adidahiya, line 85: this was a little surprising to me. it's fine for now, but I'm surprised we don't already have a public API to do this without applying custom CSS. something like `<Text intent="...">`

packages/core/src/blueprint-modern.scss
```diff
@@ -0,0 +1,830 @@
...
+      input:not(:disabled):active:checked ~ .#{$ns}-control-indicator {
+        background: linear-gradient(180deg, rgba($white, 0.1) 0%, rgba($white, 0) 100%),
+          linear-gradient(0deg, $blue1, $blue1);
+      }
+    }
+  }
+}
+
+// TODO
+// Contrast for date pickers
```
adidahiya: this won't live in this file, so this TODO should be removed

packages/core/src/blueprint-modern.scss
adidahiya: Icons _are_ supposed to inherit text color in this way, so yeah, I think this could break something further up the application stack.

What was the goal of setting this particular line with `color`? Where do we see it in effect?

packages/core/src/blueprint-modern.scss
```diff
@@ -0,0 +1,830 @@
...
+        color: $hover-text-color;
+      }
+
+      &:active {
+        background-color: rgba($background-color, 0.3);
+        color: $active-text-color;
+      }
+    }
+
+    .#{$ns}-tag-remove >.#{$ns}-icon:first-child {
```
adidahiya: I think this may be missing a space after the chevron... `> .`

packages/core/src/blueprint-modern.scss
```diff
@@ -0,0 +1,830 @@
...
+          color: $hover-text-color;
+        }
+
+        &:active {
+          background-color: rgba($background-color, 0.35);
+          color: $active-text-color;
+        }
+      }
+
+      .#{$ns}-tag-remove >.#{$ns}-icon:first-child {
```
adidahiya: same here; I'm surprised it's not enforced by stylelint? `> .`

packages/core/src/blueprint-modern.scss
```diff
@@ -0,0 +1,830 @@
...
+
+        &:active {
+          color: $active-text-color;
+        }
+      }
+    }
+  }
+
+  // Contrast for default intent
+  color: $white;
```
adidahiya, line 645: nit: need an empty line after this one

packages/core/src/blueprint-modern.scss
```diff
@@ -0,0 +1,830 @@
...
+        @include pt-tag-minimal-dark-intent-modern($color...);
+      }
+    }
+  }
+}
+
+// Contrast for toasts
+.#{$ns}-toast {
+  // HACK needs !important to override the existing hack
+  /* stylelint-disable declaration-no-important */
```
adidahiya, line 780: nit: does this only disable the rule for this Sass block? do we need to re-enable it?

adidahiya (commented):
overall looks pretty good

packages/node-build-scripts/generate-css-variables.js
adidahiya: ok cool, that checks out 👍🏽

packages/demo-app/package.json
adidahiya: nvm, outdated comment

packages/core/src/blueprint-modern.scss
adidahiya: sounds good to me

aycai:
Some dots on the last demo link above:
- Checkbox and radio: in dark mode has an inner border on static and hover modes, I believe should be an outer border (like active state and selected states)
- Editable text: disabled text should use disable styling that matches disable text elsewhere
- HTML table: dark mode hover state might look like it's using the old blue? I could be wrong here (couldn't catch it in the inspector)
- Input: dark mode intent borders should be using $color4
- Tag: dark mode hover state on solid backgrounds changes the text color to a gray
- Tree: dark mode selected state for intent icon doesn't appear to change (right now blends into the selected BG)

packages/demo-app/src/styles/_examples.scss
johnly13: me too. maybe this can be added in the 5.x release - shouldn't be breaking, right?

packages/core/src/blueprint-modern.scss
johnly13: This was originally to fix the info icon for callouts, dialogs, and toasts. I was able to fix these in each of those components so this is no longer necessary

Screen Shot 2021-09-23 at 14 17 07

Screen Shot 2021-09-23 at 14 16 57
Screen Shot 2021-09-23 at 14 16 35
Screen Shot 2021-09-23 at 14 18 15

packages/core/src/blueprint-modern.scss
johnly13: ah yes, it needs to re-enabled

packages/core/src/blueprint-modern.scss
```diff
@@ -218,7 +218,7 @@ $toast-intent-button-colors-modern: (
   .#{$ns}-dark & {
     background-color: rgba($gray3, 0.2);
 
-    &.#{$ns}-callout-icon:not([class*="#{$ns}-intent-"])>.#{$ns}-icon:first-child {
+    &.#{$ns}-callout-icon:not([class*="#{$ns}-intent-"]) > .#{$ns}-icon:first-child {
```
johnly13, line 231: Fixed the rest of the instances of `>.`

packages/demo-app/src/styles/_examples.scss
adidahiya: Just filed https://github.com/palantir/blueprint/issues/4930. it doesn't need to be tied to 5.0, since it's non-breaking

johnly13:
@aycai addressed your design comments. Only thing that I didn't directly touch was your comment about the HTML table - the blue text I used there is just hard coded, since we don't have any default intent colors for `Text`.

aycai:
Ahh shoot I misspoke re: html table dark mode - I was looking at the background color for the hover state that looked closer to the previous gray in hue. Sorry about that!

aycai:
Final nit on tree: not sure how icon color is defined in that context but it should be $color5 for dark mode, looks like it's $color2 (which is the color for light mode) in the test app.

Otherwise updates look good!

johnly13:
@aycai fixed!

adidahiya (approved):

PR #4922, opened 2021-09-21
Add `asyncControl` to `EditableText`

### LINKED ISSUE #4678: Support `asyncControl` in the EditableText component
InputGroup has an `asyncControl` prop available. This functionality should be supported in EditableText too

felixgabler:
#### Checklist

- [ ] Includes tests
- [ ] Update documentation

Similarly to `InputGroup`, `EditableText` should support the `asyncControl` prop.

Currently, I reuse `AsyncControllableInput` but also had to make that class support the `multiline` scenario. I wonder whether it is better to copy this class to a `AsyncControllableTextArea` and figure out another way of sharing code instead of having one class. What do you think?

After that decision is made, I will gladly write more tests if required. Any pointers here would also be welcome.


palantirtech:
Thanks for your interest in [palantir/blueprint](https://github.com/palantir/blueprint), @felixgabler! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

PR #4925, opened 2021-09-22
[datetime] fix(DateInput): make popover focusable via keyboard

michael-yx-wu:

Pressing Tab when focused on a `DateInput`'s text field will focus the popover content. Previously, pressing tab would focus the next keyboard-focusable element on the page. Since the popover is rendered with a portal by default, this would usually result in the popover closing.

Does _not_ autofocus the popover to continue to allow users to manually type in a date string if they prefer not to use the popover.

#### Screenshot

Before:

After:

Worth noting that the element tab order in the DatePicker seems weird
1. Left/right arrows
2. Month dropdown
3. Year dropdown

I would expect this order instead:
1. Left arrow
2. Month dropdown
3. Year dropdown
4. Right arrow

### DIFF
--- a/packages/datetime/src/dateInput.tsx
+++ b/packages/datetime/src/dateInput.tsx
@@ -423,7 +423,9 @@ export class DateInput extends AbstractPureComponent2<DateInputProps, IDateInput
         if (e.which === Keys.ENTER) {
             const nextDate = this.parseDate(this.state.valueString);
             this.handleDateChange(nextDate, true, true);
-        } else if (e.which === Keys.TAB && !e.shiftKey && this.state.isOpen) {
+        } else if (e.which === Keys.TAB && e.shiftKey) {
+            this.handleClosePopover();
+        } else if (e.which === Keys.TAB && this.state.isOpen) {
             this.getFirstTabbableElement()?.focus();
             e.preventDefault();
         } else if (e.which === Keys.ESCAPE) {
--- a/packages/datetime/src/dateInput.tsx
+++ b/packages/datetime/src/dateInput.tsx
@@ -419,14 +419,17 @@ export class DateInput extends AbstractPureComponent2<DateInputProps, IDateInput
         if (e.which === Keys.ENTER) {
             const nextDate = this.parseDate(this.state.valueString);
             this.handleDateChange(nextDate, true, true);
-        } else if (e.which === Keys.TAB) {
-            this.setState({ isOpen: false });
+        } else if (e.which === Keys.TAB && this.state.isOpen) {
+            this.getFirstTabbableElement()?.focus();
         } else if (e.which === Keys.ESCAPE) {
             this.setState({ isOpen: false });
             this.inputElement?.blur();
         }
         this.safeInvokeInputProp("onKeyDown", e);
     };
+
+    private getFirstTabbableElement = (): HTMLElement | null =>
+        this.popoverContentElement?.querySelector("input, [tabindex]:not([tabindex='-1'])");
 
     private getLastTabbableElement = () => {
         // Popover contents are well structured, but the selector will need
--- a/packages/datetime/src/dateInput.tsx
+++ b/packages/datetime/src/dateInput.tsx
@@ -423,7 +423,7 @@ export class DateInput extends AbstractPureComponent2<DateInputProps, IDateInput
         if (e.which === Keys.ENTER) {
             const nextDate = this.parseDate(this.state.valueString);
             this.handleDateChange(nextDate, true, true);
-        } else if (e.which === Keys.TAB && this.state.isOpen) {
+        } else if (e.which === Keys.TAB && !e.shiftKey && this.state.isOpen) {
             this.getFirstTabbableElement()?.focus();
             e.preventDefault();
         } else if (e.which === Keys.ESCAPE) {
--- a/packages/datetime/src/dateInput.tsx
+++ b/packages/datetime/src/dateInput.tsx
@@ -187,6 +187,10 @@ export class DateInput extends AbstractPureComponent2<DateInputProps, IDateInput
     public popoverContentElement: HTMLDivElement | null = null;
 
     // Last element in popover that is tabbable, and the one that triggers popover closure
+    // when the user presses shift+TAB on it
+    private firstTabbableElement: HTMLElement | null = null;
+
+    // Last element in popover that is tabbable, and the one that triggers popover closure
     // when the user press TAB on it
     private lastTabbableElement: HTMLElement | null = null;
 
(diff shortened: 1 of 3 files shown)


packages/docs-app/src/examples/core-examples/popoverDismissExample.tsx
```diff
@@ -36,6 +36,8 @@ export class PopoverDismissExample extends React.PureComponent<
         return (
             <Example options={false} {...this.props}>
                 <Popover
+                    autoFocus={false}
+                    enforceFocus={false}
```
michael-yx-wu, line 40: This Popover example continuously reopens by design, but now that the `enforceFocus` behavior is fixed, it will reset the scroll position to this part of the page, making it difficult to interact with the other examples. Happy to do in a separate PR if you prefer

adidahiya: yeah I was mostly just punting this problem because it's fixed in the Popover2 example. no big deal, we can make the change here

adidahiya (commented):
how do you get back to the input after tabbing into the popover? shift+tab doesn't seem to work:

michael-yx-wu:
Shift+tab interactions should be fixed now:

packages/datetime/src/dateInput.tsx
```diff
@@ -421,6 +425,7 @@ export class DateInput extends AbstractPureComponent2<DateInputProps, IDateInput
             this.handleDateChange(nextDate, true, true);
         } else if (e.which === Keys.TAB && this.state.isOpen) {
             this.getFirstTabbableElement()?.focus();
+            e.preventDefault();
```
michael-yx-wu, line 411: Necessary to prevent focusing the 2nd focusable element

adidahiya: sounds like this should be a code comment

```suggestion
            // necessary to prevent focusing the 2nd focusable element
            e.preventDefault();
```

packages/datetime/src/dateInput.tsx
```diff
@@ -419,15 +423,21 @@ export class DateInput extends AbstractPureComponent2<DateInputProps, IDateInput
         if (e.which === Keys.ENTER) {
             const nextDate = this.parseDate(this.state.valueString);
             this.handleDateChange(nextDate, true, true);
-        } else if (e.which === Keys.TAB) {
-            this.setState({ isOpen: false });
+        } else if (e.which === Keys.TAB && e.shiftKey) {
+            this.handleClosePopover();
```
adidahiya, line 407:
```suggestion
            // close popover on SHIFT+TAB key press
            this.handleClosePopover();
```

packages/datetime/src/dateInput.tsx
```diff
@@ -463,24 +473,33 @@ export class DateInput extends AbstractPureComponent2<DateInputProps, IDateInput
...
+                    this.inputElement?.focus();
+                    e.preventDefault();
+                }
                 this.handleClosePopover();
             }
         } else if (relatedTarget != null) {
-            this.unregisterPopoverBlurHandler();
+            this.unregisterPopoverBlurHandlers();
+            this.firstTabbableElement = this.getFirstTabbableElement();
+            this.firstTabbableElement?.addEventListener("blur", this.handlePopoverBlur);
```
adidahiya: this doesn't seem exactly right... I feel like the popover should stay open after a SHIFT+TAB keypress on the first tabbable element?

for example, CLICK, TAB, SHIFT+TAB interaction here:

adidahiya (changes requested):
almost there

packages/datetime/src/dateInput.tsx
adidahiya: I could go either way, though... open to other ideas

packages/datetime/src/dateInput.tsx
michael-yx-wu: ~~I moved some things around and don't need to this anymore~~

just kidding, I still need this. I'll add the comment

packages/datetime/src/dateInput.tsx
```diff
@@ -419,70 +424,55 @@ export class DateInput extends AbstractPureComponent2<DateInputProps, IDateInput
...
-        // a blur handler, so that when:
-        // * user tabs to the next element, popover closes
-        // * focus moves to element within popover, popover stays open
-        const lastTabbableElement = numOfElements > 0 ? tabbableElements[numOfElements - 1] : null;
-
-        return lastTabbableElement as HTMLElement | null;
+    private getKeyboardFocusableElements = (): HTMLElement[] => {
+        const elements: HTMLElement[] = Array.from(
+            this.popoverContentElement?.querySelectorAll(
+                "button:not([disabled]),input,[tabindex]:not([tabindex='-1'])",
```
michael-yx-wu, line 422: This is a subset of the selector used in `Overlay`. I wonder if it's worth sharing that in the future

michael-yx-wu:
Makes sense. I'm now keeping the popover open when switching focus to the input. Also fixed the issue of tabbing in the popover focusing the browser address bar.

michael-yx-wu:
Will fix tests

packages/datetime/test/dateInputTests.tsx
```diff
@@ -92,7 +92,11 @@ describe("<DateInput>", () => {
         assert.isFalse(wrapper.find(Popover).prop("isOpen"));
     });
 
-    it("Popover closes when tabbing on first day of the month", () => {
+    // Skipping because simulate just invokes the function passed to React's "on<EventName>" prop
+    // and doesn't actually simulate anything. Properly testing would require running with an actual
+    // browser and focusing specific elements via the DOM API. This would require changing the Karma
+    // config to run with Chrome instead of ChromeHeadless.
+    it.skip("Popover closes when tabbing on first day of the month", () => {
```
michael-yx-wu, line 99: I do have a working local version but could only get it to work by switching the Karma browser from `ChromeHeadless` to `Chrome`. The problem with simulate is that there's no way to get the `focusin` event handler functions to run because React doesn't provide an `onFocusIn` prop, only `onFocus`. If I use the DOM API instead and call `Element#focus()`, the browser will fire a `focusin` event on my behalf. I'm not sure if that's an anti-pattern for the unit tests.

adidahiya: ok, this is fine to skip for now. testing focus behavior is tricky in these test suites

packages/datetime/src/dateInput.tsx
```diff
@@ -301,6 +305,8 @@ export class DateInput extends AbstractPureComponent2<DateInputProps, IDateInput
         const { popoverProps = {} } = this.props;
         popoverProps.onClose?.(e);
         this.setState({ isOpen: false });
+        this.startFocusBoundaryElement?.removeEventListener("focusin", this.handleStartFocusBoundaryFocusIn);
```
adidahiya: why do we need to use this ref approach and manually bind focusin event handlers?  why can’t it be declarative like normal React code `<div onFocusIn={this.handleEndFocusBoundaryFocusIn} />` ?

michael-yx-wu: React doesn't support `onFocusIn`, but it looks like their `onFocus` prop listens to the `focusin` event under the hood (https://github.com/facebook/react/issues/6410#issuecomment-671499208). I switched to this approach and everything still works as expected.

adidahiya (approved):

PR #4932, opened 2021-09-25
[colors] feat: update modern colors

aycai:
- Grays: slight desaturation
- Primary colors: slight desaturation (except `orange`)
- Updated secondary colors to work similarly to primary colors
  - Replace `cobalt` (which is close to the new blue) with `cerulean` (which is close to our old blue)
  - Desaturated `sepia`

#### Checklist

- [ ] Update documentation

Revision of modernized color palette, plus updates to secondary colors

I replaced `cobalt` (which is close to our new blue) with `cerulean` (which is close to our old blue). I want to make sure I caught everything with the color switch 

#### Screenshot

Screen Shot 2021-09-24 at 8 58 52 PM


packages/core/src/common/_colors.scss
```diff
@@ -7,114 +7,114 @@
...
-$gray3: #8a9ba8 !default;
-$gray4: #a7b6c2 !default;
-$gray5: #bfccd6 !default;
-
-$light-gray1: #ced9e0 !default;
-$light-gray2: #d8e1e8 !default;
-$light-gray3: #e1e8ed !default;
-$light-gray4: #ebf1f5 !default;
-$light-gray5: #f5f8fa !default;
+$black: #111418 !default;
```
adidahiya: these are the legacy colors, not the modern colors... you'll have to remove your changes to this file and instead update the hex values here: https://github.com/palantir/blueprint/blob/1de7170f98c7a0b17c6c6642df11a01d806f577f/packages/colors/src/_colors.scss#L10

also, our stylelint rules require you to use lowercase hex codes. if this is inconvenient when exporting colors from Figma, we can reconsider our [stylelint-config-palantir](https://github.com/palantir/stylelint-config-palantir), but that would happen as a breaking change to the config and would need to be separate from this PR. so for now you'll have to convert all these to lowercase hex codes.

adidahiya (changes requested):

adidahiya (approved):

adidahiya:
I reverted all the changes to packages/docs-app... if you want to see the new colors in a web page, you'll need to update packages/demo-app, since docs-app is still using the legacy colors.

adidahiya:
[latest colors demo](https://48043-71939872-gh.circle-artifacts.com/0/packages/demo-app/dist/index.html) lgtm ✅

packages/colors/src/colors.ts
```diff
@@ -15,116 +15,116 @@
...
+    ORANGE5: "#F7B264",
+
+    RED1: "#942427",
+    RED2: "#B3292D",
+    RED3: "#D33136",
+    RED4: "#E96367",
+    RED5: "#F99498",
 };
 
 // Extended colors are still using the "legacy" palette
```
adidahiya: this comment is no longer accurate

PR #4935, opened 2021-09-28
[core] feat(EditableText): new HTML prop "contentId"

styu:
#### Checklist

- [x] Includes tests
- [ ] Update documentation

I would like to associate the underlying text in `EditableText` to different content outside the `EditableText` via `aria-labeledby`. Doing so requires that the span that contains the text has an `id` attribute that can be passed in, so this PR plumbs such a prop through. Open to better naming suggestions though.

### DIFF
--- a/packages/core/src/components/editable-text/editableText.tsx
+++ b/packages/core/src/components/editable-text/editableText.tsx
@@ -113,6 +113,9 @@ export interface IEditableTextProps extends IntentProps, Props {
     /** Text value of controlled input. */
     value?: string;
 
+    /** ID attribute to pass to the underlying element that contains the text contents. This allows for referencing via aria attributes */
+    textId?: string;
+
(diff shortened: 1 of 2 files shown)


packages/core/src/components/editable-text/editableText.tsx
```diff
@@ -113,6 +113,9 @@ export interface IEditableTextProps extends IntentProps, Props {
     /** Text value of controlled input. */
     value?: string;
 
+    /** ID attribute to pass to the underlying element that contains the text contents. This allows for referencing via aria attributes */
+    textId?: string;
```
adidahiya:
```suggestion
    contentId?: string;
```

adidahiya (changes requested):

adidahiya (approved):

PR #4945, opened 2021-10-01
[v4] [icons] fix: restore support for old IconName keys

adidahiya:

Restore support for icon names access via `IconNames.ERROR`, while retaining support for the new syntax (`IconNames.Error`).

This breaking change happened as a result of the switch to `fantasticon` as the icon build tool, and although IMO it looks cleaner with the pascal case syntax, the costs of this breaking change (thousands of lines of meaningless code diff) outweigh the benefits of the code looking ever so slightly cleaner.

Originally I thought I would be able to accomplish this using a custom `fantasticon` template (similar to how we have `scss` and `css` templates), but it turns out that `ts` templates are unsupported. So instead I had to resort to some hacky type mapping...

The type hacks are not very pretty, but they don't really get expose to consumers in their TS build, so I think it's ok...

#### Screenshot

N/A

### DIFF
--- a/packages/icons/src/iconNames.ts
+++ b/packages/icons/src/iconNames.ts
@@ -1,0 +1,62 @@
+/*
+ * Copyright 2021 Palantir Technologies, Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/* eslint-disable camelcase */
+
+import { snakeCase } from "change-case";
+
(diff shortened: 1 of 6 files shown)


packages/icons/src/iconNames.ts
```diff
@@ -0,0 +1,62 @@
...
+/**
+ * The CamelToSnake converter works ok for PascalCase identifiers, but it
+ * always adds a leading underscore. So we add this simple additional converter
+ * to the chain to strip that underscore.
+ */
+type StripLeadingUnderscore<T extends string> = T extends `_${infer R}` ? `${R}` : T;
+
+/**
+ * This is a hacky implementation of a camelCase to Snake_Case string literal
+ * type converter... it works up to 30 characters, which is fine for our
```
styu, line 47: is there somewhere closer to where icons are defined that we should mention this 30 character limit?

adidahiya: not really 😅 it's up to the icons reviewer (me) to enforce it.

to be fair though, it only affects the older naming convention... if you have a new icon added to the set and it's greater than 30 chars... nothing will really break, you just won't be able to reference it as `IconNames.SOMETHING_VERY_LONG_NOT_SURE_WHY_IM_SCREAMING`... you'll just need to simmer down and use pascal case

styu: ha ok. in fairness, prefer just using the string literal anyway

styu (approved):

PR #4951, opened 2021-10-05
[core] fix(Tabs): use tabIndex of -1 for non-active tabs

styu:
I learned recently that the recommendation (from https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Tab_Role#best_practices) for tab behavior is to only set `tabIndex={0}` for the active tab, and rely on left/right keyboard navigation for navigating the other tabs. This helps reduce the clutter on the screen for a user to have to tab through, while still allowing for navigation within the tablist

The blueprint tabs already support left/right key navigation, so it's mostly just the tabIndex that needed to be changed to respect the above

cc @evansjohnson

### DIFF
--- a/packages/core/src/components/tabs/tabTitle.tsx
+++ b/packages/core/src/components/tabs/tabTitle.tsx
@@ -54,7 +54,7 @@ export class TabTitle extends AbstractPureComponent2<TabTitleProps> {
                 id={generateTabTitleId(parentId, id)}
                 onClick={disabled ? undefined : this.handleClick}
                 role="tab"
-                tabIndex={disabled ? undefined : 0}
+                tabIndex={disabled ? undefined : selected ? 0 : -1}
             >
                 {title}
                 {children}


packages/core/src/components/tabs/tabTitle.tsx
```diff
@@ -54,7 +54,7 @@ export class TabTitle extends AbstractPureComponent2<TabTitleProps> {
                 id={generateTabTitleId(parentId, id)}
                 onClick={disabled ? undefined : this.handleClick}
                 role="tab"
-                tabIndex={disabled ? undefined : 0}
+                tabIndex={disabled ? undefined : selected ? 0 : -1}
```
evansjohnson, line 57: tangentially related to this PR but I think if we set `tabIndex={0}` when disabled here the element may not be visible to screen-readers

not 100% sure on that/not set up right now with a good way to test but it's been a working assumption of mine for awhile that disabled elements should remain in the tab order

if so it would possibly would be more appropriate to add to the wrapping div [here](https://github.com/palantir/blueprint/blob/syu/tabs/packages/core/src/components/tabs/tabs.tsx#L161) but again I'm not sure

evansjohnson (approved):
PR preview works for me!

packages/core/src/components/tabs/tabTitle.tsx
adidahiya: It looks like this code will never set `tabIndex={0}` when it's disabled, though.

adidahiya (approved):

packages/core/src/components/tabs/tabTitle.tsx
styu: yeah, ^ was my understanding as well so i thought this was ok

PR #4963, opened 2021-10-14
[core] fix(Overlay): improve focus management

### LINKED ISSUE #4920: New EnforceFocus on overlay causes dialogs to jump
#### Environment

- __Package version(s)__:
 "@blueprintjs/core": "^3.50.2",
    "@blueprintjs/datetime": "^3.23.12",
    "@blueprintjs/popover2": "^0.12.2",
    "@blueprintjs/select": "^3.18.4",
- __Operating System__: MacOS
- __Browser name and version__: Chrome 93

#### Code Sandbox
https://codesandbox.io/s/blueprint-sandbox-forked-5m8ok?file=/src/index.tsx

#### Steps to reproduce

Open dialog, scroll to the date picker and select a date. The user is thrown back up to the top of the screen (the first thing has can be focused)

#### Actual behavior

The user is thrown back to the first focusable element

#### Expected behavior

The user should remain at the current position

#### Possible solution

I believe this was introduced by https://github.com/palantir/blueprint/pull/4894

When you click the date picker, the dialog loses focus, then when you select a date, the dialog regains focus and the new change now automatically focuses the first element.

I would expect this only to apply when a overlay/dialog/etc was first opened, not after it's already been open.

### LINKED ISSUE #4956: default focus changes producing undesirable behaviors in Dialogs containing other popovers + popup menus
**Environment**
- **Package version(s):** core 3.50.4, popover2 0.12.4
- **Operating System:** macOS, Windows 10 VM
- **Browser name and version:** Chrome on macOS, Edge on Windows 10 VM

Noticed some new issues with popup menus + Dialogs containing popovers after a recent upgrade from `core 3.48.0` -> `core 3.50.4` and `popover2 0.11.2` -> `popover2 0.12.4` that both feel potentially related to some breaking change in the default popover focus behavior: 

1. Opening a Dialog that contains a popover with `interactionKind="hover"` will now trigger the popover to open automatically -- the popover then goes into an infinite open/close loop as the hover target gains + loses focus repeatedly

2. When opening a popover menu, the first item now steals focus from the popover target so

    * if the first `MenuItem` has a submenu then the submenu is opened automatically -- if you hover/`tab` to another `MenuItem` without first mousing over the previously-focused item the submenu won't close (even if the new `MenuItem` opens its own submenu, in which case there can be multiple submenus open at the same time)
    * we can no longer support workflows where users use the keyboard to focus a popover menu target -> `enter` to quickly check the menu contents -> `enter` again to close the menu -- this was previously possible by explicitly defining `autoFocus={false}` on the popover, but it appears that `enter` on a click-triggered popover target now causes a popover with `autoFocus={false}` to close immediately after opening

**Code Sandbox w/ repros for the ^**

https://codesandbox.io/s/bp-focus-related-popover-weirdness-kbycy?file=/src/App.js

Setting `autoFocus={false}` and `enforceFocus={false}` seems to fix the issue with the infinite popover loop for the Dialogs and prevent submenus from automatically opening in popup menus (as seen in the sandbox linked above), but the issue with the multiple submenus still exist if we `tab` between the `MenuItem`s instead of hovering with the mouse + I haven't found a way to restore the `enter` to open menu/`enter` to close menu behavior we were able to support before.

adidahiya:
#### Checklist

- [x] Includes tests
- [x] Update documentation

Adjust Overlay's focus management behavior to avoid programmatically focussing elements inside the overlay contents in most cases, opting for keeping focus on the start/end focus trap elements instead

We should still get the accessibility enhancements added by https://github.com/palantir/blueprint/pull/4894, but without the overly invasive focus management which caused issues like #4956.

### DIFF
--- a/packages/core/src/common/classes.ts
+++ b/packages/core/src/common/classes.ts
@@ -221,6 +221,8 @@ export const OVERLAY_CONTENT = `${OVERLAY}-content`;
 export const OVERLAY_INLINE = `${OVERLAY}-inline`;
 export const OVERLAY_OPEN = `${OVERLAY}-open`;
 export const OVERLAY_SCROLL_CONTAINER = `${OVERLAY}-scroll-container`;
+export const OVERLAY_START_FOCUS_TRAP = `${OVERLAY}-start-focus-trap`;
+export const OVERLAY_END_FOCUS_TRAP = `${OVERLAY}-end-focus-trap`;
 
 export const PANEL_STACK = `${NS}-panel-stack`;
 export const PANEL_STACK_HEADER = `${PANEL_STACK}-header`;
--- a/packages/core/src/components/overlay/overlay.tsx
+++ b/packages/core/src/components/overlay/overlay.tsx
@@ -231,6 +231,8 @@ export class Overlay extends AbstractPureComponent2<OverlayProps, IOverlayState>
 
     private static getLastOpened = () => Overlay.openStack[Overlay.openStack.length - 1];
 
+    private isAutoFocusing = false;
+
     private lastActiveElementBeforeOpened: Element | null | undefined;
 
     public state: IOverlayState = {
@@ -266,7 +268,7 @@ export class Overlay extends AbstractPureComponent2<OverlayProps, IOverlayState>
             return null;
         }
 
-        const { children, className, enforceFocus, usePortal, isOpen } = this.props;
+        const { autoFocus, children, className, enforceFocus, usePortal, isOpen } = this.props;
 
         // TransitionGroup types require single array of children; does not support nested arrays.
         // So we must collapse backdrop and children into one array, and every item must be wrapped in a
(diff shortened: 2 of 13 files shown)


adidahiya:
There are issues with reverse TAB key navigation with this change. Here's an example where I press TAB x4 times and then SHIFT+TAB x4 times after clicking the button to open this dialog:



When pressing SHIFT+TAB, we expect the focus to cycle back from the end of the dialog and up through the focusable elements. Instead, it appears that nothing happens and the focus stays on the end focus trap element.

adidahiya:
Things are better now, but there's still a minor issue where the _second_ to last focusable element in the dialog seems to get focussed after pressing Shift+Tab at the end of the dialog. It's turning out to be a tricky one to fix... will probably just punt it for later, since this change seems like a strict improvement over the current state of things.

adidahiya:
Ok, I've now fixed the issue I mentioned in the previous comment. This should be good to go :shipit:

PR #4966, opened 2021-10-15
Fix popover click behavior

### LINKED ISSUE #4956: default focus changes producing undesirable behaviors in Dialogs containing other popovers + popup menus

michael-yx-wu:
#### Fixes https://github.com/palantir/blueprint/issues/4956

Currently failing checks due to code coverage, which I'm happy to fix if we want to go with this approach.

#### Checklist

- [x] Includes tests
- [x] Update documentation (n/a)

There are 2 problems at play that lead to the popover auto-closing behavior reported in https://github.com/palantir/blueprint/issues/4956:
1. When handling `keydown` events triggered by enter or space, programmatically moving focus to another element (as is the case for autofocus Popovers), will cause a `click` event to be created on the new element. This is because `keydown` events occur before `click` and browsers treat enter/space keypresses as clicks for a11y reasons (devs historically implemented onClick but not onKeyDown and it seemed like a cheap way to make websites compliant without code change). Simple repro: https://codesandbox.io/s/focus-change-clicks-lpud5?file=/src/App.jsx

   **Solution:** call `Event#preventDefault` in the Popover(2) click event handler to prevent the click event from being created.

2. AbstractButton calls preventDefault for keyboard clicks (enter/space keys) on keydown and sends a fake "click" event on keyup. As it turns out, the keyup event cannot be prevented with `Event#preventDefault`; assuming you release the key eventually, it will always run! This in turn means the fake click cannot be prevented. When handling keydown and programmatically moving focus, this means that the new element will always receive a fake click event. Here is a simple repro demonstrating this bizarre behavior: https://codesandbox.io/s/cannot-prevent-keyup-1vs20?file=/src/App.jsx

   **Solution:** AbstractButton no longer prevents native clicks on keydown nor does it send a fake "click" on keyup. It seems better to let the native browser click events flow through.

#### Screenshot

No loner autocloses:

If users do not want to have the single focusable element be focused, they should set `autoFocus: false`

### DIFF
--- a/packages/popover2/src/popover2.tsx
+++ b/packages/popover2/src/popover2.tsx
@@ -435,7 +435,7 @@ export class Popover2<T> extends AbstractPureComponent2<Popover2Props<T>, IPopov
 
         return (
             <Overlay
-                autoFocus={this.props.autoFocus}
+                autoFocus={this.props.autoFocus ?? (this.isHoverInteractionKind() ? false : undefined)}
                 backdropClassName={Classes.POPOVER2_BACKDROP}
                 backdropProps={this.props.backdropProps}
                 canEscapeKeyClose={this.props.canEscapeKeyClose}
--- a/packages/popover2/src/popover2.tsx
+++ b/packages/popover2/src/popover2.tsx
@@ -400,11 +400,18 @@ export class Popover2<T> extends AbstractPureComponent2<Popover2Props<T>, IPopov
         this.popperScheduleUpdate = popperProps.update;
 
         const popoverHandlers: HTMLDivProps = {
-            // always check popover clicks for dismiss class
-            onClick: this.handlePopoverClick,
             // treat ENTER/SPACE keys the same as a click for accessibility
             // eslint-disable-next-line deprecation/deprecation
             onKeyDown: event => Keys.isKeyboardClick(event.keyCode) && this.handlePopoverClick(event),
+            // Always check popover clicks for dismiss class. Using onMouseDown instead of onClick.
+            // If the Overlay is opened with autofocus=true, it will move focus to the first
+            // keyboard-focusable element. After pressing the enter key, programmatically moving
+            // focus to another element will result in new click event being created for that
+            // element. The usual workaround is to call event#preventDefault but because the overlay
+            // autofocus behavior is event-based, we cannot prevent the click event being created.
+            // Using onMouseDown instead of onClick allows us to sidestep the problem by listening
+            // to a different event.
+            onMouseDown: this.handlePopoverClick,
         };
         if (
             interactionKind === Popover2InteractionKind.HOVER ||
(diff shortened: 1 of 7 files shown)


packages/popover2/src/popover2.tsx
```diff
@@ -639,6 +639,8 @@ export class Popover2<T> extends AbstractPureComponent2<Popover2Props<T>, IPopov
     };
 
     private handleTargetClick = (e: React.MouseEvent<HTMLElement> | React.KeyboardEvent<HTMLElement>) => {
+        // prevents default click action in case focus popover moves focus to a clickable element
+        e.preventDefault();
```
adidahiya, line 643: I don't feel great about this line... I know we're removing one instance of `e.preventDefault()` with this PR, but I'm very wary of adding it back somewhere else, afraid it might cause unintended breaks. I'd rather adjust the focus trap behavior as I've done in #4963.

adidahiya (changes requested):

adidahiya:
Closing in favor of #4963

PR #4969, opened 2021-10-15
[core] feat(MultistepDialog): Improve appearance without title

mud:
#### Fixes
When `title` is not specified to the `MultistepDialog` component a couple problems arise.

1. The top corners are no longer rounded
2. There's no close button since it's on the title bar

#### Checklist

- [ ] Includes tests
- [x] Update documentation

* Add rounded corners when the title bar is not rendered.
* Add close button in the footer, aligned to the right.

* Make sure style changes are reasonable. It uses `:first-child` to determine whether the title bar is rendered or not.
* The close button in the footer can be toggled with `isFooterCloseButtonShown` and defaults to `false` to preserve current behavior. Should `isCloseButtonShown` from `Dialog` influence this default at all?

I also want to add an option  to specify the position of the navigation (Left, Top, Right) but will do that after this PR.

##### Before (note the top corners)
Screen Shot 2021-10-15 at 12 32 05 PM

##### After

* Top corners are fixed
* There is a close button that can be toggled off
* 
Screen Shot 2021-10-15 at 1 41 12 PM


adidahiya:
For some reason CI wasn't building... I just pushed a merge commit to try and get that going

adidahiya (approved):
nice, thanks for the clean PR @mud

PR #4970, opened 2021-10-15
[core] feat(MultistepDialog): navigationPosition prop

mud:
#### Follow up to #4969 

This PR builds on top of #4969 by adding an option to position the navigation element in `MultistepDialog`. In addition to the default (`Position.LEFT`) `Position.TOP` and `Position.RIGHT` are possible.

#### Checklist

- [ ] Includes tests
- [x] Update documentation

Add a new prop to `MultistepDialog` which allows the positioning of the navigation element to the top or right.

This is all done through style changes by applying a top level class for each position. All the overrides are in `.#{$ns}-multistep-dialog-panels` so it's all in one place, but alternative is to add position overrides to each existing class name where they appear.

#### Position.TOP
Screen Shot 2021-10-15 at 3 08 08 PM
Screen Shot 2021-10-15 at 3 07 58 PM

#### Position.RIGHT
Screen Shot 2021-10-15 at 2 57 33 PM
Screen Shot 2021-10-15 at 2 57 23 PM


mud:
Hey @adidahiya I merged develop and clean up this PR in case you reconsider the position feature. Feel free to close if you don't think it's worthwhile to support -- I can create a custom component in-app instead.

aycai:
Chatted with the original designer for some of her thoughts and relaying here along with mine:
- Overall adding positioning seems reasonable. I'd say having the steps on the right side should be used with caution (for information hierarchy concerns) but some edge cases (e.g. RTL languages) might reasonably use this and at the end of the day it's up to each consumer.
- I'd want to keep the header present with `Position.TOP` and discourage removing it to keep the structure of the dialog intact, as well as to keep the "close" button present.
- `Position.TOP` makes this more apparent, but the differing background colors feel wonky when positioned next to both the header (with the white background) and the body (with the light grey background). This might be a separate change to consider outside the scope of positioning, but flagging here.

mud:
Thanks @aycai 

To address your second point, we fixed issues some issues with omitting `title` in #4969. So there is a way to close the dialog without the title.

As for discouraging `Position.RIGHT`, I just added that option because it was easy and reasonable. `Position.BOTTOM` didn't make sense since it would conflict with the footer. I'm happy to work on RTL support, especially since we'd need to flip the footer navigation (which currently is not possible.)

What I would like to get from the Blueprint team is 1) what can we do in this PR for you to accept the `Position` option and 2) any additional FLUPs to work on (ie. RTL support) which I'm happy to implement.

aycai:
Gotcha, didn't realize the hidden header is already existing behavior. Moving the close button makes sense in a regular dialog, though the hierarchy gets a little wonky if the steps are to the left and outside the footer but the button to close the entire dialog is in that footer.

As for this change no other flags on the design side!

adidahiya:
@aycai the `title` was removable before, but there were some styling issues which likely discouraged that kind of usage of the component. We only fixed those styling issues yesterday in https://github.com/palantir/blueprint/pull/4969, thereby adding "full" support for omitting a title from MultistepDialog. At this point, we still have a chance to go back and change that (we can make `title` required and always render the header); that change has not been released yet in `@blueprintjs/core`. However I feel like it's probably ok despite the small information hierarchy concern you mentioned. Lmk if you feel strongly against it, otherwise I'll go ahead and review this PR soon...

aycai:
I think we can keep our path forward here as is; the information hierarchy point is something I'm more concerned about enforcing separately.

packages/core/src/common/classes.ts
```diff
@@ -197,6 +197,8 @@ export const MULTISTEP_DIALOG_PANELS = `${MULTISTEP_DIALOG}-panels`;
 export const MULTISTEP_DIALOG_LEFT_PANEL = `${MULTISTEP_DIALOG}-left-panel`;
 export const MULTISTEP_DIALOG_RIGHT_PANEL = `${MULTISTEP_DIALOG}-right-panel`;
 export const MULTISTEP_DIALOG_FOOTER = `${MULTISTEP_DIALOG}-footer`;
+export const MULTISTEP_DIALOG_POSITION_TOP = `${MULTISTEP_DIALOG}-position-top`;
+export const MULTISTEP_DIALOG_POSITION_RIGHT = `${MULTISTEP_DIALOG}-position-right`;
```
adidahiya:
```suggestion
export const MULTISTEP_DIALOG_NAV_TOP = `${MULTISTEP_DIALOG}-nav-top`;
export const MULTISTEP_DIALOG_NAV_RIGHT = `${MULTISTEP_DIALOG}-nav-right`;
```

packages/core/src/components/dialog/_multistep-dialog.scss
```diff
@@ -23,6 +23,79 @@ $step-radius: $pt-border-radius * 2 !default;
       border-top-right-radius: $dialog-border-radius;
     }
   }
+
+  .#{$ns}-multistep-dialog-position-top & {
```
adidahiya:
```suggestion
  .#{$ns}-multistep-dialog-nav-top & {
```

packages/core/src/components/dialog/_multistep-dialog.scss
```diff
@@ -23,6 +23,79 @@ $step-radius: $pt-border-radius * 2 !default;
...
+      border-left: none;
+    }
+
+    .#{$ns}-multistep-dialog-right-panel,
+    .#{$ns}-multistep-dialog-footer {
+      border-radius: 0 0 $dialog-border-radius $dialog-border-radius;
+    }
+  }
+
+  .#{$ns}-multistep-dialog-position-right & {
```
adidahiya:
```suggestion
  .#{$ns}-multistep-dialog-nav-right & {
```

packages/core/src/components/dialog/multistepDialog.tsx
```diff
@@ -17,14 +17,16 @@
...
-import { AbstractPureComponent2, Classes, Utils } from "../../common";
+import { AbstractPureComponent2, Classes, Position, Utils } from "../../common";
 import { DISPLAYNAME_PREFIX } from "../../common/props";
 import { Button, ButtonProps } from "../button/buttons";
 import { Dialog, DialogProps } from "./dialog";
 import { DialogStep, DialogStepId, DialogStepProps, DialogStepButtonProps } from "./dialogStep";
 
 type DialogStepElement = React.ReactElement<DialogStepProps & { children: React.ReactNode }>;
 
+export type MultistepDialogNavigationPosition = typeof Position.TOP | typeof Position.LEFT | typeof Position.RIGHT;
```
adidahiya:
```suggestion
export type MultistepDialogNavPosition = typeof Position.TOP | typeof Position.LEFT | typeof Position.RIGHT;
```

packages/core/src/components/dialog/multistepDialog.tsx
```diff
@@ -45,6 +47,13 @@ export interface IMultistepDialogProps extends DialogProps {
      */
     finalButtonProps?: Partial<ButtonProps>;
 
+    /**
+     * Position of the step navigation.
+     *
+     * @default Position.LEFT
+     */
+    navigationPosition?: MultistepDialogNavigationPosition;
```
adidahiya:
```suggestion
    navPosition?: MultistepDialogNavPosition;
```

packages/core/src/components/dialog/multistepDialog.tsx
```diff
@@ -45,6 +47,13 @@ export interface IMultistepDialogProps extends DialogProps {
      */
     finalButtonProps?: Partial<ButtonProps>;
 
+    /**
+     * Position of the step navigation.
```
adidahiya:
```suggestion
     * Position of the step navigation within the dialog.
```

adidahiya (approved):

PR #4978, opened 2021-10-21
[core] docs(Button): buttons are disabled when loading={true}

danny-does-stuff:
#### Checklist

- [ ] Includes tests
- [x] Update documentation

Clarify that the `loading` prop on `IButtonProps` will also disable the button

The docs

### DIFF
--- a/packages/core/src/components/button/abstractButton.tsx
+++ b/packages/core/src/components/button/abstractButton.tsx
@@ -62,7 +62,7 @@ export interface IButtonProps<E extends HTMLButtonElement | HTMLAnchorElement = 
     large?: boolean;
 
     /**
-     * If set to `true`, the button will display a centered loading spinner instead of its contents.
+     * If set to `true`, the button will display a centered loading spinner instead of its contents, and the button will be disabled.
      * The width of the button is not affected by the value of this prop.
      *
      * @default false
--- a/packages/core/src/components/button/abstractButton.tsx
+++ b/packages/core/src/components/button/abstractButton.tsx
@@ -62,7 +62,7 @@ export interface IButtonProps<E extends HTMLButtonElement | HTMLAnchorElement = 
     large?: boolean;
 
     /**
-     * If set to `true`, the button will display a centered loading spinner instead of its contents, and the button will be disabled.
+     * If set to `true`, the button will display a centered loading spinner instead of its contents, and the button will appear disabled.
      * The width of the button is not affected by the value of this prop.
      *
      * @default false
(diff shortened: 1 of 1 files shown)


palantirtech:
Thanks for your interest in [palantir/blueprint](https://github.com/palantir/blueprint), @dannyharding10! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

packages/core/src/components/button/abstractButton.tsx
```diff
@@ -62,7 +62,7 @@ export interface IButtonProps<E extends HTMLButtonElement | HTMLAnchorElement =
     large?: boolean;
 
     /**
-     * If set to `true`, the button will display a centered loading spinner instead of its contents.
+     * If set to `true`, the button will display a centered loading spinner instead of its contents, and the button will be disabled.
```
adidahiya, line 65: we don't actually set the `disabled` attribute, so I think this change is more appropriate:

```suggestion
     * If set to `true`, the button will display a centered loading spinner instead of its contents, and the button will appear disabled.
```

adidahiya (changes requested):

packages/core/src/components/button/abstractButton.tsx
danny-does-stuff: @adidahiya According to [this line](https://github.com/palantir/blueprint/blob/758233c54de4bb707a76131507938a11d7d72e1b/packages/core/src/components/button/abstractButton.tsx#L121), the button does actually get disabled when `loading` is set. Am I reading it wrong?

adidahiya (approved):

packages/core/src/components/button/abstractButton.tsx
adidahiya: hmm wait, you are right. sorry about that, will revert my edit!

packages/core/src/components/button/abstractButton.tsx
```diff
@@ -62,7 +62,7 @@ export interface IButtonProps<E extends HTMLButtonElement | HTMLAnchorElement =
     large?: boolean;
 
     /**
-     * If set to `true`, the button will display a centered loading spinner instead of its contents.
+     * If set to `true`, the button will display a centered loading spinner instead of its contents, and the button will appear disabled.
```
```suggestion
     * If set to `true`, the button will display a centered loading spinner instead of its contents, and the button will be disabled.
```

packages/core/src/components/button/abstractButton.tsx
danny-does-stuff: Awesome thanks!

PR #4983, opened 2021-10-22
[v4] [core] fix(EditableText): do not use disabled text colors

### LINKED ISSUE #4974: [v4] EditableText uses wrong text color when disabled and alwaysRenderInput={false}
#### Environment

- __Package version(s)__: core 4.0.0-beta.4
- __Operating System__: macOS

#### Code Sandbox

#### Steps to reproduce

1. Render `<EditableText value="title" disabled="true" />`

#### Actual behavior

Text is styled as disabled text, with insufficient contrast:

image

#### Expected behavior

Text is styled the same as regular text

#### Possible solution

- Fix BP4 CSS
- Add option to EditableTextExample to render with `disabled={true}` to allow testing in the future

adidahiya:
#### Checklist

- [ ] Includes tests
- [x] Update documentation

I believe these color overrides were added in error; we don't want to render text with "disabled" appearance when `disabled={true}`; rather, that prop controls whether the interactive functionality is enabled on this component.

Was there some other intention behind this code?

#### Screenshot

image

### DIFF
--- a/packages/core/src/blueprint.scss
+++ b/packages/core/src/blueprint.scss
@@ -276,18 +276,6 @@ $tree-intent-icon-colors-modern: (
   }
 }
 
-
-// Contrast for disabled editable text
-.#{$ns}-editable-text.#{$ns}-disabled {
-  > .#{$ns}-editable-text-content {
-    color: $pt-text-color-disabled;
-
-    .#{$ns}-dark & {
-      color: $pt-dark-text-color-disabled;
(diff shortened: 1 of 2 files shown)


aycai:
The original change and intent was the placeholder text using the disabled text color and failing contrast. The placeholder color should be the only state that changes (which looks correct on the docs site), so if this is touching the disabled state then removing this is correct.

PR #4997, opened 2021-10-29
[v4] [colors] fix: color palette revision (part 2)

adidahiya:
Adds some changes I missed committing in #4996 🤦🏽‍♂️


adidahiya:
before:

image

after:

image

PR #5031, opened 2021-11-22
[core] feat(FormGroup): add support for "sub label" text

### LINKED ISSUE #5019: Helper Text for forms should have 'above input' option
I would like to know if we can support the option to make the helper text in BP forms appear above the inputs instead of below them. I find that the helper text is often used as a direction or a label for the input, in which case, it makes more sense hierarchically to appear above the input. Attached is the desired visual:
image

mud:
#### Checklist

- [ ] Includes tests
- [x] Update documentation

To resolve #5019 this PR adds a new prop `labelHelperText` to `FormGroup` which adds helperText between `label` and `children`.

We could have ignored the `labelHelperText` if `label` is empty, but decided against it as it seemed to add unnecessary complexity. Similarly, there's no restriction on specifying only one type of helperText. It looks slightly odd when both helper texts are used, but we may end up using both with one of our use case where we have a list of fields specified as `children`.

#### Screenshot
Screen Shot 2021-11-22 at 10 36 24 AM

### DIFF
--- a/packages/core/src/components/forms/formGroup.tsx
+++ b/packages/core/src/components/forms/formGroup.tsx
@@ -51,6 +51,13 @@ export interface IFormGroupProps extends IntentProps, Props {
     label?: React.ReactNode;
 
     /**
+     * Optional helper text for `label`. The given content will be wrapped in
+     * `Classes.FORM_HELPER_TEXT` and displayed beneath `label`. Helper text
+     * color is determined by the `intent`.
+     */
+    labelHelperText?: React.ReactNode;
+
+    /**
      * `id` attribute of the labelable form element that this `FormGroup` controls,
      * used as `<label for>` attribute.
      */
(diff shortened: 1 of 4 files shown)


packages/core/src/components/forms/formGroup.tsx
```diff
@@ -50,6 +50,13 @@ export interface IFormGroupProps extends IntentProps, Props {
     /** Label of this form group. */
     label?: React.ReactNode;
 
+    /**
+     * Optional helper text for `label`. The given content will be wrapped in
+     * `Classes.FORM_HELPER_TEXT` and displayed beneath `label`. Helper text
+     * color is determined by the `intent`.
+     */
+    labelHelperText?: React.ReactNode;
```
adidahiya: can we call this `sublabel` instead? and make the class name `.bp4-form-group-sublabel` accordingly

adidahiya: edit: I would make it two words actually, so the prop is `subLabel` and class `.bp4-form-group-sub-label`

mud: That makes sense. Just pushed an update.

adidahiya (approved):

PR #5050, opened 2021-12-01
fix: renamed Placement export in popover2 to PopperPlacements

jackshen:
#### Checklist

- [ ] Includes tests
- [ ] Update documentation

According to current [documentation](https://blueprintjs.com/docs/#popover2-package/popover2.placement), `PopperPlacements` "defines the full set of supported values" for the `Popover` prop `placement`. [Currently](https://github.com/palantir/blueprint/blob/23dea38f732ccc5d81a1e0c82d8e6b79e0c9f9a7/packages/popover2/src/index.ts#L34), this enum is being exported as `Placement`. This PR assumes the docs are correct and seeks to correct the export name.

N/A

#### Screenshot

N/A

### DIFF
--- a/packages/docs-app/src/examples/popover2-examples/popover2Example.tsx
+++ b/packages/docs-app/src/examples/popover2-examples/popover2Example.tsx
@@ -43,7 +43,7 @@ import {
 import {
     Classes,
     Popover2SharedProps,
-    Placement,
+    PopperPlacements,
     PlacementOptions,
     Popover2,
     Popover2InteractionKind,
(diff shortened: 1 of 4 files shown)


palantirtech:
Thanks for your interest in [palantir/blueprint](https://github.com/palantir/blueprint), @jackshen! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

adidahiya (changes requested):
we shouldn't change the API; this is an unnecessary breaking change.

instead we should fix the docs. that line in the docs should instead read:

    `import { PlacementOptions } from "@blueprintjs/popover2"` defines the full set of supported values.

adidahiya:
closing as a stale PR. I'd be happy to re-open if you implement the suggested changes.

PR #5058, opened 2021-12-03
[v4] [core] fix(Switch): fix text color and indicator border

johnly13:
#### Checklist

- [ ] Includes tests
- [ ] Update documentation

* Uses normal button `box-shadow` instead of `inset` `box-shadow` for dark switch indicators
* Fixes text colors for switches
* Removes borders for disabled switch indicator
* Fixes dark switch indicator background colors

#### Screenshot

| Before | After |
| --- | --- |
| Screen Shot 2021-12-03 at 15 32 47 | Screen Shot 2021-12-06 at 14 32 02|
| Screen Shot 2021-12-03 at 15 32 52 | Screen Shot 2021-12-06 at 14 32 08 |
| Screen Shot 2021-12-03 at 15 32 57 | Screen Shot 2021-12-06 at 17 17 44|
| Screen Shot 2021-12-03 at 15 33 03 | Screen Shot 2021-12-06 at 17 17 49 |

(note: I think the indicator looks off center vertically in some of these screenshots. I think this is an artifact of zooming in + screenshotting, rather than how they actually are styled)


packages/core/src/components/button/_common.scss
```diff
@@ -48,13 +48,13 @@ $button-box-shadow-overlay-active:
   inset 0 1px 1px rgba($black, 0.1) !default;
 
 $dark-button-box-shadow:
-  inset 0 0 0 1px rgba(17, 20, 24, 0.8) !default;
+  inset 0 0 0 $button-border-width rgba($black, 0.8) !default;
```
johnly13, line 51: using variables instead of hardcoded values

packages/core/src/components/forms/_controls.scss
```diff
@@ -300,6 +300,10 @@ $control-indicator-spacing: $pt-grid-size !default;
 
   $switch-indicator-text-font-size: 0.7em;
 
+  $switch-text-color: $pt-text-color !default;
```
johnly13, line 303: Explicitly setting text colors for switch's inner text (for light/dark and enabled/disabled)

packages/core/src/components/forms/_controls.scss
```diff
@@ -321,19 +329,23 @@ $control-indicator-spacing: $pt-grid-size !default;
   $switch-indicator-background-color: $white !default;
   $switch-indicator-background-color-disabled: rgba($white, 0.8) !default;
   $dark-switch-indicator-background-color: $dark-gray5 !default;
+  $dark-switch-indicator-box-shadow: 0 0 0 $button-border-width rgba($black, 0.8) !default;
```
johnly13: Explicitly set dark switch indicator's box shadow to not be inset, to be consistent with other indicator box shadows.

packages/core/src/components/forms/_controls.scss
```diff
@@ -407,29 +424,28 @@ $control-indicator-spacing: $pt-grid-size !default;
...
         $dark-switch-indicator-background-color-disabled
       );
 
       .#{$ns}-control-indicator::before {
         background: $dark-switch-indicator-background-color;
-        box-shadow: $dark-button-box-shadow;
-      }
-
-      input:checked ~ .#{$ns}-control-indicator::before {
-        // inset shadow so it forms a dark line instead of blurring into the blue
```
johnly13, line 431: since the box shadow is different now, we don't get this blurring effect anymore without the inset

packages/core/src/blueprint.scss
```diff
@@ -275,104 +275,7 @@ table.#{$ns}-html-table {
   }
 }
 
-// Contrast for switches
```
johnly13, line 278: Removed these overrides, now built in the backgrounds into switch variables

packages/core/src/components/forms/_controls.scss
```diff
@@ -345,29 +360,35 @@ $control-indicator-spacing: $pt-grid-size !default;
       }
 
       input#{$selector}:disabled ~ .#{$ns}-control-indicator {
-        background: $disabled-color;
+        background: $disabled-background-color;
+        color: $disabled-text-color;
 
         &::before {
           background: $disabled-indicator-color;
+          box-shadow: none;
```
johnly13, line 368: no borders for disabled switch indicator

johnly13:
@adidahiya this PR should be ready for review now!

adidahiya (approved):
this is kind of an an explosion of Sass variables... probably unnecessary complexity which makes the code harder to work with... but since that's kind of the existing code pattern for these styles, we can leave it as-is and try to refactor in a separate PR 🙃 

code changes look fine, and visuals lgtm... I'll wait for @aycai for final design approval

johnly13:
yeah, I was pretty sad adding all these new variables, but given that the switch styles have diverged slightly from the button styles, it didn't make sense to use the existing button variables. I'll think of ways to improve this in a follow up - I think low hanging fruit is to remove variables that are only used once (which I'd imagine there are a few of)

> this is kind of an an explosion of Sass variables... probably unnecessary complexity which makes the code harder to work with... but since that's kind of the existing code pattern for these styles, we can leave it as-is and try to refactor in a separate PR 🙃
> 
> code changes look fine, and visuals lgtm... I'll wait for @aycai for final design approval

aycai:
@johnly13 Shoot, didn't show switch with text in the mocks - light mode + enabled + disabled state text should be 60% opacity to match other states, but rest looks good!

johnly13:
> @johnly13 Shoot, didn't show switch with text in the mocks - light mode + enabled + disabled state text should be 60% opacity to match other states, but rest looks good!

Done - updated screenshots above as well!

PR #5059, opened 2021-12-06
[core] fix: InputGroups should inherit border radius

### LINKED ISSUE #1750: [Core/Control Group] Control group doesn't set rounded corners properly if child wrapped with <Popover>
### Bug report

- __Package version(s)__: v1.32.0
- __Browser and OS versions__: Mac Sierra, Chrome 62

#### Steps to reproduce

1. Create a `pt-control-group` with two arbitrary children, except _wrap those children with `<Popover>s`. For example, consider this date/timezone picker (which is admittedly too narrow, but ignore that):

    control-group

#### Actual behavior

Corners are not rounded because of these rules, which round corners on the `pt-popover-target` parents instead of their `pt-input`/`pt-button` children:

```scss
.pt-control-group > :first-child {
  border-radius: 3px 0 0 3px;
}

.pt-control-group > :last-child {
  border-radius: 0 3px 3px 0;
}
```

#### Expected behavior

Corners are rounded on the `pt-input`/`pt-button`/`pt-input-group` children. Maybe we can change the rule to:

```scss
.pt-control-group > :first-child {
  &.pt-button, 
  &.pt-input,
  &.pt-input-group,
  > .pt-button,
  > .pt-input,
  > .pt-input-group {
     border-radius: 3px 0 0 3px;
  }
}

// And same for :last-child
```

kmark:

Adds `border-radius: inherit;` to `input-group` so that input group corners are properly rounded when used inside of a Tooltip/Popover.

Repro:

```tsx
<ControlGroup>
    <InputGroup placeholder="Username" />
    <Tooltip>
        <InputGroup placeholder="Password" />
    </Tooltip>
</ControlGroup>
```

#### Screenshot

Before:
Screen Shot 2021-12-06 at 3 15 41 PM

After:
Screen Shot 2021-12-06 at 3 39 32 PM

### DIFF
--- a/packages/core/src/components/forms/_input-group.scss
+++ b/packages/core/src/components/forms/_input-group.scss
@@ -40,6 +40,7 @@ $input-button-height-large: $pt-button-height !default;
 $input-button-height-small: $pt-button-height-smaller !default;
 
 .#{$ns}-input-group {
+  border-radius: inherit;
   display: block;
   position: relative;
 


adidahiya:
I'm worried about the scope of this style change and its potential for unintended breaks... I'd rather make a more limited fix, I've proposed a change here: https://github.com/palantir/blueprint/pull/5067

adidahiya (changes requested):
see my comment above

kmark:
Sounds good, thanks for taking a look

PR #5067, opened 2021-12-09
[core] fix(ControlGroup): correct border radius when Tooltip wraps InputGroup

### LINKED ISSUE #1750: [Core/Control Group] Control group doesn't set rounded corners properly if child wrapped with <Popover>

adidahiya:

Limited change to add border-radius style inheritance to input groups inside control groups to fix the linked issue

#### Screenshot

See example of fix in https://codesandbox.io/s/blueprint-sandbox-forked-zslf0?file=/src/app.css

image

### DIFF
--- a/packages/core/src/components/forms/_control-group.scss
+++ b/packages/core/src/components/forms/_control-group.scss
@@ -172,6 +172,11 @@ Styleguide control-group
     }
   }
 
+  .#{$ns}-input-group {
+    // also need inheritance here, in case an input is wrapped by a popover or tooltip
+    border-radius: inherit;
+  }
+
   // input group contents appear above input always
   .#{$ns}-input-group > .#{$ns}-icon,
   .#{$ns}-input-group > .#{$ns}-button,


adidahiya:
This has some issues; it doesn't work in this case... will need more investigation:

image

PR #5080, opened 2021-12-21
[popover2] fix(Popover2): close properly when target focus lost

### LINKED ISSUE #4503: Tooltip does not close if focus is triggered by keyboard but removed mouse
#### Environment

- __Package version(s)__:  @blueprintjs/core@3.38.2, @blueprintjs/popover2@0.2.0
- __Operating System__: Ubuntu 20.04
- __Browser name and version__: Chromium Version 88.0.4324.96 (Official Build) snap (64-bit), Firefox 85.0

#### Code Sandbox

Link to a minimal repro (fork [this code sandbox](https://codesandbox.io/s/blueprint-sandbox-et9xy)): https://codesandbox.io/s/blueprint-sandbox-forked-146cf?file=/src/CoreExample.tsx

The example has been modified to add `Tooltip2` around the *Core Sandbox* buttons.

FWIW, the same happens with the deprecated `Tooltip`.

#### Steps to reproduce

1. click on a blank part of the browser
1. press tab on the keyboard to focus the buttons
1. wait for one second for the tooltip to appear
2. click on a blank area of the browser to remove the focus from the button

#### Actual behavior

The tooltip stays open

#### Expected behavior

The tooltip should close since the focus is lost

#### Possible solution

?

dlech:
The assumption that "a click handler somewhere else will close the popover if necessary" does not appear to be valid. If the user clicks on a non-focusable element. There are no other events that cause the popover to close other than mouseleave, but the mouse is not over the target when focus is lost.

Fixes #4503.

#### Checklist

- [ ] Includes tests
- [ ] Update documentation

### DIFF
--- a/packages/popover2/src/popover2.tsx
+++ b/packages/popover2/src/popover2.tsx
@@ -536,10 +536,6 @@ export class Popover2<T> extends AbstractPureComponent2<Popover2Props<T>, IPopov
 
     private handleTargetBlur = (e: React.FocusEvent<HTMLElement>) => {
         if (this.props.openOnTargetFocus && this.isHoverInteractionKind()) {
-            // e.relatedTarget ought to tell us the next element to receive focus, but if the user just
-            // clicked on an element which is not focusable (either by default or with a tabIndex attribute),
-            // it won't be set. So, we filter those out here and assume that a click handler somewhere else will
-            // close the popover if necessary.
             if (e.relatedTarget != null) {
                 // if the next element to receive focus is within the popover, we'll want to leave the
                 // popover open.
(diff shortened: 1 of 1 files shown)


palantirtech:
Thanks for your interest in [palantir/blueprint](https://github.com/palantir/blueprint), @dlech! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

adidahiya:
Thanks for the PR, this does fix the bug. I can verify this in the [docs preview](https://50367-71939872-gh.circle-artifacts.com/0/packages/docs-app/dist/index.html#popover2-package/tooltip2):



(press tab to focus & open, then click on document... the same interaction repros the bug on the [current docs page](https://blueprintjs.com/docs/#popover2-package/tooltip2))

It looks like this assumption about `e.relatedTarget` was first added to the Popover component in https://github.com/palantir/blueprint/pull/3716, and then ported to Popover2. Your change in this PR doesn't seem to cause a regression, and it passes the test suite... should be good to go 👍🏽

adidahiya (approved):

PR #5084, opened 2022-01-03
Add prop to prevent event propagation on EditableText escape

synapse:
#### Checklist

- [ ] Includes tests
- [ ] Update documentation

Added a new prop on the `EditableText` component to prevent event propagation when pressing the Esc key. This is useful when the `EditableText` component is used inside a Dialog/Modal which has `canEscapeKeyClose` as `true`

Nothing really, this is a simple change

### DIFF
--- a/packages/core/src/components/editable-text/editableText.tsx
+++ b/packages/core/src/components/editable-text/editableText.tsx
@@ -354,7 +354,7 @@ export class EditableText extends AbstractPureComponent2<EditableTextProps, IEdi
         /* eslint-disable-next-line deprecation/deprecation */
         const { altKey, ctrlKey, metaKey, shiftKey, which } = event;
         if (which === Keys.ESCAPE) {
-            if (this.props.stopPropagation) {
+            if (this.props.stopEscapeKeyPropagation) {
                 event.stopPropagation();
             }
             this.cancelEditing();
--- a/packages/core/src/components/editable-text/editableText.tsx
+++ b/packages/core/src/components/editable-text/editableText.tsx
@@ -110,7 +110,7 @@ export interface IEditableTextProps extends IntentProps, Props {
      *
      * @default false
      */
-    stopPropagation?: boolean;
+    stopEscapeKeyPropagation?: boolean;
 
     /**
      * The type of input that should be shown, when not `multiline`.
--- a/packages/core/src/components/editable-text/editableText.tsx
+++ b/packages/core/src/components/editable-text/editableText.tsx
@@ -354,7 +354,7 @@ export class EditableText extends AbstractPureComponent2<EditableTextProps, IEdi
         /* eslint-disable-next-line deprecation/deprecation */
         const { altKey, ctrlKey, metaKey, shiftKey, which } = event;
         if (which === Keys.ESCAPE) {
-            if (this.props.stopPropagation === true) {
+            if (this.props.stopPropagation) {
                 event.stopPropagation();
             }
             this.cancelEditing();
--- a/packages/core/src/components/editable-text/editableText.tsx
+++ b/packages/core/src/components/editable-text/editableText.tsx
@@ -106,7 +106,7 @@ export interface IEditableTextProps extends IntentProps, Props {
     selectAllOnFocus?: boolean;
 
     /**
-     * Whether the input should stop the evnet propagation when escape is pressed
+     * Whether the input should stop the event propagation when escape is pressed
      *
      * @default false
      */
--- a/packages/core/src/components/editable-text/editableText.tsx
+++ b/packages/core/src/components/editable-text/editableText.tsx
@@ -106,6 +106,13 @@ export interface IEditableTextProps extends IntentProps, Props {
     selectAllOnFocus?: boolean;
 
     /**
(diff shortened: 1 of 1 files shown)


synapse:
Hey @adidahiya @michael-yx-wu  is there any way you can look at this PR?

michael-yx-wu:
@adidahiya I feel like this is pretty sane default behavior. Perhaps a better alternative would be to make overlays ignore esc key events that originate from EditableText elements?

synapse:
Yes @michael-yx-wu , it could be, but the parent of the `EditableText` might not be the `Dialog`, in that case, we would need an event passed upwards through several components to change the dialogs behavior `props` (or via a state manager) but both solutions would be ugly.

Edit: I just looked a the `Overlay` and `Dialog` docs and there's not way to intercept the `Esc` key event to prevent closing based on the origin.

synapse:
Hey @michael-yx-wu, @adidahiya if there's not issues would it ok to merge this PR?

mevtho:
Quick addition to it as I think it is related and would support the case of adding that prop.

Happening on Mac :
When the browser is in Full Screen mode, pressing the Esc key when editing makes the browser exit full screen mode.

For now I have found a solution which is to wrap the EditableText in a div that captures and prevent the propagation, but that doesn't seem to be the right way to go about it.

```html
<div 
    onKeyDown={(e) => {
        if (e.key === 'Escape') {
            e.preventDefault();
       }
    }}
>
    <EditableText  ... />
</div>
```

synapse:
Hey @michael-yx-wu @adidahiya just a friendly nudge here

packages/core/src/components/editable-text/editableText.tsx
```diff
@@ -105,6 +105,13 @@ export interface IEditableTextProps extends IntentProps, Props {
      */
     selectAllOnFocus?: boolean;
 
+    /**
+     * Whether the input should stop the event propagation when escape is pressed
+     *
+     * @default false
+     */
+    stopPropagation?: boolean;
```
adidahiya: I think this prop name should be more explicit:

```suggestion
    stopEscapeKeyPropagation?: boolean;
```

packages/core/src/components/editable-text/editableText.tsx
```diff
@@ -105,6 +105,13 @@ export interface IEditableTextProps extends IntentProps, Props {
      */
     selectAllOnFocus?: boolean;
 
+    /**
+     * Whether the input should stop the event propagation when escape is pressed
```
adidahiya, line 109: I think it would be helpful to add documentation about the motivation for this feature here, like you did in the PR description.

```suggestion
     * Whether the input should stop the event propagation when escape is pressed.
     * This is useful, for example, when this component is used inside a `<Dialog>` with
     * `canEscapeKeyClose` enabled.
```

packages/core/src/components/editable-text/editableText.tsx
```diff
@@ -347,6 +354,9 @@ export class EditableText extends AbstractPureComponent2<EditableTextProps, IEdi
         /* eslint-disable-next-line deprecation/deprecation */
         const { altKey, ctrlKey, metaKey, shiftKey, which } = event;
         if (which === Keys.ESCAPE) {
+            if (this.props.stopPropagation === true) {
```
adidahiya: please add a default value of false for the new prop in `public static defaultProps`. then you can simplify this boolean check:

```suggestion
            if (this.props.stopPropagation) {
```

adidahiya (changes requested):
how does this interaction work after you press ESC once (to cancel editing) and then press ESC again? I think a user might expect the event to close a dialog in that case, since they've finished the interaction with the EditableText. have you been able to test this out? I was having trouble even hitting the `stopPropagation()` line you added in this PR when testing this locally with the docs. I think it would help to update the documentation to include an example with this specific use case, so we can test it out in the docs preview

adidahiya:
@synapse can you answer some of my questions from my last review? also can you please merge `develop` into your branch? that branch has changed quite a bit since you opened this PR; I recently merged all the v4 changes as v4.0.0 stable was released.

synapse:
Hey @adidahiya the branch should be aligned with `develop`. I would like to also add this prop to the docs but I'm having some issues running the latest version locally (do I need to run/test it against some other branch?).
To answer your question, your functionality example is correct but I would leave it up to the developer how to handle edge cases.
Otherwise, we would need to move this logic into the Dialog and not close it when Esc is hit inside an input/textarea etc.

adidahiya:
well, this PR is failing a lot of CI checks... it may be because of the CI cache... I think running `git merge origin/develop` and then a full local build (`yarn && yarn clean && yarn compile`) should get you into a good state to be able to make further changes to the docs

synapse:
Not sure why but locally all seam ok, @adidahiya check these screenshots:

synapse:
Hey @adidahiya, ok, managed to make it work locally and you're totaly right, when not using the `alwaysRenderInput` prop the `input`/`textarea` is gone but I'm guessing that the inputs event `stopPropagation` is still active. I'll search for a better solution

PR #5085, opened 2022-01-04
Fixed types of Tree Node IDs

Mike-Dax:
I noticed when trying to build a Tree with strings as the Node IDs that the static method `Tree.nodeFromPath` had an incorrect type for the path argument. 

https://github.com/Mike-Dax/blueprint/blob/0be8e951f857134afffcc57e74bff143a542d2e2/packages/core/src/components/tree/treeNode.tsx#L55

The treeNode ID is `string | number`, and it's stored in an object, so it's always cast to a string internally.

#### Checklist

- [ ] Includes tests
- [ ] Update documentation

I presume the relevant types in the documentation will be automatically updated, namely the `path` prop of `ITreeNodeProps` is incorrect, it should be `(string | number)[]`.

Changes the exposed methods to use `string | number` instead of just `number` for the IDs. Similarly paths use the array of the union instead of just `number[]` 

I'm not sure if this is all the instances of this type, there may be more.

#### Screenshot

image
to
image

### DIFF
--- a/packages/core/src/components/tree/tree.tsx
+++ b/packages/core/src/components/tree/tree.tsx
@@ -22,10 +22,12 @@ import { DISPLAYNAME_PREFIX, Props } from "../../common/props";
 import { isFunction } from "../../common/utils";
 import { TreeNodeInfo, TreeNode } from "./treeNode";
 
+type NodeID = string | number
+
 // eslint-disable-next-line @typescript-eslint/ban-types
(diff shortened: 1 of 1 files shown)


adidahiya:
Looks like this isn't passing compile or formatting checks yet. I'd be happy to take a look after it does. You may find it helpful to run the `yarn compile` script in the packages/core/ folder only, rather than running it for the entire repo.

PR #5086, opened 2022-01-04
[core] feat: allow dialog bg color to be overridden as Sass variable

dylangrandmont:
#### Checklist

- [x] Includes tests
- [x] Update documentation

Allow override of dialog `background-color` using a variable `$pt-dialog-background-color`

* Naming convention of variable
* Appropriateness of using a variable here
* Whether this needs to be documented (it appears that component-level variables like this are not documented elsewhere)


palantirtech:
Thanks for your interest in [palantir/blueprint](https://github.com/palantir/blueprint), @dylangrandmont! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

packages/core/src/components/dialog/_dialog.scss
```diff
@@ -34,6 +34,7 @@ Markup:
 Styleguide dialog
 */
 
+$pt-dialog-background-color: $light-gray4 !default;
```
adidahiya: this should match the naming convention of the other dialog variables:

```suggestion
$dialog-background-color: $light-gray4 !default;
```

adidahiya (changes requested):
sorry for the very delayed review here! please let me know if you'd like to continue working on this PR.

I'm open to this code change but have some comments on the implementation.

Note that we'll also need a corresponding dark theme color variable called `$dark-dialog-background-color`.

adidahiya (approved):

dylangrandmont:
Thanks for taking this over the line @adidahiya

PR #5098, opened 2022-01-16
[v4] [icons] fix: pre-es2017 compatibility

### LINKED ISSUE #5097: @blueprintjs/icons 4.0.0-beta does not support IE11
#### Environment

- __Package version(s)__: @blueprintjs/icons@4.0.0-beta.6
- __Operating System__: macOS
- __Browser name and version__: IE11

---

It looks like the `Object.values(...)` call in `@blueprintjs/icons` isn't getting transpiled before publishing.

https://github.com/palantir/blueprint/blob/4dac1637138f850d51e08a1f7e46a5897abcbacc/packages/icons/src/iconNames.ts#L31

Here's the contents of `node_modules/.pnpm/@blueprintjs+icons@4.0.0-beta.6/node_modules/@blueprintjs/icons/lib/esm/iconNames.js` without the copyright header.

```js
import { __assign } from "tslib";
/* eslint-disable camelcase */
import { pascalCase, snakeCase } from "change-case";
import { BlueprintIcons_16 } from "./generated/16px/blueprint-icons-16";
var IconNamesNew = {};
var IconNamesLegacy = {};
for (var _i = 0, _a = Object.values(BlueprintIcons_16); _i < _a.length; _i++) {
    var name_1 = _a[_i];
    IconNamesNew[pascalCase(name_1)] = name_1;
    IconNamesLegacy[snakeCase(name_1).toUpperCase()] = name_1;
}
export var IconNames = __assign(__assign({}, IconNamesNew), IconNamesLegacy);
//# sourceMappingURL=iconNames.js.map
```

Is this expected?

- Does Blueprint 4.x drop support for IE11? Feel free to close this issue if so, but I wasn't able to find any docs stating so.
- I'm a bit surprised myself since the `"target": "es5"` is set here on the `next` branch: https://github.com/palantir/blueprint/blob/87dd5091551625d4e30365d59b0cd3668ed9df00/config/tsconfig.base.json#L23
  The `__assign` helper is getting applied. Would have expected `__values` to be imported as well.

gluxon:

Switches an `Object.values` usage to a `for ... in ...` loop for pre-es2017 browser compatibility.

**Edit**: The `guard-for-in` ESLint rule triggered on the below. Updated the commit for `propertyIsEnumerable`.

According to MDN, [`Object.values` also filters on own enumerable properties](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_objects/Object/values), so technically we should also check for  [`Object.prototype.propertyIsEnumerable()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/propertyIsEnumerable).

TypeScript doesn't seem to generate non-own enumerable properties for enum objects though, so in practice the extra check isn't needed. https://www.typescriptlang.org/play?#code/KYOwrgtgBAYg9nKBvAsAKCpqAhAhgJygF4oAiAIwNPQF910AzOQgCgGM4QBnAFygGtgATygBLELAQBKZOixQO3OABtgAOmVwA5i0FCptIA

Happy to add it if reviewers think the extra soundness is worthwhile.

### DIFF
--- a/packages/icons/src/iconNames.ts
+++ b/packages/icons/src/iconNames.ts
@@ -28,7 +28,14 @@ export type { IconName };
 const IconNamesNew = {} as Record<PascalCase<IconName>, IconName>;
 const IconNamesLegacy = {} as Record<ScreamingSnakeCase<IconName>, IconName>;
 
-for (const name of Object.values(BlueprintIcons_16) as IconName[]) {
+const BlueprintIconsMap: Record<string, IconName> = BlueprintIcons_16;
+
+for (const key in BlueprintIconsMap) {
+    if (!BlueprintIconsMap.propertyIsEnumerable(key)) {
+        continue;
+    }
+
+    const name = BlueprintIconsMap[key];
     IconNamesNew[pascalCase(name) as PascalCase<IconName>] = name;
     IconNamesLegacy[snakeCase(name).toUpperCase() as ScreamingSnakeCase<IconName>] = name;
 }


palantirtech:
Thanks for your interest in [palantir/blueprint](https://github.com/palantir/blueprint), @gluxon! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

gluxon:
Did a quick smoke test to make sure this didn't affect the contents of `IconNames`.

```sh
❯ cd packages/icons

# Before
❯ git reset --hard ORIG_HEAD
HEAD is now at 4dac1637 Publish
❯ node -e 'console.log(Object.keys(require("./lib/cjs/iconNames").IconNames).length)'
1078
❯ node -e 'console.log(require("./lib/cjs/iconNames"))' | md5sum
8cb0237c9ed0ff2db4679aed00f76a00  -

# After
❯ git reset --hard ORIG_HEAD
HEAD is now at 1bcd43e7 [v4] [icons] fix: pre-es2017 compatibility
❯ yarn compile  
❯ node -e 'console.log(Object.keys(require("./lib/cjs/iconNames").IconNames).length)'
1078
❯ node -e 'console.log(require("./lib/cjs/iconNames"))' | md5sum
8cb0237c9ed0ff2db4679aed00f76a00  -
```

adidahiya (changes requested):
Since we already require a few [modern language features](https://blueprintjs.com/docs/versions/4/#blueprint/getting-started.language-features) to use Blueprint (including a few which are not fully supported in IE11 and thus require polyfills), I'd rather take a different approach.

Rather than add backwards-compatible code for IE11, let's instead update the docs to indicate that `Object.values()` support is now required at runtime with Blueprint v4.0. Users can [polyfill it with core-js](https://github.com/zloirock/core-js#objectvalues--objectentries).

We are planning to completely remove support for IE11 anyway soon (I'll add a deprecation notice to the docs before the v4.0 stable release).

gluxon:
Sounds good to me. Thanks for taking a look @adidahiya. 🙂

PR #5109, opened 2022-01-27
feat: Add popper modifier classnames

justinbhopper:

This adds new classes to the popover using some of the helpful modifiers from PopperJS's hide.  

These classnames can be helpful to allow consumers to hide the popper when the reference element is scrolled out of its scrolling container.  
See example: https://popper.js.org/docs/v2/modifiers/hide/

For Popover (@blueprintjs/core):
- Adds `.bp3-popover-out-of-boundaries` when PopperJS v1's `popperProps.outOfBoundaries` is true

For Popover2 (@blueprintjs/popover2):
- Adds `.bp3-popover-reference-hidden` when the reference element is not visible (scrolled off screen)
- Adds `.bp3-popper-escaped` when the popper escapes the reference element's boundaries

This PR does not change any behavior other than rendering the new classnames, which have no styles applied to them by default.


adidahiya (approved):
nice and simple, good PR description :+1:

PR #5115, opened 2022-02-03
[v4] [table] fix: prevent scrolling to ghost rows

### LINKED ISSUE #5027: Table with ghost cells continues to scroll when there are no more rows
#### Environment

- __Package version(s)__: blueprint 4 core and table

#### Code Sandbox

Link to a minimal repro:
https://codesandbox.io/s/cool-benz-es9oo?file=/src/App.js

#### Steps to reproduce

Create a scrollable Table with `enableGhostCells` set to true

#### Actual behavior

Scrollbar continues to scroll when reaching the bottom of the table. Also grabbing the scrollbar and dragging down creates more ghost rows.
scrolling
#### Expected behavior

The scrollbar stops scrolling when it reached the bottom of the table

Notes:
Scrolling over a frozen column to the bottom of the table behaves as normal

adidahiya:
#### Checklist

- [x] Includes tests
- [x] Update documentation

~Add Table2 vertical scroll logic to detect when scrolling has reached the bottom of the rendered cells, so it's not necessary to allow scrolling further down to the ghost cells.~

Edit: that approach was quite buggy, so I figured out a better one. We don't need to render ghost rows or columns at all, when there is vertical or horizontal overflow, respectively.

This should fix the linked bug.

#### Screenshot

with ghost cells enabled in the table dev app:


adidahiya:
if you scroll too fast, you get blank sections of the table on the bottom / right where the viewport does not update:

image

packages/table/src/table2.tsx
```diff
@@ -761,23 +775,29 @@ export class Table2 extends AbstractComponent2<TableProps, TableState, TableSnap
...
         });
 
-        const columnIndices = this.grid.getColumnIndicesInRect(viewportRect, enableGhostCells);
+        if (this.grid === null || this.locator === undefined || viewportRect === undefined) {
+            // if we haven't mounted yet (which we need in order for grid/viewport calculations),
+            // we still want to hand a DOM ref over to TableQuadrantStack for later
+            return <div className={classes} ref={refHandler} />;
+        }
+
+        // if we have horizontal overflow, no need to render ghost rows
```
adidahiya:
```suggestion
        // if we have horizontal overflow, no need to render ghost columns
```

packages/table/test/table2Tests.tsx
```diff
@@ -226,6 +247,65 @@ describe("<Table2>", function (this) {
...
+                `.${Classes.TABLE_CELL_GHOST}.${Classes.columnCellIndexClass(0)}`
+            ).length;
+            expect(numGhostCellsInFirstColumn).to.be.eq(0);
+
+            // cleanup
+            document.body.removeChild(containerElement);
+        });
+
+        function runTestToEnsureScrollingIsEnabled(enableGhostCells: boolean) {
+            it(`isn't disabled when there is half a row left to scroll to and enableGhostCells is set to ${enableGhostCells}`, () => {
```
adidahiya, line 278: note that these two tests were ported over from https://github.com/palantir/blueprint/commit/7460d1035d500d41c17b78b1f68eb4f5b0065567

packages/table/test/table2Tests.tsx
```diff
@@ -749,7 +829,7 @@ describe("<Table2>", function (this) {
         });
     });
 
-    describe("Resizing", () => {
+    describe.only("Resizing", () => {
```
adidahiya:
```suggestion
    describe("Resizing", () => {
```

PR #5158, opened 2022-03-04
[core] fix: Set aria-hidden for purely decorative icons

styu:
#### Checklist

- [ ] Includes tests
- [ ] Update documentation

For icons that are purely decorative because they have associated text, add `aria-hidden={true}` and `tabIndex={-1}` to the `Icon` props


adidahiya (approved):
nice 👍🏽

bvandrc:
I am pretty sure that Icons were not being tabbed to in the first place so I really don't think adding `tabIndex=-1` was necessary. Doesn't hurt though, but yeah, probably unnecessary.

PR #5162, opened 2022-03-09
[v4] [table] feat: experimental cellRendererDependencies prop

### LINKED ISSUE #4094: call cellRenderer sometimes like skipped
#### Environment

The version of @blueprintjs/table I'm using is 3.8.4.

#### Question

Render a state(number) in cells. 
When increasing the state. cells content isn't always updated.

```js
const [currentIndex, setCurrentIndex] = React.useState<number>(0);

  const increase = () => {
    setCurrentIndex(currentIndex + 1);
  };

  const getCellData = (
    rowIndex: number,
    columnIndex: number,
    key: string
  ): ICellLookup => {
    return `Current: ${currentIndex} value: ${Items[rowIndex]}` as any;
  };

  const table = (
    <Table
      ref={tableRef}
      numRows={Items.length}
      defaultRowHeight={35}
      defaultColumnWidth={400}
    >
      {keyColumn("Text", "text").getColumn(getCellData)}
    </Table>
  );
```

I use a [codesandbox](https://codesandbox.io/s/blueprint-sandbox-0twvp?file=/src/TableExample.tsx) to demo it.
Yes, I  just can't firgure out the problem in my code.

adidahiya:

New experimental feature for Table2: `cellRendererDependencies` prop

### DIFF
--- a/packages/table/src/common/errors.ts
+++ b/packages/table/src/common/errors.ts
@@ -39,3 +39,5 @@ export const TABLE_NUM_COLUMNS_COLUMN_WIDTHS_MISMATCH =
     ns + ` Table requires columnWidths.length to equal the number of <Column>s if provided.`;
 
 export const TABLE_UNMOUNTED_RESIZE_WARNING = `${ns} Table resize method called while component is unmounted, this is a no-op.`;
+
+export const TABLE_INVALID_CELL_RENDERER_DEPS = `${ns} cellRendererDependencies should either always be defined or undefined; this feature cannot be enabled during the compoennt lifecycle`;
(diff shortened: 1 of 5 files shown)


packages/table/src/common/errors.ts
```diff
@@ -39,3 +39,5 @@ export const TABLE_NUM_COLUMNS_COLUMN_WIDTHS_MISMATCH =
     ns + ` Table requires columnWidths.length to equal the number of <Column>s if provided.`;
 
 export const TABLE_UNMOUNTED_RESIZE_WARNING = `${ns} Table resize method called while component is unmounted, this is a no-op.`;
+
+export const TABLE_INVALID_CELL_RENDERER_DEPS = `${ns} cellRendererDependencies should either always be defined or undefined; this feature cannot be enabled during the compoennt lifecycle`;
```
adidahiya:
```suggestion
export const TABLE_INVALID_CELL_RENDERER_DEPS = `${ns} cellRendererDependencies should either always be defined or undefined; this feature cannot be enabled during the component lifecycle`;
```

PR #5146, opened 2022-02-27
fix: node_modules relative paths in sourcemaps

zachkirsch:
#### Checklist

- [ ] Includes tests
- [ ] Update documentation

#### Changes proposed in this pull request:

Taking a stab at fixing #3500.

The source maps published with blueprint are incorrect. For example:
```json
app/node_modules/@blueprintjs/select/lib/css/blueprint-select.css.map
{
    "version":3,
    "sources":[
        "../../src/components/omnibar/_omnibar.scss",
        "../../../../node_modules/@blueprintjs/core/src/common/_react-transition.scss",
        ...
```

The resolved path of that second source is `app/node_modules/node_modules/@blueprintjs/core/src/common/_react-transition.scss` (notice the double `node_modules`). This leads to webpack errors like:
```
Failed to parse source map from '/Users/zachkirsch/Dropbox/Mac/Documents/trace-app/node_modules/node_modules/@blueprintjs/core/src/common/_react-transition.scss' file: Error: ENOENT: no such file or directory, open '/Users/zachkirsch/Dropbox/Mac/Documents/trace-app/node_modules/node_modules/@blueprintjs/core/src/common/_react-transition.scss'
```

**Where does `"../../../../node_modules` come from?**

When the sass is compiled, the path `"~@blueprintjs/core/src/common/react-transition"` is resolved to `"node_modules/@blueprintjs/core/src/common/_react-transition.scss"`, and `node_modules` is 4 directories up, hence the final result of `../../../../node_modules/@blueprintjs/core/src/common/_react-transition.scss`. Ideally node-sass-chokidar could resolve/follow the symlink to figure out that there's a more direct route to that file (`../../../core/src/common/_react-transition.scss`), but it doesn't seem like that's possible.

I tried moving to `dart-sass` since `node-sass-chokidar` is [deprecated](https://github.com/michaelwayman/node-sass-chokidar), but it looks like the source maps have the same issue (paths going through node_modules).

So this PR is a "cheap" solution - after writing the source maps files, convert the "roundabout" `node_modules` paths to more direct paths:

```json
{
	"version": 3,
	"sources": [
		"../../src/components/omnibar/_omnibar.scss",
		"../../../core/src/common/_react-transition.scss",
``` 

#### Reviewers should focus on:

Is there a place I can add tests?

PR #5163, opened 2022-03-09
chore: Fix sass compile script for OS X

mattskrobola:
Was running into `yarn compile` errors locally: `sed: 1: "lib/css//blueprint-icon ...": extra characters at the end of l command
error Command failed with exit code 1.` I think this was introduced with https://github.com/palantir/blueprint/pull/5146. 

Found this solution `Unlike Ubuntu, OS X requires the extension to be explicitly specified. The workaround is to set an empty string:` which fixed the issue

#### Checklist

- [ ] Includes tests
- [ ] Update documentation


adidahiya (approved):
thanks @mattskrobola!

PR #5165, opened 2022-03-14
[core] fix(AsyncControllableInput): handle compound composition events

Athenodoros:
There's an issue with the AsyncControllableInput, where Korean input can lead to incorrect values being stored.

#### The Bug

The cause is the way that Korean handles keyboard events: letters in the Korean language are combined into symbols, which are bubbled up as [CompositionEvents](https://developer.mozilla.org/en-US/docs/Web/API/CompositionEvent). For instance, when typing the characters `ㄷ` -> `ㅣ` -> `ㄱ` -> `ㅏ`, they would successively render as `ㄷ`, `디`, `딕`, `디가` - note that with the last input the value splits out into two characters, which manifests as one CompositionEvent (`디`) ending and then a new one (`가`) immediately starting. I put together some logging to show the flow (times in seconds, logging code [here](https://github.com/Athenodoros/blueprint/blob/hms/composition-logging/packages/core/src/components/forms/asyncControllableInput.tsx)):

Screen Shot 2022-03-14 at 11 49 44 AM

Currently, this gives incorrect results with AsyncControllableInput: because the `value` prop can be updated on a delay (eg. [example from the docs](https://github.com/palantir/blueprint/blob/develop/packages/docs-app/src/examples/core-examples/inputGroupExample.tsx)), when the first composition ends the stored value (`디`) is blown away by the old state (`딕`), which results in the incorrect final result `닥가` (note the extra `ㄱ` in the first character).

#### The solution

As far as I can tell, there's no way in the onCompositionEnd callback to determine whether another CompositionEvent is about to start. If that's the case, the best solution I can think of is to manually wait to unlock the component for edits after some time delay - I've chosen 10ms, because in my experimentation it was always <=1ms before the next event started.

#### Checklist

- [x] Includes tests
- [ ] Update documentation

Is there documentation I should add?

### DIFF
--- a/packages/core/src/components/forms/asyncControllableInput.tsx
+++ b/packages/core/src/components/forms/asyncControllableInput.tsx
@@ -146,6 +146,9 @@ export class AsyncControllableInput extends React.PureComponent<
     };
 
     private handleCompositionEnd = (e: React.CompositionEvent<HTMLInputElement>) => {
+        // In some non-latin languages, a keystroke can end a composition event and immediately afterwards open another.
+        // This timeout unlocks nextValue to be overwritten by the `value` prop, if and only if a delay (10ms) has
+        // passed without a new composition event starting.
         this.compositionStatusTimeoutID = window.setTimeout(() => this.setState({ isComposing: false }), 10);
         this.props.onCompositionEnd?.(e);
     };
--- a/packages/core/src/components/forms/asyncControllableInput.tsx
+++ b/packages/core/src/components/forms/asyncControllableInput.tsx
@@ -17,7 +17,7 @@
 import * as React from "react";
 import { polyfill } from "react-lifecycles-compat";
 
-import { DISPLAYNAME_PREFIX } from "../../common/props";
+import { AbstractPureComponent2, DISPLAYNAME_PREFIX } from "../../common";
 
 export interface IAsyncControllableInputProps
     extends React.DetailedHTMLProps<React.InputHTMLAttributes<HTMLInputElement>, HTMLInputElement> {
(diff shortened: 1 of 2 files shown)


palantirtech:
Thanks for your interest in [palantir/blueprint](https://github.com/palantir/blueprint), @Athenodoros! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

packages/core/src/components/forms/asyncControllableInput.tsx
```diff
@@ -134,17 +142,21 @@ export class AsyncControllableInput extends React.PureComponent<
...
     };
 
     private handleCompositionEnd = (e: React.CompositionEvent<HTMLInputElement>) => {
-        this.setState({ isComposing: false });
+        // In some non-latin languages, a keystroke can end a composition event and immediately afterwards start another.
+        // This can lead to unexpected characters showing up in the text input. In order to circumvent this problem, we
+        // use a timeout which creates a delay which merges the two composition events, creating a more natural and predictable UX.
+        // `this.state.nextValue` will become "locked" (it cannot be overwritten by the `value` prop) until a delay (10ms) has
+        // passed without a new composition event starting.
+        this.cancelPendingCompositionEnd = this.setTimeout(
```
adidahiya, line 156: @Athenodoros we have a mechanism in `AbstractPureComponent2` to cancel outstanding timeouts / intervals when components unmount, so I've gone ahead and used that here instead of the lower-level `window.setTimeout`

packages/core/src/components/forms/asyncControllableInput.tsx
```diff
@@ -134,17 +142,21 @@ export class AsyncControllableInput extends React.PureComponent<
...
+            () => this.setState({ isComposing: false }),
+            AsyncControllableInput.COMPOSITION_END_DELAY,
+        );
         this.props.onCompositionEnd?.(e);
```
adidahiya, line 160: @Athenodoros looks like we still fire the `onCompositionEnd` callback without the new delay... that sounds fine to me for now, but I just want to make sure that doesn't cause any problems for you using the component with korean characters?

adidahiya (approved):

adidahiya:
docs preview: https://52033-71939872-gh.circle-artifacts.com/0/packages/docs-app/dist/index.html#core/components/text-inputs.input-group

first input group in that example uses `asyncControl={true}`. could you verify the bug is fixed with korean characters there @Athenodoros?

packages/core/src/components/forms/asyncControllableInput.tsx
Athenodoros: I think it's fine - anyone overriding onCompositionEnd should probably have the native behaviour (and I think React's synthetic events don't work well with timeouts anyway?). If it becomes a problem then I think they're probably doing something specific enough that they can manage the composition events themselves...

I can't see anywhere onCompositionEnd is actually supplied - most of the benefit of using this component is in not having to think about it.

Athenodoros:
Yep, looks good to me - thanks @adidahiya!

PR #5168, opened 2022-03-15
[core] fix(OverflowList): improve spacer width check

### LINKED ISSUE #5167: Overflow list does not detect popover size correctly due to scale
#### Code Sandbox

Link to a minimal repro https://codesandbox.io/s/blueprint-sandbox-forked-xcw0jp?file=/src/index.tsx

#### Steps to reproduce

Use an overflow list in a popover with the minimal prop set to false

#### Actual behavior

The overflow list will never render more than one element

#### Expected behavior

The overflow list renders as many elements as fit in the container

#### Possible solution
- We currently scale the popover in css [here](https://github.com/palantir/blueprint/blob/33c3af317ba77238bf0ea66ccb33f6043d714781/packages/popover2/src/_common.scss#L90)
- If we switch to using `element.offsetWidth` instead of `element.getBoundingClientRect().width` it will remove the scale factor from the spacer repartition check.

dylanrcooke:
#### Checklist

- [ ] Includes tests
- [ ] Update documentation

Fix the OverflowList spacer width check to account for css scaling

Any risks to using `offsetWidth` instead of `getBoundingClientRect()`

#### Screenshot

No user facing change

### DIFF
--- a/packages/core/src/components/overflow-list/overflowList.tsx
+++ b/packages/core/src/components/overflow-list/overflowList.tsx
@@ -151,7 +151,7 @@ export class OverflowList<T> extends React.Component<OverflowListProps<T>, IOver
     /** A cache containing the widths of all elements being observed to detect growing/shrinking */
     private previousWidths = new Map<Element, number>();
 
-    private spacer: Element | null = null;
+    private spacer: HTMLElement | null = null;
 
     public componentDidMount() {
         this.repartition(false);
@@ -255,7 +255,7 @@ export class OverflowList<T> extends React.Component<OverflowListProps<T>, IOver
                 overflow: [],
                 visible: this.props.items,
             }));
-        } else if (this.spacer.getBoundingClientRect().width < 0.9) {
+        } else if (this.spacer.offsetWidth < 0.9) {
             // spacer has flex-shrink and width 1px so if it's much smaller then we know to shrink
             this.setState(state => {
                 if (state.visible.length <= this.props.minVisibleItems!) {


palantirtech:
Thanks for your interest in [palantir/blueprint](https://github.com/palantir/blueprint), @dylanrcooke! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

packages/core/src/components/overflow-list/overflowList.tsx
```diff
@@ -151,7 +151,7 @@ export class OverflowList<T> extends React.Component<OverflowListProps<T>, IOver
     /** A cache containing the widths of all elements being observed to detect growing/shrinking */
     private previousWidths = new Map<Element, number>();
 
-    private spacer: Element | null = null;
+    private spacer: HTMLElement | null = null;
```
invliD, line 154: Technically this no longer allows non-HTML usage of the overflow list, but given that it's relatively specific for HTML layout, I'm ok with this.

packages/core/src/components/overflow-list/overflowList.tsx
```diff
@@ -255,7 +255,7 @@ export class OverflowList<T> extends React.Component<OverflowListProps<T>, IOver
                 overflow: [],
                 visible: this.props.items,
             }));
-        } else if (this.spacer.getBoundingClientRect().width < 0.9) {
+        } else if (this.spacer.offsetWidth < 0.9) {
```
invliD, line 258: I tested this works in our internal repro.

invliD (approved):
@adidahiya to confirm that this minor break is ok. We'll also need this on 4.x.

adidahiya (approved):

PR #5171, opened 2022-03-16
[select] fix: mark items arrays as readonly where appropriate

### LINKED ISSUE #4976: Select's types don't accept readonly item arrays
#### Environment

- __Package version(s)__:
  - @blueprintjs/core: 3.48.0
  - @blueprintjs/select: 3.17.0
- __Operating System__: macOS 11.6
- __Browser name and version__: Firefox 92

#### Code Sandbox

https://codesandbox.io/s/blueprint-sandbox-forked-8sm3h?file=/src/index.tsx

#### Steps to reproduce

1.  open above link

#### Actual behavior

The code does not compile

#### Expected behavior

Select should be able to receive readonly array since it does not mutate the passed array.

#### Possible solution

Annotate the props with `readonly`

maclockard:
#### Checklist

- [x] Includes tests
- [x] Update documentation

Added readonly annotations to select array props

The types

### DIFF
--- a/packages/docs-app/src/examples/select-examples/multiSelectExample.tsx
+++ b/packages/docs-app/src/examples/select-examples/multiSelectExample.tsx
@@ -209,7 +209,7 @@ export class MultiSelectExample extends React.PureComponent<IExampleProps, IMult
         this.selectFilms([film]);
     }
 
-    private selectFilms(filmsToSelect: readonly IFilm[]) {
+    private selectFilms(filmsToSelect: IFilm[]) {
         const { createdItems, films, items } = this.state;
 
         let nextCreatedItems = createdItems.slice();
@@ -259,7 +259,7 @@ export class MultiSelectExample extends React.PureComponent<IExampleProps, IMult
         }
     };
 
-    private handleFilmsPaste = (films: readonly IFilm[]) => {
+    private handleFilmsPaste = (films: IFilm[]) => {
         // On paste, don't bother with deselecting already selected values, just
         // add the new ones.
         this.selectFilms(films);
--- a/packages/docs-app/src/examples/select-examples/multiSelectExample.tsx
+++ b/packages/docs-app/src/examples/select-examples/multiSelectExample.tsx
@@ -209,7 +209,7 @@ export class MultiSelectExample extends React.PureComponent<IExampleProps, IMult
         this.selectFilms([film]);
     }
 
-    private selectFilms(filmsToSelect: IFilm[]) {
+    private selectFilms(filmsToSelect: readonly IFilm[]) {
         const { createdItems, films, items } = this.state;
 
         let nextCreatedItems = createdItems.slice();
@@ -259,7 +259,7 @@ export class MultiSelectExample extends React.PureComponent<IExampleProps, IMult
         }
     };
 
-    private handleFilmsPaste = (films: IFilm[]) => {
+    private handleFilmsPaste = (films: readonly IFilm[]) => {
         // On paste, don't bother with deselecting already selected values, just
         // add the new ones.
         this.selectFilms(films);
--- a/packages/docs-theme/src/components/navigator.tsx
+++ b/packages/docs-theme/src/components/navigator.tsx
@@ -85,7 +85,7 @@ export class Navigator extends React.PureComponent<INavigatorProps> {
     }
 
     private filterMatches: ItemListPredicate<INavigationSection> = (query, items) =>
-        filter([...items], query, {
+        filter(items, query, {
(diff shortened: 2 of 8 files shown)


packages/select/src/common/predicate.ts
```diff
@@ -18,7 +18,7 @@
  * A custom predicate for returning an entirely new `items` array based on the provided query.
  * See usage sites in `IListItemsProps`.
  */
-export type ItemListPredicate<T> = (query: string, items: T[]) => T[];
+export type ItemListPredicate<T> = (query: string, items: readonly T[]) => T[];
```
maclockard: Changing the types of callback arguments may cause compile breaks for consumers, but should be easily fixed, since if they were mutating this argument here, it was probably causing problems

packages/docs-theme/src/components/navigator.tsx
```diff
@@ -85,7 +85,7 @@ export class Navigator extends React.PureComponent<INavigatorProps> {
     }
 
     private filterMatches: ItemListPredicate<INavigationSection> = (query, items) =>
-        filter(items, query, {
+        filter([...items], query, {
```
maclockard: could also cast as a writable array instead. Problem here is the upstream types for `filter` from `fuzzaldrin-plus` don't use `readonly` for this argument

packages/select/src/common/listItemsProps.ts
```diff
@@ -155,7 +155,7 @@ export interface IListItemsProps<T> extends Props {
     /**
      * Callback invoked when multiple items are selected at once via pasting.
      */
-    onItemsPaste?: (items: T[]) => void;
+    onItemsPaste?: (items: readonly T[]) => void;
```
maclockard: this one is optional since `items` here appears to be a new array when this callback is invoked.

packages/docs-app/src/examples/select-examples/multiSelectExample.tsx
```diff
@@ -209,7 +209,7 @@ export class MultiSelectExample extends React.PureComponent<IExampleProps, IMult
         this.selectFilms([film]);
     }
 
-    private selectFilms(filmsToSelect: IFilm[]) {
+    private selectFilms(filmsToSelect: readonly IFilm[]) {
```
adidahiya: this seems like a breaking change for consumers, no? looks like you were getting TS compilation errors until you changed this.

maclockard: Yeah, that was what I was talking about here: https://github.com/palantir/blueprint/pull/5171#discussion_r828450664 

It is a compile time break, but if people were actually modifying the array I would imagine the select wouldn't be working.

However, it's easy enough to make a mutable copy of the item list and pass that to the callback instead if you'd prefer to avoid the compile time break

adidahiya: yeah, i'd prefer that to avoid upgrade surprises right now. I don't want to force consumers to add the `readonly ` even if it's easy to do so. we can make the compile break in v5.0 (v4 is already closed for breaking changes)

adidahiya (approved):

adidahiya:
I have to revert this, it's causing downstream compilation errors:

```
[error] TypeScript error packages/[redacted]/filterable-select.tsx:25:9 - error TS2345: Argument of type 'readonly T[]' is not assignable to parameter of type 'T[]'.
  The type 'readonly T[]' is 'readonly' and cannot be assigned to the mutable type 'T[]'.

25         items,
           ~~~~~
```

maclockard:
@adidahiya curious if that's actually coming from using a `Select`/`QueryList`/etc component or if its coming from downstream consumer reusing `IListItemsProps`/`ISelectProps`/etc for their own components?

I suspect it's probably the second looking at the file name. If that's the case, seems easy enough to only apply the change for types blueprint components use, and leave the exported types alone?

adidahiya:
It is the latter, "downstream consumer reusing `ISelectProps` for their own components". I'm open to an alternative fix if you want to send a PR. I'll test it downstream before merging.

maclockard:
hmmm, now that I think about it, my potential solution to that problem kinda falls under 'no one wins' territory since folks trying to get the actual prop types will have trouble.

I think it makes sense to do this as a breaking change as a part of v5. That way we could also fix the callback types as well. What does queuing a change for v5 look like?

adidahiya:
> What does queuing a change for v5 look like?

I'm not ready for this yet, but i'll have a process for this ready in the next few weeks. Stay tuned

PR #5172, opened 2022-03-16
chore: upgrade typescript, replace resize-observer-polyfill

adidahiya:
- Upgrade TypeScript to v4.6
- Replace resize-observer-polyfill with @juggle/resize-observer
  - The former has type conflict errors with TS 4.2 and is unmaintained, see https://github.com/que-etc/resize-observer-polyfill/issues/83
  - [@juggle/resize-observer](https://github.com/juggle/resize-observer) looks like a better maintained alternative


adidahiya:
confirmed that copy-pasting cells in the table component still works :+1:

PR #5202, opened 2022-03-27
[table] fix(Table2): uncontrolled resizing of cols/rows

mattskrobola:
Fixes resizing rows/cols not visually updating for Tables that don't use the `columnWidths` / `rowHeights` prop.

I think this was introduced [here](https://github.com/palantir/blueprint/commit/e9e562060e6f2266b059ccc69bbe795f886e4422#diff-043a1f4cfcb847ca6d0d77077219b447e26786e280bbf7fe3428ddb0a9f7ef84L557) where before we were comparing `props.columnWidths` to `prevState.columnWidths` where `props.columnWidths` was  always `undefined` for Tables not using the prop. Now that there is an undefined check `didColumnWidthsChange` will always be false for Tables not using the prop.

Maybe a better spot for this update is in `handleColumnWidthChanged` ? We call `validateGrid` there, but that doesn't seem to visually update the table with the new sizes.
#### Checklist

- [ ] Includes tests
- [ ] Update documentation

Added a check that prevState !== currentState for column widths and row heights when the prop is not set

#### Screenshot

before:
before

after:
after

### DIFF
--- a/packages/table/src/table2.tsx
+++ b/packages/table/src/table2.tsx
@@ -573,14 +573,14 @@ export class Table2 extends AbstractComponent2<Table2Props, TableState, TableSna
                 (dep, index) => dep !== (prevProps.cellRendererDependencies ?? [])[index],
             );
 
-        const didColumnWidthsChange =
-            this.props.columnWidths !== undefined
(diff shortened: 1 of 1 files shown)


adidahiya (approved):

adidahiya:
thanks @mattskrobola!

PR #5215, opened 2022-04-01
migrate to dart-sass

pgoldberg:
#### Checklist

- [ ] Includes tests
- [ ] Update documentation


palantirtech:
Thanks for your interest in [palantir/blueprint](https://github.com/palantir/blueprint), @pgoldberg! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

packages/core/src/components/menu/_submenu.scss
```diff
@@ -17,11 +17,11 @@
       // keep a trail of hovered items as submenus are opened
       /* stylelint-disable scss/at-extend-no-missing-placeholder */
       &:not([class*="#{$ns}-intent-"]) {
-        @extend .#{$ns}-menu-item:hover;
+        @extend .#{$ns}-menu-item, :hover;
```
pgoldberg, line 20: https://sass-lang.com/documentation/at-rules/extend#disallowed-selectors

> If `.message.info` could be extended, the definition of `@extend` says that elements matching the extender would be styled as though they matched `.message.info`. That’s just the same as matching both `.message` and `.info`, so there wouldn’t be any benefit in writing that instead of `@extend .message, .info`.

packages/node-build-scripts/sass-compile.ts
```diff
@@ -0,0 +1,40 @@
+import fs from "fs-extra";
```
pgoldberg, line 1: see https://github.com/palantir/blueprint/pull/5215/files#diff-8c508100032d0fe222dc9679ae03f037d137a2323c18decd3775ccea41dddf2eR14

> the `dart-sass` CLI doesn't support custom functions or importers, but the JS API does, so delegate to node

Very frustrating, but basically the only way we can migrate and keep `sass-inline-svg` and the importer (which allows us to resolve tilde imports outside of webpack), we need to delegate to node

pgoldberg:
merged in https://github.com/palantir/blueprint/pull/5216

PR #5235, opened 2022-04-05
[core] feat(FocusStyleManager): add escape hatch

### LINKED ISSUE #5229: Disabling focus manager takes two clicks
#### Environment

- __Package version(s)__: @blueprintjs/core 4.0.4
- __Operating System__: Ubuntu 20.04
- __Browser name and version__: Firefox 98, Chromium 100

#### Code Sandbox

https://codesandbox.io/s/blueprint-sandbox-forked-4x77pn?file=/src/index.tsx

#### Steps to reproduce

1. Click on the *Always focus* button 
   - note the focus outline on this button - this is expected 
   - note the state text - this is expected
1. Click on the *Focus on tabs only* button -
   - note the focus outline on this button - this is _not_ expected
   - note the state text - this _is_ expected and tell us that `FocusStyleManager.onlyShowFocusOnTabs()` was called
1. Click on the *Focus on tabs only* button 
   - note the focus outline is gone - this should have happened on the previous click, not this one

#### Actual behavior

The style change is not applied until the click after `FocusStyleManager.onlyShowFocusOnTabs()` is called.

#### Expected behavior

The style change should be applied immediately when `FocusStyleManager.onlyShowFocusOnTabs()` is called.

#### Possible solution

I think a `mousedown` event is already queued when `FocusStyleManager.onlyShowFocusOnTabs()` is called so it immediately fires and applies the style change again. We could add an `if (this.isRunning)` in the `mousedown` handler to make sure it doesn't do this if the style manager has been disabled in the meantime.

dlech:
#### Checklist

- [ ] Includes tests
- [x] Update documentation

Not sure how to do tests for css.

This adds a new `Classes.FOCUS_MANAGER_IGNORE` for allowing individual components to chose to ignore the focus manager and always show the focus outline, even when clicked.

### DIFF
--- a/packages/core/src/accessibility/_focus-states.scss
+++ b/packages/core/src/accessibility/_focus-states.scss
@@ -5,8 +5,8 @@
   @include focus-outline();
 }
 
-// override any focus outline anywhere
-.#{$ns}-focus-disabled :focus {
+// override any focus outline anywhere unless specifically ignored
+.#{$ns}-focus-disabled :focus:not(.#{$ns}-focus-manager-ignore *) {
   /* stylelint-disable declaration-no-important */
   outline: none !important;
 
--- a/packages/core/src/accessibility/_focus-states.scss
+++ b/packages/core/src/accessibility/_focus-states.scss
@@ -6,7 +6,7 @@
 }
 
 // override any focus outline anywhere unless specifically ignored
-.#{$ns}-focus-disabled :focus:not(.#{$ns}-focus-manager-ignore *) {
+.#{$ns}-focus-disabled :focus:not(.#{$ns}-focus-style-manager-ignore *) {
   /* stylelint-disable declaration-no-important */
   outline: none !important;
 
--- a/packages/core/src/common/classes.ts
+++ b/packages/core/src/common/classes.ts
@@ -61,6 +61,7 @@ export const INTENT_WARNING = intentClass(Intent.WARNING)!;
 export const INTENT_DANGER = intentClass(Intent.DANGER)!;
 
 export const FOCUS_DISABLED = `${NS}-focus-disabled`;
+export const FOCUS_MANAGER_IGNORE = `${NS}-focus-manager-ignore`;
 
 // text utilities
 export const UI_TEXT = `${NS}-ui-text`;
--- a/packages/core/src/common/classes.ts
+++ b/packages/core/src/common/classes.ts
@@ -61,7 +61,7 @@ export const INTENT_WARNING = intentClass(Intent.WARNING)!;
 export const INTENT_DANGER = intentClass(Intent.DANGER)!;
 
 export const FOCUS_DISABLED = `${NS}-focus-disabled`;
-export const FOCUS_MANAGER_IGNORE = `${NS}-focus-manager-ignore`;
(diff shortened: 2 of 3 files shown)


packages/core/src/accessibility/_focus-states.scss
```diff
@@ -5,8 +5,8 @@
   @include focus-outline();
 }
 
-// override any focus outline anywhere
-.#{$ns}-focus-disabled :focus {
+// override any focus outline anywhere unless specifically ignored
+.#{$ns}-focus-disabled :focus:not(.#{$ns}-focus-manager-ignore *) {
```
adidahiya:
```suggestion
.#{$ns}-focus-disabled :focus:not(.#{$ns}-focus-style-manager-ignore *) {
```

packages/core/src/common/classes.ts
```diff
@@ -61,6 +61,7 @@ export const INTENT_WARNING = intentClass(Intent.WARNING)!;
 export const INTENT_DANGER = intentClass(Intent.DANGER)!;
 
 export const FOCUS_DISABLED = `${NS}-focus-disabled`;
+export const FOCUS_MANAGER_IGNORE = `${NS}-focus-manager-ignore`;
```
adidahiya:
```suggestion
export const FOCUS_STYLE_MANAGER_IGNORE = `${NS}-focus-style-manager-ignore`;
```

packages/core/src/docs/accessibility.md
```diff
@@ -37,6 +37,25 @@ __@blueprintjs/core__ package. It supports the following public methods:
 - `FocusStyleManager.onlyShowFocusOnTabs(): void`: Enable behavior which hides focus styles during mouse interaction.
 - `FocusStyleManager.alwaysShowFocus(): void`: Stop this behavior (focus styles are always visible).
 
+@### Selectivly ignoring the focus manager
+
+There is an escape hatch to allow components to ignore the focus manager. This
+can be useful when you do want to always show the focus outline, but only for certain
+components, like a tree. This is done by applying `Classes.FOCUS_MANAGER_IGNORE`
+to a container.
```
adidahiya:
```suggestion
@### Selectively ignoring the focus style manager

There is an escape hatch to allow components to ignore the focus style manager. This
can be useful when you do want to always show the focus outline, but only for certain
components, like a tree. This is done by applying `Classes.FOCUS_STYLE_MANAGER_IGNORE`
to a container element.
```

packages/core/src/docs/accessibility.md
```diff
@@ -37,6 +37,25 @@ __@blueprintjs/core__ package. It supports the following public methods:
...
+
+```tsx
+import { Classes } from "@blueprintjs/core";
+
+
+const MyComponent = () => ({
+    <div classname={Classes.FOCUS_MANAGER_IGNORE}>
```
adidahiya:
```suggestion
    <div classname={Classes.FOCUS_STYLE_MANAGER_IGNORE}>
```

adidahiya (changes requested):
looks good, let's just make the naming a little more explicit so that it's clear this corresponds to `FocusStyleManager`

adidahiya:
In fact I would go one step further and rename the docs section from "Focus management" to "Focus style management". but I can do that in a separate PR.

adidahiya (approved):

PR #5242, opened 2022-04-06
[core] allow MenuItem keyboard focus when disabled

dlech:
According to the [aria spec](https://www.w3.org/TR/wai-aria-practices-1.2/#menu):

> Disabled menu items are focusable but cannot be activated.

In this context, "focusable" includes keyboard focusable.

#### Checklist

- [ ] Includes tests
- [ ] Update documentation


dlech:
This should probably wait until the rest of the keyboard accessibility is implemented.

PR #5246, opened 2022-04-07
[node-build-scripts] fix(generate-css-variables): LibSass compat

### LINKED ISSUE #5244: variables.scss is incompatible with node-sass
#### Environment

- __Package version(s)__: core v4.1.1

#### Bug

@blueprintjs/core v4.1.1 contains a `lib/scss/variables.scss` file which uses `@use "sass:math"` and `math.div(...)`. This language feature is only available in Dart Sass v1.33+, and not available in LibSass / node-sass ([see docs](https://sass-lang.com/documentation/modules/math#div)). This means that node-sass compatibility is broken in the latest release.

#### Solution

Update `generate-css-variables` script, use `get-sass-vars` functionality (we are already doing this to generate Less variables) to simplify the variable values and compute the math which `math.div()` is currently doing.

pgoldberg:
#### Fixes #5244

Uses `get-sass-vars` for scss variable generation as well as less generation

#### Checklist

- [ ] Includes tests
- [ ] Update documentation

### DIFF
--- a/packages/node-build-scripts/generate-css-variables.js
+++ b/packages/node-build-scripts/generate-css-variables.js
@@ -33,94 +33,139 @@ function main() {
         }).argv;
 
     const outputFilename = args["outputFileName"];
-    const scssVariables = generateScssVariables(args["_"], outputFilename, args["retainDefault"]);
-    generateLessVariables(scssVariables, outputFilename);
+    const parsedInput = getParsedVars(args["_"], args["retainDefault"]);
+    generateScssVariables(parsedInput, outputFilename);
+    generateLessVariables(parsedInput, outputFilename);
+}
+
+/**
+ * Pulls together variables from the specified Sass source files, sanitizes them for consumption,
+ * and gets compiled output from `get-sass-vars`.
+ *
+ * @param {string[]} inputSources
+ * @param {boolean} retainDefault
+ * @returns {{parsedVars: object, varsInBlocks: Set<string>[]}} output compiled variable values and grouped variable blocks
+ */
+function getParsedVars(inputSources, retainDefault) {
+    // concatenate sources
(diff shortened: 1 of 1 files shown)


packages/node-build-scripts/generate-css-variables.js
```diff
@@ -33,94 +33,139 @@ function main() {
...
         .replace(/(@import|\/\/).*\n+/g, "")
         .replace(/@use "sass:math";\n/g, "")
         .replace(/border-shadow\((.+)\)/g, "0 0 0 1px rgba($black, $1)")
         .replace(/\n{3,}/g, "\n\n");
     if (!retainDefault) {
-        variablesScss = variablesScss.replace(/\ \!default/g, "");
+        cleanedInput = cleanedInput.replace(/\ \!default/g, "");
     }
-    variablesScss = [COPYRIGHT_HEADER, USE_MATH_RULE, variablesScss].join("\n");
+    cleanedInput = [USE_MATH_RULE, cleanedInput].join("\n");
```
adidahiya, line 64: no need for the USE_MATH_RULE now, we can remove that constant completely

adidahiya (dismissed):

packages/node-build-scripts/generate-css-variables.js
pgoldberg: It is actually needed here so it can properly run through `getSassVars` (which compiles with `sass`, so without it, it wouldn't know how to resolve `math.div`s). When it goes through `generateScssVariables` there's no outputted `@use "sass:math"`

adidahiya: oh wait, I guess we need this to compute the values... sorry, disregard

adidahiya (approved):
lgtm, I verified the variables.scss and variables.less output locally

PR #5247, opened 2022-04-08
Disable today button if today's date cannot be chosen

### LINKED ISSUE #5227: Add support for "minDate" and "maxDate" to "Today" button on datepicker action bar
#### Environment

- __Package version(s)__: 3.23.1
- __Browser and OS versions__: Google Chrome 99.0.4844.74 on Mac OS 12.3.1

#### Feature request

It would be really nice if the "Today" button on the action bar respected "minDate" and "maxDate"

#### Examples

When I have "minDate" set in the future, "Today" button should be disabled, or at least shouldn't set the date.
When I have "maxDate" set in the past, "Today" button should be disabled, or at least shouldn't set the date.

scooobs:
#### Checklist

- [x] Includes tests
- [ ] Update documentation

Disable the Today button in the Actions bar if it does not fall between minDate and maxDate.

#### Screenshot
(here minDate is April 8th 2023 and maxDate is April 8th 2025)
Screen Shot 2022-04-08 at 3 26 23 pm

With regards to documentation, is there anything that needs to be updated?

### DIFF
--- a/packages/datetime/src/common/classes.ts
+++ b/packages/datetime/src/common/classes.ts
@@ -34,7 +34,6 @@ export const DATEPICKER_DAY_IS_TODAY = `${DATEPICKER_DAY}--isToday`;
 export const DATEPICKER_DAY_WRAPPER = `${DATEPICKER}-day-wrapper`;
 export const DATEPICKER_FOOTER = `${DATEPICKER}-footer`;
 export const DATEPICKER_MONTH_SELECT = `${DATEPICKER}-month-select`;
-export const DATEPICKER_TODAY_BUTTON_DISABLED = `${NS}-disabled`;
 export const DATEPICKER_YEAR_SELECT = `${DATEPICKER}-year-select`;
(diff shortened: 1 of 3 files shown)


packages/datetime/src/common/classes.ts
```diff
@@ -34,6 +34,7 @@ export const DATEPICKER_DAY_IS_TODAY = `${DATEPICKER_DAY}--isToday`;
 export const DATEPICKER_DAY_WRAPPER = `${DATEPICKER}-day-wrapper`;
 export const DATEPICKER_FOOTER = `${DATEPICKER}-footer`;
 export const DATEPICKER_MONTH_SELECT = `${DATEPICKER}-month-select`;
+export const DATEPICKER_TODAY_BUTTON_DISABLED = `${NS}-disabled`;
```
adidahiya: no need for this class, it's unused. also it's just an alias for the disabled class from the core package

```suggestion
```

adidahiya (changes requested):
lgtm otherwise :+1:

adidahiya (approved):

PR #5251, opened 2022-04-11
[table] fix: vertical scroll bug when ghostCells enabled attempt 2

### LINKED ISSUE #4765: Table vertical scroll is improperly disabled when ghost cells are visible
#### Environment

- __Package version(s)__: at least `"@blueprintjs/table": "3.5.0"` - `3.8.33`
- __Operating System__: OSCatalina 10.15.7 
- __Browser name and version__: Chrome 91.0.4472.101

#### Code Sandbox

https://codesandbox.io/s/blueprint-sandbox-forked-kpx0r?file=/src/index.tsx

#### Steps to reproduce

1. create a table
2. enable ghost cells
3. restrict the table height such that the last row is only partially visible

#### Actual behavior

Vertical scrolling is disabled

#### Expected behavior

Scrolling should be enabled, to reveal the rest of the row.

#### Possible solution

Some tweak to the shouldDisableVerticalScroll function here 
https://github.com/palantir/blueprint/blob/63d5de4c91f6dcfbd9828ed3a40aa1015100df9e/packages/table/src/table.tsx#L1307

mattskrobola:
#### Checklist

- [ ] Includes tests
- [ ] Update documentation

Adds a parameter `rectHeightCorrection` to `getRowIndicesInRect`. This is utilized in `shouldDisableVerticalScroll` to account for the `columnHeaderHeight` when checking for which cells are visible to determine whether or not to disable vertical scrolling. 

Also reverts the table change made in https://github.com/palantir/blueprint/pull/5113 that attempted to fix this issue (which it did for the case when the table used the default column header) but didn't address the root cause of the problem - that header heights weren't being accounted for when determining which rows were currently visible.

Added a smoke test table in the table-app features
#### Screenshot

before:
scroll-before
after:
scroll-after

### DIFF
--- a/packages/table/src/common/grid.ts
+++ b/packages/table/src/common/grid.ts
@@ -264,7 +264,7 @@ export class Grid {
      * `Rect` argument.
      */
     public getRowIndicesInRect(options: GetRowIndicesInRectOptions): RowIndices {
-        const { rect, includeGhostCells = false, columnHeaderHeight = 0, limit = 0 } = options;
+        const { rect, includeGhostCells = false, columnHeaderHeight = 0, limit = Grid.DEFAULT_MAX_ROWS } = options;
         if (rect == null) {
             return { rowIndexEnd: 0, rowIndexStart: 0 };
         }
--- a/packages/table/src/common/grid.ts
+++ b/packages/table/src/common/grid.ts
@@ -253,7 +253,12 @@ export class Grid {
      * Returns the start and end indices of rows that intersect with the given
      * `Rect` argument.
      */
-    public getRowIndicesInRect(rect: Rect, includeGhostCells = false, limit = Grid.DEFAULT_MAX_ROWS): RowIndices {
+    public getRowIndicesInRect(
+        rect: Rect,
+        includeGhostCells = false,
+        rectHeightCorrection = 0,
+        limit = Grid.DEFAULT_MAX_ROWS,
+    ): RowIndices {
         if (rect == null) {
             return { rowIndexEnd: 0, rowIndexStart: 0 };
         }
@@ -262,7 +267,7 @@ export class Grid {
         const { top = 0, height } = rect;
         const { start, end } = this.getIndicesInInterval(
             top,
-            top + height,
+            top + (height - rectHeightCorrection),
             searchEnd,
             !includeGhostCells,
             this.getCumulativeHeightAt,
--- a/packages/table/src/common/grid.ts
+++ b/packages/table/src/common/grid.ts
@@ -35,6 +35,16 @@ export interface ColumnIndices {
     columnIndexEnd: number;
 }
 
+interface GetRowIndicesInRectOptions {
+    rect: Rect;
+    /**
+     * height to subtract from the rect height, as rows hidden behind the columnHeader should not be returned.
+     */
+    columnHeaderHeight?: number;
+    limit?: number;
+    includeGhostCells?: boolean;
+}
+
 /**
  * This class manages the sizes of grid cells using arrays of individual row/column sizes.
  */
@@ -203,7 +213,7 @@ export class Grid {
             return results;
         }
 
-        const { rowIndexStart, rowIndexEnd } = this.getRowIndicesInRect(rect);
+        const { rowIndexStart, rowIndexEnd } = this.getRowIndicesInRect({ rect });
         const { columnIndexStart, columnIndexEnd } = this.getColumnIndicesInRect(rect);
         for (let rowIndex = rowIndexStart; rowIndex <= rowIndexEnd; rowIndex++) {
             for (let columnIndex = columnIndexStart; columnIndex <= columnIndexEnd; columnIndex++) {
@@ -224,7 +234,7 @@ export class Grid {
             return results;
         }
 
-        const { rowIndexStart, rowIndexEnd } = this.getRowIndicesInRect(rect);
+        const { rowIndexStart, rowIndexEnd } = this.getRowIndicesInRect({ rect });
         for (let rowIndex = rowIndexStart; rowIndex <= rowIndexEnd; rowIndex++) {
             results.push(callback(rowIndex));
         }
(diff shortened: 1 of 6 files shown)


packages/table/src/table2.tsx
```diff
@@ -669,11 +669,15 @@ export class Table2 extends AbstractComponent2<Table2Props, TableState, TableSna
             return false;
         }
 
-        const rowIndices = this.grid.getRowIndicesInRect(viewportRect, enableGhostCells!);
+        const rowIndices = this.grid.getRowIndicesInRect(
+            viewportRect,
+            enableGhostCells!,
+            this.columnHeaderHeight || Grid.MIN_COLUMN_HEADER_HEIGHT,
```
mattskrobola: used `||` vs `??` bc `this.columnHeaderHeight` was inconsistently being `0` vs the actual height when running tests

packages/table/src/common/grid.ts
```diff
@@ -253,7 +253,12 @@ export class Grid {
      * Returns the start and end indices of rows that intersect with the given
      * `Rect` argument.
      */
-    public getRowIndicesInRect(rect: Rect, includeGhostCells = false, limit = Grid.DEFAULT_MAX_ROWS): RowIndices {
+    public getRowIndicesInRect(
```
adidahiya: this args list is getting a little unwieldy, could we refactor it to use an options object?

also, when you do, can you document this new argument/option and leave a comment about why we need it?

packages/table/src/table2.tsx
adidahiya: ok, that's fair, but I think I'd rather not make an exception to the [strict-boolean-expressions](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/strict-boolean-expressions.md) rule here (we don't actually lint for it, but I try to enforce it manually). I find the `||` hard to read, how about something like this instead:

```suggestion
            this.columnHeaderHeight > 0 ? this.columnHeaderHeight : Grid.MIN_COLUMN_HEADER_HEIGHT,
```

packages/table/src/table2.tsx
```diff
@@ -669,11 +669,15 @@ export class Table2 extends AbstractComponent2<Table2Props, TableState, TableSna
...
+        );
 
         const isViewportUnscrolledVertically = viewportRect != null && viewportRect.top === 0;
         const areRowHeadersLoading = hasLoadingOption(this.props.loadingOptions, TableLoadingOption.ROW_HEADERS);
-        const areGhostRowsVisible = enableGhostCells! && this.grid.isGhostIndex(rowIndices.rowIndexEnd - 1, 0);
+        const areGhostRowsVisible = enableGhostCells! && this.grid.isGhostIndex(rowIndices.rowIndexEnd, 0);
```
adidahiya, line 680: I came to the same conclusion a little while ago when I was playing around with this code – that the `- 1` is not strictly necessary – so this change lgtm 👍🏽

adidahiya:
PR preview links: [table-dev-app](https://output.circle-artifacts.com/output/job/00ba2e64-4236-4318-aea0-d8608fe6424c/artifacts/0/packages/table-dev-app/dist/index.html), [docs-app](https://output.circle-artifacts.com/output/job/00ba2e64-4236-4318-aea0-d8608fe6424c/artifacts/0/packages/docs-app/dist/index.html)

packages/table/src/table2.tsx
adidahiya: actually, on second thought... we never really want `this.columnHeaderHeight` to be `0`.

we could change the ref handler to prevent that from happening, and keep the min-height logic scoped to that part of this component's code:

```ts
private refHandlers = {
        columnHeader: (ref: HTMLElement | null) => {
            this.columnHeaderElement = ref;
            if (ref != null) {
                this.columnHeaderHeight = Math.max(ref.clientHeight, Grid.MIN_COLUMN_HEADER_HEIGHT);
            }
```

packages/table/src/common/grid.ts
```diff
@@ -253,7 +253,12 @@ export class Grid {
...
+        rect: Rect,
+        includeGhostCells = false,
+        rectHeightCorrection = 0,
```
adidahiya: I think naming this `columHeaderHeight` would be simpler & more explicit

adidahiya (changes requested):
functionality looks good in the preview build, just left some comments for code style

packages/table/src/common/grid.ts
mattskrobola: Makes sense, there are some other functions using `getRowIndicesInRect` like `invokeOnVisibleCellsChangeCallback` that aren't using this arg but maybe they should be? since a row is technically not visible after it goes behind the header column but we still consider it visible since it's in the rect

adidahiya: yeah, I guess `invokeOnVisibleCellsChangeCallback` should know about `columnHeaderHeight` as well...

if we're finding that the Grid needs to know the column header height so often, then maybe we should try to pass it `columnHeaderHeight` from Table / Table2 when we construct/validate the grid? hopefully this value won't get out of sync 🤞🏽, and that change might reduce some code complexity. before you do this, however, can you verify that `invokeOnVisibleCellsChangeCallback` is buggy right now? i.e. that it returns the wrong range of indices when the column header is larger than the default height.

adidahiya:
@mattskrobola please try to avoid force-pushing; I like to see all the changes in a PR branch, and sometimes a force-push will make the CI build confused (it got canceled and had to be manually re-triggered here)

packages/table/src/common/grid.ts
mattskrobola: It's hard to test the accuracy of `invokeOnVisibleCellsChangeCallback` since its using a bleed buffer on both sides when ghost cells are enabled, but it does seem to be a bit off without accounting for the header height.
without accounting for header height:
beforevisible

with accounting for header height:
aftervisible

adidahiya (approved):

packages/table/src/common/grid.ts
adidahiya: It still seems off after accounting for header height... the "start" should be closer to 5 or 6 when you scroll all the way down, no? we can address this in a separate PR, but it looks like it's not a high priority for this particular callback since we haven't gotten any bug reports for `onVisibleCellsChange` accuracy

PR #5252, opened 2022-04-11
[select] fix(a11y): update active item on focus changes

jakejrichards:
Pass handleFocus to itemRenderer to make select & query list components more accessible

#### Checklist

- [ ] Includes tests
- [X] Update documentation

Pass `handleFocus` to itemRenderer to make `Select` & `QueryList` components more accessible. Previously, tabbing through a query list would not update the `activeItem`, this caused unexpected behavior when selecting the actively focused item as it would actually select whatever was in `activeItem`.

These changes don't completely solve the problem as the up/down key driven state can still get out of sync with the focused item, but tabbing again will resume and we'll never select an _unexpected_ item. I think the fully-baked solution here would go both ways focus <-> active, but this would be much greater lift.

The focus here should be whether or not we consider this a viable intermediary state, and if it makes sense to update the active item "on focus" in this way. Another note is that this isn't relevant for `Suggest`/`Multiselect` components as the input is outside of the popover in those - tabbing in general is just completely broken there.

#### Screenshot

Before:
bp-bug-before

After:
bp-bug-after

### DIFF
--- a/packages/docs-app/src/common/films.tsx
+++ b/packages/docs-app/src/common/films.tsx
@@ -132,7 +132,7 @@ export const TOP_100_FILMS: IFilm[] = [
     { title: "Monty Python and the Holy Grail", year: 1975 },
 ].map((m, index) => ({ ...m, rank: index + 1 }));
 
-export const renderFilm: ItemRenderer<IFilm> = (film, { handleClick, modifiers, query }) => {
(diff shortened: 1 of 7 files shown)


adidahiya (approved):
I tried this out in the preview build; this feels like a strict improvement over the current UX, so I'm good with merging this. thanks for the PR :+1:

PR #5254, opened 2022-04-12
[core] feat(useHotkeys): warn when used outside of HotkeysProvider

### LINKED ISSUE #3755: [table] tracking skipped unit tests for console warnings
Added in https://github.com/palantir/blueprint/pull/3702/

adidahiya:
Fixes #3755 (for Table2, not touching Table since it's deprecated)

- Log a console warning when `useHotkeys()` is used outside of a `<HotkeysProvider>` tree.
- Clean up some docs and tests which still referenced the possibility of React 15 (which was dropped in v4.0)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -102,6 +102,8 @@ jobs:
   dist:
     docker: *docker-node-image
     resource_class: large
+    environment:
+      NODE_ENV: test
     steps:
       - checkout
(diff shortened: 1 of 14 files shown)


adidahiya:
I realized that table-dev-app wasn't using the provider. This is now fixed, and the dialog works:

image

Issue #5225, opened 2022-04-05
Unboxing Blueprint4 video: Upgrading from BP3 -> BP4 and sharing thoughts.


vogievetsky:
Writing on behalf of myself and the [Apache Druid](https://github.com/apache/druid) project, I just wanted to say THANK YOU to the blueprint team for the excellent library and all the love y'all pour into it.

Recently @jgoz and I [upgraded](https://github.com/apache/druid/pull/12391) the Blueprint version used by the [Druid web console](https://druid.apache.org/docs/latest/operations/druid-console.html) from 3 to 4 and [recorded the experience](https://www.youtube.com/watch?v=Ew8BgdCfMDg), along with fun memories from the early days of using Blueprint.

We decided to do it sort of like an "unboxing video" that we would do whenever version 4 was released and it happened to coincide with the anniversary of the time we [updated Druid to use BP3](https://github.com/apache/druid/pull/7398) which was cool.

The upgrade itself went smoothly thanks to all the work that was put into version 3 to add depreciation messages and such. I was afraid that the video is going to be totally boring until we ran into one little issue concerning the color overrides (the Druid console overrides the BP colors). The issue we encountered was that `./node_modules/@blueprintjs/core/src/common/_mixins.scss` imports color definitions from the "lib" folder in `@blueprintjs/colors` but we need to import it from the "src" folder. The "src" version includes `!default` in variable definitions, which allows  us to override color variables, but the "lib" version does not.

You can see us struggle with this in the video or look at the diff in `webpack.config.js` to see how we solved it. I am not sure what to make of it... is it a bug or are we stepping outside of the bounds of how BP should be used? In any-case we had a great time and maybe this video can help guide someone else in a similar position.

Thank you again for BP4, I heartily look forward to making another video of us upgrading to BP5 when it is time!

adidahiya:
@vogievetsky thanks for the kind words and sharing your experience working with Blueprint. After spending a lot of time working through this upgrade in our internal code bases, it’s really helpful for me to see how BP4 has landed for dev teams outside of Palantir. There’s often room for improvement in the migration paths for Blueprint upgrades, and you’ve identified some of it here. Here are my findings after reviewing your unboxing video:

[...]

PR #5260, opened 2022-04-15
[core,colors] fix !default modifiers in Sass variables

adidahiya:

Fixes the Sass variables issues I described in https://github.com/palantir/blueprint/issues/5225#issuecomment-1098162039.

The `! default` modifier was accidentally removed from Sass variables in https://github.com/palantir/blueprint/pull/4941. Adding back the CLI argument brings those back, both for core and colors packages.

Also I've removed the ability to override `$ns`, as it is meaningless to do so.

### DIFF
--- a/packages/colors/package.json
+++ b/packages/colors/package.json
@@ -8,7 +8,7 @@
         "compile": "run-p \"compile:*\"",
         "compile:esm": "tsc -p src/",
         "compile:css": "sass-compile ./src",
-        "compile:css-colors": "generate-css-variables --outputFileName colors _colors.scss",
+        "compile:css-colors": "generate-css-variables --retainDefault true --outputFileName colors _colors.scss",
         "dev": "run-p \"compile:esm -- --watch\" \"compile:css -- --watch\"",
         "lint": "npm-run-all -p lint:scss lint:es",
         "lint:scss": "sass-lint",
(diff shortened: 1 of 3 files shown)


bradchristensen:
I don't think this has had the intended effect - the output files (`lib/less/colors.less` and `lib/less/colors.scss`) are still being generated without the `!default` flags.

I started digging and it looks like the [`getSassVarsSync`](https://github.com/palantir/blueprint/blob/c1d7849832e2aab208aaf1191f5fac0d64fe543e/packages/node-build-scripts/generate-css-variables.js#L69) method being used in the `generate-css-variables` script is just throwing away those flags, even though they're being retained first by the `retainDefault` flag passed into the script. `cleanedInput` is correct, but `parsedVars` is not. This actually seems to match up with the example on the `get-sass-vars` README where it shows the `!default` flag being thrown away there too: https://github.com/niksy/get-sass-vars#indexscss

It looks like this might have only become an issue quite recently in https://github.com/palantir/blueprint/pull/5246 when the script was rewritten for interop with node-sass and dart-sass.

I can't see any simple fix, but perhaps the `!default` flags need to be merged back into `parsedVars` afterwards? The variable names are being tracked alongside, so default flags could be tracked with them too.

EDIT: I've opened [this PR](https://github.com/palantir/blueprint/pull/5262) with the fix described above 🙂

PR #5262, opened 2022-04-16
[core,colors] fix !default flags on generated SCSS variables

bradchristensen:
#### Fixes the issue described in the comment: https://github.com/palantir/blueprint/pull/5260#issuecomment-1100586558

The previous fix (https://github.com/palantir/blueprint/pull/5260) partially fixed the issue with the `!default` flag being stripped from generated SCSS variable files, but didn't fully fix it due to another issue introduced just recently (described in my comment on that PR).

This PR implements the idea I had for a fix mentioned in that comment. A regex is used to search for variables in the input SCSS and match whether each variable has a `!default` flag or not. The flags are then restored afterwards, when generating the new output (for SCSS only, since it is not applicable to LESS). While this is just a regex and therefore imperfect, it does seem to work reliably for every use case that currently exists (though I presume, for example, it would fail if it encountered a string value containing a semicolon). I've tested it on the variables files generated in both `colors` and `core` and they both appear to have the correct output now.

- Whether the regex is reliable enough for production usage 😄 (However my thinking is that this should be fine - it improves on the previous situation significantly and the impact of a failed match is obviously very low; it wouldn't break anything around it, but it could be confusing and difficult to track down for anyone unlucky enough to encounter it)

### DIFF
--- a/packages/node-build-scripts/generate-css-variables.js
+++ b/packages/node-build-scripts/generate-css-variables.js
@@ -142,7 +142,7 @@ function generateScssVariables(parsedInput, outputFilename, retainDefault) {
         const variablesArray = Object.entries(parsedVars)
             .filter(([varName, _value]) => varsInBlock.has(varName))
             .map(([varName, value]) => {
-                const defaultFlag = varsWithDefaultFlag.has(varName) ? "!default" : "";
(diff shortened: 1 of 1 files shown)


adidahiya (approved):
lgtm 👍🏽  I tried this out locally and it seems to do the right thing. The regex is pretty reasonable in this situation.

PR #5266, opened 2022-04-21
fix: declare children prop explicitly for class components

### LINKED ISSUE #5250: Getting type errors after upgrading to @types/react v18.0.0.
I upgraded react and @types/react packages to the new v18.0.0 yesterday and I started to get type errors in regards to the removed "children" prop in @types/react version 18.0.0.

Can you please apply a fix for this?

Thanks

adidahiya:
#### Fixes #5250

Follow-up to https://github.com/palantir/blueprint/pull/5258

Audit all remaining usage of `props.children`. Where components _do_ use children, we now declare that prop type explicitly for compatibility with `@types/react@18`.

### DIFF
--- a/packages/core/src/components/forms/radioGroup.tsx
+++ b/packages/core/src/components/forms/radioGroup.tsx
@@ -26,7 +26,9 @@ import { RadioProps, Radio } from "./controls";
 export type RadioGroupProps = IRadioGroupProps;
 /** @deprecated use RadioGroupProps */
 export interface IRadioGroupProps extends Props {
-    /** Drawer contents. */
+    /**
+     * Radio elements. This prop is mutually exclusive with `options`.
+     */
     children?: React.ReactNode;
 
     /**
--- a/packages/core/src/components/forms/radioGroup.tsx
+++ b/packages/core/src/components/forms/radioGroup.tsx
@@ -26,6 +26,9 @@ import { RadioProps, Radio } from "./controls";
 export type RadioGroupProps = IRadioGroupProps;
 /** @deprecated use RadioGroupProps */
 export interface IRadioGroupProps extends Props {
+    /** Drawer contents. */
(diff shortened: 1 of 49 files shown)


packages/core/src/components/forms/radioGroup.tsx
```diff
@@ -26,6 +26,9 @@ import { RadioProps, Radio } from "./controls";
 export type RadioGroupProps = IRadioGroupProps;
 /** @deprecated use RadioGroupProps */
 export interface IRadioGroupProps extends Props {
+    /** Drawer contents. */
```
adidahiya:
```suggestion
```

packages/core/src/components/popover/popover.tsx
```diff
@@ -45,6 +45,8 @@ export interface IPopoverProps extends IPopoverSharedProps {
     /** HTML props for the backdrop element. Can be combined with `backdropClassName`. */
     backdropProps?: React.HTMLProps<HTMLDivElement>;
 
+    children?: React.ReactNode;
```
adidahiya: this should probably go into IPopoverSharedProps

andreash:
I am using @types/react 18.0.21 and @blueprintjs/table 4.7.4. I am still getting a type-script error in vscode for Column components:
```
'Column' cannot be used as a JSX component.
  Its instance type 'Column' is not a valid JSX element.
    The types returned by 'render()' are incompatible between these types.
      Type 'ReactNode' is not assignable to type 'false | Element'.
        Type 'string' is not assignable to type 'false | Element'.ts(2786)
```

Any idea?

Code example:
```
ReactDOM.render(
    <HotkeysProvider>
        <Table2 numRows={5}>
            <Column />
            <Column />
            <Column />
        </Table2>
    </HotkeysProvider>,
    document.querySelector("#app"),
);
```

PR #5277, opened 2022-04-29
[WIP] [datetime] feat: add prop to ignore max/min date bounds

adidahiya:
#### Checklist

- [x] Includes tests
- [x] Update documentation

This partially addresses the linked issue by allowing selection of dates that fall outside of the `minDate` - `maxDate` bounds.

Note that `minDate` and `maxDate` are still used to narrow the dropdown options for the available years in the date picker, but a user can to select a year outside the min/max by typing into a DateInput.

Further work is required if we want to extend the month/year options in DatePickerCaption to allow navigating to those out-of-bounds dates; we'll have to address the issue of performance with a very large `<HTMLSelect>` list.

Is this UX improvement worthwhile on its own? Or do we need to allow navigating to the out-of-bounds months/years?

#### Screenshot

Coming soon


adidahiya:
_Note: I tested in one direction (exceeding the max date boundary) but all of this also applies in the other direction (exceeding the min date boundary)._

Here's what happens when you enable `ignoreBounds={true}` and go past the max date (example uses the default of current date +6mo, which is October 29 right now):



First off, there's a major bug after clicking the next month button where the month displayed in DatePickerCaption is mislabeled until you navigate to the next year. Instead of November and December, the caption shows January because November/December aren't available as options in the dropdown. The calendar days are correct, and you can select them as November/December 2022 dates, but the caption is wrong.

Fixing this will require expanding the month dropdown options to include all the months of the year:

image

In addition, it would also be nice to fix the discontinuity in the year dropdown where years past the max date are not shown:

image

@robertcoopercode are you interested in picking this up again? If not, I will get back to this myself some time in the next couple months...

PR #5278, opened 2022-05-02
Visual design updates, cleaner dark theme

adidahiya:

This PR implements a number of visual design updates to core components. Most of these changes affect the dark theme, with the goal of improving clarity across the UI.

Dark theme styles have shifted one step darker on the color palette, so the default background color is now $dark-gray2 instead of $dark-gray3. This darker background allows us to more consistently use white borders across our components; you'll notice this change in Cards, Dialogs, Buttons (default intent), and Popovers.

The border radius for rounded corners across all components has been reduced from 3px to 2px; this along with the flattening of Button backgrounds creates a more modern look & feel.

A few components got additional layout changes:

- Dialog headers now use a smaller font size, `<h5>` instead of `<h4>`
- Non-ideal state icons and text are more muted, and there is now a horizontal layout option

#### Important things to note:

These changes may produce some inconsistencies in application UIs where dark theme elements were designed with the older existing visual paradigms. To mitigate these issues and bring your application up to date with Blueprint's more modern appearance, the most important things you can do are:

- In dark theme
  - Darken background colors one step on the color palette. Even better if you use variables like `$pt-dark-app-background-color`.
  - Switch to semi-transparent `$white` borders / dividers instead of `$black` ones. Consider using the `$pt-dark-divider-white` variable.
- In both light + dark themes
  - Ensure you are using the `$pt-border-radius` variable wherever you have rounded corners.
  - Use `Card` and elevation styles/classes/variables where appropriate.

- Matching the design spec
- No visual regressions or unexpected interactive states
- Special case styles for light theme toasts look OK? (they don't use the default elevation border/shadow styles)
- NonIdealState muted icons with strokes look OK?
- 
### Screenshot

[Compare against demo-app on develop branch here](https://output.circle-artifacts.com/output/job/a2a4cd18-3d01-4d98-a4b8-7721565b5799/artifacts/0/packages/demo-app/dist/index.html).

#### Light theme

| Before | After |
|-|-|
|image |image|
| image | image |
| image | image |
| image | image |

New horizontal NonIdealState:

image

#### Dark theme

| Before | After |
|-|-|
| image | image |
| image | image |
| image | image |
| image | image |
| image | image |
| image | image |
| image | image |
| image | image |
| image | image |
| image | image |

New horizontal NonIdealState:

image


packages/core/src/components/dialog/_dialog.scss
```diff
@@ -15,7 +16,7 @@ Markup:
   <div class="#{$ns}-dialog">
     <div class="#{$ns}-dialog-header">
       <span class="#{$ns}-icon-large #{$ns}-icon-inbox"></span>
-      <h4 class="#{$ns}-heading">Dialog header</h4>
+      <h5 class="#{$ns}-heading">Dialog header</h5>
```
adidahiya, line 19: This is a minor change to the generated DOM, but it should generally be backwards compatible (won't break existing layouts), and anyone using the React component will get the new styles.

packages/core/src/components/non-ideal-state/_non-ideal-state.scss
```diff
@@ -12,8 +12,10 @@ Markup:
   <div class="#{$ns}-non-ideal-state-visual">
     <span class="#{$ns}-icon #{$ns}-icon-folder-open"></span>
   </div>
-  <h4 class="#{$ns}-heading">This folder is empty</h4>
-  <div>Create a new file to populate the folder.</div>
+  <div class="#{$ns}-non-ideal-state-text">
```
adidahiya, line 15: This is another minor change to the generated DOM, but I confirmed that the styles are compatible with the old markup when both title and description are present:

image

I'll have to make a small change to make it work when there's only a title... or we might just accept the small break to the CSS API in that case (we'll just see a 10px margin instead of a 20px one iff using the old markup).

packages/core/src/components/tooltip/_tooltip.scss
```diff
@@ -30,15 +33,51 @@
     padding: $tooltip-padding-vertical $tooltip-padding-horizontal;
   }
 
+  // override typography because we use a dark background
```
adidahiya, line 36: I'm planning to implement a more comprehensive typography refactor here to make all the text styles work better inside tooltips... coming in a separate PR

packages/datetime/src/blueprint-datetime.scss
```diff
@@ -1,66 +1,8 @@
-@import "~@blueprintjs/colors/lib/scss/colors";
-
-$datepicker-day-background-color-hover: rgba($gray3, 0.15);
```
adidahiya, line 3: N.B. all these styles/variables got merged into the appropriate component files

adidahiya (commented):
there's a bug with the NonIdealState CSS API, I need to bump up the font size:

image

PR #5300, opened 2022-05-13
[popover2] feat: Popover2 popupkind

dlech:
#### Checklist

- [x] Includes tests
- [x] Update documentation

This adds a new `popupKind` property to the `Popover2` component for controlling the `aria-haspopup` attribute of the target element.

The current implementation hard-codes `aria-haspopup` to "true" which means that the popup is a menu. However, this is not true in all cases. Furthermore, the `aria-haspopup` attribute should not be set for tooltips since they are not interactive elements. So if the interaction kind is a hover interaction, this new property is ignored.

Reference: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-haspopup

#### Screenshot

Before:
image

After:
image

### DIFF
--- a/packages/popover2/src/popover2.tsx
+++ b/packages/popover2/src/popover2.tsx
@@ -51,6 +51,20 @@ export const Popover2InteractionKind = {
 // eslint-disable-next-line @typescript-eslint/no-redeclare
 export type Popover2InteractionKind = typeof Popover2InteractionKind[keyof typeof Popover2InteractionKind];
 
+/** Specifies the popup kind for aria-haspopup. */
+export enum Popover2PopupKind {
+    /** The popup is a menu. */
+    MENU = "menu",
+    /** The popup is a listbox. */
+    LISTBOX = "listbox",
+    /** The popup is a tree. */
+    TREE = "tree",
+    /** The popup is a grid. */
+    GRID = "grid",
+    /** The popup is a dialog. */
+    DIALOG = "dialog",
+}
+
 // eslint-disable-next-line deprecation/deprecation
 export type Popover2Props<TProps = React.HTMLProps<HTMLElement>> = IPopover2Props<TProps>;
 /** @deprecated use Popover2Props */
@@ -76,6 +90,17 @@ export interface IPopover2Props<TProps = React.HTMLProps<HTMLElement>> extends P
      * @default "click"
      */
     interactionKind?: Popover2InteractionKind;
+
+    /**
+     * The kind of popup displayed by the popover. This property is ignored if
+     * `interactionKind` is a hover interaction. This controls the
+     * `aria-haspopup` attribute of the target element. The default is "menu"
+     * (technically, `aria-haspopup` will be set to "true", which is the same
+     * as "menu", for backwards compatibility).
+     *
+     * @default "menu" or undefined
+     */
+    popupKind?: Popover2PopupKind;
 
     /**
      * Enables an invisible overlay beneath the popover that captures clicks and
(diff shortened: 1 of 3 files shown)


packages/popover2/src/popover2.tsx
```diff
@@ -50,14 +50,28 @@ export const Popover2InteractionKind = {
 // eslint-disable-next-line @typescript-eslint/no-redeclare
 export type Popover2InteractionKind = typeof Popover2InteractionKind[keyof typeof Popover2InteractionKind];
 
+/** Specifies the popup kind for aria-haspopup. */
+export enum Popover2PopupKind {
```
adidahiya: I think we can generalize this enum name and just call it `PopupKind`, and move it to a new file `popupKind.ts` in this package. Also please include a link to the [docs](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-haspopup) in the JSDoc for this symbol.

packages/popover2/src/popover2.tsx
```diff
@@ -50,14 +50,28 @@ export const Popover2InteractionKind = {
...
+
 // eslint-disable-next-line deprecation/deprecation
 export type Popover2Props<TProps = React.HTMLProps<HTMLElement>> = IPopover2Props<TProps>;
 /** @deprecated use Popover2Props */
 export interface IPopover2Props<TProps = React.HTMLProps<HTMLElement>> extends Popover2SharedProps<TProps> {
     /**
      * Whether the popover/tooltip should acquire application focus when it first opens.
      *
-     * @default true for click interations, false for hover interactions
+     * @default true for click interactions, false for hover interactions
```
adidahiya, line 62: thanks for fixing the typos :)

packages/popover2/src/popover2.tsx
```diff
@@ -327,7 +352,7 @@ export class Popover2<T> extends AbstractPureComponent2<Popover2Props<T>, IPopov
         // Ensure target is focusable if relevant prop enabled
         const targetTabIndex = openOnTargetFocus && isHoverInteractionKind ? 0 : undefined;
         const targetProps = {
-            "aria-haspopup": "true",
+            "aria-haspopup": this.props.popupKind ?? (isHoverInteractionKind ? undefined : "true"),
```
adidahiya: I think this should be only `undefined` if `this.props.interactionKind === Popover2InteractionKind.HOVER_TARGET_ONLY`.

in the case of `Popover2InteractionKind.HOVER` popovers, you _can_ still interact with the content, and we have instances of that in our applications. so haspopup should be set to something.

when you make this change, please make sure to update the new prop JSDoc.

packages/popover2/test/popover2Tests.tsx
```diff
@@ -192,6 +192,16 @@ describe("<Popover2>", () => {
...
             assert.isTrue(wrapper.find("[aria-haspopup='true']").exists());
         });
+
+        it("sets aria-haspopup attr base on popupKind", () => {
+            wrapper = renderPopover({ isOpen: true, popupKind: Popover2PopupKind.DIALOG });
+            assert.isTrue(wrapper.find("[aria-haspopup='dialog']").exists());
+        });
+
+        it("renders without aria-haspopup attr for hover interaction", () => {
+            wrapper = renderPopover({ isOpen: true, interactionKind: Popover2InteractionKind.HOVER });
```
adidahiya: this test will have to be updated if you make my change suggested above.

adidahiya (changes requested):

packages/popover2/src/popover2.tsx
dlech: done

packages/popover2/src/popover2.tsx
dlech: done

packages/popover2/test/popover2Tests.tsx
dlech: done

adidahiya (approved):

adidahiya:
thanks @dlech!

PR #5302, opened 2022-05-15
[popover2] fix: hide Popover2 arrow from a11y API

dlech:
The popover arrow is purely decorative and should be hidden from the accessibility DOM.

#### Checklist

- [ ] Includes tests
- [ ] Update documentation


packages/popover2/src/popover2Arrow.tsx
```diff
@@ -81,6 +81,7 @@ export interface IPopoverArrowProps {
 export const Popover2Arrow: React.FC<IPopoverArrowProps> = ({ arrowProps: { ref, style }, placement }) => (
     // data attribute allows popper.js to position the arrow
     <div
+        aria-hidden
```
adidahiya:
```suggestion
        aria-hidden={true}
```

adidahiya (changes requested):

adidahiya (approved):

PR #5305, opened 2022-05-16
[core] fix: use solid focus outline style

dlech:
Using `auto` for the outline style allows the user agent to override the outline style, leading to different results on different systems. This changes the style to solid so that all users will see the same style.

#### Checklist

- [ ] Includes tests
- [ ] Update documentation

#### Screenshot

Before:

image

After:

image

### DIFF
--- a/packages/core/src/common/_mixins.scss
+++ b/packages/core/src/common/_mixins.scss
@@ -80,7 +80,7 @@ $pt-dark-intent-text-colors: (
 }
 
 @mixin focus-outline($offset: 2px) {
-  outline: $pt-outline-color auto 2px;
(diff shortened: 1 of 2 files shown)


adidahiya:
@dlech which platforms are you testing on?

this doesn't seem to change anything for Buttons in Firefox (AnchorButtons are good, they use the solid style):

image

... but I don't see any regressions with this change, so it seems like a net positive 👍🏽

adidahiya (approved):

dlech:
I don't remember if I tried Firefox (pretty sure I did though). The screenshot is from Chromium installed from the Ubuntu snap on 22.04. So apparently the white outline is the system default selected by `auto` on my computer.

PR #5309, opened 2022-05-18
[icons] feat: new icon "floating-point"

float-icon:
#### Checklist

- [ ] Update documentation

Adds a float icon reading "1.0".

#### Screenshot

20px_float
16px_float


palantirtech:
Thanks for your interest in [palantir/blueprint](https://github.com/palantir/blueprint), @float-icon! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

packages/icons/icons.json
```diff
@@ -3883,5 +3883,12 @@
         "tags": "open, application, new-tab, external",
         "group": "interface",
         "codepoint": 62251
+    },
+    {
+        "displayName": "Float",
+        "iconName": "float",
+        "tags": "number, float, double, decimal",
+        "group": "interface",
```
adidahiya: let's put it into Miscellaneous with the other (array) data types. the icons list is due for a re-organization some time; we should have a separate category for data types...

```suggestion
        "group": "miscellaneous",
```

packages/icons/icons.json
```diff
@@ -3883,5 +3883,12 @@
         "tags": "open, application, new-tab, external",
         "group": "interface",
         "codepoint": 62251
+    },
+    {
+        "displayName": "Float",
+        "iconName": "float",
```
adidahiya: I think we should be more specific and call this "floating-point". You will need to rename the SVG files to match the iconName here after you do that.

```suggestion
        "displayName": "Floating point number",
        "iconName": "floating-point",
```

adidahiya (changes requested):
SVGs look good, but I'd like to change the metadata

float-icon:
> SVGs look good, but I'd like to change the metadata

done

adidahiya (approved):

PR #5314, opened 2022-05-23
[core] feat(Dialog): add containerRef prop

### LINKED ISSUE #5273: Dialog should allow attaching ref to container element
#### Environment

- __Package version(s)__: core v4.2.1

#### Feature request

Currently there's no easy way to get a ref to the DOM element which is inside a Dialog's Overlay, or the `.bp4-overlay` element itself. This attempt in a code sandbox doesn't work: https://codesandbox.io/s/blueprint-sandbox-forked-7fy0f9?file=/src/CoreExample.tsx

We should forward `<Dialog ref={...}>` somehow, or add a prop to `OverlayableProps` like `overlayRef?: React.Ref<HTMLElement | null>`.

dlichen:
#### Checklist

- [x] Includes tests
- [x] Update documentation

Adds a ref prop to dialog.

I'm not entirely sure if this is the correct approach, also not completely sure how to test, so any feedback is welcomed.

#### Screenshot

n/a


packages/core/src/components/dialog/dialog.tsx
```diff
@@ -80,6 +80,11 @@ export interface IDialogProps extends OverlayableProps, IBackdropProps, Props {
      */
     transitionName?: string;
 
+    /**
+     * Ref supplied to the `Classes.DIALOG_CONTAINER` element.
+     */
+    dialogRef?: IRef<HTMLDivElement>;
```
adidahiya: Let's name this containerRef:

```suggestion
    containerRef?: IRef<HTMLDivElement>;
```

adidahiya (changes requested):
Please add a unit test. You should be able to add a test for this in core/test/dialog/dialogTests.tsx... something like:

```tsx
    it("supports ref objects attached to container", () => {
        const containerRef = React.createRef<HTMLDivElement>();
        mount(
            <Dialog containerRef={containerRef} usePortal={false}>
                body
            </Dialog>,
        );
        assert.isTrue(containerRef.current?.classList.contains(Classes.DIALOG_CONTAINER));
    });
```

dlichen:
> Please add a unit test. You should be able to add a test for this in core/test/dialog/dialogTests.tsx... something like:
> 
> ```tsx
>     it("supports ref objects attached to container", () => {
>         const containerRef = React.createRef<HTMLDivElement>();
>         mount(
>             <Dialog containerRef={containerRef} usePortal={false}>
>                 body
>             </Dialog>,
>         );
>         assert.isTrue(containerRef.current?.classList.contains(Classes.DIALOG_CONTAINER));
>     });
> ```

added the test, but still need to test locally. can't seem to figure out yarn test. can hold off on reviewing

packages/core/test/dialog/dialogTests.tsx
```diff
@@ -192,6 +192,16 @@ describe("<Dialog>", () => {
...
         });
+
+        it("supports ref objects attached to container", () => {
+            const containerRef = React.createRef<HTMLDivElement>();
+            mount(
+                <Dialog containerRef={containerRef} usePortal={false}>
+                    {createDialogContents()}
+                </Dialog>,
+            );
+            assert.isTrue(containerRef.current?.classList.contains(Classes.DIALOG_CONTAINER));
```
adidahiya: you may need to use a setTimeout to wait for the lifecycle to complete and the ref object to be updated, like we do for button tests here:

https://github.com/palantir/blueprint/blob/1c7fa364e956073d8894edf01d1361fd0c07fa44/packages/core/test/buttons/buttonTests.tsx#L116-L120

packages/core/test/dialog/dialogTests.tsx
```diff
@@ -192,6 +192,18 @@ describe("<Dialog>", () => {
             // test existence here because id is generated
             assert.notExists(dialog.find(".no-default-if-no-title").hostNodes().prop("aria-labelledby"));
         });
+
+        it("supports ref objects attached to container", () => {
```
adidahiya: you didn't call `done` correctly here: since your test assertion is async, you need to use that additional test parameter (just like the button test code I linked earlier)

packages/core/test/dialog/dialogTests.tsx
```diff
@@ -192,6 +192,18 @@ describe("<Dialog>", () => {
...
+            const containerRef = React.createRef<HTMLDivElement>();
+            mount(
+                <Dialog containerRef={containerRef} usePortal={false}>
```
adidahiya: you also have to render the dialog with `isOpen={true}` in order for the container & ref to get mounted. I have a fix, will push a commit

adidahiya (approved):

Issue #4390, opened 2020-10-27
TagInput should have a string-only variant with simpler prop types


adidahiya:
Working with `onRemove: (value: React.ReactNode, index: number)` is inconvenient under TypeScript's strict compiler options. Even when the `values` prop is `string[]`, you are either forced to type cast the removed value to `string` or use the `index` parameter to index into your values array. We should make the simple case easier to work with, either with a new component API or changing the existing component to be parameterized by its data type `<T>`.

PR #5316, opened 2022-05-24
[core] docs(TagInput): remove comment about typed usage

alexdlm:
#### Checklist

- [ ] Includes tests
- [ ] Update documentation

Original comment:
> Note about typed usage: If you know your values will always be of a certain ReactNode subtype, such as string or ReactChild, you can use that type on all your handlers to simplify type logic.

Not sure that this was ever true, but I only went back 5 years. 
Alternately instead of removing this comment the props could be typed to make it true, which would certainly be an improvement.


adidahiya (approved):
I think this comment became irrelevant when the TypeScript compiler tightened up its type checks. There's definitely room for improvement in this component when it comes to a well-typed API. See https://github.com/palantir/blueprint/issues/4390

PR #5322, opened 2022-05-25
[datetime2] feat: new package with timezone-aware components

vhellem:
> edited by @adidahiya 

#### Checklist

- [x] Includes tests
- [x] Update documentation

- Create a new package `@blueprintjs/datetime`, which includes two new components:
  - DateInput2 (replacement for DateInput from `@blueprintjs/datetime`)
  - TimezoneSelect (replacement for TimezonePicker from `@blueprintjs/timezone`)
- Refer to new documentation for more info about these components and the package motivation

- [x] API for date input component

#### Not done in this PR

These new components are going to require a decent amount of iteration in real products before they are considered good from a UX/DX perspective and their APIs stable. As such, there are a number of improvements which we will not squeeze into this initial PR but leave for future PRs. These include:

1. comparison of date-fns library with alternatives like Luxon or the [temporal-polyfill](https://github.com/js-temporal/temporal-polyfill) ([see Moment.js project status post](https://momentjs.com/docs/#/-project-status/))
2. performance bottleneck with [formatInTimeZone](https://github.com/marnusw/date-fns-tz#formatintimezone) (https://github.com/marnusw/date-fns-tz#formatintimezone)
3. a smooth migration API from `<DateInput>`, which currently accepts a `value: Date`
    - there's a clear migration API we can provide here on the new `<DateInput2>` where we accept a `Date` _or_ an ISO string, it will just require a bit more code
    - we should probably set `hideTimezoneSelect: true` by default
    - I'm confident that we can do it, and it's important that we do this so that consumers can quickly move over to the new component which uses Popover2, if they want to prioritize that migration over the timezone-awareness improvements in the new package
5. adding DateInput to the `no-deprecated-components` rule to encourage migration -- this should be done after the item 4 is addressed

#### Screenshot

image
image


adidahiya:
The new timezone picker has much worse performance than the old one:

| TimezonePicker | TimezonePickerV2 |
| - | - |
|  |  |

we'll need to dig into this with browser performance profiling...

vhellem:
> 

Done a number of improvements. Main drawback was probably that we rendered all items initially vs just a subset list, which I've added. Also done a lot more preprocessing.

packages/datetime/src/datePicker.tsx
```diff
@@ -143,13 +143,14 @@ export class DatePicker extends AbstractPureComponent2<DatePickerProps, IDatePic
...
-        const { className, dayPickerProps, locale, localeUtils, maxDate, minDate, showActionsBar } = this.props;
+        const { className, dayPickerProps, footerComponent, locale, localeUtils, maxDate, minDate, showActionsBar } =
+            this.props;
         const { displayMonth, displayYear } = this.state;
 
         return (
             <div className={classNames(Classes.DATEPICKER, className)}>
                 {this.maybeRenderShortcuts()}
-                <div>
+                <div className={Classes.DATEPICKER_CONTENT}>
```
adidahiya, line 153: note to self: consider a different name for this class, like "datepicker main panel" or something

packages/datetime/src/datePickerCore.tsx
```diff
@@ -50,6 +50,11 @@ export interface IDatePickerBaseProps {
      */
     dayPickerProps?: DayPickerProps;
 
+    /**
+     * An additional component to show below the date picker
+     */
+    footerComponent?: JSX.Element;
```
adidahiya: this isn't a component, it's an element:

```suggestion
    footerElement?: JSX.Element;
```

comment should be updated to match

packages/docs-app/src/examples/datetime-examples/datePickerExample.tsx
```diff
@@ -50,6 +52,16 @@ export class DatePickerExample extends React.PureComponent<IExampleProps, IDateP
 
     private toggleActionsBar = handleBooleanChange(showActionsBar => this.setState({ showActionsBar }));
 
+    private toggleFooterComponent = () => {
+        this.setState(prevState => {
+            const footerComponent =
+                prevState.footerComponent !== undefined ? undefined : (
+                    <Callout>This additional footer component can be displayed below the date picker</Callout>
```
adidahiya: this implementation is a little convoluted. you can just store a constant outside this component which is the sample footer element

```tsx
const exampleFooterElement = <Callout>This additional footer element can be displayed below the date picker</Callout>
```

store a boolean toggle flag instead of `footerComponent` in state

and then set the DatePicker prop:

```tsx
<DatePicker
    // ...
    footerElement={this.showFooterElement ? exampleFooterElement : undefined}
/>
```

packages/docs-app/src/examples/datetime-examples/datePickerExample.tsx
```diff
@@ -83,6 +95,11 @@ export class DatePickerExample extends React.PureComponent<IExampleProps, IDateP
                     label="Reverse month and year menus"
                     onChange={this.toggleReverseMenus}
                 />
+                <Switch
+                    checked={this.state.footerComponent !== undefined}
+                    label="Show a footer component"
```
adidahiya:
```suggestion
                    label="Show custom footer element"
```

packages/docs-app/src/examples/datetime-examples/datePickerExample.tsx
```diff
@@ -101,6 +118,12 @@ export class DatePickerExample extends React.PureComponent<IExampleProps, IDateP
                     label="Use AM/PM"
                     onChange={this.toggleUseAmPm}
                 />
+                <Switch
+                    disabled={!showTimePicker}
+                    checked={this.state.useAmPm}
+                    label="Use AM/PM"
+                    onChange={this.toggleUseAmPm}
+                />
```
adidahiya: remove duplicated code

```suggestion
```

packages/datetime-timezone-aware/src/components/timezone-picker/timezone-picker-v2.md
```diff
@@ -0,0 +1,51 @@
+@# TimezonePickerV2
+
+`TimezonePickerV2` allows the user to select from a list of timezones. The list of times is coded into the library itself, and it does not depend on any external packages for updating the this list of timezones.
```
adidahiya:
```suggestion
`TimezonePickerV2` allows the user to select from a list of timezones. The list is coded into the library itself, so it does not depend on any external packages for the list of timezones.
```

packages/datetime-timezone-aware/src/components/timezone-picker/timezone-picker-v2.md
```diff
@@ -0,0 +1,51 @@
...
+<TimezonePicker value={...} onChange={...}>
+    <Icon icon="globe" />
+</TimezontPicker>
+```
+
+<div class="@ns-callout @ns-intent-warning @ns-icon-warning-sign">
+    <h4 class="@ns-heading">Local timezone detection</h4>
+
+We detect the local timezone when the `showLocalTimezone` prop is enabled and cannot guarantee correctness in all browsers.
+In supported browsers, the [i18n API](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/DateTimeFormat/resolvedOptions) is used.
```
adidahiya: we don't need to reference outdated browsers. the only one that doesn't support this part of the i18n API is IE11, and we are dropping support for that in Blueprint v5. Since this is a new package, we can mention that IE11 is not supported with this package. so this documentation can then just say:

```suggestion
We use the [browser `Intl` API](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/DateTimeFormat/resolvedOptions).
```

packages/datetime-timezone-aware/src/components/timezone-picker/timezone-picker-v2.md
```diff
@@ -0,0 +1,51 @@
...
+In other browsers, `Date` methods and a population heuristic are used.
```
adidahiya:
```suggestion
```

adidahiya (dismissed):
(1) The package name "datetime-timezone-aware" is fairly verbose, I’d like to avoid that. I think we can just go with "datetime2" and make things easier for consumers, with the assumption that many of our users do care about timezones and want their datetime UIs to be timezone-aware. The new dependencies [jstimezonedetect](https://bundlephobia.com/package/jstimezonedetect@1.0.7), [date-fns-tz](https://bundlephobia.com/package/date-fns-tz@1.3.4), and [date-fns](https://bundlephobia.com/package/date-fns@2.28.0) are not too large, especially with bundle modularity & tree-shaking applied.

Then we would also drop “timezone aware” from the component names... `TimezoneAwareDateInput` becomes simply `DateInput2`.

(2) Note that I’m getting this warning in the console:

```
Warning: Encountered two children with the same key, America/New_York. Keys should be unique so that components maintain their identity across updates. Non-unique keys may cause children to be duplicated and/or omitted — the behavior is unsupported and could change in a future version.
```

(3) It would be great if the TimezonePicker2 example could display the offset in the button like TimezonePicker does:

image

(4) I’m noticing the timezone-aware date input example has similar performance issues like I noted earlier with TimezonePicker2, we should fix this:



(5) API feedback... I think we should support controlled usage of `timezone` in addition to uncontrolled usage via `defaultTimezone`... a user may want to _always_ show the picker in a certain timezone. And we should allow both _hiding_ and _disabling_ of the timezone picker (`disableTimezonePicker` + `hideTimezonePicker`). Otherwise, this overall API seems ok, using ISO strings only instead of Date values. We will have to add some props validation after adding a `timezone` prop... for example if the controlled `timezone` prop value doesn’t match the timezone offset present in the `value` ISO string, we should log an error and choose one to take precedence over the other.

(6) I don’t love the fact that this component has two popovers. I generally try to discourage multi-overlay UIs wherever possible. But this approach is fine as a first draft. I can take a stab at improving the embedded timezone selector UI after this PR is merged.

(7) The timezone tag and calendar icon appear interactive, but they don’t trigger anything on click. Clicking the timezone tag should go directly to the timezone select. The calendar button feels superfluous, it could just be a `leftIcon` on the InputGroup, I think.

packages/datetime/src/datePicker.tsx
vhellem: Happy to update it myself as well?

packages/datetime-timezone-aware/src/components/timezone-picker/timezone-picker-v2.md
vhellem: Done, also removed `jstz` as what was what was being used to guess

vhellem:
Thanks so much for the review! A couple follow up questions

> (1) The package name "datetime-timezone-aware" is fairly verbose, I’d like to avoid that. I think we can just go with "datetime2" and make things easier for consumers,

Can do, although it feel's slightly weird to me that `datetime2` would depent on `datetime`? Although I guess we may eventually want to replace more of those? 

> (4) I’m noticing the timezone-aware date input example has similar performance issues like I noted earlier with TimezonePicker2, we should fix this:

Hmm, I can't really repro this here, as things are quite fast on my end. I imagine it has something to do with the initial render. The below suggestion of moving it to the right element may solve it. 

> (5) API feedback... I think we should support controlled usage of `timezone` in addition to uncontrolled usage via `defaultTimezone`... a user may want to _always_ show the picker in a certain timezone. And we should allow both _hiding_ and _disabling_ of the timezone picker (`disableTimezonePicker` + `hideTimezonePicker`). Otherwise, this overall API seems ok, using ISO strings only instead of Date values. We will have to add some props validation after adding a `timezone` prop... for example if the controlled `timezone` prop value doesn’t match the timezone offset present in the `value` ISO string, we should log an error and choose one to take precedence over the other.

Are you speaking specifically about the `DateInput` or the `TimezonePicker` now? We support borth hiding and disabling of the timezone picker in the `DateInput` component, so currently there's a way to achieve all of these by hiding/disabling the timezone picker and then passing `defaultTimezone`. I'm not sure if it's worth passing in an additional `timezoneValue` in addition to this, as it doesn't feel like what we want to control here? Could add it though, it just feel's like a place where we might want to be more opinionated about how it's being used. 


> (7) The timezone tag and calendar icon appear interactive, but they don’t trigger anything on click. Clicking the timezone tag should go directly to the timezone select. The calendar button feels superfluous, it could just be a `leftIcon` on the InputGroup, I think.

Removed the calendar all together. I see now that the `dateInput` component doesn't trigger an opening on the right elements, only on input focus itself. What do you think about rendering the timezone-picker itself as the right element? I think that would solve the performance issues (I'm struggling a bit to repro that),Then we could potentially just remove it from within the time picker itself as well to avoid having two popovers. Some styling changes would probably be needed

adidahiya:
The new timezone picker target in DateInput2 looks good!

(1) I find myself selecting the timezone text accidentally too often. I think we should disable that with `user-select: none`

image

(2) "disabling" the timezone picker prevents it from popping up, but it still appears interactive. I think you need to set `disabled` on the Tag or apply some custom styling:



(3) Disabling the overall DateInput2 does not disable the timezone picker; this is a bug:



(4) as for my earlier request about setting `timezone` in a controlled manner instead of only allowing `defaultTimezone`... if you think that's not needed for now, I'm more than happy to punt it for a future PR and keep this one smaller

PR #5325, opened 2022-05-25
[core] fix(Toast): add alert role to toast message for a11y

### LINKED ISSUE #5324: add role="alert" to toast messages for better a11y
#### Environment

#### Feature request

add `role="alert"` to toast messages for a11y purposes.

bvandrc:
#### Checklist

- [N/A] Includes tests
- [N/A] Update documentation

add `role="alert"` to toast message for a11y purposes.

### DIFF
--- a/packages/core/src/components/toast/toast.tsx
+++ b/packages/core/src/components/toast/toast.tsx
@@ -72,11 +72,12 @@ export class Toast extends AbstractPureComponent2<IToastProps> {
                 onFocus={this.clearTimeouts}
                 onMouseEnter={this.clearTimeouts}
                 onMouseLeave={this.startTimeout}
-                role="alert"
                 tabIndex={0}
             >
                 <Icon icon={icon} />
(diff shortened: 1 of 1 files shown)


bvandrc:
issue https://github.com/palantir/blueprint/issues/5324

adidahiya:
I think this should only be applied to the toast message, according to the [docs](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/alert_role):

> The alert role should only be used for text content, not interactive elements such as links or buttons.

bvandrc:
> I think this should only be applied to the toast message, according to the [docs](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/alert_role):
> 
> > The alert role should only be used for text content, not interactive elements such as links or buttons.

updated. or would you prefer to kill this PR and have the user pass say, `<span role="alert">Users Message</span>` as the `message` themselves?

adidahiya (approved):
lgtm

Issue #4221, opened 2020-07-10
`yarn verify` throw an error on Windows 10, docs-data package fails to build


lmk123:
#### Environment

- __Operating System__: Windows 10
- __Node version__:  v12.18.1
- __Npm version__: v6.14.5
- __Yarn version__: v1.22.4

#### Code Sandbox

Link to a minimal repro: https://github.com/lmk123/blueprint/tree/patch-1

#### Steps to reproduce

1. Run `yarn global add windows-build-tools` in PowerShell
2. Run `git clone https://github.com/lmk123/blueprint.git` and checkout `patch-1` branch
3. Run `yarn install` in repo
4. Run `npm config set script-shell "C:\\Program Files\\git\\bin\\bash.exe"`
5. Run `yarn verify`

#### Actual behavior

`yarn verify` throw an error whin it run into `node ./compile-docs-data.js`:

```bash
C:\Users\Mica\WebstormProjects\blueprint>yarn verify
yarn run v1.22.4
$ npm-run-all -s compile dist:libs dist:apps -p test lint format-check
$ lerna run compile
lerna notice cli v3.22.0
lerna info versioning independent
lerna info Executing command in 11 packages: "yarn run compile"
lerna info run Ran npm script 'compile' in '@blueprintjs/test-commons' in 4.2s:
$ tsc -p ./src
lerna info run Ran npm script 'compile' in '@blueprintjs/tslint-config' in 4.9s:
$ tsc -p ./src
lerna info run Ran npm script 'compile' in '@blueprintjs/eslint-plugin' in 5.0s:
$ tsc -p src/
lerna info run Ran npm script 'compile' in '@blueprintjs/icons' in 6.8s:
$ npm-run-all -s "compile:pre" -p "compile:*"
$ generate-icons-source
$ generate-icons-source
$ tsc -p ./src
$ tsc -p ./src -m commonjs --outDir lib/cjs
$ tsc -p ./src -t esnext --outDir lib/esnext
$ sass-compile ./src
Rendering Complete, saving .css file...
Wrote CSS to C:\Users\Mica\WebstormProjects\blueprint\packages\icons\lib\css\blueprint-icons.css
Wrote Source Map to C:\Users\Mica\WebstormProjects\blueprint\packages\icons\lib\css\blueprint-icons.css.map
lerna info run Ran npm script 'compile' in '@blueprintjs/core' in 8.1s:
$ run-p "compile:*"
$ tsc -p ./src
[...]

PR #5326, opened 2022-05-25
[core] fix(Menu): add a11y roles and aria attributes

### LINKED ISSUE #5320: Need ability to modify the li props in a MenuItem in order to create a proper a11y menu
#### Environment

#### Feature request

Need a way to pass props to `MenuItem` that get assigned to the `li` element rather than being applied to the inner `a` element. Per the a11y menu example here: https://w3c.github.io/aria-practices/examples/menubar/menubar-navigation.html , I need to be able to set `<li role="none">`. The reasons are also given in this page:
- `<li role="none">` Removes the implied listitem role of the li element.
- Necessary because the parent ul is serving as a menu so the li elements are not in their required list context.

#### Examples

See "HTML Source Code" section of this page: https://w3c.github.io/aria-practices/examples/menubar/menubar-navigation.html

bvandrc:
#### Checklist

- [N/A ] Includes tests
- [N/A] Update documentation

Fix `Menu` and `MenuItem` a11y roles for a menu per this example: https://w3c.github.io/aria-practices/examples/menubar/menubar-navigation.html

the above example aria menu

### DIFF
--- a/packages/core/src/components/menu/menuItem.tsx
+++ b/packages/core/src/components/menu/menuItem.tsx
@@ -177,7 +177,6 @@ export class MenuItem extends AbstractPureComponent2<MenuItemProps & React.Ancho
         const target = React.createElement(
             tagName,
             {
-                "aria-expanded": hasSubmenu,
                 role: "menuitem",
                 tabIndex: 0,
                 ...htmlProps,
--- a/packages/core/src/components/menu/menuItem.tsx
+++ b/packages/core/src/components/menu/menuItem.tsx
@@ -177,6 +177,7 @@ export class MenuItem extends AbstractPureComponent2<MenuItemProps & React.Ancho
         const target = React.createElement(
             tagName,
             {
+                role: "menuitem",
                 tabIndex: 0,
                 ...htmlProps,
                 ...(disabled ? DISABLED_PROPS : {}),
--- a/packages/core/src/components/menu/menuItem.tsx
+++ b/packages/core/src/components/menu/menuItem.tsx
@@ -177,6 +177,7 @@ export class MenuItem extends AbstractPureComponent2<MenuItemProps & React.Ancho
         const target = React.createElement(
             tagName,
             {
+                "aria-expanded": hasSubmenu,
                 role: "menuitem",
                 tabIndex: 0,
                 ...htmlProps,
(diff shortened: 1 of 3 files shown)


bvandrc:
I am on Windows 10 and am getting errors when running `yarn verify`/`yarn compile`, and cannot run `yarn test` either-- can one of the Devs please check out this branch and run the test I've written plus whatever else needs to be done?

adidahiya:
Thanks for your contributions, and sorry about the trouble building the project on Windows. Could you share some of your build errors in this issue? Other contributors have had similar problems: https://github.com/palantir/blueprint/issues/4221

packages/core/src/components/menu/menuItem.tsx
```diff
@@ -177,6 +177,7 @@ export class MenuItem extends AbstractPureComponent2<MenuItemProps & React.Ancho
         const target = React.createElement(
             tagName,
             {
+                role: "menuitem",
```
adidahiya, line 180: let's also add `"aria-expanded": hasSubmenu` here as suggested in the linked [docs](https://w3c.github.io/aria-practices/examples/menubar/menubar-navigation.html).

```suggestion
                "aria-expanded": hasSubmenu,
                role: "menuitem",
```

there's a separate issue where `aria-hassubmenu` is being added to the wrong element (the wrapping `<span>` of the popover target rather than the `<a role="menuitem">` element), but that's a problem I will fix later when I create a MenuItem2 variant which uses Popover2 instead of Popover:

image

adidahiya (changes requested):
mostly looks good, just one small comment

packages/core/src/components/menu/menuItem.tsx
adidahiya: oops, this is wrong, sorry, "expanded" state is different from simply _having_ a submenu. setting "aria-expanded" will not be this simple. I will revert my suggestion.

packages/core/src/components/menu/menuItem.tsx
```diff
@@ -177,6 +177,8 @@ export class MenuItem extends AbstractPureComponent2<MenuItemProps & React.Ancho
         const target = React.createElement(
             tagName,
             {
+                "aria-expanded": hasSubmenu,
```
```suggestion
```

adidahiya (approved):

PR #5336, opened 2022-05-31
[core] feat(MenuItem): forward MenuProps to submenu

### LINKED ISSUE #5335: need ability to pass custom props to the Menu created in the Menu popover of a MenuItem
#### Environment

#### Feature request

For A11y purposes, need ability to pass `aria-label` prop to the Menu that gets created in the Popover when a MenuItem is hovered over

bvandrc:
#### Checklist

- [X] Includes tests
- [X] Update documentation

Allow `menuProps` to be passed to `MenuItem` that get forwarded to the `Menu` created within the `Popover` button.


packages/core/src/components/menu/menuItem.tsx
```diff
@@ -77,6 +77,11 @@ export interface IMenuItemProps extends ActionProps, LinkProps {
      */
     labelElement?: React.ReactNode;
 
+    /**
+     * Props to spread to the `Menu` within the `Popover`.
+     */
+    menuProps?: Partial<MenuProps> & object;
```
adidahiya: we don't need `& object`... that's a weird quirk from an older version of TypeScript that's no longer necessary.

packages/core/src/components/menu/menuItem.tsx
```diff
@@ -126,6 +131,7 @@ export class MenuItem extends AbstractPureComponent2<MenuItemProps & React.Ancho
     public static defaultProps: MenuItemProps = {
         active: false,
         disabled: false,
+        menuProps: {},
```
adidahiya: I don't think we need this

```suggestion
```

packages/core/src/components/menu/menuItem.tsx
```diff
@@ -77,6 +77,11 @@ export interface IMenuItemProps extends ActionProps, LinkProps {
      */
     labelElement?: React.ReactNode;
 
+    /**
+     * Props to spread to the `Menu` within the `Popover`.
```
```suggestion
     * Props to spread to the child `Menu` component if this item has a submenu.
```

packages/core/src/components/menu/menuItem.tsx
adidahiya: also I think we should call this prop `submenuProps`

```suggestion
    submenuProps?: Partial<MenuProps>;
```

adidahiya (changes requested):

packages/core/src/components/menu/menuItem.tsx
bvandrc: I agree with the name change.

So the issue with just `Partial<MenuProps>` is that `MenuProps` doesn't have the commonly used prop `data-testid`. Any Ideas on how we can get that in there?

adidahiya: hmm, there's no good way to do that in the object type. `data-*` attributes work in JSX because they're built into React. I guess you can leave the `& object` in this type and leave a note in the JSDoc about why it's there.

bvandrc: How would you describe that? Like this?
    /**
     * Props to spread to the child `Menu` component if this item has a submenu.
     * `& object` allows for `data-*` attributes to be passed.
     */

packages/core/src/components/menu/menuItem.tsx
```diff
@@ -77,6 +77,11 @@ export interface IMenuItemProps extends ActionProps, LinkProps {
      */
     labelElement?: React.ReactNode;
 
+    /**
+     * Props to spread to the child `Menu` component if this item has a submenu.
+     */
+    menuProps?: Partial<MenuProps> & object;
```
adidahiya: actually, it's an implementation detail, so it should be a code comment for devs rather than JSDoc:

```suggestion
     */
     // note that `& object` is required to allow attributes built-in to JSX like "data-testid"
    submenuProps?: Partial<MenuProps> & object;
```

bvandrc: I'm just going to remove the `& object`, that actually didn't solve it. Just  ` Partial<MenuProps>` is fine.

adidahiya (approved):

PR #5338, opened 2022-05-31
a11y tooltip role

bvandrc:
#### Checklist

- [ ] Includes tests
- [ ] Update documentation

add `role="tooltip"` to tooltip and tooltip2 for a11y purposes. Wanted to see if the devs would support this change. If not, no worries, users of blueprint can always manually include this role in the child they pass to these elements.

### DIFF
--- a/packages/core/src/components/tooltip/tooltip.tsx
+++ b/packages/core/src/components/tooltip/tooltip.tsx
@@ -107,7 +107,7 @@ export class Tooltip extends AbstractPureComponent2<TooltipProps> {
                 portalContainer={this.props.portalContainer}
                 ref={ref => (this.popover = ref)}
             >
-                {children}
+                <div role="tooltip">{children}</div>
             </Popover>
         );
     }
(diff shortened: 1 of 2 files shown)


bvandercar-vt:
Closing this PR. Given that `tooltip2tests` has:

```
<Tooltip2 content={<p>Text</p>} hoverOpenDelay={0} {...props} usePortal={false}>
                <Button id={TEST_TARGET_ID} text="target" />
            </Tooltip2>,
```

It seems that users may sometimes desire to have an interactive element within their tooltip, and per the rules of the role (https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/tooltip_role) :

"Because the tooltip itself never receives focus and is not in the tabbing order, a tooltip can not contain interactive elements like links, inputs, or buttons."

So, probably best to keep the role out of their by default in case a user really wants to have interactive elements. They can always add `role="tooltip"` to their child themselves.

bvandrc:
new PR: https://github.com/palantir/blueprint/pull/5339

PR #5339, opened 2022-05-31
Bv/tooltip role

bvandrc:
#### Checklist

- [ ] Includes tests
- [ ] Update documentation

add `role="tooltip"` to tooltip and tooltip2 for a11y purposes. Wanted to see if the devs would agree with / support this change. If not, no worries, BP users can always manually include this role in the tooltip contents they pass.

### DIFF
--- a/packages/core/src/components/tooltip/tooltip.tsx
+++ b/packages/core/src/components/tooltip/tooltip.tsx
@@ -108,7 +108,7 @@ export class Tooltip extends AbstractPureComponent2<TooltipProps> {
                 portalContainer={this.props.portalContainer}
                 ref={ref => (this.popover = ref)}
             >
-                children
+                {children}
             </Popover>
         );
     }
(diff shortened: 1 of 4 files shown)


packages/core/src/components/tooltip/tooltip.tsx
```diff
@@ -101,6 +101,7 @@ export class Tooltip extends AbstractPureComponent2<TooltipProps> {
                 {...restProps}
                 autoFocus={false}
                 canEscapeKeyClose={false}
+                content={isContentEmpty(content) ? content : <div role="tooltip">{content}</div>}
```
adidahiya, line 104: this is a breaking change to DOM layout, and I'd like to avoid extra wrapper elements where possible (note that one of the big changes in Popover2 is the removal of a wrapper element around the target). If we can make this a11y improvement without adding an extra element, then I'm open to it, but otherwise I would suggest that consumers add the `role` themselves.

adidahiya (changes requested):

packages/core/src/components/tooltip/tooltip.tsx
bvandrc: I don't think there's any way to do it without adding an extra element. Plust I also realized, a user may want to add more props to the `tooltip` div. 

Can close this PR.

PR #5343, opened 2022-06-01
[eslint-plugin] feat: no-deprecated-components rule

styu:
#### Checklist

- [x] Includes tests
- [ ] Update documentation

Adds an ESLint rule that flags particular deprecated blueprint components and recommends their v2 versions instead

What other components am I missing?


packages/eslint-plugin/src/rules/no-deprecated-components.ts
```diff
@@ -0,0 +1,167 @@
...
+import { TSESTree } from "@typescript-eslint/utils";
+
+import { createRule } from "./utils/createRule";
+
+const DEPRECATED_TO_NEW_MAPPING: { [deprecated: string]: string } = {
+    AbstractComponent: "AbstractComponent2",
+    AbstractPureComponent: "AbstractPureComponent2",
+    CollapsibleList: "OverflowList",
+    Popover: "Popover2",
+    Tooltip: "Tooltip2",
```
adidahiya:
```suggestion
    MultiSelect: "MultiSelect2",
    PanelStack: "PanelStack2",
    Popover: "Popover2",
    Select: "Select2",
    Suggest: "Suggest2",
    Tooltip: "Tooltip2",
```

packages/eslint-plugin/src/rules/no-deprecated-components.ts
```diff
@@ -0,0 +1,167 @@
...
+                deprecatedImports.some(
+                    deprecatedImport => deprecatedImport.type === "namespace" && deprecatedImport.namespace === name,
+                ) && DEPRECATED_TO_NEW_MAPPING[property] != null
+            );
+        }
+
+        // Get the list of all deprecated imports from blueprint
+        return {
+            "Program > ImportDeclaration": (node: TSESTree.ImportDeclaration) => {
+                if (node.source.value !== "@blueprintjs/core") {
```
adidahiya: This needs to check for other packages as well, like `@blueprintjs/select`. We could maintain a list of packages which have deprecated exports at the top of this file.

packages/eslint-plugin/test/no-deprecated-components.test.ts
```diff
@@ -0,0 +1,118 @@
...
+                },
+            ],
+        },
+        {
+            code: dedent`
+                import * as BP from "@blueprintjs/core";
+
+                export class MyClass extends BP.AbstractComponent {
+                }
+                class MyClass2 extends BP.AbstractComponent {}
```
adidahiya, line 78: looks like two errors should be expected here instead of just one

adidahiya (changes requested):

packages/eslint-plugin/test/no-deprecated-components.test.ts
styu: hm.. it passes both if i pass 1 error or 2 errors below 🤔

packages/eslint-plugin/src/rules/no-deprecated-components.ts
```diff
@@ -0,0 +1,172 @@
...
+    MultiSelect: "MultiSelect2",
+    PanelStack: "PanelStack2",
+    Popover: "Popover2",
+    Select: "Select2",
+    Suggest: "Suggest2",
+    Tooltip: "Tooltip2",
+};
+const PACKAGES_WITH_DEPRECATED_IMPORTS = ["@blueprintjs/core", "@blueprintjs/select"];
+
+type MessageIds = "nonDeprecated" | "deprecated";
```
adidahiya: I don't think we need two message IDs, if only one ("nonDeprecated") is used.

packages/eslint-plugin/src/rules/no-deprecated-components.ts
```diff
@@ -0,0 +1,172 @@
...
+
+// tslint:disable object-literal-sort-keys
+export const noDeprecatedComponentsRule = createRule<unknown[], MessageIds>({
+    name: "no-deprecated-components",
+    meta: {
+        type: "suggestion",
+        docs: {
+            description: "Recommends using non-deprecated versions of the component or constant instead",
```
adidahiya:
```suggestion
            description: "Reports on usage of deprecated Blueprint components and recommends migrating to their corresponding newer non-deprecated API alternatives.",
```

packages/eslint-plugin/src/rules/no-deprecated-components.ts
```diff
@@ -0,0 +1,172 @@
...
+            requiresTypeChecking: false,
+            recommended: "error",
+        },
+        messages: {
+            nonDeprecated: "Replace deprecated component {{ deprecated }} with {{ nonDeprecated }}",
```
adidahiya:
```suggestion
            nonDeprecated: "Replace usage of deprecated component {{ deprecated }} with {{ nonDeprecated }}",
```

packages/eslint-plugin/src/rules/no-deprecated-components.ts
```diff
@@ -0,0 +1,172 @@
...
+    Select: "Select2",
+    Suggest: "Suggest2",
+    Tooltip: "Tooltip2",
+};
+const PACKAGES_WITH_DEPRECATED_IMPORTS = ["@blueprintjs/core", "@blueprintjs/select"];
+
+type MessageIds = "nonDeprecated" | "deprecated";
+
+// tslint:disable object-literal-sort-keys
+export const noDeprecatedComponentsRule = createRule<unknown[], MessageIds>({
```
adidahiya, line 31: I think there should be some mention here (even if just for devs inspecting the rule source) of the fact that we're not actually reading the `@deprecated` JSDoc annotation, and thus the rule is quite distinct from the `deprecation/deprecation` ESLint rule.

adidahiya (changes requested):

packages/eslint-plugin/src/rules/no-deprecated-components.ts
styu: ah yeah my original thought was if we had components that were removed with no replacement, but i didn't end up capturing those here

adidahiya (approved):

adidahiya:
thanks @styu!

PR #5346, opened 2022-06-02
[select] feat(Select2, MultiSelect2): popoverContentProps

bvandrc:
#### Checklist

- [ ] Includes tests
- [ ] Update documentation

allow `popoverContentProps` to be passed to `MultiSelect2` and `Select2`. Allows user to apply props such as `aria-label` or `data-testid` to the popover that is created by these elements.

### DIFF
--- a/packages/select/src/components/multi-select/multiSelect2.tsx
+++ b/packages/select/src/components/multi-select/multiSelect2.tsx
@@ -73,12 +73,15 @@ export interface MultiSelect2Props<T> extends IListItemsProps<T> {
      */
     placeholder?: string;
 
+    /** Props to spread to `Popover2` popover content. */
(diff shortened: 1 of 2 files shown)


adidahiya (approved):
nice, clean 👍🏽

adidahiya:
I'm adding the same for Suggest2 here: https://github.com/palantir/blueprint/pull/5347

bvandrc:
@adidahiya why was `PopoverContentProps` removed from the props interfaces after this was merged? We need to re-add those

adidahiya:
@bvandercar-vt it's inherited from here in the shared interface: https://github.com/palantir/blueprint/blob/7ea317e054565a23dc4407f7feb5bef4553c195c/packages/select/src/common/selectPopoverProps.ts#L25

PR #5347, opened 2022-06-02
[select] standardize shared popover props

adidahiya:

- :x: BREAK: move MultiSelect2 popover ref prop
  - before: `<MultiSelect2 popoverProps={ ref: ... }>`
  - after: `<MultiSelect2 popoverRef={...}>`
- Add `popoverContentProps` for `<Suggest2>`, following up from https://github.com/palantir/blueprint/pull/5346
- Move shared props to a shared interface for DRYness

No regressions

### DIFF
--- a/packages/select/src/common/selectPopoverProps.ts
+++ b/packages/select/src/common/selectPopoverProps.ts
@@ -1,0 +1,38 @@
+/*
+ * Copyright 2022 Palantir Technologies, Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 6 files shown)


packages/select/src/common/selectPopoverProps.ts
```diff
@@ -0,0 +1,38 @@
...
+    /** Props to spread to `Popover2` content wrapper eleemnt. */
+    popoverContentProps?: React.HTMLAttributes<HTMLDivElement>;
+
+    /**
+     * Props to spread to `Popover2`.
+     * Note that `content` cannot be changed aside from utilizing `popoverContentProps`.
+     */
+    popoverProps?: Partial<Omit<Popover2Props, "content">>;
+
+    /*
```
adidahiya:
```suggestion
    /**
```

PR #5348, opened 2022-06-02
[select] fix: ARIA listbox popup type and combobox list role

bvandrc:
#### Checklist

- [ ] Includes tests
- [ ] Update documentation

Apply `role="listbox"` for the `Menu` created by `QueryList`-- `Menu` has `role="menu"` by default, and `listbox` is more applicable here. Also apply `role="combobox"` to the item that creates the `listbox` popover

Trying to model after https://www.digitala11y.com/combobox-role/ for Select2 and MultiSelect2, https://www.w3.org/TR/2021/NOTE-wai-aria-practices-1.2-20211129/examples/combobox/combobox-autocomplete-list.html for Suggest2.

with `menuProps`, user can pass a custom `id` (or other props) to the `Menu` created within the Select/Multiselect popover.

### DIFF
--- a/packages/select/src/components/multi-select/multiSelect2.tsx
+++ b/packages/select/src/components/multi-select/multiSelect2.tsx
@@ -29,7 +29,7 @@ import {
     TagInputProps,
 } from "@blueprintjs/core";
 import { Popover2 } from "@blueprintjs/popover2";
-import { PopupKind } from "@blueprintjs/popover2/popupKind";
+import { PopupKind } from "@blueprintjs/popover2/src/popupKind";
 
 import { Classes, IListItemsProps, SelectPopoverProps } from "../../common";
 import { IQueryListRendererProps, QueryList } from "../query-list/queryList";
--- a/packages/select/src/components/multi-select/multiSelect2.tsx
+++ b/packages/select/src/components/multi-select/multiSelect2.tsx
@@ -204,7 +204,7 @@ export class MultiSelect2<T> extends AbstractPureComponent2<MultiSelect2Props<T>
                 }
             >
                 <div
-                    {...wrapperProps}
+                    {...popoverTargetProps}
                     aria-expanded={this.state.isOpen}
                     onKeyDown={this.getTagInputKeyDownHandler(handleKeyDown)}
                     onKeyUp={this.getTagInputKeyUpHandler(handleKeyUp)}
--- a/packages/select/src/components/multi-select/multiSelect2.tsx
+++ b/packages/select/src/components/multi-select/multiSelect2.tsx
@@ -83,6 +83,11 @@ export interface MultiSelect2Props<T> extends IListItemsProps<T>, SelectPopoverP
 
     /** Custom renderer to transform an item into tag content. */
     tagRenderer: (item: T) => React.ReactNode;
+
+    /**
+     * Props to add to the `div` that wraps the TagInput
+     */
+    wrapperProps?: React.HTMLAttributes<HTMLDivElement>;
 }
 
 export interface MultiSelect2State {
@@ -153,6 +158,7 @@ export class MultiSelect2<T> extends AbstractPureComponent2<MultiSelect2Props<T>
             popoverRef,
             selectedItems = [],
             placeholder,
+            wrapperProps = {},
         } = this.props;
         const { handlePaste, handleKeyDown, handleKeyUp } = listProps;
 
@@ -199,13 +205,15 @@ export class MultiSelect2<T> extends AbstractPureComponent2<MultiSelect2Props<T>
                 }
             >
                 <div
+                    {...wrapperProps}
+                    aria-expanded={this.state.isOpen}
                     onKeyDown={this.getTagInputKeyDownHandler(handleKeyDown)}
                     onKeyUp={this.getTagInputKeyUpHandler(handleKeyUp)}
+                    role="combobox"
                 >
                     <TagInput
                         placeholder={placeholder}
                         {...tagInputProps}
-                        aria-expanded={this.state.isOpen}
                         className={classNames(Classes.MULTISELECT, tagInputProps.className)}
                         inputRef={this.refHandlers.input}
                         inputProps={inputProps}
@@ -214,7 +222,6 @@ export class MultiSelect2<T> extends AbstractPureComponent2<MultiSelect2Props<T>
                         onAdd={handleTagInputAdd}
                         onInputChange={listProps.handleQueryChange}
                         onRemove={this.handleTagRemove}
-                        role="combobox"
                         values={selectedItems.map(this.props.tagRenderer)}
                     />
                 </div>
--- a/packages/select/src/components/multi-select/multiSelect2.tsx
+++ b/packages/select/src/components/multi-select/multiSelect2.tsx
@@ -145,7 +145,7 @@ export class MultiSelect2<T> extends AbstractPureComponent2<MultiSelect2Props<T>
 
         const queryListMenuProps = {
             id: listboxId,
-            ...restProps.menuProps,
+            ...(restProps.menuProps ?? {}),
         };
 
         return (
--- a/packages/select/src/components/multi-select/multiSelect2.tsx
+++ b/packages/select/src/components/multi-select/multiSelect2.tsx
@@ -32,14 +32,11 @@ import {
 import { Popover2, PopupKind } from "@blueprintjs/popover2";
 
 import { Classes, IListItemsProps, SelectPopoverProps } from "../../common";
-import { IQueryListRendererProps, QueryList, QueryListProps } from "../query-list/queryList";
+import { IQueryListRendererProps, QueryList } from "../query-list/queryList";
 
 // N.B. selectedItems should really be a required prop, but is left optional for backwards compatibility
 
-export interface MultiSelect2Props<T>
-    extends IListItemsProps<T>,
-        Pick<QueryListProps<T>, "menuProps">,
-        SelectPopoverProps {
+export interface MultiSelect2Props<T> extends IListItemsProps<T>, SelectPopoverProps {
     /**
      * Whether the component should take up the full width of its container.
      * This overrides `popoverProps.fill` and `tagInputProps.fill`.
@@ -143,15 +140,10 @@ export class MultiSelect2<T> extends AbstractPureComponent2<MultiSelect2Props<T>
         // omit props specific to this component, spread the rest.
         const { openOnKeyDown, popoverProps, tagInputProps, ...restProps } = this.props;
 
-        const queryListMenuProps = {
-            id: this.listboxId,
-            ...(restProps.menuProps ?? {}),
-        };
-
         return (
             <this.TypedQueryList
                 {...restProps}
-                menuProps={queryListMenuProps}
+                menuProps={{ id: this.listboxId }}
                 onItemSelect={this.handleItemSelect}
                 onQueryChange={this.handleQueryChange}
                 ref={this.refHandlers.queryList}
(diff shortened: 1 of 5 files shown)


packages/select/src/components/multi-select/multiSelect2.tsx
```diff
@@ -167,6 +167,7 @@ export class MultiSelect2<T> extends AbstractPureComponent2<MultiSelect2Props<T>
         const inputProps = {
             ...tagInputProps.inputProps,
             className: classNames(tagInputProps.inputProps?.className, Classes.MULTISELECT_TAG_INPUT_INPUT),
+            role: "combobox",
```
adidahiya: I'm pretty sure we also need to set `"aria-expanded": this.state.isOpen` here

packages/select/src/components/select/select2.tsx
```diff
@@ -178,6 +178,7 @@ export class Select2<T> extends AbstractPureComponent2<Select2Props<T>, Select2S
                 <div
                     onKeyDown={this.state.isOpen ? handleKeyDown : this.handleTargetKeyDown}
                     onKeyUp={this.state.isOpen ? handleKeyUp : undefined}
+                    role="combobox"
```
adidahiya, line 201: same here, this needs `"aria-expanded": this.state.isOpen`.

and the Popover2 should get `popupKind="listbox"`

packages/select/src/components/multi-select/multiSelect2.tsx
adidahiya: actually, I think these attributes need to go on the `<div>` wrapping the TagInput on L203. That's where we set `aria-haspopup`, so it would make sense to add the other aria attributes on the same element.

Related to this, note the recent PR that merged https://github.com/palantir/blueprint/pull/5300. You should merge `develop` to get that change, and then explicitly set `<Popover2 popupKind="listbox">` below

adidahiya (changes requested):
can you please add these attributes for Suggest2 as well? thanks

packages/select/src/components/multi-select/multiSelect2.tsx
bvandrc: > I'm pretty sure we also need to set `"aria-expanded": this.state.isOpen` here

done. Agreed.

bvandrc:
> can you please add these attributes for Suggest2 as well? thanks

Updated Suggest2 to match the suggest combobox example here https://www.w3.org/TR/2021/NOTE-wai-aria-practices-1.2-20211129/examples/combobox/combobox-autocomplete-list.html

packages/select/src/components/multi-select/multiSelect2.tsx
```diff
@@ -203,6 +204,7 @@ export class MultiSelect2<T> extends AbstractPureComponent2<MultiSelect2Props<T>
                     <TagInput
                         placeholder={placeholder}
                         {...tagInputProps}
+                        aria-expanded={this.state.isOpen}
```
adidahiya: this needs to go on the wrapping div... see how we are setting another aria attribute here:
https://github.com/palantir/blueprint/blob/7ea317e054565a23dc4407f7feb5bef4553c195c/packages/popover2/src/popover2.tsx#L344

packages/select/src/components/select/select2.tsx
```diff
@@ -179,11 +186,15 @@ export class Select2<T> extends AbstractPureComponent2<Select2Props<T>, Select2S
                 onOpened={this.handlePopoverOpened}
                 onOpening={this.handlePopoverOpening}
                 popoverClassName={classNames(Classes.SELECT_POPOVER, popoverProps.popoverClassName)}
+                popupKind="listbox"
```
adidahiya: hmm that's unfortunate that you can't use a string literal... I will look into that. in the meantime you can import the `PopupKind` enum.

packages/select/src/components/multi-select/multiSelect2.tsx
```diff
@@ -29,6 +29,7 @@ import {
     TagInputProps,
 } from "@blueprintjs/core";
 import { Popover2 } from "@blueprintjs/popover2";
+import { PopupKind } from "@blueprintjs/popover2/src/popupKind";
```
adidahiya: is it not exported? sorry, that's a bug. feel free to add an ~import from~ export to `popover2/src/index.ts`

packages/select/src/components/multi-select/multiSelect2.tsx
```diff
@@ -82,6 +82,11 @@ export interface MultiSelect2Props<T> extends IListItemsProps<T>, SelectPopoverP
 
     /** Custom renderer to transform an item into tag content. */
     tagRenderer: (item: T) => React.ReactNode;
+
+    /**
+     * Props to add to the `div` that wraps the TagInput
+     */
+    wrapperProps?: React.HTMLAttributes<HTMLDivElement>;
```
adidahiya:
```suggestion
    popoverTargetProps?: React.HTMLAttributes<HTMLDivElement>;
```

packages/select/src/components/multi-select/multiSelect2.tsx
```diff
@@ -152,6 +157,7 @@ export class MultiSelect2<T> extends AbstractPureComponent2<MultiSelect2Props<T>
             popoverRef,
             selectedItems = [],
             placeholder,
+            wrapperProps = {},
```
adidahiya:
```suggestion
            popoverTargetProps = {},
```

packages/select/src/components/multi-select/multiSelect2.tsx
```diff
@@ -190,15 +196,19 @@ export class MultiSelect2<T> extends AbstractPureComponent2<MultiSelect2Props<T>
...
                 popoverClassName={classNames(Classes.MULTISELECT_POPOVER, popoverProps.popoverClassName)}
+                popupKind={PopupKind.LISTBOX}
                 ref={
                     popoverRef === undefined
                         ? this.refHandlers.popover
                         : mergeRefs(this.refHandlers.popover, popoverRef)
                 }
             >
                 <div
+                    {...wrapperProps}
```
adidahiya:
```suggestion
                    {...popoverTargetProps}
```

packages/select/src/components/select/select2.tsx
```diff
@@ -77,6 +77,12 @@ export interface Select2Props<T> extends IListItemsProps<T>, SelectPopoverProps
      * @default false
      */
     resetOnClose?: boolean;
+
+    /**
+     * Props to add to the `div` that wraps the element that is clicked to display
+     * the select popover
+     */
+    wrapperProps?: React.HTMLAttributes<HTMLDivElement>;
```
adidahiya:
```suggestion
    popoverTargetProps?: React.HTMLAttributes<HTMLDivElement>;
```

packages/select/src/components/select/select2.tsx
```diff
@@ -77,6 +77,12 @@ export interface Select2Props<T> extends IListItemsProps<T>, SelectPopoverProps
      * @default false
      */
     resetOnClose?: boolean;
+
+    /**
+     * Props to add to the `div` that wraps the element that is clicked to display
+     * the select popover
```
adidahiya:
```suggestion
     * Props to add to the popover target wrapper element.
```

packages/select/src/components/select/select2.tsx
```diff
@@ -140,6 +146,7 @@ export class Select2<T> extends AbstractPureComponent2<Select2Props<T>, Select2S
             popoverContentProps = {},
             popoverProps = {},
             popoverRef,
+            wrapperProps = {},
```
adidahiya:
```suggestion
            popoverTargetProps = {},
            popoverRef,
```

packages/select/src/components/select/select2.tsx
```diff
@@ -179,11 +186,15 @@ export class Select2<T> extends AbstractPureComponent2<Select2Props<T>, Select2S
                 onOpened={this.handlePopoverOpened}
                 onOpening={this.handlePopoverOpening}
                 popoverClassName={classNames(Classes.SELECT_POPOVER, popoverProps.popoverClassName)}
+                popupKind={PopupKind.LISTBOX}
                 ref={popoverRef}
             >
                 <div
+                    {...wrapperProps}
```
adidahiya:
```suggestion
                    {...popoverTargetProps}
```

adidahiya (changes requested):
Almost there... thanks for iterating with me on this.

I don't like to use the term "wrapper" in the API; it's a little too ambiguous. And besides we had a "wrapper" element in Popover v1 which we've gotten rid of in Popover2, so I don't want to re-introduce that term.

adidahiya:
Quick note for using Github's suggestions feature - it's possible to batch suggestions into a single commit, [see the docs here](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/incorporating-feedback-in-your-pull-request). Not a big deal or important for my review, just something for your convenience.

bvandrc:
> Quick note for using Github's suggestions feature - it's possible to batch suggestions into a single commit, [see the docs here](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/incorporating-feedback-in-your-pull-request). Not a big deal or important for my review, just something for your convenience.

Thanks, great tip!! I also re-ordered the props in alphabetical order given the change

packages/select/src/components/multi-select/multiSelect2.tsx
```diff
@@ -28,20 +28,29 @@ import {
...
      * Whether the component should take up the full width of its container.
      * This overrides `popoverProps.fill` and `tagInputProps.fill`.
      */
     fill?: boolean;
 
+    /**
+     * Adds `id={listboxId}` to the list Menu and `aria-controls={listboxId}` to the
+     * combobox element that opens it
+     */
+    listboxId?: string;
```
adidahiya: we should auto-generate this ID like `<Dialog>` does here, to simplify the API for consumers:

https://github.com/palantir/blueprint/blob/7ea317e054565a23dc4407f7feb5bef4553c195c/packages/core/src/components/dialog/dialog.tsx#L115-L116

then, we wouldn't need to add this `listboxId` prop.

adidahiya (changes requested):

adidahiya:
Latest code looks great.

> with menuProps, user can pass a custom id (or other props) to the Menu created within the Select/Multiselect popover.

do you still need this feature now that we auto-generate the ID?

bvandrc:
> Latest code looks great.
> 
> > with menuProps, user can pass a custom id (or other props) to the Menu created within the Select/Multiselect popover.
> 
> do you still need this feature now that we auto-generate the ID?

yes we need it so that we can pass the `id` prop to the `Menu` created in the `queryList`

packages/select/src/components/multi-select/multiSelect2.tsx
```diff
@@ -27,15 +27,19 @@ import {
...
 import { Classes, IListItemsProps, SelectPopoverProps } from "../../common";
-import { IQueryListRendererProps, QueryList } from "../query-list/queryList";
+import { IQueryListRendererProps, QueryList, QueryListProps } from "../query-list/queryList";
 
 // N.B. selectedItems should really be a required prop, but is left optional for backwards compatibility
 
-export interface MultiSelect2Props<T> extends IListItemsProps<T>, SelectPopoverProps {
+export interface MultiSelect2Props<T>
+    extends IListItemsProps<T>,
+        Pick<QueryListProps<T>, "menuProps">,
```
adidahiya: I meant this part -- do you still need to add this to public API?

packages/select/src/components/query-list/queryList.tsx
```diff
@@ -40,6 +40,11 @@ export interface IQueryListProps<T> extends IListItemsProps<T> {
      */
     initialActiveItem?: T;
 
+    /**
+     * Additional props to apply to the `Menu` that is created within the `QueryList`
+     */
+    menuProps?: React.HTMLAttributes<HTMLUListElement>;
```
adidahiya, line 46: I know that _this_ part is necessary to implement the feature.

adidahiya (approved):
cool, thanks for removing that extra addition to the public API. I am open to adding it in the future if necessary, but I prefer to be conservative about adding to the API for now.

PR #5354, opened 2022-06-03
[select] fix: use renderTarget API to fix fill prop

### LINKED ISSUE #5353: Fill prop doesn't work on MultiSelect 2 (when coming from MultiSelect)
#### Environment

- __Package version(s)__: 4.4.0

#### Steps to reproduce

1. Can see this doesn't work by visiting MultiSelect2 example page https://blueprintjs.com/docs/#select/multi-select2 . "Fill container width" does not work

#### Actual behavior

MultiSelect2 `fill` prop doesn't work.

#### Expected behavior

MultiSelect2 `fill` prop to work the same as the original MultiSelect's did

adidahiya:
#### Checklist

- [x] Includes tests - covered by existing test suite
- [x] Update documentation

Select2, Suggest2, MultiSelect2:
- Fix `fill` prop, which did not work and was a regression from the V1 components
- This was done by migrating to the Popover2 `renderTarget` API, and has the added benefit of collapsing the DOM rendered by the components
  - :warning: This is a slightly breaking change, but we don't expect any real consumers using these component in production yet, so it's ok
- Also improved the docs examples for these components so that they clearly demonstrate `disabled`, `fill`, `matchTargetWidth` props

Select2, MultiSelect2:
- Fix `matchTargetWidth` styling so that the width of the popup menu is no longer constrained

MultiSelect2:
- Add support for `disabled?: boolean` prop

No regressions, passing test suite

#### Screenshot

Screen Shot 2022-06-03 at 4 13 54 PM

### DIFF
--- a/packages/docs-app/src/examples/select-examples/multiSelectExample.tsx
+++ b/packages/docs-app/src/examples/select-examples/multiSelectExample.tsx
@@ -108,7 +108,9 @@ export class MultiSelectExample extends React.PureComponent<IExampleProps, IMult
         const maybeCreateNewItemRenderer = allowCreate ? renderCreateFilmOption : null;
 
         const clearButton =
-            films.length > 0 ? <Button icon="cross" minimal={true} onClick={this.handleClear} /> : undefined;
+            films.length > 0 ? (
+                <Button disabled={flags.disabled} icon="cross" minimal={true} onClick={this.handleClear} />
+            ) : undefined;
 
         return (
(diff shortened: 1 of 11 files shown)


adidahiya:
there's a bug with the min width of the popover when `matchTargetWidth={true}`:

image

adidahiya:
also `disabled` is not working for MultiSelect2:

image

PR #5355, opened 2022-06-06
[core] fix(OverflowList): improve performance, allow more items

### LINKED ISSUE #3792: [OverflowList] Implement more efficient logic to avoid triggering React setState depth limit
#### Environment

- __Package version(s)__: @blueprintjs/core: ^3.18.1 (but potentially seen on all versions since the `OverflowList` was introduced) 

#### Feature request

The `OverflowList` is very convenient and useful for auto-magically figuring out how many elements to display in cases where the width is dynamic, but we've run into a few unexpected issues with the list, due to how it determines how many items fit. Specifically, in cases where we potentially have a long list of repeating items (e.g. circle avatars, tags, etc.), we've run into unexpected application crashes when the `setState` calls end up exceeding the React `setState` depth limit of 50, causing the application to crash with an error message of: 

> Maximum update depth exceeded. This can happen when a component repeatedly calls setState inside componentWillUpdate or componentDidUpdate. React limits the number of nested updates to prevent infinite loops.

Note that this depth limit was changed from 1000 to 50 in React 16.5.0 (https://github.com/facebook/react/pull/13163)

Typically in these cases, we didn't anticipate that the overflow list would attempt to render more than 50 items, so we usually end up resolving by pre-filtering the items so the length is below 50, but it can often be difficult to figure out the `OverflowList` is the cause, given the generic React error that's shown.

Potential ideas for resolving include:

1. Changing the "search" for the visible items to take a binary search approach, which should make it much harder to hit the depth limit of 50.
2. As an alternative, the debugging confusion could potentially be alleviated if we were able to detect the issue in the component itself and throwing a clearer error.

matt-gold:
Use binary search instead of sequential search in the OverflowList repartition cycle to increase collapsible elements from 50 to 2^50.

#### Checklist

- [x] Includes tests
- [x] Update documentation

This pull request changes the algorithm used to repartition the OverflowList so that it can support a much higher number of items. The previous overflow list would break if there were 50 or more items collapsed into the Overflow element as described in issue #3792.

The new OverflowList partitioning logic.

### DIFF
--- a/packages/core/src/components/overflow-list/overflowList.tsx
+++ b/packages/core/src/components/overflow-list/overflowList.tsx
@@ -14,7 +14,6 @@
  * limitations under the License.
  */
 
-import type { ResizeObserverEntry } from "@juggle/resize-observer";
 import classNames from "classnames";
 import * as React from "react";
 
@@ -24,13 +23,6 @@ import { OVERFLOW_LIST_OBSERVE_PARENTS_CHANGED } from "../../common/errors";
 import { DISPLAYNAME_PREFIX, Props } from "../../common/props";
 import { shallowCompareKeys } from "../../common/utils";
 import { ResizeSensor } from "../resize-sensor/resizeSensor";
-
-/** @internal - do not expose this type */
-export enum OverflowDirection {
-    NONE,
-    GROW,
-    SHRINK,
-}
 
 // eslint-disable-next-line deprecation/deprecation
 export type OverflowListProps<T> = IOverflowListProps<T>;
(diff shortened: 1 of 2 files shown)


palantirtech:
Thanks for your interest in [palantir/blueprint](https://github.com/palantir/blueprint), @matt-gold! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

packages/core/src/components/overflow-list/overflowList.tsx
```diff
@@ -116,16 +108,15 @@ export interface IOverflowListProps<T> extends Props {
...
-    direction: OverflowDirection;
+    /** Whether repartitioning is still active. An overflow can take several frames to settle. */
+    repartitioning: boolean;
     /** Length of last overflow to dedupe `onOverflow` calls during smooth resizing. */
     lastOverflowCount: number;
     overflow: readonly T[];
     visible: readonly T[];
+    /** Pointer for the binary search algorithm used to find the finished non-overflowing state */
+    chopSize: number;
+    lastChopSize: number | null;
```
adidahiya, line 119: I don't think either of these need to live in state, they can just be private members of the component class. They are not referenced in any lifecycle method, just `repartition()`, so it should be save to move out of state.

adidahiya (commented):
[docs preview](https://output.circle-artifacts.com/output/job/98052d5f-c824-4217-8f2d-07b99860b7ac/artifacts/0/packages/docs-app/dist/index.html#core/components/overflow-list)

looks good! no regressions here:


and thanks for fixing the test suite, nice to see those tests back in action :)

I just have a small code style comment, but otherwise this is pretty much ready to merge :shipit:

packages/core/src/components/overflow-list/overflowList.tsx
adidahiya: This is not a huge deal, though, so I can fix it in a follow-up PR if necessary. Thanks for the contribution @matt-gold!

adidahiya (approved):

PR #5356, opened 2022-06-06
allow for additional props on nonIdealState

bvandrc:
#### Checklist

- [ ] Includes tests
- [ ] Update documentation

Allow additional html props to be passed to nonIdealState. Allows user to set `role`, or other props.

### DIFF
--- a/packages/docs-app/src/examples/core-examples/nonIdealStateExample.tsx
+++ b/packages/docs-app/src/examples/core-examples/nonIdealStateExample.tsx
@@ -126,6 +126,7 @@ export class NonIdealStateExample extends React.PureComponent<IExampleProps, INo
                     description={this.state.showDescription ? description : undefined}
                     action={this.state.showAction ? action : undefined}
                     layout={this.state.layout}
+                    role="alert"
                 />
             </Example>
         );
--- a/packages/core/src/components/non-ideal-state/nonIdealState.tsx
+++ b/packages/core/src/components/non-ideal-state/nonIdealState.tsx
@@ -17,7 +17,7 @@
 import classNames from "classnames";
 import * as React from "react";
 
-import { AbstractPureComponent2 } from "../../common";
+import { AbstractPureComponent2, removeNonHTMLProps } from "../../common";
 import * as Classes from "../../common/classes";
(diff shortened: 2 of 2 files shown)


adidahiya (changes requested):
this doesn't work in TypeScript:

image

bvandrc:
> this doesn't work in TypeScript:
> 
> image

update the props interface, maybe that will fix that

adidahiya:
> maybe that will fix that

I think it will, but just to be sure, it's good practice to add code to this repo which demonstrates usage of your new feature, and checks that it compiles correctly at build time. Please add an representative example of this new feature to one of the NonIdealState examples, perhaps `role` if you think that's useful in your application.

bvandrc:
> > maybe that will fix that
> 
> I think it will, but just to be sure, it's good practice to add code to this repo which demonstrates usage of your new feature, and checks that it compiles correctly at build time. Please add an representative example of this new feature to one of the NonIdealState examples, perhaps `role` if you think that's useful in your application.

great point. added to one of the examples.

packages/docs-app/src/examples/core-examples/nonIdealStateExample.tsx
```diff
@@ -126,6 +126,7 @@ export class NonIdealStateExample extends React.PureComponent<IExampleProps, INo
                     description={this.state.showDescription ? description : undefined}
                     action={this.state.showAction ? action : undefined}
                     layout={this.state.layout}
+                    role="alert"
```
adidahiya, line 129: this example does not show an [urgent, time-sensitive message](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/alert_role), so this role is inappropriate.

to be honest I don't see any role in the [list here](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles) which would be appropriate for a NonIdealState. so that begs the question, why are we adding this feature or demonstrating a `<NonIdealState>` example using the `role` attribute?

bvandrc: we can just cancel this PR

PR #5357, opened 2022-06-07
[select2] fix(a11y): add aria-multiselectable, aria-autocomplete

bvandrc:
#### Checklist

- [ ] Includes tests
- [ ] Update documentation

### DIFF
--- a/packages/select/src/components/multi-select/multiSelect2.tsx
+++ b/packages/select/src/components/multi-select/multiSelect2.tsx
@@ -157,7 +157,7 @@ export class MultiSelect2<T> extends AbstractPureComponent2<MultiSelect2Props<T>
         return (
             <this.TypedQueryList
                 {...restProps}
-                menuProps={{ id: this.listboxId }}
+                menuProps={{ "aria-multiselectable": true, id: this.listboxId }}
                 onItemSelect={this.handleItemSelect}
                 onQueryChange={this.handleQueryChange}
                 ref={this.refHandlers.queryList}
@@ -241,6 +241,7 @@ export class MultiSelect2<T> extends AbstractPureComponent2<MultiSelect2Props<T>
             };
             return (
                 <div
+                    aria-autocomplete="list"
                     aria-controls={this.listboxId}
                     {...popoverTargetProps}
                     {...targetProps}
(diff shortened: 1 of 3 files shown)


adidahiya (changes requested):
the [`aria-multiselectable` docs](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-multiselectable#used_with_aria-selected) say that we should use this with `aria-selected`.

I looked into this... at first glance, it seems like the simplest thing we can do to get that working is apply `"aria-selected": selected === true` here:

https://github.com/palantir/blueprint/blob/d9d20dc73dc598ea5c1220a6f42acc8654b60ef9/packages/core/src/components/menu/menuItem.tsx#L185-L191

BUT -- that's not right because of some weirdness in the current MenuItem API around "selected" vs. "active" appearance. MenuItem's styling API sort of assumes single (not multi) selection. I will follow up on this problem later.

for the time being, we must push the `aria-selected` responsibility onto consumers. you should set `aria-selected={this.isFilmSelected(film)}` here:

https://github.com/palantir/blueprint/blob/d9d20dc73dc598ea5c1220a6f42acc8654b60ef9/packages/docs-app/src/examples/select-examples/multiSelectExample.tsx#L206-L214

packages/select/src/components/multi-select/multiSelect2.tsx
```diff
@@ -241,6 +241,7 @@ export class MultiSelect2<T> extends AbstractPureComponent2<MultiSelect2Props<T>
             };
             return (
                 <div
+                    aria-autocomplete="list"
```
adidahiya, line 244: this looks like a good addition... I think we should also set this in other places we use `role="combox"`... Select2 and Suggest2

bvandrc: We already have that on `Suggest2`. On `Select2`, it didn't seem appropriate on the combobox since you can't type into that. I put it on the filter input though.

bvandrc:
> aria-selected={this.isFilmSelected(film)}

done.

adidahiya (approved):

PR #5362, opened 2022-06-07
[core] fix(ResizeSensor): compatibility with React 18

### LINKED ISSUE #5313: [React 18] `ResizeSensor` cannot be used as a JSX component
#### Environment

#### Steps to reproduce

1. Upgrade to React 18 & its `@types/react` packages
1. Got this TS error related to `ResizeSensor`

```
TS2786: 'ResizeSensor' cannot be used as a JSX component.
  Its instance type 'ResizeSensor' is not a valid JSX element.
    The types returned by 'render()' are incompatible between these types.
      Type 'boolean | {} | ReactPortal | ReactChild | null | undefined' is not assignable to type 'ReactNode'.
        Type '{}' is not assignable to type 'ReactNode'.

```

#### Possible solution

Looking at the return type of `render()` method inside `resizeSensor.d.ts`, looks like the new type of React 18 no longer consider `{}` as a valid `ReactNode`

Screen Shot 2022-05-24 at 2 56 52 AM

PurkkaKoodari:
#### Checklist

- [ ] Includes tests: This is not testable easily without testing against React 18, which would be a pretty big change given that the build is currently hard-coded against React 16.
- [ ] Update documentation: I didn't touch the docs, as this doesn't change the user-facing API at all.

(If you think this needs tests or docs, I'll update the PR.)

Firstly, this fixes #5313 by defining an explicit return type for `render()`.

The difference between React 18 and 16 here seems to be that React 18 specifies `ReactFragment = Iterable<ReactNode>`, while React 16 specifies `{} | ReactNodeArray`. The definition of `only` drops the array, so the compiled type declaration is an union of what's left of `ReactNode`. That breaks when the newer `@types/react` doesn't allow `{}`. Specifying `ReactNode` should allow the union at compile time, and then the compiled `render(): ReactNode` is compatible with any version of React.

A quick grep for `React.Children.only` suggests that `packages/table/src/common/loadableContent.tsx` and `packages/table/src/interactions/draggable.tsx` are also affected, but I didn't have time to test them right now.

Secondly, this makes `ResizeSensor` work with React 18 StrictMode and its ["reusable state"](https://github.com/reactwg/react-18/discussions/19), which essentially runs effects twice - against their [own docs for `componentWillUnmount`](https://reactjs.org/docs/react-component.html#componentwillunmount). (The docs make this very enjoyable to debug.)

The current code assumes that the component is dumped after `componentWillUnmount` so it doesn't bother to clear `this.element`. This permanently disables the observer since `observeElement` will then think the element is already being observed. I added `this.element = null` to make it "reusable".

If this breaks something subtle I didn't notice, I guess?

### DIFF
--- a/packages/core/src/components/resize-sensor/resizeSensor.tsx
+++ b/packages/core/src/components/resize-sensor/resizeSensor.tsx
@@ -62,7 +62,7 @@ export class ResizeSensor extends AbstractPureComponent2<ResizeSensorProps> {
 
     private observer = new ResizeObserver(entries => this.props.onResize?.(entries));
 
-    public render() {
+    public render(): React.ReactNode {
         // pass-through render of single child
         return React.Children.only(this.props.children);
     }
@@ -77,6 +77,7 @@ export class ResizeSensor extends AbstractPureComponent2<ResizeSensorProps> {
 
     public componentWillUnmount() {
         this.observer.disconnect();
+        this.element = null;
     }
 
     /**


palantirtech:
Thanks for your interest in [palantir/blueprint](https://github.com/palantir/blueprint), @PurkkaKoodari! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

adidahiya (approved):
Sweet, this looks like a clean solution to the problem, thanks @PurkkaKoodari!

PetrusAsikainen:
@adidahiya Seems this change never made it to `v5`?

adidahiya:
@PetrusAsikainen good catch! Fixing that here: https://github.com/palantir/blueprint/pull/6191

PR #5365, opened 2022-06-09
[core] fix(Icon): reference a11y name via aria-labelledby

bvandrc:
#### Checklist

- [ ] Includes tests
- [ ] Update documentation

Improve a11y for `Icon` elements (when has title, label the `svg` so screen readers can read the title)

Examples outlined here: https://www.tpgi.com/using-aria-enhance-svg-accessibility/ and here: https://www.unimelb.edu.au/accessibility/techniques/accessible-svgs


packages/core/src/components/icon/icon.tsx
```diff
@@ -142,8 +145,16 @@ export class Icon extends AbstractPureComponent2<IconProps & Omit<React.HTMLAttr
...
             },
-            <svg fill={color} data-icon={icon} width={size} height={size} viewBox={viewBox}>
-                {title && <desc>{title}</desc>}
+            <svg
+                fill={color}
+                data-icon={icon}
+                width={size}
+                height={size}
+                viewBox={viewBox}
+                aria-labelledby={title ? titleId : undefined}
```
adidahiya, line 154: why not just use [`aria-label`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-label)? the `<title>` is not visible on the page anyway, so I'm not sure that `aria-labelledby` is the best approach here.

adidahiya, line 154: why not just use [`aria-label`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-label)? the `<title>` is not visible on the page anyway, so I'm not sure that `aria-labelledby` is the best approach here.

bvandrc: an item does not necessarily have to be visible for `aria-labelledby` to reference it and its text contents. In this case, I am following the most common examples of how labelling an SVG is done.

Example 1:
image

https://www.tpgi.com/using-aria-enhance-svg-accessibility/

Example 2:
image

https://www.unimelb.edu.au/accessibility/techniques/accessible-svgs

adidahiya: alright, that 1st example seems quite outdated but the 2nd seems alright. still I prefer to go by what MDN says. I've verified that this is legit, according to item 5 [here](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-labelledby#benefits_and_drawbacks):

> The `aria-labelledby` property value can include content from elements that aren't even visible.

adidahiya (approved):

PR #5371, opened 2022-06-14
[docs-app] feat(DateInput2Example): use date-fns formatter

adidahiya:
- [datetime2] fix: loosen date-fns version dependency range, upgrade to latest to fix some bugs
- [docs-app] feat(DateInput2Example): use date-fns to format & parse dates instead of moment.js
  - one of the benefits of datetime2 is supposed to be that we're dropping moment.js... so we need to do this in the documentation as well.


adidahiya:
@vhellem for SA

PR #5377, opened 2022-06-15
[core] feat(MenuItem): new prop 'roleStructure'

bvandrc:
#### Checklist

- [ ] Includes tests
- [ ] Update documentation

allow for setting MenuItem to have proper role structure for when the MenuItem is an item of a "listbox" or "select" element. 

Default role structure for a MenuItem is:

```
      <li role="none"
         <a role="menuitem"
 ```

which is correct when it is a child of a Menu that has `role="menu"`, which is the default role for the Menu element.

However, for situations where we change the `Menu` role to `role="listbox"` (such as in Select2, Suggest2, and MultiSelect2), we need a different role structure for its child MenuItems:
```
  <li role="option"
         <a role=undefined
```

### CODE REVIEW

**Review by adidahiya — CHANGES_REQUESTED** [review]
When would you use an alternative role for the `<li>`? The child element is already getting `role="menuitem"`.

Also what other attributes would you imagine passing along? Again I must point you to the comment I made on another PR: https://github.com/palantir/blueprint/pull/5356#issuecomment-1148910921  -- if this PR is to be merged, it _must_ add an example usage of the new feature to either docs-app or demo-app.

**Comment by bvandrc** [comment]
> When would you use an alternative role for the `<li>`? The child element is already getting `role="menuitem"`.
> 
> Also what other attributes would you imagine passing along? Again I must point you to the comment I made on another PR: [#5356 (comment)](https://github.com/palantir/blueprint/pull/5356#issuecomment-1148910921) -- if this PR is to be merged, it _must_ add an example usage of the new feature to either docs-app or demo-app.

The usage for an alternative role for the `<li>` can be seen in the change to the `select2.md` file in this PR-- for most cases that we use `MenuItem` , it would be under a `Menu` that has the default `role="menu"`, in which case we need the `<li role="none"><a role="menuitem">` setup, which is also the default. But for other cases where the `Menu` is assigned a different role than `menu`, such as in this case with the menu of `Select2` where it assigned `role="listbox"`, we need it to have `<li role="option"><a>` since the parent menu is a `listbox`.

As you can see in all these examples:
https://www.w3.org/TR/2017/WD-wai-aria-practices-1.1-20170628/examples/listbox/listbox.html
when the `ul` (the `Menu`) has `role="listbox"`, the `li`'s should have `role="option"`.


I'm not sure how this would be represented in any of the docs or examples, it's just used to add additional props if the user wants/needs to, but it doesn't affect anything functionally or visually

**Comment by bvandrc** [comment]
@adidahiya I completely revamped this PR, changed the description. Same goal as before. Please re-review and let me know what you think. I think these changes indicate what I'm going for here.

**Comment by bvandrc** [comment]
@adidahiya so basically our a11y improvements to Select2/MultiSelect2 (ie adding role="combobox", aria-controls="listbox", etc.) are still kindof useless until we make this a11y change. Please review and consider this PR

File: packages/core/src/components/menu/menuItem.tsx [hunk 1]
```diff
@@ -77,6 +77,28 @@ export interface IMenuItemProps extends ActionProps, LinkProps {
...
+     * If `listoption`, role structure becomes:
+     *
+     * `<li role="option"`
+     *     `<a role=undefined`
+     *
+     *  which is proper role structure for a `<ul role="listbox"` parent, or a `<select>` parent.
+     *
+     * @default "menuitem"
+     */
+    roleConfig?: "menuitem" | "listoption";
```
adidahiya: prefer to avoid names like "config" in prop / variable names, as this is all essentially "configuration". instead we use "kind" in other places in Blueprint, and I think it could work well here:

```suggestion
    roleKind?: "menuitem" | "listoption";
```

packages/core/src/components/menu/menuItem.tsx
```diff
@@ -180,10 +203,15 @@ export class MenuItem extends AbstractPureComponent2<MenuItemProps & React.Ancho
             className,
         );
 
+        const [liRole, targetRole, ariaSelected] =
+            roleConfig === "listoption"
+                ? ["option", undefined, active || selected] // for when is item of a listbox role parent, or a select parent
+                : ["none", "menuitem", undefined]; // for when is item of a menu role parent
```
```suggestion
                ? ["option", undefined, active || selected] // parent has listbox role, or is a <select>
                : ["none", "menuitem", undefined]; // parent has menu role
```

packages/docs-app/src/examples/select-examples/multiSelectExample.tsx
```diff
@@ -124,7 +124,7 @@ export class MultiSelectExample extends React.PureComponent<IExampleProps, IMult
                     // we may customize the default filmSelectProps.items by
                     // adding newly created items to the list, so pass our own
                     items={this.state.items}
-                    noResults={<MenuItem disabled={true} text="No results." />}
+                    noResults={<MenuItem disabled={true} text="No results." roleConfig="listoption" />}
```
adidahiya: this seems to be common enough that we should add a note to the `noResults` prop documentation so that users remember to set `roleKind`

packages/select/src/components/select/select2.md
```diff
@@ -286,6 +288,7 @@ const renderMenu: ItemListRenderer<Film> = ({ items, itemsParentRef, query, rend
             <MenuItem
                 disabled={true}
                 text={`Found ${renderedItems.length} items matching "${query}"`}
+				roleConfig="listoption"
```
adidahiya: bad indentation; these should be spaces instead of tabs

adidahiya (changes requested):
overall this looks good, just some minor code style concerns

packages/core/src/components/menu/menuItem.tsx
bvandrc: used `roleStructure`, thinking that `roleKind` is a little too close to having the user think that the value should be an aria role, when that's not the case. 

Let me know if you'd still rather have it be `roleKind`, no big deal.

packages/docs-app/src/examples/select-examples/multiSelectExample.tsx
bvandrc: good idea, added

packages/core/src/components/menu/menuItem.tsx
bvandrc: update made

packages/select/src/components/select/select2.md
bvandrc: fixed i think

packages/core/src/components/menu/menuItem.tsx
adidahiya: Alright, I don't have a strong opinion between the two options, let's go with what you have now 👍🏽

adidahiya (approved):

PR #5390, opened 2022-06-17
[datetime2] feat: DateRangeInput2 component

adidahiya:
- New DateRangeInput2 component uses Popover2 instead of Popover. It does not have built-in TimezoneSelect support yet.
- DateRangeInput is now deprecated.


adidahiya:
This is good to go, but waiting on some bugfixes / patches required on the latest set of releases before merging.

packages/datetime2/karma.conf.js
```diff
@@ -7,6 +7,10 @@ const { createKarmaConfig } = require("@blueprintjs/karma-build-scripts");
 module.exports = function (config) {
     const baseConfig = createKarmaConfig({
         dirname: __dirname,
+        coverageExcludes: [
+            // HACKHACK: needs coverage
```
adidahiya, line 11: actually, strike that -- we need to add a test suite and remove this hack

adidahiya:
There are some styling issues with the time picker inside the DateRangePicker, but those are unrelated and will be addressed in a separate PR.

PR #5396, opened 2022-06-21
[core] fix(NumericInput): improve a11y markup, add spinbutton role

bvandrc:
#### Checklist

- [N/A] Includes tests
- [N/A] Update documentation

Add `aria-controls` prop for the increment/decrement buttons of the numericInput.

Add other aria props to make a proper spinbox a11y-wise-- see example https://w3c.github.io/aria-practices/examples/spinbutton/datepicker-spinbuttons.html


adidahiya (approved):

adidahiya:
For future reference: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/spinbutton_role

PR #5397, opened 2022-06-22
[datetime] fix(TimePicker): improve a11y markup

bvandrc:
#### Checklist

- [N/A] Includes tests
- [N/A] Update documentation

Spinbutton example: https://w3c.github.io/aria-practices/examples/spinbutton/datepicker-spinbuttons.html

### DIFF
--- a/packages/datetime/src/timePicker.tsx
+++ b/packages/datetime/src/timePicker.tsx
@@ -302,7 +302,7 @@ export class TimePicker extends React.Component<TimePickerProps, ITimePickerStat
                 onFocus={this.getInputFocusHandler(unit)}
                 onKeyDown={this.getInputKeyDownHandler(unit)}
                 onKeyUp={this.getInputKeyUpHandler(unit)}
-                role={this.props.showArrowButtons ? "spinbutton" : "textinput"}
+                role={this.props.showArrowButtons ? "spinbutton" : undefined}
                 value={value}
                 disabled={this.props.disabled}
                 autoFocus={isHour && this.props.autoFocus}
--- a/packages/datetime/src/timePicker.tsx
+++ b/packages/datetime/src/timePicker.tsx
@@ -17,7 +17,16 @@
 import classNames from "classnames";
 import * as React from "react";
 
-import { Classes as CoreClasses, DISPLAYNAME_PREFIX, HTMLSelect, Icon, Intent, Keys, Props } from "@blueprintjs/core";
+import {
+    Classes as CoreClasses,
+    Utils as CoreUtils,
+    DISPLAYNAME_PREFIX,
+    HTMLSelect,
+    Icon,
+    Intent,
+    Keys,
+    Props,
+} from "@blueprintjs/core";
 
 import * as Classes from "./common/classes";
 import * as DateUtils from "./common/dateUtils";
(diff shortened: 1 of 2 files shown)


packages/datetime/src/timePicker.tsx
```diff
@@ -257,21 +280,25 @@ export class TimePicker extends React.Component<TimePickerProps, ITimePickerStat
...
 
     private renderInput(className: string, unit: TimeUnit, value: string) {
-        const isValid = isTimeUnitValid(unit, parseInt(value, 10));
+        const valueNumber = parseInt(value, 10);
+        const isValid = isTimeUnitValid(unit, valueNumber);
         const isHour = unit === TimeUnit.HOUR_12 || unit === TimeUnit.HOUR_24;
 
         return (
             <input
+                aria-valuenow={valueNumber}
```
adidahiya, line 293: let's leave a comment about why we're doing this -- it's a little unintuitive since this is _almost_ the correct semantic element for this use case, which would obviate the need for this attribute. also I think we should set min/max attributes...

```suggestion
                // we use a type="text" input here, so we must set some a11y attributes
                // which we would otherwise get for free with a type="number" input
                aria-valuemin={0}
                aria-valuenow={valueNumber}
                aria-valuemax={getTimeUnitMax(unit)}
```

packages/datetime/src/timePicker.tsx
```diff
@@ -257,21 +280,25 @@ export class TimePicker extends React.Component<TimePickerProps, ITimePickerStat
...
                     { [CoreClasses.intentClass(Intent.DANGER)]: !isValid },
                     className,
                 )}
+                id={this.timeInputIds[unit]}
                 onBlur={this.getInputBlurHandler(unit)}
                 onChange={this.getInputChangeHandler(unit)}
                 onFocus={this.getInputFocusHandler(unit)}
                 onKeyDown={this.getInputKeyDownHandler(unit)}
                 onKeyUp={this.getInputKeyUpHandler(unit)}
+                role={this.props.showArrowButtons ? "spinbutton" : "textinput"}
```
adidahiya: I don't see "textinput" as a valid ARIA role anywhere, do you have documentation for that?

I think we can just leave it undefined when there are no arrow buttons, and inherit the input type="text" role...

```suggestion
                role={this.props.showArrowButtons ? "spinbutton" : undefined}
```

packages/datetime/src/timePicker.tsx
```diff
@@ -243,7 +260,13 @@ export class TimePicker extends React.Component<TimePickerProps, ITimePickerStat
         const onClick = () => (isDirectionUp ? this.incrementTime : this.decrementTime)(timeUnit);
         // set tabIndex=-1 to ensure a valid FocusEvent relatedTarget when focused
         return (
-            <span tabIndex={-1} className={classes} onClick={onClick}>
+            <span
+                aria-controls={this.timeInputIds[timeUnit]}
+                aria-label={`${isDirectionUp ? "Increase" : "Decrease"} ${timeUnit}`}
```
adidahiya: it's good that you included the time unit here, but it looks like it will not print nicely for some of the units. can you create a helper function in `timeUnit.ts` to pretty print the time unit enum? something like:

```ts
/** Prints a descriptive label representing the plural of th given time unit. */
export function printTimeUnit(unit: TimeUnit) {
    return {
        [TimeUnit.HOUR_24]: "hours (24hr clock)",
        [TimeUnit.HOUR_12]: "hours (12hr clock)",
        [TimeUnit.MINUTE]: "minutes",
        [TimeUnit.SECOND]: "seconds",
        [TimeUnit.MS]: "milliseconds",
    }[unit];
}
```

... and then use that helper function in this label value.

adidahiya (changes requested):

packages/datetime/src/timePicker.tsx
bvandrc: done

adidahiya (approved):

PR #5409, opened 2022-07-01
[datetime2] fix(DateInput2): clearing input calls onChange(null)

### LINKED ISSUE #5408: [DateInput2] Clearing value does not propagate to consumer
#### Steps to reproduce

1. On a DateInput2 component, remove the value

#### Actual behavior

No `onChange` handler is called

#### Expected behavior

`onChange` handler is called with `null` value, like previous `DateInput`

vhellem:

`DateInput2` will now also propagate changes that set's the date as null. 

with_revert

### DIFF
--- a/packages/datetime2/package.json
+++ b/packages/datetime2/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@blueprintjs/datetime2",
-    "version": "0.3.0",
+    "version": "0.2.0",
     "description": "Components for interacting with dates and times with timezones",
     "main": "lib/cjs/index.js",
     "module": "lib/esm/index.js",
--- a/packages/datetime2/package.json
+++ b/packages/datetime2/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@blueprintjs/datetime2",
-    "version": "0.2.0",
+    "version": "0.3.0",
     "description": "Components for interacting with dates and times with timezones",
     "main": "lib/cjs/index.js",
     "module": "lib/esm/index.js",
--- a/packages/datetime2/src/components/date-input2/dateInput2.tsx
+++ b/packages/datetime2/src/components/date-input2/dateInput2.tsx
@@ -31,8 +31,8 @@ export interface DateInput2Props extends Omit<DateInputProps, "onChange" | "valu
     /** The default timezone selected. Defaults to the user local timezone */
     defaultTimezone?: string;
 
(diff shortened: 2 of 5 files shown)


packages/datetime2/package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "@blueprintjs/datetime2",
-    "version": "0.2.0",
+    "version": "0.3.0",
```
adidahiya: we only bump versions during release. this needs to be reverted.

```suggestion
    "version": "0.2.0",
```

adidahiya (changes requested):

packages/datetime2/src/components/date-input2/dateInput2.tsx
```diff
@@ -88,7 +88,7 @@ export const DateInput2: React.FC<DateInput2Props> = React.memo(function _DateIn
     const handleDateChange = React.useCallback(
         (newDate: Date | null, isUserChange: boolean) => {
             if (newDate == null) {
-                return;
+                return onChange?.(newDate, isUserChange);
```
adidahiya: prefer to be even more explicit here:

```suggestion
                onChange?.(null, isUserChange);
                return;
```

packages/datetime2/src/components/date-input2/dateInput2.tsx
```diff
@@ -31,8 +31,8 @@ export interface DateInput2Props extends Omit<DateInputProps, "onChange" | "valu
     /** The default timezone selected. Defaults to the user local timezone */
     defaultTimezone?: string;
 
-    /** Callback invoked whenever the date or timezone has changed. ISO string */
-    onChange: (newDate: string, isUserChange?: boolean) => void;
+    /** Callback invoked whenever the date or timezone has changed. ISO string. Null if's not currently a valid date */
```
adidahiya: I think we should update the datetime package types to reflect the `| null` possibility (that package doesn't use strict null checks yet so it hasn't been required thus far). I would also slightly tweak this language:

```suggestion
    /**
     * Callback invoked whenever the date or timezone has changed.
     *
     * @param newDate ISO string or `null` (if the date is invalid or text input has been cleared)
     * ...
     */
```

adidahiya (changes requested):

adidahiya (approved):

PR #5420, opened 2022-07-08
[core] fix(NumericInput): don't apply textbox role

bvandrc:
#### Checklist

- [N/A] Includes tests
- [N/A] Update documentation

update the `type` prop for the `<input>` element created in a `numericInput`. `role` prop no longer needed due to given `type`

### DIFF
--- a/packages/core/src/components/forms/numericInput.tsx
+++ b/packages/core/src/components/forms/numericInput.tsx
@@ -427,7 +427,7 @@ export class NumericInput extends AbstractPureComponent2<HTMLInputProps & Numeri
                 asyncControl={this.props.asyncControl}
                 autoComplete="off"
                 id={this.numericInputId}
-                type={this.props.allowNumericCharactersOnly ? "number" : "text"}
+                role={this.props.allowNumericCharactersOnly ? "spinbutton" : "textbox"}
                 {...inputGroupHtmlProps}
                 intent={this.state.currentImeInputInvalid ? Intent.DANGER : this.props.intent}
                 inputRef={this.inputRef}
--- a/packages/core/src/components/forms/numericInput.tsx
+++ b/packages/core/src/components/forms/numericInput.tsx
@@ -421,7 +421,6 @@ export class NumericInput extends AbstractPureComponent2<HTMLInputProps & Numeri
 
     private renderInput() {
         const inputGroupHtmlProps = removeNonHTMLProps(this.props, NON_HTML_PROPS, true);
-        const valueAsNumber = this.getCurrentValueAsNumber();
 
         return (
             <InputGroup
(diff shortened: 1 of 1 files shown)


packages/core/src/components/forms/numericInput.tsx
```diff
@@ -421,19 +421,15 @@ export class NumericInput extends AbstractPureComponent2<HTMLInputProps & Numeri
...
                 asyncControl={this.props.asyncControl}
                 autoComplete="off"
                 id={this.numericInputId}
-                role={hasSpinButtons ? "spinbutton" : "textbox"}
+                role={this.props.allowNumericCharactersOnly ? "spinbutton" : "textbox"}
+                type={this.props.allowNumericCharactersOnly ? "number" : "text"}
                 {...inputGroupHtmlProps}
-                aria-valuemax={this.props.max}
-                aria-valuemin={this.props.min}
-                aria-valuenow={valueAsNumber}
```
bvandrc: Realized that these are probably not needed due to the fact that the `<input>` already has `min` and `max` props, and a value.

packages/core/src/components/forms/numericInput.tsx
```diff
@@ -421,19 +421,15 @@ export class NumericInput extends AbstractPureComponent2<HTMLInputProps & Numeri
...
-        const valueAsNumber = this.getCurrentValueAsNumber();
-        const hasSpinButtons = this.props.buttonPosition !== undefined && this.props.buttonPosition !== "none";
 
         return (
             <InputGroup
                 asyncControl={this.props.asyncControl}
                 autoComplete="off"
                 id={this.numericInputId}
-                role={hasSpinButtons ? "spinbutton" : "textbox"}
+                role={this.props.allowNumericCharactersOnly ? "spinbutton" : "textbox"}
```
nerdstep: The `role` would be implicit here based on setting the `type` correctly

bvandrc: got it, removed the role prop

packages/core/src/components/forms/numericInput.tsx
adidahiya: Is there any documentation for this? Would it hurt to leave them in?

bvandrc: I don't think it hurts, just unneccessary

adidahiya: Does the browser read min/max attributes when the input type is "text"? I feel like they could potentially be useful there... but again, I don't have specific documentation to support this idea.

bvandrc: that is a very good point actually. We should probably keep these since the type is "text".

adidahiya (approved):

PR #5421, opened 2022-07-08
[table] feat(EditableCell2): new placeholder prop

### LINKED ISSUE #5083: EditableCell2.editableTextProps.placeholder is not working
#### Environment

- __Package version(s)__: 3.52.0
- __Operating System__: Win10
- __Browser name and version__: Chrome 96.0.4664

#### Code Sandbox

Link to a minimal repro (fork [this code sandbox](https://codesandbox.io/s/blueprint-sandbox-et9xy)): https://codesandbox.io/s/blueprintjs-table-forked-u9up5?file=/src/DataTable.js

#### Steps to reproduce

1. check the sandbox example, it's just not working

#### Actual behavior

placeholder remains empty

#### Expected behavior

show placeholder

ayxliu19:

Added a placeholder feature that can be inputted to the editableTextProps field in EditableCell2. The placeholder shows the placeholder string if the cell is blank.

Another Design:
Since this approach uses the editableTextProps field, it is awkward calling it outside of the editable text field that is rendered when isEditing is true in https://github.com/palantir/blueprint/blob/bf4c8bead8d6ff0f50acf3167ca63b9a8d19fe76/packages/table/src/cell/editableCell2.tsx#L148
 
A better approach may be to create a new placeholder field in the EditableCell2 class and use that instead of calling the one that is in editableTextProps.

#### Screenshot
Example:

```tsx
      <EditableCell2
          value={value == null ? "" : value}
          intent={this.state.sparseCellIntent[dataKey]}
          onCancel={this.cellValidator(rowIndex, columnIndex)}
          onChange={this.cellValidator(rowIndex, columnIndex)}
          onConfirm={this.cellSetter(rowIndex, columnIndex)}
          editableTextProps={{ placeholder: "42" }}
      />
```

Screen Shot 2022-07-08 at 5 08 32 PM


packages/table/src/cell/editableCell2.tsx
```diff
@@ -143,6 +143,7 @@ export class EditableCell2 extends React.Component<EditableCell2Props, EditableC
 
         const { isEditing, dirtyValue, savedValue } = this.state;
         const interactive = spreadableProps.interactive || isEditing;
+        const hasValue = this.props.value != null && this.props.value !== "";
```
michael-yx-wu, line 146: Is the empty string not a valid value?

yarn.lock
```diff
@@ -5255,7 +5255,7 @@ first-mate@^7.0.2:
     event-kit "^2.2.0"
     fs-plus "^3.0.0"
     grim "^2.0.1"
-    oniguruma "7.2.1"
+    oniguruma "7.2.3"
```
michael-yx-wu: This was fixed in https://github.com/palantir/blueprint/pull/5407. Just merge in develop

adidahiya:
Hey @ayxliu19, please revert your changes to `yarn.lock` and merge in develop to this branch. In the future, if you commit code to Blueprint which edits the lockfile, you'll need to remember to change your NPM config to avoid pointing to our internal NPM mirror while working in this repo. I usually do this by running `npm config list` to find where the `registry` config is being set, and then editing that file to comment out the `registry` line. This will reset NPM to use the default registry at `https://registry.yarnpkg.com`.

packages/table/src/cell/_cell.scss
```diff
@@ -64,3 +64,7 @@
   right: 0;
   top: 0;
 }
+
+.#{$ns}-table-cell-text-placeholder {
+  color: $input-placeholder-color;
```
adidahiya: we'll need to add dark theme styles as well, using `$pt-dark-text-color-muted`

packages/table/src/common/_variables.scss
```diff
@@ -33,6 +33,7 @@ Color aliases
 
 $table-background-color: $light-gray5 !default;
 $dark-table-background-color: $dark-gray4 !default;
+$input-placeholder-color: $pt-text-color-muted !default;
```
adidahiya: I don't think you need this variable, especially since it's the same name as another (non-public) variable in core/src/components/forms. You can just reference the `$pt-text-color-muted` and `$pt-dark-text-color-muted` variables directly where you need the colors.

adidahiya (changes requested):
I think your intuition is right, that the `editableTextProps={ placeholder }` API feels a bit weird, especially since you've implemented it in a way that the placeholder is rendered even when `<EditableText>` is not present. It would be better to add a `placeholder` prop to `<EditableCell2>`, and mention that this new prop overrides `editableTextProps.placeholder` in its documentation.

ayxliu19:
Hi @adidahiya . Thank you for the feedback. I have made the changes and recompiled the documentation, but it seems like currently EditableCell2Props are not being displayed in the Blueprint documentation page: https://blueprintjs.com/docs/#table/api.editablecell 

Would you like me to incorporate the same changes to EditableCell1 so that the changes are reflected in the documentation?

packages/table/src/cell/editableCell2.tsx
```diff
@@ -36,6 +36,12 @@ export interface EditableCell2Props extends Omit<ICellProps, "onKeyDown" | "onKe
      */
     isFocused?: boolean;
 
+    /**
+     * Optional placeholder value for when the cell is empty (overrides the
+     * placeholder in editableTextProps)
```
michael-yx-wu: nit 
```suggestion
     * placeholder in {@link EditableTextProps})
```

packages/table/src/cell/editableCell2.tsx
```diff
@@ -167,9 +174,10 @@ export class EditableCell2 extends React.Component<EditableCell2Props, EditableC
             const textClasses = classNames(Classes.TABLE_EDITABLE_TEXT, {
                 [Classes.TABLE_TRUNCATED_TEXT]: truncated,
                 [Classes.TABLE_NO_WRAP_TEXT]: !wrapText,
+                [Classes.TABLE_CELL_TEXT_PLACEHOLDER]: !hasValue,
             });
 
-            cellContents = <div className={textClasses}>{savedValue}</div>;
+            cellContents = <div className={textClasses}>{hasValue ? savedValue : this.props.placeholder}</div>;
```
adidahiya, line 180: This is good but I think we need to bring back the change to `cellContents` a few lines above here... `<EditableText placeholder={this.props.placeholder}>` should be set on L168, as suggested in the documentation of the new prop you just added.

adidahiya (changes requested):
@ayxliu19 good point about EditableCell2 not being shown in the docs -- I will fix that in a separate PR. No need to block on that for now, though.

adidahiya:
I'm adding EditableCell2Props to the docs in this PR: https://github.com/palantir/blueprint/pull/5449

adidahiya (approved):
nice, thanks @ayxliu19 👍🏽

PR #5424, opened 2022-07-10
[select] feat(MultiSelect2): allow creating multiple items

### LINKED ISSUE #5384: [MultiSelect2] allow to create multiple items
#### Environment

- __Package version(s)__:  4.2.0

#### Feature request

Multi select should allow the same behaviour we have in the tag input component where we can create multiple values at once. 
Multi select doesn't need to implement the same behaviour, it just needs to support creating multiple items at once. 

Existing typing
```
createNewItemFromQuery | undefined \| (query: string) => T
```
Proposed typing
```
createNewItemFromQuery | undefined \| (query: string) => T | T[]
```

#### Examples

How tag input works:

tag input

How multi select currently works

multi select

I would like to split the input by comma and add both `test1` and `hello` as new values.

ZeRego:
#### Checklist

- [x] Includes tests
- [x] Update documentation

- Change `createNewItemFromQuery` to support array of new items

- Unsure how `wouldCreatedItemMatchSomeExistingItem` should behave. Since the logic for splitting the query into multiple items is on the user implementation side, it will display the create item option when there is only a partial match. E.g: query "test, test1" items are "test" and "test1" and test already exists in the list. We would still show the create item option 
- I wrote a test but I don't know why it is failing and how to debug tests. Can you provide more info on how to run tests in debug mode ? 

#### Screenshot

create multiple


palantirtech:
Thanks for your interest in [palantir/blueprint](https://github.com/palantir/blueprint), @ZeRego! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

packages/select/src/common/listItemsProps.ts
```diff
@@ -170,7 +170,7 @@ export interface IListItemsProps<T> extends Props {
      * created, either by pressing the `Enter` key or by clicking on the "Create
      * Item" option. It transforms a query string into an item type.
```
adidahiya: I think you should add a bit to the JSDoc here about creating multiple items by returning an array.

adidahiya (commented):
The best way to run unit tests here would be to focus on one component, let's say Select2 to start, and scope the test runner to just those tests by temporarily editing `select2Tests.tsx` to say `describe.only("<Select2>", ...` You can scope down the tests further by editing `selectComponentSuite.tsx` to say `describe.only("create", ..`, etc. Make sure to revert these edits before pushing to the PR later.

Then you can go into packages/select and run `yarn test:karma:debug`. This will run the tests once, and start a watcher to re-run them when source or test files change.

adidahiya:
ping @ZeRego, any interest in continuing this PR?

ZeRego:
> ping @ZeRego, any interest in continuing this PR?

@adidahiya Sorry for the delay. 

Testing was a bit of a pain since I couldn't figure out how to run the tests with breakpoints in Webstorm IDE. 
I had to add `.skip` to all the tests beside the select2 and use console logs to figure out what was happening. 

Looks like I was missing the event keydown even though the other tests only have the keyup event so I'm now confused on how those tests are passing 🤷

adidahiya:
@ZeRego I'll take a look at your updates. In the future, please don't force-push to PR branches; I like to see all the commits in their original timeline. I squash all PRs into a single commit when merging.

packages/docs-app/src/common/films.tsx
```diff
@@ -178,6 +178,27 @@ export const renderCreateFilmOption = (
     />
 );
 
+export const renderCreateFilmOptions = (
```
adidahiya: minor: I think the pluralization in this name is slightly off, it should be:

```suggestion
export const renderCreateFilmsOption = (
```

adidahiya (commented):
[docs preview](https://output.circle-artifacts.com/output/job/ce08dbdd-9b7c-492b-adcd-a03c605c11b4/artifacts/0/packages/docs-app/dist/index.html#select/multi-select2) looks good!



the test coverage failure looks like a bug in our test coverage configuration. we shouldn't be checking docs-app code for coverage. please update packages/select/karma.conf.js:

```
        dirname: __dirname,
        coverageExcludes: [
            "../docs-app/*",
        ],
    });
```

adidahiya:
> please update packages/select/karma.conf.js:

My suggestion above didn't work, for some reason Karma doesn't allow me to match those globs on files outside the current working directory. I've added a commit here which uses `/* istanbul ignore next */` instead.

adidahiya (approved):

PR #5432, opened 2022-07-11
fix(MenuItem2): improve selected appearance and a11y

bvandrc:
#### Checklist

- [N/A] Includes tests
- [N/A] Update documentation

allow `liProps` to be passed to `MenuItem` so that user can set `aria-selected` prop.


adidahiya:
Why is this better than the previous/existing approach?

bvandrc:
> Why is this better than the previous/existing approach?

So, I previously thought that the `active || selected` were true if the item was checked-- I didn't realize that it actually means the item is currently highlighted for selection. So here:

image

currently the item in blue would have `aria-selected=true`, because apparently that's what the `active`/`selected` props are for. But what we really need, is for those checked items to all have `aria-selected=true`. As far as I can tell, there's no way to determine that an item is checked within MenuItem, so `aria-selected` must be manually passed to the MenuItem, hence the addition of the `liProps` allowing you to pass that.

Since `aria-selected` must be manually passed, I added that to the multiSelectExample -- see that implementation for why we need to do it this way

adidahiya:
> As far as I can tell, there's no way to determine that an item is checked within MenuItem

You're right. The semantics around active/selected for MenuItems has been messy for a while (there is some documentation about this in the deprecated `active` prop).

I think we can be smarter about this going forward, make the design of MenuItems more prescriptive, and avoid adding this plumbing of `aria-selected` to the API. Here's my proposal:

## `selected` prop

MenuItems ought to use "tick" icons to represent selected status when `roleStructure="listoption"`. This would apply to Select2, Suggest2, and MultiSelect2.

For MultiSelect2, this matches the existing behavior we have implemented in the docs example:

image

It would be great if we could do this out of the box, so users don't have to set `icon` themselves. We'll have to render a blank icon for non-selected items. Here's how it would look with Select2:

image

Users should still be able to _override_ this icon if they choose with the `icon` prop.

Once you implement this, setting `aria-selected` is trivial... just take the value of the `selected` prop.

## `active` prop

`active` can be used to indicate keyboard focus, like it currently does for @blueprintjs/select components. We should un-deprecate this prop, at least for `roleStructure="listoption"`, since we are now disambiguating it from `selected`.

image

## MenuItems with `roleStructure="menuitem"`

In the future, I think it would be worth reconsidering `selected` prop behavior for regular MenuItems as well (so that `selected={true}` renders a "tick" icon in _all_ contexts), but for now, I think we should preserve the existing behavior to avoid breaking things too much.

packages/core/src/components/menu/menuItem.tsx
```diff
@@ -199,16 +213,11 @@ export class MenuItem extends AbstractPureComponent2<MenuItemProps & React.Ancho
                 [Classes.DISABLED]: disabled,
                 // prevent popover from closing when clicking on submenu trigger or disabled item
                 [Classes.POPOVER_DISMISS]: shouldDismissPopover && !disabled && !hasSubmenu,
-                [Classes.SELECTED]: selected || (active && intentClass === undefined),
+                [Classes.SELECTED]: active && intentClass === undefined,
```
bvandercar-vt, line 216: I'm not sure if this is correct (because I'm not sure how these classes look), needs review

packages/core/src/components/menu/menuItem.tsx
```diff
@@ -170,7 +165,7 @@ export class MenuItem extends AbstractPureComponent2<MenuItemProps & React.Ancho
             className,
             children,
             disabled,
-            icon,
+            icon: passedIcon,
```
adidahiya: I think you can avoid this awkward naming and make the code a little clearer just by omitting this line from the destructuring statement. Then you can reference `this.props.icon` wherever you mean to use the "passed icon".

packages/core/src/components/menu/menuItem.tsx
```diff
@@ -187,6 +182,25 @@ export class MenuItem extends AbstractPureComponent2<MenuItemProps & React.Ancho
...
 
+        const [liRole, targetRole, icon, ariaSelected] =
+            roleStructure === "listoption"
+                ? // "listoption": parent has listbox role, or is a <select>
+                  [
+                      "option",
+                      undefined, // target should have no role
+                      selected
+                          ? passedIcon ?? IconNames.SMALL_TICK // if selected, set icon to check by default
+                          : IconNames.BLANK, // if not selected, no icon
```
adidahiya: I think `this.props.icon` should always take precedence, even if `selected === true`

packages/docs-app/src/examples/select-examples/multiSelectExample.tsx
```diff
@@ -248,10 +248,11 @@ export class MultiSelectExample extends React.PureComponent<IExampleProps, IMult
         if (!modifiers.matchesPredicate) {
             return null;
         }
+
         return (
             <MenuItem
-                selected={modifiers.active}
-                icon={this.isFilmSelected(film) ? "tick" : "blank"}
+                active={modifiers.active}
+                selected={this.isFilmSelected(film)}
```
adidahiya, line 240: ~the comment above this method needs to be updated on L246 -- we are not using the default renderer so that we can set `selected` now, not `icon`.~

actually, after removing `renderFilm` from `filmSelectProps` like I suggested above, the comment will no longer be relevant since each component will be implementing its own `renderFilm` method -- the comment on L246 should be removed entirely.

packages/docs-app/src/common/films.tsx
```diff
@@ -136,17 +136,17 @@ export const renderFilm: ItemRenderer<IFilm> = (film, { handleClick, handleFocus
     if (!modifiers.matchesPredicate) {
         return null;
     }
-    const text = `${film.rank}. ${film.title}`;
     return (
         <MenuItem
+            text={highlightText(`${film.rank}. ${film.title}`, query)}
             active={modifiers.active}
+            // selected={modifiers.selected} // TODO: add selected prop to modifiers? or does this need to be handled outside of QueryList?
```
adidahiya: Select2 has no controlled "selected item", so QueryList cannot set `modifiers.selected` for us.

The changes to `selected` semantics we're implementing in this PR make this standalone `renderFilm` function hard to re-use across the docs examples via `filmSelectProps`. We're going to have to remove it from `filmSelectProps` and copy it over to each of the example implementations in `filmSelect.tsx`, `omnibarExample.tsx`, and `suggestExample.tsx`.

adidahiya (changes requested):

packages/docs-app/src/common/films.tsx
```diff
@@ -132,15 +132,27 @@ export const TOP_100_FILMS: IFilm[] = [
...
-export const renderFilm: ItemRenderer<IFilm> = (film, { handleClick, handleFocus, modifiers, query }) => {
+export const renderFilm: ItemRenderer<IFilm> = (
+    film: IFilm,
+    {
+        handleClick,
+        handleFocus,
+        modifiers,
+        query,
+    }: {
+        handleClick: () => void;
```
adidahiya: you don't need any of these type defs. they are inferred through `: ItemRenderer<IFilm`. please remove them, as this code is now overly verbose.

packages/docs-app/src/common/films.tsx
```diff
@@ -166,7 +178,7 @@ export const renderCreateFilmOption = (
     />
 );
 
-export const filterFilm: ItemPredicate<IFilm> = (query, film, _index, exactMatch) => {
+export const filterFilm: ItemPredicate<IFilm> = (query: string, film: IFilm, _index: number, exactMatch: boolean) => {
```
adidahiya: same here, no need for these verbose type defs.

adidahiya (changes requested):
Select2 items aren't getting `selected={true}` or the tick icon yet:

image

Also, I think Suggest2 & Omnibar menu items in the docs _should not_ get tick icons. They behave differently, it doesn't feel necessary to visually show selection status on the items there. They should probably use the default `roleStructure="menuitem"`.

bvandrc:
> 
> Also, I think Suggest2 & Omnibar menu items in the docs _should not_ get tick icons. They behave differently, it doesn't feel necessary to visually show selection status on the items there. They should probably use the default `roleStructure="menuitem"`.

they should still have `roleSctructure="listoption"` because they are children of a Menu `ul` that has `role="listbox"`-- that is the key function of  `roleSctructure="listoption"`, creating the proper role structure for that. If `selected` isn't passed, they don't receive the checkmark, so all one has to do is not pass `selected` for those.

> Select2 items aren't getting `selected={true}` or the tick icon yet:
> 

I was thinking that for a Select2, the checkmark isn't really that useful anyways-- you can only select 1 option, and you already see the selected option in the input field, so seeing a check next to that option doesn't really provide much. If you disagree, and you want it in the example, I could implement a new itemRenderer for the select2 example, sure. The consumer could always pass `selected` in their own itemRenderer if they want a checkmark next to their selected option in Select2, but I was thinking about keeping that out of the example, for the sake of not having to write a new itemRenderer.

adidahiya:
If it's not going to get the tick icon, then there shouldn't be a space here:

image

but... I think I _would_ prefer to add the icon, even if it's potentially redundant with the Select2 target. Some users may not show the selected item text in the target, and the tick icon can be a helpful indicator in those cases.

There's a similar situation with Omnibar and Suggest2... I don't think there should be a blank space on the left since we're not showing tick icons:

image

... this is why I was suggesting changing the `roleStructure`, but if that's not appropriate markup, then we must figure out an alternative solution that allows us to omit the blank icons for these components... open to ideas.

bvandrc:
> If it's not going to get the tick icon, then there shouldn't be a space here:
> 
> image
> 
> but... I think I _would_ prefer to add the icon, even if it's potentially redundant with the Select2 target. Some users may not show the selected item text in the target, and the tick icon can be a helpful indicator in those cases.
> 
> There's a similar situation with Omnibar and Suggest2... I don't think there should be a blank space on the left since we're not showing tick icons:
> 
> image
> 
> ... this is why I was suggesting changing the `roleStructure`, but if that's not appropriate markup, then we must figure out an alternative solution that allows us to omit the blank icons for these components... open to ideas.

idea 1: we add a new `roleStructure` called maybe, `"listoptionCheckSelected"` that enables the checkmark/blank icon -- and passing the normal already-existing `"listoption"` will have no effect on the icon.

idea 2:
passed `selected` prop value:
- undefined: no icon
- false: blank icon
- true: check icon
`icon` prop passed: overrides

idea 3: new boolean prop called `checkSelected` that works in conjunction with the `selected` prop. If `checkSelected=false` and `selected=true`, `aria-selected` would be set to true but there would be no effect on the icon.

Which of these do you like best?

adidahiya:
@bvandercar-vt I like idea 2!

adidahiya:
I've caused some conflicts here with the latest PR I merged #5471. If you don't mind, I'll push an update to your branch to resolve them @bvandercar-vt

adidahiya:
I've updated the PR with my suggestions, and an implementation of your idea (2) from your last comment. I decided to port the `selected` prop semantic changes to MenuItem2 and leave MenuItem as-is, because in the time while this PR was open, I deprecated MenuItem and froze its code.

I think the docs examples are in a good state now:

Select:



MultiSelect:



Suggest:



Omnibar:

adidahiya (approved):

PR #5433, opened 2022-07-12
add `title` prop to button

bvandrc:
#### Checklist

- [N/A] Includes tests
- [N/A] Update documentation

Weird that `title` prop is not present on `HTMLButtonElement` when this prop is allowed, applicable, and common. Manually added it.


adidahiya:
`title` is already supported via `React.ButtonHTMLAttributes<HTMLButtonElement>` / `React.AnchorHTMLAttributes<HTMLAnchorElement>`, it just doesn't show up in the props interface documentation. I don't think we need this addition.

PR #5438, opened 2022-07-12
[core] fix(Slider, MultiSlider): improve a11y markup

bvandrc:
#### Checklist

- [ ] Includes tests
- [X] Update documentation

improve slider a11y per this official ARIA example: https://www.w3.org/WAI/ARIA/apg/example-index/slider/slider-multithumb.html

### DIFF
--- a/packages/core/src/components/slider/handle.tsx
+++ b/packages/core/src/components/slider/handle.tsx
@@ -67,14 +67,14 @@ export class Handle extends AbstractPureComponent2<IInternalHandleProps, IHandle
     }
 
     public render() {
-        const { className, disabled, label, min, max, value, vertical, handleHtmlProps } = this.props;
+        const { className, disabled, label, min, max, value, vertical, htmlProps } = this.props;
         const { isMoving } = this.state;
 
         return (
             <span
                 role="slider"
                 tabIndex={0}
-                {...handleHtmlProps}
+                {...htmlProps}
                 className={classNames(Classes.SLIDER_HANDLE, { [Classes.ACTIVE]: isMoving }, className)}
                 onKeyDown={disabled ? undefined : this.handleKeyDown}
                 onKeyUp={disabled ? undefined : this.handleKeyUp}
--- a/packages/core/src/components/slider/handle.tsx
+++ b/packages/core/src/components/slider/handle.tsx
@@ -67,7 +67,7 @@ export class Handle extends AbstractPureComponent2<IInternalHandleProps, IHandle
     }
 
     public render() {
-        const { className, disabled, label } = this.props;
+        const { className, disabled, label, min, max, value, vertical, ariaLabel } = this.props;
         const { isMoving } = this.state;
 
         return (
@@ -80,6 +80,12 @@ export class Handle extends AbstractPureComponent2<IInternalHandleProps, IHandle
                 ref={this.refHandlers.handle}
                 style={this.getStyleProperties()}
                 tabIndex={0}
+                role="slider"
+                aria-valuemin={min}
+                aria-valuemax={max}
+                aria-valuenow={value}
+                aria-label={ariaLabel}
+                aria-orientation={vertical ? "vertical" : "horizontal"}
             >
                 {label == null ? null : <span className={Classes.SLIDER_LABEL}>{label}</span>}
             </span>
--- a/packages/core/src/components/slider/handle.tsx
+++ b/packages/core/src/components/slider/handle.tsx
@@ -67,11 +67,14 @@ export class Handle extends AbstractPureComponent2<IInternalHandleProps, IHandle
     }
 
     public render() {
-        const { className, disabled, label, min, max, value, vertical, ariaLabel } = this.props;
+        const { className, disabled, label, min, max, value, vertical, handleHtmlProps } = this.props;
         const { isMoving } = this.state;
 
         return (
             <span
+                role="slider"
+                tabIndex={0}
+                {...handleHtmlProps}
                 className={classNames(Classes.SLIDER_HANDLE, { [Classes.ACTIVE]: isMoving }, className)}
                 onKeyDown={disabled ? undefined : this.handleKeyDown}
                 onKeyUp={disabled ? undefined : this.handleKeyUp}
@@ -79,12 +82,9 @@ export class Handle extends AbstractPureComponent2<IInternalHandleProps, IHandle
                 onTouchStart={disabled ? undefined : this.beginHandleTouchMovement}
                 ref={this.refHandlers.handle}
                 style={this.getStyleProperties()}
-                tabIndex={0}
-                role="slider"
                 aria-valuemin={min}
                 aria-valuemax={max}
                 aria-valuenow={value}
-                aria-label={ariaLabel}
                 aria-orientation={vertical ? "vertical" : "horizontal"}
             >
                 {label == null ? null : <span className={Classes.SLIDER_LABEL}>{label}</span>}
--- a/packages/core/src/components/slider/handleProps.tsx
+++ b/packages/core/src/components/slider/handleProps.tsx
@@ -90,6 +90,6 @@ export interface IHandleProps extends Props {
      */
     type?: HandleType;
 
-    /** Applied to `aria-label` prop of the handle `span` element */
+    /** Accessible label text for this handle element. */
     ariaLabel?: string;
 }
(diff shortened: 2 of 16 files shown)


packages/core/src/components/slider/handle.tsx
```diff
@@ -80,6 +80,12 @@ export class Handle extends AbstractPureComponent2<IInternalHandleProps, IHandle
                 ref={this.refHandlers.handle}
                 style={this.getStyleProperties()}
                 tabIndex={0}
+                role="slider"
+                aria-valuemin={min}
+                aria-valuemax={max}
+                aria-valuenow={value}
```
bvandrc, line 87: `value` gives the current value of this Handle, correct?

adidahiya: yes

packages/docs-app/src/examples/core-examples/collapsibleListExample.tsx
```diff
@@ -58,6 +58,7 @@ export class CollapsibleListExample extends React.PureComponent<IExampleProps, I
                     onChange={this.handleChangeCount}
                     showTrackFill={false}
                     value={this.state.visibleItemCount}
+                    handleAriaLabel="number of visible items"
```
adidahiya: This slider (and the other sliders in the docs examples) may be better labeled by their visible label text inside the preceding `<Label>` elements... something like:

```
<Label id={labelId}>Visible items</Label>
<Slider handleProps={{ ariaLabelledBy: labelId }}>
```

what do you think?

packages/core/src/components/slider/handleProps.tsx
```diff
@@ -89,4 +89,7 @@ export interface IHandleProps extends Props {
      * @default "full"
      */
     type?: HandleType;
+
+    /** Applied to `aria-label` prop of the handle `span` element */
```
adidahiya:
```suggestion
    /** Accessible label text for this handle element. */
```

packages/core/src/components/slider/slider.tsx
```diff
@@ -44,6 +44,9 @@ export interface ISliderProps extends ISliderBaseProps {
 
     /** Callback invoked when the handle is released. */
     onRelease?(value: number): void;
+
+    /** Applied to `aria-label` prop of the slider Handle. */
+    handleAriaLabel?: string;
```
adidahiya: According to the [MDN ARIA: slider role docs](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/slider_role), we should also support `aria-labelledby` in addition to `aria-label`... I'm thinking the API would look better if we forwarded a subset of handle props from Slider...

```suggestion
    /** Subset of handle props to apply to the child Handle. */
    handleProps?: Pick<HandleProps, "ariaLabel" | "ariaLabelledBy">;
```

packages/core/src/components/slider/rangeSlider.tsx
```diff
@@ -44,6 +44,9 @@ export interface IRangeSliderProps extends ISliderBaseProps {
 
     /** Callback invoked when a handle is released. */
     onRelease?(value: NumberRange): void;
+
+    /** Applied to `aria-label` props of the slider Handles */
+    handleAriaLabels?: { startAriaLabel: string; endAriaLabel: string };
```
adidahiya: following my comment on `slider.tsx`... my API suggestion here would be

```suggestion
    /** Subset of handle props to apply to the start and end Handles. */
    handleProps?: {
    	start: Pick<HandleProps, "ariaLabel" | "ariaLabelledBy">;
    	end: Pick<HandleProps, "ariaLabel" | "ariaLabelledBy">;
    };
```

adidahiya (changes requested):

packages/docs-app/src/examples/core-examples/collapsibleListExample.tsx
bvandrc: yep, I agree. Implemented.

packages/core/src/components/slider/slider.tsx
bvandrc: yep you're right, we should support `aria-labelledby`. Did it a little differently, see changes

packages/core/src/components/slider/rangeSlider.tsx
bvandrc: yep, see new changes

packages/core/src/components/slider/handleProps.tsx
```diff
@@ -47,6 +47,8 @@ export const HandleInteractionKind = {
 // eslint-disable-next-line @typescript-eslint/no-redeclare
 export type HandleInteractionKind = typeof HandleInteractionKind[keyof typeof HandleInteractionKind];
 
+export type HandleHtmlProps = Pick<React.HTMLProps<HTMLSpanElement>, "aria-label" | "aria-labelledby">;
```
bvandrc, line 50: do we actually want to `Pick` these specific props or should we just let any `span` props be passed? (Should this just be `React.HTMLProps<HTMLSpanElement>`?)

packages/core/src/components/slider/handleProps.tsx
```diff
@@ -89,4 +91,9 @@ export interface IHandleProps extends Props {
      * @default "full"
      */
     type?: HandleType;
+
+    /**
+     * HTML props to apply to the Handle
+     */
+    handleHtmlProps?: HandleHtmlProps;
```
adidahiya:
```suggestion
     * A limited subset of HTML props to apply to the rendered `<span>` element.
     */
    htmlProps?: HandleHtmlProps;
```

adidahiya (changes requested):
Looks good! Just one small edit. `<Handle handleHtmlProps={...}>` feels verbose/redundant so I prefer `<Handle htmlProps={...}>`.

`<Slider handleHtmlProps={...}>` is sufficiently verbose, though, let's keep that prop name.

packages/core/src/components/slider/handleProps.tsx
adidahiya: Let's just pick these two for now. We can add more later if necessary

bvandrc: sounds good

packages/core/src/components/slider/handleProps.tsx
bvandrc: done, changed

packages/core/src/components/slider/rangeSlider.tsx
```diff
@@ -56,11 +60,20 @@ export class RangeSlider extends AbstractPureComponent2<RangeSliderProps> {
...
+        const { value, handleHtmlProps, ...props } = this.props;
         return (
             <MultiSlider {...props}>
-                <MultiSlider.Handle value={value![RangeIndex.START]} type="start" intentAfter={props.intent} />
-                <MultiSlider.Handle value={value![RangeIndex.END]} type="end" />
+                <MultiSlider.Handle
+                    value={value![RangeIndex.START]}
+                    type="start"
+                    intentAfter={props.intent}
+                    htmlProps={handleHtmlProps ? handleHtmlProps.start : undefined}
```
adidahiya:
```suggestion
                    htmlProps={handleHtmlProps?.start}
```

packages/core/src/components/slider/rangeSlider.tsx
```diff
@@ -56,11 +60,20 @@ export class RangeSlider extends AbstractPureComponent2<RangeSliderProps> {
...
+                />
+                <MultiSlider.Handle
+                    value={value![RangeIndex.END]}
+                    type="end"
+                    htmlProps={handleHtmlProps ? handleHtmlProps.end : undefined}
```
adidahiya:
```suggestion
                    htmlProps={handleHtmlProps?.end}
```

packages/core/src/components/slider/rangeSlider.tsx
```diff
@@ -44,6 +45,9 @@ export interface IRangeSliderProps extends ISliderBaseProps {
 
     /** Callback invoked when a handle is released. */
     onRelease?(value: NumberRange): void;
+
+    /** HTML props to apply to the slider Handles */
+    handleHtmlProps?: { start: HandleHtmlProps; end: HandleHtmlProps };
```
adidahiya: I just realized that these should probably be optional, so you can only set one of them if you want:

```suggestion
    handleHtmlProps?: { start?: HandleHtmlProps; end?: HandleHtmlProps };
```

adidahiya (changes requested):

adidahiya (approved):

PR #5440, opened 2022-07-13
[datetime2] feat(DateInput2): use Popover2, port DateInput impl

adidahiya:
#### Checklist

- [x] Includes tests
- [x] Update documentation

Despite what the docs stated before this change, we were not using Popover2 properly in DateInput2. It was delegating to DateInput for the text input + popover implementation, which used Popover.

This PR ports over the remaining code from DateInput into DateInput2 and migrates to Popover2.

- No regressions from DateInput
- Passing test suite

#### Screenshot

image

### DIFF
--- a/packages/datetime/src/dateInput.tsx
+++ b/packages/datetime/src/dateInput.tsx
@@ -14,6 +14,12 @@
  * limitations under the License.
  */
 
+/**
+ * @fileoverview This component is DEPRECATED, and the code is frozen.
+ * All changes & bugfixes should be made to DateInput2 in the datetime2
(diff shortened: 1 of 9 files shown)


adidahiya:
There's a bug in controlled mode -- the "Invalid date" message does not appear after typing in an invalid date:

| DateInput2 | DateInput |
|-|-|
|  |  |

adidahiya:
Things are much better now, but there is still this weird state you can get into when typing an incomplete date and the timezone is interpreted as a very unexpected offset:

image

Will fix that in a follow-up PR.

PR #5454, opened 2022-07-20
refactor: use IconNames constant

bvandrc:
Style refactor, no functional change

Better to use consts than string literals, since there is a const for the IconNames available

### DIFF
--- a/packages/core/src/components/dialog/dialog.tsx
+++ b/packages/core/src/components/dialog/dialog.tsx
@@ -17,6 +17,8 @@
 import classNames from "classnames";
 import * as React from "react";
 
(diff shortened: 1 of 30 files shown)


adidahiya (changes requested):
Why is this better? I prefer the string literals. They auto-complete, type-check, and are less verbose (you don't need the `IconNames` import).

bvandrc:
> Why is this better? I prefer the string literals. They auto-complete, type-check, and are less verbose (you don't need the `IconNames` import).

I suppose I just prefer using the const, or at least that's what we do in our own project so I thought it would help. Good point that they type check and so the string literals autocomplete. All good, I'll close this PR

PR #5456, opened 2022-07-21
[table] fix: scrolling bug when grid size == viewport size

mattskrobola:
Fixes a scrolling bug that is caused by unnecessarily rendering ghost cells when the grid height/width is equal to the viewports height/width (commonly happens if the container element has `height: "auto"`). Similar behavior in this past issue https://github.com/palantir/blueprint/issues/5027

#### Checklist

- [x] Includes tests
- [ ] Update documentation

Changes `hasVerticalOverflow` to `hasVerticalOverflowOrExactFit` to account for the case where height/width is equal to the viewport where we don't need to render ghost cells.

#### Screenshot

before:
beforeBPScroll
after:
afterBPScroll

### DIFF
--- a/packages/table-dev-app/src/features.html
+++ b/packages/table-dev-app/src/features.html
@@ -148,6 +148,10 @@
   <div id="table-11"></div>
   <br /><br />
 
+  <h4>Table with auto height and ghost cells enabled</h4>
+  <div id="table-12"></div>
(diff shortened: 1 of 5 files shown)


adidahiya (approved):
nice, thanks for the thorough PR @mattskrobola!

PR #5459, opened 2022-07-22
[core] fix(MultiSlider): loosen children validation

### LINKED ISSUE #4456: Slider style props support or innerStyle
#### Environment

- __Package version(s)__: `3.22.3`
- __Browser and OS versions__: `Windows 10 WSL`

#### Feature request

Add style props support for Slider component to modify style property on specific element

#### Examples

```jsx
<Slider
  style={{ height: "100%" }}
  // OR
  innerStyle={{ height: "100%" }}
/>
```

adidahiya:
#### Checklist

- [x] Includes tests
- [ ] Update documentation

Log a console error instead of throwing a hard error when MultiSlider children fail to validate as Handles. This allows users to use styled-components HOCs on Handle children elements.

### DIFF
--- a/packages/core/src/components/slider/multiSlider.tsx
+++ b/packages/core/src/components/slider/multiSlider.tsx
@@ -218,7 +218,7 @@ export class MultiSlider extends AbstractPureComponent2<MultiSliderProps, ISlide
             throw new Error(Errors.SLIDER_ZERO_STEP);
         }
         if (props.labelStepSize !== undefined && props.labelValues !== undefined) {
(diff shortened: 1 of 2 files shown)


adidahiya:
Hmm, actually this won't work because we filter out non `MultiSlider.Handle` children and need to inspect their props directy:

https://github.com/palantir/blueprint/blob/0f4f0e60a364c3447eb2bcff1c82a1c5f19ee3ab/packages/core/src/components/slider/multiSlider.tsx#L495

will need to find an alternative solution...

PR #5462, opened 2022-07-25
[select] fix(Suggest2): get root node correctly in Shadow DOM

### LINKED ISSUE #5461: Suggest2 not working with Shadow DOM
#### Environment

- __Package version(s)__: 4.5.1
- __Operating System__: Win 11
- __Browser name and version__: Chrome 103

#### Code Sandbox

Link to a minimal repro : (https://codesandbox.io/s/recursing-panna-w2dqsr?file=/src/App.js)

#### Steps to reproduce

1. Create React app inside Shadow DOM
2. Use Suggest2 component
3. Try to search something in Suggest2 component

#### Actual behavior

Entering value inside Suggest2 isn't possible. 
Dropdown doesn't open. When you do focus out of Suggest2 component and then focus it again it opens dropdown for a second and immediately closes it.

#### Expected behavior

User should be able to enter value to Suggest2 components and see results dropdown

#### Possible solution

This happens because in suggest2.tsx in function "handlePopoverInteraction" there's a code
`const isInputFocused = this.inputElement === document.activeElement;`
It checks is input element focused by comparing it with document.activeElement.
While when Suggest2 component is inside Shadom DOM it doesn't work. Therefore when it thinks that input isn't focused it clear input and closes dropdown.
In that case we need to check shadowRoot.activeElement instead.

I propose next code to be used instead:
```
const root = (this.inputElement?.getRootNode?.() ?? document) as DocumentOrShadowRoot & Node;
const isInputFocused = this.inputElement === root.activeElement;
```
With that change it will look actual root whether it document or shadowRoot

kutensky:

Changed detecting is input focused by comparing with input's rootNode rather than with window

### DIFF
--- a/packages/select/src/components/suggest/suggest2.tsx
+++ b/packages/select/src/components/suggest/suggest2.tsx
@@ -316,7 +316,8 @@ export class Suggest2<T> extends AbstractPureComponent2<Suggest2Props<T>, Sugges
     // Note that we defer to the next animation frame in order to get the latest document.activeElement
     private handlePopoverInteraction = (nextOpenState: boolean, event?: React.SyntheticEvent<HTMLElement>) =>
         this.requestAnimationFrame(() => {
-            const isInputFocused = this.inputElement === document.activeElement;
+            const root = (this.inputElement?.getRootNode?.() ?? document) as DocumentOrShadowRoot & Node;
+            const isInputFocused = this.inputElement === root.activeElement;
 
             if (this.inputElement != null && !isInputFocused) {
                 // the input is no longer focused, we should close the popover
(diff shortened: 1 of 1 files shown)


palantirtech:
Thanks for your interest in [palantir/blueprint](https://github.com/palantir/blueprint), @kutensky! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

adidahiya (approved):
nice, thanks @kutensky! I think we'll have to audit other places where we use `document.activeElement`. it's probably worth having a shared helper function to get the active element which we can use across various components -- I'll implement that in a follow-up PR.

kutensky:
Thank you @adidahiya for such a quick response and for adding support on other components. That helps a lot.

PR #5468, opened 2022-07-26
[core] feat(MultistepDialog): new DialogStep "disabled" prop

### LINKED ISSUE #4871: [MultistepDialog] Navigation to later steps is allowed via side navigation when `nextButtonProps` disallows it
#### Environment

- __Package version(s)__: @blueprintjs/core@3.48.0

#### Code Sandbox

https://codesandbox.io/s/blueprint-sandbox-multistepdialog-4di0d

#### Steps to reproduce

(Using the example in the Code Sandbox)
1. Advance through each step of the `<MultistepDialog>` to the end.
2. See how each step must be "valid" before you can progress.
2. Before submitting, go back to an earlier step via the "Back" buttons or via the side navigation.
3. Uncheck the checkbox to make the step invalid.
4. ✅ See that the "Next" button on the step is now disabled (according to `nextButtonProps` on that step)
5. ❌ See that you can still navigate to the next step via the side navigation, as it was previously visited.

#### Actual behavior

image

#### Expected behavior

image

If the "Next" button of the current `<DialogStep>` is disabled through the `nextButtonProps`, I do not expect to be able to navigate further forwards through the steps via the side navigation; even if I have advanced further previously.

### LINKED ISSUE #4872: [MultistepDialog] Allow side navigation to be disabled (e.g. when submitting)
#### Environment

- __Package version(s)__: @blueprintjs/core@3.48.0

#### Feature Request

1. Add a `disabled` prop to `<MultistepDialog>` that would allow the side navigation to be disabled (e.g when the dialog is being submitted)

1. Perhaps if specified `true` (and ignored as `false | undefined`), it could also combine & override any individual `disabled` props for all the other steps' `nextButtonProps` and `backButtonProps`? (You can see in my Code Sandbox example, I am passing `|| submitting` through to most of them)

#### Code Sandbox

https://codesandbox.io/s/blueprint-sandbox-multistepdialog-4di0d

#### Steps to reproduce

(Using the example in the Code Sandbox)
1. Advance through each step of the `<MultistepDialog>` to the end.
2. See how each step must be "valid" before you can progress.
3. On the last step, click the "Submit" button to trigger a fake submit state.
4. ❌ See that you can still navigate around between past steps via the side navigation whilst the form submit is ongoing.

### LINKED ISSUE #5041: [Multistep Dialog] Allow navigation via sidebar steps to be disabled
#### Feature request

Add the ability to disable navigation to different steps in the sidebar.

##### Use case

In complex wizards where validation of each step is dependant on the previous results, we do not necessarily want users to be able to navigate through the wizard using the side panel. The navigation should purely be done via the next button to allow proper validation on each step

#### Examples

`<MultiStepDialog ... sidebarNavigation={false} />`

christian1koch:
#### Checklist

- [x] Includes tests
- [x] Update documentation (happens automatically since the prop description was added)

Prop: panelDisabled is added to DialogStep. 

This allows to disable navigation on a specific panel of a DialogStep, which by adding the props to all DialogSteps, can disable the navigation through the panels altogether. This would solve https://github.com/palantir/blueprint/issues/5041 and potentially allow the fixes for : https://github.com/palantir/blueprint/issues/4872 & https://github.com/palantir/blueprint/issues/4871, since the dialogSteps could be disabled when submitting the Dialog on the last step.

The decision for having a step by step disable instead of an complete props.disableNavigation on the MultiStepDialog, is that there could be cases to where only some steps would like to be disabled and others not, thus having more control over the Dialog overall, and having a full disabled navigation state could be easily implemented using this prop.

- Code Quality for the implementation, mainly on the css implementation (any improvements are welcome).
- That disabled spanel should look exactly like when the step has been visited.

#### Screenshot

Step 2 disabled:
image

Dark Theme:
image


palantirtech:
Thanks for your interest in [palantir/blueprint](https://github.com/palantir/blueprint), @christian1koch! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

christian1koch:
This PR is ready for review.

adidahiya:
Thanks for the update @christian1koch, I will review this week.

packages/core/src/components/dialog/dialogStep.tsx
```diff
@@ -57,6 +57,10 @@ export interface IDialogStepProps extends Props, Omit<HTMLDivProps, "id" | "titl
      * Props for the next button.
      */
     nextButtonProps?: DialogStepButtonProps;
+    /**
+     * Disable panel for side navigation
+     */
+    panelDisabled?: boolean;
```
adidahiya: I think this could simply be called `disabled`

```suggestion
    disabled?: boolean;
```

packages/core/src/components/dialog/dialogStep.tsx
```diff
@@ -57,6 +57,10 @@ export interface IDialogStepProps extends Props, Omit<HTMLDivProps, "id" | "titl
      * Props for the next button.
      */
     nextButtonProps?: DialogStepButtonProps;
+    /**
+     * Disable panel for side navigation
```
adidahiya:
```suggestion

    /**
     * Setting this to `false` will disable navigation to this step's panel via the steps list in the sidebar.
```

packages/core/src/components/dialog/_multistep-dialog.scss
```diff
@@ -160,6 +160,12 @@ $step-radius: $pt-border-radius * 2 !default;
       background: $dark-gray4;
     }
   }
+  &.#{$ns}-dialog-step-viewed.#{$ns}-dialog-step-disabled {
```
adidahiya: I think some of this Sass implementation could be cleaned up a bit to reduce duplication... I will take a stab at it by pushing a commit to the PR

adidahiya (changes requested):
I think we should also add a demo for this prop to the docs example, something like a "Disable second step" toggle. Then we could easily test what the dialog looks like when `disable={true}` for one of the steps

adidahiya:
I pushed some of my suggested improvements and added a toggle to the docs example so that we can interactively test this new feature. There's still a problem where you can navigate to the disabled step in the example. See my code comment about this. I chose not to disable navigation via `nextButtonProps` / `backButtonProps` because that would defeat the purpose of adding the new toggle to the docs example (we wouldn't be able to move past the first step to see the new styling).

We'll have to think some more about how this `disabled` prop interacts with the footer actions rendered by MultistepDialog... users may want the option to either:
1. skip disabled steps ("Next" should go to the next available non-disabled step), or
1. automatically disable the "Next" button if the next step is `disabled={true}`

christian1koch:
I agree with your previous comment. I would vote for the first option and my reasoning behind this, is that there could be cases where you would want to "skip a step" depending on something selected on the current step. I believe the first option gives this component more possibilities to be used in different cases, while the second one would make the this prop too restrictive.

The first option also allows the user to recreate the second option, by disabling the following steps with the same conditions that disable the next button on the current step.

Overall IMO, the first option would make this component more "powerful". Any thoughts about this, @adidahiya?

adidahiya:
@christian1koch sounds good, let's go with the first option, to skip disabled steps. Can you add that to this PR?

adidahiya:
friendly ping @christian1koch, are you interested in finishing up this PR?

christian1koch:
@adidahiya Yes, sorry I've been kinda busy, I will start working on it this weekend most likely.

adidahiya:
closing PR as stale for now, but these issues are still open for PRs

PR #5471, opened 2022-07-27
feat: MenuItem2 component

adidahiya:
#### Checklist

- [x] Includes tests
- [x] Update documentation

- Create new MenuItem2 component in @blueprintjs/popover2, a replacement for MenuItem which uses Popover2 instead of Popover to render submenus
  - Note: it would be more convenient (and less of an upgrade burden on consumers) if we could only deprecate usage of MenuItem with `children` defined (those which have submenus), but we cannot do that with TypeScript's JSX type checking features, so we must deprecate MenuItem completely to be safe.
- Deprecate MenuItem
- Migrate MenuItem -> MenuItem2 across all non-deprecated Blueprint components in this repo

No regressions in MenuItem2 docs example compared to MenuItem

#### Screenshot

MenuItem2 example:

image

Updated Popover2 example:

image

Updated Select2 example:

image

image

Updated icon docs page:

image

Updated ContextMenu2 example:

image

Updated table example:

image


packages/table/src/interactions/menus/copyCellsMenuItem.tsx
```diff
@@ -16,14 +16,14 @@
...
-import { MenuItem, MenuItemProps } from "@blueprintjs/core";
+import { MenuItem2, MenuItem2Props } from "@blueprintjs/popover2";
 
 import { Clipboard } from "../../common/clipboard";
 import { TABLE_COPY_FAILED } from "../../common/errors";
 import { Regions } from "../../regions";
 import { IMenuContext } from "./menuContext";
 
-export interface ICopyCellsMenuItemProps extends MenuItemProps {
+export interface ICopyCellsMenuItemProps extends MenuItem2Props {
```
adidahiya, line 26: Note that this is a small breaking change, but in practice I think it should be ok for consumers since I didn't see any usage of `popoverProps: Partial<IPopoverProps>` for this component in any of our code bases.

PR #5473, opened 2022-08-01
[datetime2] feat(DateInput2): remove target wrapper, fix fill prop

### LINKED ISSUE #5472: DateInput2 fill prop not working
#### Environment

- __Package version(s)__: @blueprintjs/datetime2 v0.4.0 - v0.4.2

#### Steps to reproduce

render `<DateInput2 fill={true}>`, or try the option in the [docs example](https://blueprintjs.com/docs/#datetime2/date-input2)

#### Actual behavior

image

#### Expected behavior

Input should take up `width: 100%`

adidahiya:
#### Checklist

- [ ] Includes tests
- [x] Update documentation

- Refactor DateInput2 to use Popover2's `renderTarget` API, like we do in Select2, Suggest2, and MultiSelect2. Note that this is a breaking change to the DOM layout, as we have combined the popover target and input group elements into one:
```diff
- <span aria-haspopup="true" class="bp4-date-input bp4-popover2-target">
-   <div class="bp4-input-group">
-     <input>
- </span>
+ <div class="bp4-input-group bp4-date-input bp4-popover2-target">
+   <input aria-haspopup="true">
+ </div>
```
- Fix `fill` prop so it now works as expected

N/A

#### Screenshot

image

### DIFF
--- a/packages/datetime2/src/components/date-input2/dateInput2.tsx
+++ b/packages/datetime2/src/components/date-input2/dateInput2.tsx
@@ -35,7 +35,7 @@ import {
     DatePickerShortcut,
     DatePickerUtils,
 } from "@blueprintjs/datetime";
-import { Popover2, Popover2Props } from "@blueprintjs/popover2";
+import { Popover2, Popover2Props, Popover2TargetProps } from "@blueprintjs/popover2";
 
 import * as Classes from "../../common/classes";
 import { isDateValid, isDayInRange } from "../../common/dateUtils";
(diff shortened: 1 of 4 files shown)


packages/datetime2/src/components/date-input2/dateInput2.tsx
```diff
@@ -570,40 +571,75 @@ export const DateInput2: React.FC<DateInput2Props> = React.memo(function _DateIn
     const shouldShowErrorStyling =
         !isInputFocused || inputValue === props.outOfRangeMessage || inputValue === props.invalidDateMessage;
 
+    // We use the renderTarget API to flatten the rendered DOM and make it easier to implement features like
+    // the "fill" prop. Note that we must take `isOpen` as an argument to force this render function to be called
```
adidahiya: this copy-pasted comment is a little wrong, the `isOpen` part is no longer necessary to point out where we have a dependency list (compared to the class component where this comment came from)

PR #5475, opened 2022-08-01
[core] fix(Tag): remove aria-labelledby attribute

bvandrc:

`aria-labelledby` should only apply when a `role` is passed-- should not be on the Tag because the Tag is just a `span` and has no `role` prop

### DIFF
--- a/packages/core/src/components/tag/tag.tsx
+++ b/packages/core/src/components/tag/tag.tsx
@@ -147,6 +147,7 @@ export class Tag extends AbstractPureComponent2<TagProps> {
             ...htmlProps
         } = this.props;
         const isRemovable = Utils.isFunction(onRemove);
+        const hasChildren = !isReactNodeEmpty(children);
         const tagClasses = classNames(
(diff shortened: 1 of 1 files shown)


adidahiya:
> aria-labelledby should only apply when a role is passed

documentation link for this?

bvandrc:
> > aria-labelledby should only apply when a role is passed
> 
> documentation link for this?

we are now failing the aXe `aria-allowed-attr` rule because we are applying an `aria-labelledby` prop to a `span` element. https://accessibilityinsights.io/info-examples/web/aria-allowed-attr/

If this `span` were to have a different `role` assigned, this would be allowed, but as of now, the Tag span is just a span.

adidahiya (approved):

adidahiya:
I couldn't find any [role](https://www.w3.org/TR/wai-aria/#roles_categorization) which is suitable for all/any Tags in general, so I guess it's better to remove this attribute if we can't set `role` ourselves.

PR #5477, opened 2022-08-03
[eslint-plugin] feat(no-deprecated-components): more components

adidahiya:

- Add `@blueprintjs/datetime` package to the list of packages which the rule checks for deprecated imports
- Add components to list of deprecations / disallowed imports:
  - DateInput
  - DateTimePicker
  - MenuItem

Difficulty of rolling this out, as it's kind of a breaking change for the lint rule (builds with the rule set to "error" will start failing).


adidahiya:
> Difficulty of rolling this out, as it's kind of a breaking change for the lint rule (builds with the rule set to "error" will start failing).

The plan here is to introduce rule configuration which allows users to opt-out of strict "error" violations for specific Blueprint packages (downgrading them to "warn" violations). This could allow, for example, someone to migrate away from deprecated `@blueprintjs/core` components while continuing to use deprecated `@blueprintjs/datetime` components during an incremental migration effort.

PR #5478, opened 2022-08-03
[eslint-plugin] feat: split no-deprecated-components rule

adidahiya:
This PR creates new ESLint rules which flag usage of deprecated components specific to each `@blueprintjs` package:
  - `@blueprintjs/no-deprecated-core-components`
  - `@blueprintjs/no-deprecated-datetime-components`
  - `@blueprintjs/no-deprecated-select-components`
  - `@blueprintjs/no-deprecated-timezone-components`

__Rationale__ (also in README): In migrations of large code bases, it may be useful to apply more granular rule configuration of "no-deprecated-components" to make incremental progress towards the newer APIs. This allows you, for example, to flag deprecated `@blueprintjs/core` component usage as :x: errors while allowing deprecated components from other packages
to pass as lint ⚠️ warnings.

Note that `@blueprintjs/no-deprecated-components` has not been removed, it is still available, and is enabled in the `"@blueprintjs/recommended"` config (this config is not used in our internal monorepos, btw, just provided for convenience and seems to be an ESLint plugin convention). I refactored the implementation of this rule into a reusable function so that all 5 rules share the same implementation.

Following up from https://github.com/palantir/blueprint/pull/5477#issuecomment-1204213250 and other conversations... the other idea I had of introducing a configuration object for `@blueprintjs/no-deprecated-components` to set "error"/"warn" levels for the different Blueprint packages turned out to be messy and less legible than the approach taken in this PR. It's better to let ESLint's infrastructure handle the "error"/"warn" option toggling, and splitting into separate rules also helps with legibility because we can search the codebase for `// eslint-disable-next-line @blueprintjs/no-deprecated-core-components` lint flags.


packages/eslint-plugin/src/rules/no-deprecated-components/no-deprecated-components.ts
```diff
@@ -0,0 +1,33 @@
...
+        DateRangeInput: "DateRangeInput2",
+        DateTimePicker: "DatePicker",
+        MenuItem: "MenuItem2",
+        MultiSelect: "MultiSelect2",
+        PanelStack: "PanelStack2",
+        Popover: "Popover2",
+        Select: "Select2",
+        Suggest: "Suggest2",
+        TimezonePicker: "TimezoneSelect",
+        Tooltip: "Tooltip2",
```
styu: should the different sets be variables exported from the other rules and spread into here instead? so we don't need to keep 2 lists up to date

adidahiya:
@styu this is ready for another round of review when you get the chance

packages/datetime/test/dateInputTests.tsx
```diff
@@ -14,6 +14,14 @@
  * limitations under the License.
  */
 
+/**
+ * @fileoverview This component is DEPRECATED, and the code is frozen.
+ * All changes & bugfixes should be made to DateInput2 in the datetime2
+ * package instead.
+ */
+
+/* eslint-disable deprecation/deprecation, @blueprintjs/no-deprecated-components */
```
adidahiya, line 23: I realize it is redundant to have both of these rules on as "error" violations, but I'm keeping them both so we can dogfood our own lint rule.

styu (approved):

PR #5481, opened 2022-08-04
[table] feat: ColumnHeaderCell2, TruncatedFormat2, JSONFormat2

### LINKED ISSUE #5480: @blueprintjs/table TruncatedFormat and ColumnHeaderCell still use Popover
Need v2 variants that use Popover2 instead of Popover.

adidahiya:

- Create ColumnHeaderCell2, TruncatedFormat2, and JSONFormat2 components which use Popover2 instead of Popover
- Deprecate ColumnHeaderCell, TruncatedFormat, and JSONFormat components
- Import `blueprint-popover2.css` in table-dev-app, add documentation for this new requirement

This PR also fixes an undocumented issue in the table "Formatting" example where the cell truncation was not being detected, and thus the overflow popover was never shown. It's now fixed with `detectTruncation={true}`:

image

~Minor breaking change in table, requiring users to import `blueprint-popover2.css`. We expect most consumers to be using @blueprintjs/popover2 already at this point, so this should not be too much of a burden (likely a no-op).~

We're avoiding the breaking change by splitting these out into new components. Consumers of the new components will need to add the CSS import when they migrate.

#### Screenshot

ColumnHeaderCell2

image

TruncatedFormat2

image


adidahiya:
Changed my mind -- I think we should split these out into new v2 components, as the breaking change is not worth it. If we make the change directly to ColumnHeaderCell and TruncatedFormat, then some consumers who have not migrated to Popover2 yet will get broken column header menus immediately after this upgrade, and that feel unnecessary.

PR #5486, opened 2022-08-06
[core] fix(OverflowList): edge case in shouldComponentUpdate

braeden:
#### Checklist

- [ ] Includes tests
- [ ] Update documentation

#### Changes proposed in this pull request (update):

- Ensure we compare props reference equality in addition to the existing logic to **ensure** a re-render when props change. This is an addition to the logic introduced here: https://github.com/palantir/blueprint/pull/3278.

#### Context:
Through some testing, I've been able to repro the problematic condition: props change on the same tick as both `state !== prevState` AND `shallowCompare(state, prevState) === true`.  

This means a re-render doesn't happen when it should. The original code was written to aggressively re-render (even in the case that the state doesn't change). But it failed to account for the case where: props change, state changes, but shallow compare is truthy (all in a single `shouldComponentUpdate`). 

----------
Here's a log demonstrating this happen:
```
 console.log("SCU called! [props, state, shallowCompare, return]", this.props === _nextProps, this.state === nextState, shallowCompareKeys(this.state, nextState), !(this.state !== nextState && shallowCompareKeys(this.state, nextState)));
```
image

---------
Note: This component is reasonable suited returning to a PureComponent in the future, but that will be a breaking change.


palantirtech:
Thanks for your interest in [palantir/blueprint](https://github.com/palantir/blueprint), @braeden! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

adidahiya (approved):

PR #5488, opened 2022-08-08
[icons] fix: align 16px "layers" icon to pixel grid

quentinlepape:
Updated the 16px layers icon to make sure its shape is aligned with the pixel grid.


quentinlepape:
@invliD I updated the 16px `layers` icon per our discussion (the 20px version is actually aligned to the pixel grid already).

adidahiya (approved):

PR #5495, opened 2022-08-10
[eslint-plugin] feat: only lint MenuItem with popoverProps attr

adidahiya:
#### Checklist

- [x] Includes tests
- [ ] Update documentation

Improve `@blueprintjs/no-deprecated-core-components` & `@blueprintjs/no-deprecated-components` lint rule to understand components usage which is deprecated iff a specific prop is used. This is particularly useful for `<MenuItem>`, where we don't want to force consumers to unnecessarily migrate to `<MenuItem2>` in the overwhelming majority use case... we only need them to migrate when `popoverProps` is used. This will greatly reduce the code diff required in migrations.

Implementation & test cases

### DIFF
--- a/packages/eslint-plugin/src/rules/no-deprecated-components/createNoDeprecatedComponentsRule.ts
+++ b/packages/eslint-plugin/src/rules/no-deprecated-components/createNoDeprecatedComponentsRule.ts
@@ -87,6 +87,38 @@ export function createNoDeprecatedComponentsRule(
                 );
             }
 
+            function checkDeprecatedComponentAndProp(
+                jsxOpeningElementChildNode: TSESTree.Node,
+                elementName: string,
+                openingElementNode: TSESTree.JSXOpeningElement,
+            ) {
+                const deprecatedProp = openingElementNode.attributes.find(
+                    attribute =>
+                        attribute.type === TSESTree.AST_NODE_TYPES.JSXAttribute &&
+                        attribute.name.type === TSESTree.AST_NODE_TYPES.JSXIdentifier &&
+                        deprecatedToNewComponentMapping[`${elementName}.${attribute.name.name}`] != null,
+                );
+
+                if (deprecatedProp === undefined) {
+                    return;
+                }
+
+                const deprecatedComponentKey = Object.keys(deprecatedToNewComponentMapping).find(
+                    key => key.includes(".") && key.split(".")[0] === elementName,
+                );
+                const deprecatedPropName = ((deprecatedProp as TSESTree.JSXAttribute).name as TSESTree.JSXIdentifier)
+                    .name;
+                context.report({
+                    data: {
+                        deprecatedComponentName: elementName,
+                        deprecatedPropName,
+                        newComponentName: deprecatedToNewComponentMapping[deprecatedComponentKey!],
+                    },
+                    messageId: "migrationWithPropUsage",
+                    node: jsxOpeningElementChildNode,
+                });
+            }
+
             // Get the list of all deprecated imports from packages included in the provided list
             return {
                 "Program > ImportDeclaration": (node: TSESTree.ImportDeclaration) => {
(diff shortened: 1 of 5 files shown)


packages/eslint-plugin/src/rules/no-deprecated-components/createNoDeprecatedComponentsRule.ts
```diff
@@ -112,10 +129,34 @@ export function createNoDeprecatedComponentsRule(
...
                         });
+                    } else if (isOpeningElement(node.parent)) {
+                        // check <DeprecatedComponent withDeprecatedProp={...}> syntax
+                        const deprecatedProp = node.parent.attributes.find(
+                            attribute =>
+                                attribute.type === TSESTree.AST_NODE_TYPES.JSXAttribute &&
+                                attribute.name.type === TSESTree.AST_NODE_TYPES.JSXIdentifier &&
+                                deprecatedToNewComponentMapping[`${node.name}.${attribute.name.name}`] != null,
+                        );
+                        if (deprecatedProp !== undefined) {
```
styu: nit: can unnest if we return early instead
```suggestion
                        if (deprecatedProp === undefined) {
                            return;
                        }
```

packages/eslint-plugin/src/rules/no-deprecated-components/createNoDeprecatedComponentsRule.ts
```diff
@@ -112,10 +129,34 @@ export function createNoDeprecatedComponentsRule(
...
+                            const deprecatedComponentKey = Object.keys(deprecatedToNewComponentMapping).find(
+                                key => key.split(".")[0] === node.name,
```
styu: it is technically possible to have both `MenuItem` and `MenuItem.popoverProps` as keys, in which case this would return the wrong key, right?
```suggestion
                                key => key.indexOf(".") >= 0 && key.split(".")[0] === node.name,
```

packages/eslint-plugin/src/rules/no-deprecated-components/createNoDeprecatedComponentsRule.ts
```diff
@@ -137,6 +178,30 @@ export function createNoDeprecatedComponentsRule(
...
+                                    deprecatedComponentName: node.property.name,
+                                    deprecatedPropName: (
+                                        (deprecatedProp as TSESTree.JSXAttribute).name as TSESTree.JSXIdentifier
+                                    ).name,
+                                    newComponentName: deprecatedToNewComponentMapping[deprecatedComponentKey!],
+                                },
+                                messageId: "migrationWithPropUsage",
+                                node,
+                            });
+                        }
```
styu: is this code repeated, just for a different node type?

packages/eslint-plugin/src/rules/no-deprecated-components/createNoDeprecatedComponentsRule.ts
adidahiya: Yeah, I think it would fail in that case. The rule doesn't gracefully handle bad configuration right now, but I figure it's ok to be a little simplistic here since the configuration is not exposed to consumers -- it only lives inside the rule implementation in this package.

packages/eslint-plugin/src/rules/no-deprecated-components/createNoDeprecatedComponentsRule.ts
adidahiya: it's a little different, it uses `node.property.name` instead of `node.name`. but yes it's very similar to the other block. I think it would be a little convoluted / unreadable if we tried to DRY this code between the two syntax handlers.

packages/eslint-plugin/src/rules/no-deprecated-components/createNoDeprecatedComponentsRule.ts
adidahiya: fixed it up a little to guard against the edge case you mentioned

packages/eslint-plugin/src/rules/no-deprecated-components/createNoDeprecatedComponentsRule.ts
adidahiya: on second thought, I've refactored this bit into a function shared between the two handlers, it's still legible

PR #5509, opened 2022-08-23
[eslint-plugin] feat: no-deprecated-type-references rule

adidahiya:

New ESLint rule `@blueprintjs/no-deprecated-type-references` to flag & fix usage of deprecated interfaces in preparation for Blueprint v5.0.


packages/eslint-plugin/src/index.ts
```diff
@@ -26,6 +26,7 @@ module.exports = {
                 "@blueprintjs/classes-constants": "error",
                 "@blueprintjs/html-components": "error",
                 "@blueprintjs/no-deprecated-components": "error",
+                "@blueprintjs/no-deprecated-type-references": "error",
```
styu, line 29: since this is autofixable, can/should we promote it in the default eslint config?

packages/eslint-plugin/src/rules/no-deprecated-type-references.ts
```diff
@@ -0,0 +1,339 @@
...
+
+for (const [packageName, deprecatedTypeNames] of Object.entries(DEPRECATED_TYPE_REFERENCES_BY_PACKAGE)) {
+    for (const typeName of deprecatedTypeNames) {
+        const newTypeName = typeName.slice(1);
+        // "IProps": "Props"
+        DEPRECATED_TYPES_MAPPING[typeName] = newTypeName;
+        // "Props": "@blueprintjs/core"
+        NEW_TYPE_PACKAGE_MAPPING[newTypeName] = packageName;
+    }
+}
```
styu: i kind of wonder if this is logic we should be running unilaterally as part of importing this rule, or if it should instead be worked into the logic within the rule but without the for loop.

packages/eslint-plugin/test/no-deprecated-type-references.test.ts
```diff
@@ -0,0 +1,153 @@
...
+                    data: { deprecatedTypeName: "ITimezoneItem", newTypeName: "TimezoneItem" },
+                },
+            ],
+            output: dedent`
+                import { Select } from "@blueprintjs/select";
+                import { TimezoneItem } from "@blueprintjs/timezone";
+                const MySelect = Select.ofType<TimezoneItem>();
+            `,
+        },
+        {
```
styu, line 126: can you add a test for when there are multiple imports from the same package?

packages/eslint-plugin/src/rules/no-deprecated-type-references.ts
adidahiya: moved it into the `create` callback

packages/eslint-plugin/test/no-deprecated-type-references.test.ts
adidahiya: I tried to do this, but for some reason the rule tester fixer doesn't like to run multiple auto-fixes. so when I write a test like this:

```tsx
                import { IItemRendererProps, ISelectProps } from "@blueprintjs/select";
                const mySelectProps: ISelectProps = { items: [], renderItem };
                const renderItem = (_item: any, { handleClick, modifiers }: IItemRendererProps) => {
                    return <MenuItem {...modifiers} onClick={handleClick} />;
                };
```

expecting it to be autofixed to:

```tsx
                import { ItemRendererProps, SelectProps } from "@blueprintjs/select";
                const mySelectProps: SelectProps = { items: [], renderItem };
                const renderItem = (_item: any, { handleClick, modifiers }: ItemRendererProps) => {
                    return <MenuItem {...modifiers} onClick={handleClick} />;
                };
```

then it doesn't pass, since the rule tester only runs the first auto-fix. I don't want to write a test with weird expected output showing only one of the deprecated type symbols being fixed. I can look into if this is a bug on ESLint's part...

styu: is it possible to test without the fixer, but show that both symbols are flagged as errors instead?

styu (approved):

packages/eslint-plugin/test/no-deprecated-type-references.test.ts
adidahiya: discussed offline, ESLint does not allow testing fixable rules without running their fixers. probably another FR for ESLint given the limitations of fixer output in RuleTester (see https://github.com/eslint/eslint/issues/11187#issuecomment-470990425)

PR #5515, opened 2022-08-25
fix: declare @types/react as optional peer dependency

adidahiya:
#### Changes proposed in this pull request:

Add `@types/react` as an optional peer dependency in package.json for each package that depends on those types in its generated `.d.ts` files.

This is in line with the approach taken by other modern OSS React UI library packages like [@mui/material](https://github.com/mui/material-ui/blob/64e5dc1656e2c70e1b692618a5121ff67c645e78/packages/mui-material/package.json#L47) and [react-bootstrap](https://github.com/react-bootstrap/react-bootstrap/blob/5dc9ed4cd991a50da14b7e004a3d4e7bb0709bc6/package.json#L149).

#### Reviewers should focus on:

If this actually fixes the problem we've been having in PNPM installs, where sometimes `@types/react` is not available to `@blueprintjs/core` to compile with `tsc` with `skipLibCheck: false`.

PR #5518, opened 2022-08-26
fix: widen versions allowed for @types/react peer dep

adidahiya:
based on feedback from https://github.com/palantir/blueprint/pull/5515#issuecomment-1228658796


gluxon:
Thanks!

PR #5524, opened 2022-09-02
[core] feat: support windows high contrast mode

styu:

This adds windows high contrast mode support for most components in Blueprint:
- Buttons
- Form controls (radio buttons, switches, checkboxes)
- Inputs
- Callouts
- Cards + elevations
- HTML tables
- Menus
- Popovers 
- Tooltips 
- Navbar 
- Slider
- Tags

For the components affected, do the following show up correctly in windows high contrast mode:
- disabled states
- active states
- hover states
- default states

#### Screenshot

image


adidahiya:
Based on this [Microsoft Edge blog post](https://blogs.windows.com/msedgedev/2020/09/17/styling-for-windows-high-contrast-with-new-standards-for-forced-colors/), I think it would be more appropriate to use the following media queries:

```css
@media (forced-colors: active) and (prefers-color-scheme: dark) {
  // Windows High Contrast dark theme 
}

@media (forced-colors: active) and (prefers-color-scheme: light) {
  // Windows High Contrast light theme 
}
```

see https://developer.mozilla.org/en-US/docs/Web/CSS/@media/forced-colors

styu:
@adidahiya sounds good, i'm picking this back up today!

adidahiya:
Visual review only (code review coming separately):

This is a huge improvement over the current state of Blueprint in high contrast mode, nice work! I reviewed all the components in the docs preview and found some areas we could still do better.

## To address in this PR

### Minimal button groups have extra borders

We should make these look the same as regular buttons, without the double border between adjacent buttons:

Screen Shot 2022-09-26 at 4 31 37 PM

### Time picker inputs have no borders

Screen Shot 2022-09-26 at 4 40 15 PM



## To address in separate PRs

Listed below roughly in order of importance:

### Dialogs and Drawers should get a border to separate them from underlying content

(should look similar to popover/tooltips and dark theme dialogs)

Screen Shot 2022-09-26 at 4 38 54 PM

Screen Shot 2022-09-26 at 4 39 15 PM

### Active tab indicator is hard/impossible to see

Screen Shot 2022-09-26 at 4 35 45 PM

### Progress bar is invisible

Screen Shot 2022-09-26 at 4 34 51 PM

### Slider filled track is invisible

Screen Shot 2022-09-26 at 4 37 48 PM

### Non ideal state icon stroke has low contrast

we shouldn't apply the opacity adjustment in high contrast mode:

Screen Shot 2022-09-26 at 4 34 15 PM

### Skeleton styling has no effect

maybe we should make the text color transparent?

Screen Shot 2022-09-26 at 4 35 18 PM

### Tables

Not 100% sure what we should do here, but adding borders between all cells probably makes the most sense. Also a thicker border between row/col headers and the table body

Screen Shot 2022-09-26 at 4 41 34 PM

packages/core/src/common/_variables.scss
```diff
@@ -145,3 +145,11 @@ $pt-dark-popover-box-shadow: 0 0 0 1px $pt-dark-popover-border-color,
 $pt-dark-tooltip-box-shadow: 0 2px 4px rgba($black, $pt-dark-drop-shadow-opacity),
                              0 8px 24px rgba($black, $pt-dark-drop-shadow-opacity) !default;
 $pt-dark-toast-box-shadow: $pt-dark-elevation-shadow-3 !default;
+
+// Colors used for Windows high contrast mode
+// Because high contrast mode doesn't actually obey any colors we define, but uses system colors,
+// we define these variables in terms of System colors so that it's easier to understand the intent.
```
adidahiya, line 151: I think it might be helpful to link to documentation here, like https://developer.mozilla.org/en-US/docs/Web/CSS/color_value/system_color_keywords

packages/core/src/components/editable-text/_editable-text.scss
```diff
@@ -31,6 +31,12 @@
     box-shadow: input-transition-shadow($input-shadow-color-focus, true), $input-box-shadow-focus;
   }
 
+  @media (forced-colors: active) and (prefers-color-scheme: dark) {
+    &::before {
```
adidahiya: I think we should add disabled styles as well; we don't want to keep showing a border in this case:

image

packages/core/src/components/forms/_controls.scss
```diff
@@ -34,10 +34,21 @@ $control-indicator-spacing: $pt-grid-size !default;
...
+      border: 1px solid $pt-high-contrast-mode-active-background-color;
+    }
   }
 
   &:hover input#{$selector} ~ .#{$ns}-control-indicator {
     background-color: $control-checked-background-color-hover;
+
+    @media (forced-colors: active) and (prefers-color-scheme: dark) {
+      // Windows High Contrast dark theme
+      background-color: $pt-high-contrast-mode-active-background-color;
```
adidahiya, line 50: not exactly related to this line of code, but I noticed we seem to be getting a nice background color for disabled & active radio indicators:

Screen Shot 2022-09-26 at 5 23 06 PM

but we don't get the same for checkboxes:

Screen Shot 2022-09-26 at 5 23 20 PM

not a huge priority, but there may be an easy fix for this?

packages/core/src/components/forms/_controls.scss
```diff
@@ -412,6 +468,13 @@ $control-indicator-spacing: $pt-grid-size !default;
         position: absolute;
         transition: left $pt-transition-duration $pt-transition-ease;
         width: $switch-indicator-size;
+
+        @media (forced-colors: active) and (prefers-color-scheme: dark) {
+          // Windows High Contrast dark theme
+          border: 1px solid $pt-high-contrast-mode-border-color;
+          // Because we're using a border for the outline, we need to decrease the top margin
+          margin-top: 1px;
```
adidahiya: should this be `$switch-indicator-margin - 1px`?

packages/core/src/components/popover/_common.scss
```diff
@@ -161,6 +161,11 @@ $dark-popover-background-color: $dark-gray3 !default;
 
   .#{$ns}-popover-arrow-fill {
     fill: $background-color;
+
+    @media (forced-colors: active) and (prefers-color-scheme: dark) {
+      // Windows High Contrast dark theme
+      fill: $pt-high-contrast-mode-border-color;
```
adidahiya, line 167: minor: we may want to apply the same style to tooltips with intents, otherwise we get this:

image

adidahiya (changes requested):

packages/core/src/components/forms/_controls.scss
styu: hm i can fix - in any case, the disabled color is wrong for both

packages/core/src/components/popover/_common.scss
styu: whoops good call

styu:
I can address dialog + drawer styling in this same PR (I tried to hit the commonly-used components, but for some reason missed those).

I did try to fix the active tab indicator, but that one's a bit tricky since it throws off all the spacing as a border

adidahiya:
latest changes look good :+1:

image

image

image

image

image

adidahiya (approved):

adidahiya:
I filed #5622 for the follow-up issues

PR #5530, opened 2022-09-07
fix(no-deprecated-type-references): no overlapping import fixes

adidahiya:
#### Checklist

- [x] Includes tests
- [ ] Update documentation

Fix the `@blueprintjs/no-deprecated-type-references` rule to return fixes properly so that multiple instances/uses of a deprecated type are all fixed, and not just the first one.

Check out the new test case. That was failing before this change, as only the first type reference of `IItemRendererProps` was getting replaced with `ItemRendererProps`.

The problem was that we were reporting a fix on the type reference _and_ the import of that type for _every_ instance of deprecated usage (two fixes for a single violation). ESLint will not apply overlapping fixes, so only the first fix list was being applied, even though all violations were being reported.


packages/eslint-plugin/test/no-deprecated-type-references.test.ts
```diff
@@ -138,6 +140,38 @@ ruleTester.run("no-deprecated-type-references", noDeprecatedTypeReferencesRule,
                 const mySelectProps: SelectProps = { items: [] };
             `,
         },
+
+        // Ensure that multiple references/uses of the _same_ deprecated type are all fixed
```
styu, line 144: can we add a test case that mixes both type reference + interface extending?

adidahiya: done

styu (approved):

PR #5534, opened 2022-09-10
[core] feat(MultistepDialog): tooltips on navigation buttons

### LINKED ISSUE #5519: Add tooltip to MultistepDialog final button
#### Environment

- __Package version(s)__: 

#### Feature request

I would like to be able to add a tooltip to the final button for the MultistepDialog. I am performing validation within the panel to determine if the button is disabled, but there is no way currently to provide an error message to the user explaining why the button is disabled.

ohcnivek:
#### Checklist

- [ ] Includes tests
- [ ] Update documentation

- adding an extra prop, ```tooltipContent```, to ```abstractButton.tsx``` props, which by association enables the prop  to exist on ```DialogStepButtonProps``` 
- renders tooltip on dialog step button if ```tooltipContent``` prop exists (user should define when it exists / doesnt, checkout the example. ```packages/docs-app/src/examples/core-examples/multistepDialogExample.tsx```)
- also adjusted the example in docs to reflect changes

- typescript quality / implementation logic — followed what @adidahiya suggested in  #5519 

#### Screenshot
https://user-images.githubusercontent.com/64056046/189471022-580de65c-4790-4e95-ae6e-0a862ee989e4.mov

### DIFF
--- a/packages/core/src/components/dialog/multistepDialog.tsx
+++ b/packages/core/src/components/dialog/multistepDialog.tsx
@@ -17,7 +17,8 @@
 import classNames from "classnames";
 import * as React from "react";
 
-import { Tooltip2 } from "../../../../popover2/src/tooltip2";
+import { Tooltip } from "@blueprintjs/core";
+
 import { AbstractPureComponent2, Classes, Position, Utils } from "../../common";
 import { DISPLAYNAME_PREFIX } from "../../common/props";
 import { Button, ButtonProps } from "../button/buttons";
@@ -233,63 +234,46 @@ export class MultistepDialog extends AbstractPureComponent2<MultistepDialogProps
 
         if (this.state.selectedIndex > 0) {
             const backButtonProps = steps[selectedIndex].props.backButtonProps ?? this.props.backButtonProps;
-
-            if (backButtonProps?.tooltipContent) {
-                buttons.push(
-                    <Tooltip2 content={backButtonProps?.tooltipContent} hoverOpenDelay={100}>
-                        <Button
-                            key="back"
-                            onClick={this.getDialogStepChangeHandler(selectedIndex - 1)}
-                            text="Back"
-                            {...backButtonProps}
-                        />
-                    </Tooltip2>,
-                );
-            } else {
+            const backButton = (
                 <Button
                     key="back"
                     onClick={this.getDialogStepChangeHandler(selectedIndex - 1)}
                     text="Back"
                     {...backButtonProps}
-                />;
+                />
+            );
+
+            if (backButtonProps?.tooltipContent != null) {
+                // eslint-disable-next-line deprecation/deprecation
+                buttons.push(<Tooltip content={backButtonProps?.tooltipContent}>{backButton}</Tooltip>);
+            } else {
+                buttons.push(backButton);
             }
         }
 
         if (selectedIndex === this.getDialogStepChildren().length - 1) {
-            if (this.props.finalButtonProps?.tooltipContent) {
-                buttons.push(
-                    <Tooltip2 content={this.props.finalButtonProps?.tooltipContent} hoverOpenDelay={100}>
-                        <Button intent="primary" key="final" text="Submit" {...this.props.finalButtonProps} />
-                    </Tooltip2>,
-                );
+            const finalButton = <Button intent="primary" key="final" text="Submit" {...this.props.finalButtonProps} />;
+            if (this.props.finalButtonProps?.tooltipContent != null) {
+                buttons.push(<Tooltip content={this.props.finalButtonProps?.tooltipContent}>{finalButton}</Tooltip>);
             } else {
-                buttons.push(<Button intent="primary" key="final" text="Submit" {...this.props.finalButtonProps} />);
+                buttons.push(finalButton);
             }
         } else {
             const nextButtonProps = steps[selectedIndex].props.nextButtonProps ?? this.props.nextButtonProps;
+            const nextButton = (
+                <Button
+                    intent="primary"
+                    key="next"
+                    onClick={this.getDialogStepChangeHandler(selectedIndex + 1)}
+                    text="Next"
+                    {...nextButtonProps}
+                />
+            );
 
-            if (nextButtonProps?.tooltipContent) {
-                buttons.push(
-                    <Tooltip2 content={nextButtonProps?.tooltipContent} hoverOpenDelay={100}>
-                        <Button
-                            intent="primary"
-                            key="next"
-                            onClick={this.getDialogStepChangeHandler(selectedIndex + 1)}
-                            text="Next"
-                            {...nextButtonProps}
-                        />
-                    </Tooltip2>,
-                );
+            if (nextButtonProps?.tooltipContent != null) {
+                buttons.push(<Tooltip content={nextButtonProps?.tooltipContent}>{nextButton}</Tooltip>);
             } else {
-                buttons.push(
-                    <Button
-                        intent="primary"
-                        key="next"
-                        onClick={this.getDialogStepChangeHandler(selectedIndex + 1)}
-                        text="Next"
-                        {...nextButtonProps}
-                    />,
-                );
+                buttons.push(nextButton);
             }
         }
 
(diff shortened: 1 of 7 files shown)


palantirtech:
Thanks for your interest in [palantir/blueprint](https://github.com/palantir/blueprint), @ohcnivek! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

packages/core/src/components/dialog/multistepDialog.tsx
```diff
@@ -17,6 +17,7 @@
 import classNames from "classnames";
 import * as React from "react";
 
+import { Tooltip2 } from "../../../../popover2/src/tooltip2";
```
adidahiya: ah, you can't reach across packages in the monorepo to import modules like this. you have to do it through the public package APIs.

furthermore, since @blueprintjs/core does not depend on @blueprintjs/popover2 (it's designed not to), we can't use `Tooltip2` in this component. you'll have to use `Tooltip` even though it's deprecated. this shouldn't be a problem, though, since `Tooltip` will work fine for this simple use case, and in Blueprint v5 it will get automatically/seamlessly upgraded to use the `Tooltip2` implementation.

packages/core/src/components/dialog/multistepDialog.tsx
```diff
@@ -233,30 +234,63 @@ export class MultistepDialog extends AbstractPureComponent2<MultistepDialogProps
         if (this.state.selectedIndex > 0) {
             const backButtonProps = steps[selectedIndex].props.backButtonProps ?? this.props.backButtonProps;
 
-            buttons.push(
+            if (backButtonProps?.tooltipContent) {
```
adidahiya: please avoid implicit type coercion here. instead use an explicit check:

```suggestion
            if (backButtonProps?.tooltipContent != null) {
```

packages/core/src/components/dialog/multistepDialog.tsx
```diff
@@ -233,30 +234,63 @@ export class MultistepDialog extends AbstractPureComponent2<MultistepDialogProps
...
+                buttons.push(
+                    <Tooltip2 content={backButtonProps?.tooltipContent} hoverOpenDelay={100}>
+                        <Button
```
adidahiya: please reduce the code repetition here by storing the button as a variable outside this conditional block:

```ts
const backButton = (
    <Button
        key="back"
        onClick={this.getDialogStepChangeHandler(selectedIndex - 1)}
        text="Back"
        {...backButtonProps}
    />
);
```

packages/core/src/components/dialog/multistepDialog.tsx
```diff
@@ -233,30 +234,63 @@ export class MultistepDialog extends AbstractPureComponent2<MultistepDialogProps
...
+                />;
+            }
         }
 
         if (selectedIndex === this.getDialogStepChildren().length - 1) {
-            buttons.push(<Button intent="primary" key="final" text="Submit" {...this.props.finalButtonProps} />);
+            if (this.props.finalButtonProps?.tooltipContent) {
+                buttons.push(
+                    <Tooltip2 content={this.props.finalButtonProps?.tooltipContent} hoverOpenDelay={100}>
+                        <Button intent="primary" key="final" text="Submit" {...this.props.finalButtonProps} />
```
adidahiya: same here, I'd rather have this button defined in a `finalButton` variable and used twice

packages/core/src/components/dialog/multistepDialog.tsx
```diff
@@ -233,30 +234,63 @@ export class MultistepDialog extends AbstractPureComponent2<MultistepDialogProps
...
                     {...backButtonProps}
-                />,
-            );
+                />;
+            }
         }
 
         if (selectedIndex === this.getDialogStepChildren().length - 1) {
-            buttons.push(<Button intent="primary" key="final" text="Submit" {...this.props.finalButtonProps} />);
+            if (this.props.finalButtonProps?.tooltipContent) {
```
adidahiya: same here, please avoid implicit type coercion, and use an explicit check instead:

```suggestion
            if (this.props.finalButtonProps?.tooltipContent != null) {
```

packages/core/src/components/dialog/multistepDialog.tsx
```diff
@@ -233,30 +234,63 @@ export class MultistepDialog extends AbstractPureComponent2<MultistepDialogProps
...
-            buttons.push(
-                <Button
-                    intent="primary"
-                    key="next"
-                    onClick={this.getDialogStepChangeHandler(selectedIndex + 1)}
-                    text="Next"
-                    {...nextButtonProps}
-                />,
-            );
+            if (nextButtonProps?.tooltipContent) {
```
adidahiya:
```suggestion
            if (nextButtonProps?.tooltipContent != null) {
```

packages/core/src/components/dialog/multistepDialog.tsx
```diff
@@ -233,30 +234,63 @@ export class MultistepDialog extends AbstractPureComponent2<MultistepDialogProps
...
+                buttons.push(
+                    <Tooltip2 content={nextButtonProps?.tooltipContent} hoverOpenDelay={100}>
+                        <Button
```
adidahiya: same here, I'd rather have this button defined in a `nextButton` variable and used twice

packages/docs-app/src/examples/core-examples/multistepDialogExample.tsx
```diff
@@ -115,7 +115,10 @@ export class MultistepDialogExample extends React.PureComponent<
                     icon="info-sign"
                     navigationPosition={position}
                     onClose={this.handleClose}
-                    nextButtonProps={{ disabled: this.state.value === undefined }}
+                    nextButtonProps={{
+                        disabled: this.state.value === undefined,
+                        tooltipContent: this.state.value === undefined ? "Needs an input." : null,
```
adidahiya: no need to use `null` here:

```suggestion
                        tooltipContent: this.state.value === undefined ? "Needs an input" : undefined,
```

adidahiya (changes requested):

packages/core/src/components/dialog/multistepDialog.tsx
ohcnivek: interesting — out of curiosity, what's the reason behind not being able to import across monorepos like that (other than @blueprint/core not depending on @blueprint/popover2)?

also, after editing the import to ```import Tooltip from @blueprint/core```,  it continues to fail to compile — is there something im missing?

adidahiya: One of the main ideas/principles in monorepos is to create clear API boundaries between packages / projects in the monorepo. Code from package A should only be able to use public APIs exported from package B. This ensures that the code will work when the projects are packaged for distribution and published to NPM.

packages/core/src/components/dialog/multistepDialog.tsx
```diff
@@ -17,6 +17,8 @@
 import classNames from "classnames";
 import * as React from "react";
 
+import { Tooltip } from "@blueprintjs/core";
```
adidahiya: You don't need to import from `@blueprintjs/core` -- this _is_ the source code for that package. Within this package, you can import other components in the `src/` directory, just like you see below on L24 with `Button`, etc.

This import should be `from "../tooltip/tooltip";`

adidahiya:
There's another issue here -- Tooltips don't play nicely with disabled Buttons. This edge case will be common with the `tooltipContent`, as users will likely be showing something in the tooltip when the button is disabled. See the Button docs for more info. It's better to use an AnchorButton.



I'll push a commit here to make these fixes.

adidahiya (approved):

adidahiya:
[docs preview build](https://output.circle-artifacts.com/output/job/5d2de8fc-66a5-484c-b3a0-5718a95875fc/artifacts/0/packages/docs-app/dist/index.html#core/components/dialog.multistep-dialog) looks good 👍🏽

PR #5538, opened 2022-09-13
Fz/date range input footer element

### LINKED ISSUE #5537: [DateRangeInput2]: footerElement prop does not render
#### Environment

- __Package version(s)__: "@blueprintjs/core": "^4.7.0",
    "@blueprintjs/datetime": "^4.3.4",
    "@blueprintjs/datetime2": "^0.5.0",
    "@blueprintjs/popover2": "^1.5.0",
    "@blueprintjs/select": "^4.5.2",
    "@blueprintjs/table": "^4.5.0",
    "@blueprintjs/timezone": "^4.4.4",
- __Operating System__: MacOs Monterey 12.1
- __Browser name and version__: Chrome Version 104.0.5112.101

#### Code Sandbox
[Example](https://codesandbox.io/s/blueprint-sandbox-forked-crmux7?file=/src/DatetimeExample.tsx)

#### Steps to reproduce

1. add footerElement prop to DateRangeInput2 
1. ensure that it does not render

#### Actual behavior

No footer component rendered

#### Expected behavior

Render footer component

#### Possible solution

It seems that the prop is not used at all by DateRangeInput or by DateRangeInput2. Both DateInput and DateRangeInput, do render it. I think it just needs to be added

fnziman:
#### Checklist

- [ ] Includes tests
- [ ] Update documentation

Adds the footerElement to the DateRangePicker

Seeing that the footer now works

#### Screenshot

image

### DIFF
--- a/packages/datetime/src/dateRangePicker.tsx
+++ b/packages/datetime/src/dateRangePicker.tsx
@@ -225,7 +225,7 @@ export class DateRangePicker extends AbstractPureComponent2<DateRangePickerProps
     }
 
     public render() {
-        const { className, contiguousCalendarMonths, singleMonthOnly, footerElement } = this.props;
+        const { className, contiguousCalendarMonths, singleMonthOnly, footerElement = <div>footer</div> } = this.props;
         const isShowingOneMonth = singleMonthOnly || DateUtils.areSameMonth(this.props.minDate, this.props.maxDate);
 
         const classes = classNames(DateClasses.DATEPICKER, DateClasses.DATERANGEPICKER, className, {
@@ -237,11 +237,10 @@ export class DateRangePicker extends AbstractPureComponent2<DateRangePickerProps
         return (
             <div className={classes}>
                 {this.maybeRenderShortcuts()}
-                <div className={DateClasses.DATEPICKER_CONTENT}>
-                    <div>
-                        {this.renderCalendars(isShowingOneMonth)}
-                        {this.maybeRenderTimePickers()}
-                    </div>
+
+                <div>
+                    {this.renderCalendars(isShowingOneMonth)}
+                    {this.maybeRenderTimePickers()}
                     {footerElement}
                 </div>
             </div>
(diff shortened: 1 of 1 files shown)


palantirtech:
Thanks for your interest in [palantir/blueprint](https://github.com/palantir/blueprint), @fnziman! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

packages/datetime/src/dateRangePicker.tsx
```diff
@@ -237,9 +237,12 @@ export class DateRangePicker extends AbstractPureComponent2<DateRangePickerProps
         return (
             <div className={classes}>
                 {this.maybeRenderShortcuts()}
-                <div>
-                    {this.renderCalendars(isShowingOneMonth)}
-                    {this.maybeRenderTimePickers()}
+                <div className={DateClasses.DATEPICKER_CONTENT}>
+                    <div>
```
adidahiya: This doesn't seem like the ideal DOM structure... it probably makes more sense for the calendar(s) and time picker(s) to live adjacent to the footer element. I would like to see more options added to the docs example so we can test out this new layout. I'll try to push my ideas as a commit on your branch.

packages/datetime/src/dateRangePicker.tsx
```diff
@@ -237,9 +237,12 @@ export class DateRangePicker extends AbstractPureComponent2<DateRangePickerProps
         return (
             <div className={classes}>
                 {this.maybeRenderShortcuts()}
-                <div>
-                    {this.renderCalendars(isShowingOneMonth)}
-                    {this.maybeRenderTimePickers()}
+                <div className={DateClasses.DATEPICKER_CONTENT}>
```
adidahiya: also this CSS class was named for and designed for "datepicker", not "date range picker", so we'll need to make a new class to keep the naming semantics clear

packages/datetime/src/dateRangePicker.tsx
fnziman: I made an update that I think is more what you are looking for..

adidahiya:
Rendering the footer element exposed a few layout issues with the multiple calendars and time pickers rendered by DateRangePicker. I also discovered an unrelated bug where it renders two time pickers even when `singleMonthOnly={true}`. The fix for all these things was not trivial, so I took your first commit and added my changes on it in a new PR: #5539

adidahiya:
Thanks for the attempted PR -- I'm going to close this in favor of my new one, but you'll still get credited as a collaborator on the commit when that PR merges :)

PR #5641, opened 2022-10-04
[core] fix: improve contrast for disabled inputs in iOS

mduczek:
#### Checklist

No additional documentation/tests necessary

Fix or iOS specific styling, based on:
https://stackoverflow.com/questions/262158/disabled-input-text-color-on-ios
and:
https://stackoverflow.com/questions/30102792/css-media-query-to-target-only-ios-devices

##### Before

Screenshot 2022-10-04 at 12 06 38

Screenshot 2022-10-04 at 12 05 21

##### After

Screenshot 2022-10-04 at 12 07 42

Screenshot 2022-10-04 at 12 02 13

### DIFF
--- a/packages/core/src/components/forms/_common.scss
+++ b/packages/core/src/components/forms/_common.scss
@@ -142,11 +142,6 @@ $control-group-stack: (
   cursor: not-allowed;
   resize: none;
 
-  @supports (-webkit-touch-callout: none) {
-    opacity: 1;
-    -webkit-text-fill-color: $input-color-disabled;
-  }
-
   &::placeholder {
     color: $input-color-disabled;
   }
--- a/packages/core/src/components/forms/_common.scss
+++ b/packages/core/src/components/forms/_common.scss
@@ -142,6 +142,11 @@ $control-group-stack: (
   cursor: not-allowed;
   resize: none;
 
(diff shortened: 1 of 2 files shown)


adidahiya:
regular text inputs look fine but the text in TagInput tags is now unreadable in iOS:

| Before | After |
| - | - |
| image | image |

packages/core/src/components/forms/_common.scss
```diff
@@ -142,6 +142,11 @@ $control-group-stack: (
   cursor: not-allowed;
   resize: none;
 
+  @supports (-webkit-touch-callout: none) {
```
adidahiya: please add a code comment explaining that this is for iOS support

adidahiya (changes requested):
:-1: for TagInput regression described in my comment above

mduczek:
Thanks for finding the issue with TagInput!

Given iOS renders only disabled `<input/>` in non-standard way, I changed the selector to only target `<input />`:

Screenshot 2022-10-04 at 22 40 03

while testing I also noticed we should have an override for dark-mode, so added it as well :)

Screenshot 2022-10-04 at 23 01 32

adidahiya (approved):
lgtm, thanks @mduczek 👍🏽

PR #5668, opened 2022-10-12
[core] fix(Tabs, Switch): improve high contrast mode support

styu:

Adds high contrast mode support for Blueprint tabs
I also ended up fixing the high contrast mode for switches in dark theme, since I noticed that as I was testing out the dark theme + high contrast combo for the tabs

Do active/disabled/normal states look ok?


adidahiya:
Looks good in light theme.

Minor feedback: vertical tabs could probably use a highlighted background, but this is low priority since we don't see vertical tabs used often.



Dark theme indicators are still not getting the highlight color:

image

styu:
The vertical tab background is kind of rough to deal with because we have `!important` here:

https://github.com/palantir/blueprint/blob/980c6348eda1279a0b0f63972b2c226dbf48b1c2/packages/core/src/components/tabs/_tabs.scss#L108-L115

adidahiya (approved):

PR #5687, opened 2022-10-21
Downgrade classnames dependency to 2.3.1

styu:
In 2.3.2, classnames changed their types to require esModuleInterop and/or changing to `import * as classNames from "classnames"`. This relaxes the dependency in blueprint so that consumers that are not ready to bump to 2.3.2 have the choice to stay on 2.3.1

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -3251,9 +3251,9 @@ class-utils@^0.3.5:
     isobject "^3.0.0"
     static-extend "^0.1.1"
 
-classnames@^2.3.2:
+classnames@^2.3.1:
   version "2.3.2"
-  resolved "https://registry.yarnpkg.com/classnames/-/classnames-2.3.2.tgz#351d813bf0137fcc6a76a16b88208d2560a0d924"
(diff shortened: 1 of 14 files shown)


adidahiya (approved):

yarn.lock
```diff
@@ -3251,9 +3251,9 @@ class-utils@^0.3.5:
     isobject "^3.0.0"
     static-extend "^0.1.1"
 
-classnames@^2.3.2:
+classnames@^2.3.1:
   version "2.3.2"
-  resolved "https://registry.yarnpkg.com/classnames/-/classnames-2.3.2.tgz#351d813bf0137fcc6a76a16b88208d2560a0d924"
+  resolved "https://artifactory.palantir.build/artifactory/api/npm/all-npm/classnames/-/classnames-2.3.2.tgz#351d813bf0137fcc6a76a16b88208d2560a0d924"
```
adidahiya, line 3256: really strange that this did not fail the build with an `artifactory` URL 😮

styu: whooooops

PR #5698, opened 2022-10-26
[DRAFT] Ban Sass imports from @blueprintjs/*/src/ folders

### LINKED ISSUE #5639: FR: stylelint rule to ban Sass imports from @blueprintjs/*/src/ folders
#### Feature request

I'd like to ban imports from `@blueprintjs/core/src/**/*` files in Sass/Scss code. There doesn't seem to be an existing Stylelint rule to do this, so I propose that we ship our own with `@blueprintjs/stylelint-plugin`. It may be useful for other projects / packages unrelated to Blueprint... maybe we can configure this new rule to take a "denylist" of import specifier globs like:

```
{
  "rules": {
	"@blueprintjs/ban-imports": [
	  "*@blueprintjs/*/src/**/*"
	]
  }
}
```

We enforce similar rules for TS code with TSLint ([no-submodule-imports](https://palantir.github.io/tslint/rules/no-submodule-imports/)) and ESLint ([import/no-internal-modules](https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/no-internal-modules.md)). The shape of the package API is a little different in Sass because there's no "index" file to export the public API from the package root (we have to use `@import "@blueprintjs/core/lib/scss/variables"`), but the idea is mostly the same.

SoulPancake:
#### Checklist

- [ ] Includes tests
- [ ] Update documentation


adidahiya:
Closing this for now. Let me know if you'd like to continue working on this at some point.

PR #5708, opened 2022-10-31
[eslint-plugin] fix(no-deprecated-type-references): handle name conflicts with import aliases for new types

adidahiya:
#### Checklist

- [x] Includes tests
- [ ] Update documentation

Improve `@blueprintjs/no-deprecated-type-references` lint rule implementation to detect name conflicts for the new symbols. In these cases the rule will now alias the new type name with a "Blueprint" prefix, such as `{ Props as BlueprintProps }`.

New test cases

### DIFF
--- a/packages/eslint-plugin/src/rules/utils/replaceImportInFile.ts
+++ b/packages/eslint-plugin/src/rules/utils/replaceImportInFile.ts
@@ -17,35 +17,45 @@
 import { AST_NODE_TYPES, TSESLint, TSESTree } from "@typescript-eslint/utils";
 
 /**
- * Return a function which when provided with a fixer will produce a RuleFix to replace the
+ * Return a function which, when provided with a fixer, will produce a RuleFix to replace the
  * specified `fromImportName` with `toImportName` from the specified `packageName` at the top of the file.
  */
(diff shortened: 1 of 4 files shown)


packages/eslint-plugin/src/rules/utils/replaceImportInFile.ts
```diff
@@ -17,35 +17,45 @@
...
+        ) as TSESTree.ImportSpecifier;
 
         if (nodeToReplace === undefined) {
             throw new Error(
-                `Unable to find import { ${fromImportName} } from "${packageName}" while fixing lint error`,
+                `Unable to find import { ${options.fromName} } from "${options.moduleSpecifier}" while fixing lint error`,
+            );
+        } else if (options.alias !== undefined && nodeToReplace.local.name !== nodeToReplace.imported.name) {
+            throw new Error(
+                `Existing import '${nodeToReplace.imported.name}' is already aliased as '${nodeToReplace.local.name}', refusing to rename with new alias '${options.alias}'`,
```
gluxon, line 54: Good catch of this situation!

gluxon (approved):
This is fantastic. Thank you. ❤️

PR #5730, opened 2022-11-07
[select] popoverTargetProps and inputProps fixes & docs

adidahiya:

- Move `popoverTargetProps` (prop available on Select2 and MultiSelect2) typedef & docs into `SelectPopoverProps`
  - Note that `popoverTargetProps` is not supported on Suggest2, and users should use `inputProps` instead, as it does the same thing.
- fix(`MultiSelect2`): invoke `popoverTargetProps` `onKeyDown`/`oKeyUp` event handlers if provided
- ~chore(`Suggest2`): clean up some code to make event handlers more legible~
- fix(`Suggest2`): fix support for `inputProps.className` (it no longer overrides the popover class name, which caused all sorts of things to break)

No regressions (should be covered by test suites)

### DIFF
--- a/packages/select/src/components/select/select2.tsx
+++ b/packages/select/src/components/select/select2.tsx
@@ -54,8 +54,7 @@ export interface Select2Props<T> extends ListItemsProps<T>, SelectPopoverProps {
 
     /**
      * Whether the component should take up the full width of its container.
-     * This overrides `popoverProps.fill`. You also have to ensure that the child
-     * component has `fill` set to `true` or is styled appropriately.
+     * You also have to ensure that the child component has `fill` set to `true` or is styled appropriately.
      */
     fill?: boolean;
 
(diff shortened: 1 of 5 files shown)


packages/select/src/components/select/select2.tsx
```diff
@@ -251,6 +250,24 @@ export class Select2<T> extends AbstractPureComponent2<Select2Props<T>, Select2S
...
 
+    private withPopoverTargetPropsHandler = (
+        eventType: "keydown" | "keyup",
+        handler: React.KeyboardEventHandler<HTMLElement> | undefined,
+    ): React.KeyboardEventHandler<HTMLElement> => {
+        switch (eventType) {
+            case "keydown":
+                return event => {
+                    handler?.(event);
+                    this.props.popoverTargetProps?.onKeyDown?.(event);
```
adidahiya, line 261: this is a bugfix -- `popoverTargetProps.onKey{Down|Up}` handlers were never getting called before.

packages/select/src/components/suggest/suggest2.tsx
```diff
@@ -252,6 +254,7 @@ export class Suggest2<T> extends AbstractPureComponent2<Suggest2Props<T>, Sugges
                     {...inputProps}
                     aria-autocomplete="list"
                     aria-expanded={isOpen}
+                    className={classNames(targetProps.className, inputProps.className)}
```
adidahiya, line 257: this is also a bugfix -- `inputProps.className` was clobbering `targetProps.className` before, which would totally break things.

PR #5713, opened 2022-11-04
[select,datetime2] support popoverProps.targetTagName

adidahiya:
#### Checklist

- [x] Includes tests
- [x] Update documentation

#### Changes proposed in this pull request:

While migrating a large codebase to Popover2, I noticed some v2 components no longer supported `popoverProps.targetTagName` like they did with Popover. Although this use case is rare, I think we can continue supporting it in Popover2.

This PR does a few things:
- 🌟 feat(`InputGroup`): new prop `tagName` allows customizing the tag name of the HTML element that contains the input group
- 🔧 fix(`Select2`, `Suggest2`, `MultiSelect2`): restore support for `popoverProps={{ targetTagName }}`, which was previously available in the v1 components with `<Popover>`
- 🔧 fix(`DateInput2`, `DateRangeInput2`): restore support for `popoverProps={{ targetTagName }}`, which was previously available in the v1 components with `<Popover>`
- 🌟 feat(`DateInput2`, `DateRangeInput2`): new prop `popoverRef` allows access to the Popover2 instance, sometimes useful for repositioning the popover

#### Reviewers should focus on:

No regressions

PR #5734, opened 2022-11-09
[popover2] feat(Popover2, Tooltip2): make component type param optional

adidahiya:

It turns out https://github.com/palantir/blueprint/pull/5713 also had a typings regression for the `popoverRef` prop on datetime2 & select components, specifically the bit with `React.HTMLProps<unknown>`. In trying to improve this type definition, I found myself writing `React.HTMLProps<HtmlElement>` in a lot of places, since that's the default assumed in the definition of `Popover2Props<T = ...>`. It seems like we can actually just define this as the default in the `Popover2` class definition and avoid specifying the type param in most cases. I'm not exactly sure why this wasn't the case before (`T` was required if referencing `Popover2` as a type), but this will make these type references a lot simpler in our codebase and for consumers.

### DIFF
--- a/packages/datetime2/src/common/datetimePopoverProps.ts
+++ b/packages/datetime2/src/common/datetimePopoverProps.ts
@@ -22,7 +22,7 @@ import type { Popover2, Popover2Props } from "@blueprintjs/popover2";
  */
 export interface DatetimePopoverProps {
     /**
-     * Props to spread to `Popover2`.
+     * Props to spread to Popover2.
      */
     popoverProps?: Partial<
         Omit<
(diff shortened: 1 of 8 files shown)


adidahiya:
Ok, it looks like we can't actually do this. Providing a default value for the `Popover2<T>` type param makes it hard to deal with the target HTML props interface in `renderTarget`, leading to issues like this:

Screen Shot 2022-11-08 at 7 29 39 PM

... which _might_ be easily solved by using `<Popover2<React.HTMLProps<HTMLButtonElement>` in the opening tag of that code example, but we don't want to introduce a type regression which forces all consumers to instantiate the type of their Popover2s.

Opening an alternative PR to fix the type regression: #5735

PR #5753, opened 2022-11-10
[core] feat(Dialog): modernise visuals, body & footer components

CPerinet:
#### Checklist

- [x] Update documentation

- Reduced title from H5 to H6
- Header min-height is now based on the button size + padding.
- Use light-gray5 as default background colour for light mode
- Enable scrolling of the body past a given height
- Add a fixed footer. This is especially useful when the dialog body is likely to scroll.
  - Made it use the same class for both normal and multi-step dialog
  - Kept current footer classes

#### Screenshot

Screenshot 2022-11-10 at 14 50 02


packages/core/src/components/dialog/dialog.tsx
```diff
@@ -121,14 +126,19 @@ export class Dialog extends AbstractPureComponent2<DialogProps> {
             <Overlay {...this.props} className={Classes.OVERLAY_SCROLL_CONTAINER} hasBackdrop={true}>
                 <div className={Classes.DIALOG_CONTAINER} ref={this.props.containerRef}>
                     <div
-                        className={classNames(Classes.DIALOG, this.props.className)}
+                        className={classNames(Classes.DIALOG, this.props.footer != null && Classes.DIALOG_NO_PADDING, this.props.className)}
```
adidahiya: classNames has a slightly nicer syntax for conditional classes:

```suggestion
                        className={classNames(Classes.DIALOG, this.props.className, {
                            [Classes.DIALOG_NO_PADDING]: this.props.footer != null,
                        })}
```

packages/core/src/common/classes.ts
```diff
@@ -197,7 +200,6 @@ export const MULTISTEP_DIALOG = `${NS}-multistep-dialog`;
 export const MULTISTEP_DIALOG_PANELS = `${MULTISTEP_DIALOG}-panels`;
 export const MULTISTEP_DIALOG_LEFT_PANEL = `${MULTISTEP_DIALOG}-left-panel`;
 export const MULTISTEP_DIALOG_RIGHT_PANEL = `${MULTISTEP_DIALOG}-right-panel`;
-export const MULTISTEP_DIALOG_FOOTER = `${MULTISTEP_DIALOG}-footer`;
```
adidahiya, line 202: this is part of the public API, so unfortunately we can't remove it so easily in this PR. we'll have to keep this constant here and continue adding the class name to the DOM in `packages/core/src/components/dialog/multistepDialog.tsx` on L220 (this will also fix the failing tests in this PR), but we can remove its Sass styles and rely on a new class name for the updated styling.

packages/core/src/components/dialog/dialog.tsx
```diff
@@ -121,14 +126,19 @@ export class Dialog extends AbstractPureComponent2<DialogProps> {
...
                         aria-labelledby={this.props["aria-labelledby"] || (this.props.title ? this.titleId : undefined)}
                         aria-describedby={this.props["aria-describedby"]}
                         style={this.props.style}
                     >
                         {this.maybeRenderHeader()}
-                        {this.props.children}
+
+                        <div className={Classes.DIALOG_BODY_SCROLL_CONTAINER}>
+                            {this.props.children}
+                        </div>
```
adidahiya: I'm afraid this may be a breaking change for some consumers who are applying their own scroll/overflow styles to dialogs. We will have to add it in a non-breaking way with a new prop, something like `useScrollableBodyContainer?: boolean` where the default is `false` to preserve existing behavior.

Or: see my comment below with an API proposal with 3 new components.

packages/core/src/components/dialog/dialog.tsx
```diff
@@ -121,14 +126,19 @@ export class Dialog extends AbstractPureComponent2<DialogProps> {
...
+
+                        {this.maybeRenderFixedFooter()}
```
adidahiya: The ergonomics / developer experience of this new footer feel a little awkward to me right now.

We've kept the Dialog component pretty simple so far, but if we're trying to add more functionality to improve & evolve its layout, we may want to reconsider the design choice of using only a single React component.

I think if we add 3 new components, we can create an API that's more legible while still preserving backwards-compatibility:

```ts
<Dialog isOpen={...} onClose={...}>
  <DialogHeader title="..." icon="..." isCloseButtonShown={...} />
  <DialogBody useOverflowScrollContainer={...}> // <-- scroll container would be enabled by default
    {/* children go here */}
  </DialogBody>
  <DialogFooter fixed={...} actions={...}>
    {/* footer left content goes here */}
  </DialogFooter>
</Dialog>
```

If you like this direction, I can help you implement these changes in this PR.

adidahiya (dismissed):

packages/core/src/components/dialog/dialog.tsx
CPerinet: I don't have a strong preference here. I've also seen the following pattern:

```
<Dialog
  header={IDialogHeader.basic(
    {
      title: "Title",
      icon: IconNames.FOLDER,
      isCloseButtonShown: false,
    })
  footer={IDialogFooter.fixed(
    {
      left: "",
      actions: ""
    })
  useOverflowScrollContainer={true}
>
  // Content
</Dialog>
```

That may help with discoverability of features? Not sure how back-compatible that would be.

(I'm going to hold off on making the other changes while we discuss the API)

packages/core/src/common/classes.ts
```diff
@@ -116,8 +116,14 @@ export const CONTROL_GROUP = `${NS}-control-group`;
 export const DIALOG = `${NS}-dialog`;
 export const DIALOG_CONTAINER = `${DIALOG}-container`;
 export const DIALOG_BODY = `${DIALOG}-body`;
+export const DIALOG_BODY_NO_PADDING = `${DIALOG}-body-no-padding`;
```
CPerinet: <DialogBody /> not only provides padding but also scroll. Thus, I'm adding a way to remove padding as just not using <DialogBody /> is not practical.
Adding a "subtractive" class to keep DIALOG_BODY as close from current.

packages/core/src/common/classes.ts
```diff
@@ -116,8 +116,14 @@ export const CONTROL_GROUP = `${NS}-control-group`;
...
+export const DIALOG_BODY_SCROLL_CONTAINER = `${DIALOG}-body-scroll-container`;
```
CPerinet, line 121: Class applied when making <DialogBody /> scrollable.

packages/core/src/common/classes.ts
```diff
@@ -116,8 +116,14 @@ export const CONTROL_GROUP = `${NS}-control-group`;
...
 export const DIALOG_CLOSE_BUTTON = `${DIALOG}-close-button`;
 export const DIALOG_FOOTER = `${DIALOG}-footer`;
+export const DIALOG_FOOTER_CONTAINER = `${DIALOG}-footer-container`;
```
CPerinet: Needed to apply flex properties on left and right side of the footer. Did not directly added to DIALOG_FOOTER to keep it as current.

packages/core/src/common/classes.ts
```diff
@@ -116,8 +116,14 @@ export const CONTROL_GROUP = `${NS}-control-group`;
...
+export const DIALOG_FIXED_FOOTER = `${DIALOG}-fixed-footer`;
```
CPerinet: Styling for the fixed footer (white background, border &c.)

packages/core/src/common/classes.ts
```diff
@@ -116,8 +116,14 @@ export const CONTROL_GROUP = `${NS}-control-group`;
...
+export const DIALOG_INLINE_FOOTER = `${DIALOG}-inline-footer`;
```
CPerinet: Same for inline, in this case, it is just spacing.

packages/core/src/components/dialog/_dialog-body.scss
```diff
@@ -0,0 +1,13 @@
+.#{$ns}-dialog-body {
+  flex: 1 1 auto;
+  padding: $dialog-padding;
```
CPerinet, line 6: Changed margin to padding so the scroll works better. Should not cause regressions.
Removed line-height, did not seem appropriate default.

packages/core/src/components/dialog/_dialog-footer.scss
```diff
@@ -0,0 +1,40 @@
...
+  align-items: center;
+  justify-content: space-between;
+  gap: $dialog-padding;
+}
+
+.#{$ns}-dialog-footer-left {
+  flex: 1 0 auto;
+}
+
+.#{$ns}-dialog-footer-actions {
```
CPerinet, line 29: Moved, no changes

packages/core/src/components/dialog/_dialog-footer.scss
```diff
@@ -0,0 +1,40 @@
+.#{$ns}-dialog-footer {
+  flex: 0 0 auto;
```
CPerinet, line 5: Removed side margin. Will break things. I can add a class to keep it and then suppress it if using the new <DialogBody /> component.

adidahiya: Yeah, we'll need to keep the existing margin and add any new functionality with a modifier class.

packages/core/src/components/dialog/_dialog.scss
```diff
@@ -135,23 +136,3 @@ $dialog-padding: $pt-grid-size * 1.5 !default;
     border-bottom: 1px solid $pt-high-contrast-mode-border-color;
   }
 }
-
-.#{$ns}-dialog-body {
-  flex: 1 1 auto;
-  line-height: $pt-grid-size * 1.8;
```
adidahiya, line 141: unfortunately we need to keep this style too, it's too risky to change, and it's part of our API contract not to break styles on public class names like this.

adidahiya (commented):
I'd like to push a commit to address my comments about the breaking changes. I also want to clean up a few things, fix lint, etc.

packages/core/src/common/classes.ts
adidahiya: it seems like this class is not being demonstrated in the example... do we need to add it to the public API right now? maybe we can hold off on adding it until we find a concrete use case for it. until then, users can apply custom styles to get `padding: 0`.

packages/core/src/common/classes.ts
adidahiya: I think a better name for this would be "minimal", as it's a kind of minimal appearance. "Inline" is a bit confusing because it's not using inline layout; it's a "block" element with `display: flex`.

packages/core/src/common/classes.ts
adidahiya: Hmm, I don't think this is quite working like we want it to. I think we can do this without an additional level of DOM nesting, by using a modifier class instead. I'll take a stab at that approach.

| before | after |
| - | - |
| image | image |

packages/core/src/components/dialog/_dialog-footer.scss
```diff
@@ -0,0 +1,40 @@
...
+}
+
+.#{$ns}-dialog-fixed-footer {
+  background-color: $white;
+  border-radius: 0  0 $dialog-border-radius $dialog-border-radius;
+  border-top: 1px solid $pt-divider-black;
+  padding: $pt-grid-size;
+  padding-left: $dialog-padding;
```
adidahiya: why the extra padding here? is this intentional? if so, should we add the extra (5px) padding on the right side as well?

| before | after |
| - | - |
| image | image |

adidahiya: I see now that it's a little tricky to design this style to work for both button and non-button content. In the docs example we have some text in the "left" section for Dialog, but there's a button in that same place for MultiStepDialog...

packages/core/src/components/dialog/_dialog.scss
adidahiya: actually, I take this back -- this is the default line height, and it's applied by the `base-typography()` Sass mixin on `body` already, so it's probably ok to remove this style :+1:

adidahiya (approved):

adidahiya:
There's a bug in the preview script which is now failing on builds from forks. I'll fix that separately. For now you can view the latest docs preview here: https://output.circle-artifacts.com/output/job/a9d924bc-9ef0-4993-8939-2eb3fcabbd68/artifacts/0/packages/docs-app/dist/index.html

PR #5762, opened 2022-11-19
[core] feat(MultiStepDialog): allow multiple close buttons

### LINKED ISSUE #5758: [Multistep Dialog] Allow showing close buttons both in the header and footer for customizability
#### Feature request

It would be useful if this `showCloseButtonInFooter` prop being true did not mean that the close button in the dialog title will not be shown. Instead of being exclusive, both buttons can appear.

```
showCloseButtonInFooter: (undefined | false | true = false) Whether the footer close button is shown. The button will only appear if isCloseButtonShown is true. The close button in the dialog title will not be shown when this is true.
```

#### Usecase

Since the `closeButtonInFooter` has `closeButtonProps` it can be used to do other custom things, the most important use case for a multistep dialog button being _Skip Dialog_, for which developers can then have a custom callback

#### Examples

```
...
const handleSkipDialog = React.useCallback(...);
return <MultiStepDialog ... isCloseButtonShown={true} showCloseButtonInFooter{true} closeButtonProps={... onClick={handleSkipDialog}/>
```

tinopetros:
#### Checklist

- [ ] Includes tests
- [x] Update documentation

- `isCloseButtonShown` only affects the close button in the header
- `showCloseButtonInFooter` independently affects the close button in the footer (and now allows the exciting possibility of having a third custom button e.g. to skip the multiple steps of the dialog.
- Updated inline prop docs and omitted tests since there were none for close buttons previously

- The separation of `isCloseButtonShown` and `showCloseButtonInFooter` and that it works as expected depending on the set props.

#### Screenshot

Screen Shot 2022-11-19 at 3 40 15 PM

### DIFF
--- a/packages/core/src/components/dialog/multistepDialog.tsx
+++ b/packages/core/src/components/dialog/multistepDialog.tsx
@@ -40,7 +40,7 @@ export interface IMultistepDialogProps extends DialogProps {
     children?: React.ReactNode;
 
     /**
-     * Props for the close button that appears in the footer when there is no title.
+     * Props for the close button that appears in the footer.
      */
     closeButtonProps?: DialogStepButtonProps;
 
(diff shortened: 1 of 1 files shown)


adidahiya (approved):

adidahiya:
thanks @tinopetros!

PR #5784, opened 2022-12-01
docs: add JSDoc to all components linking to docs page

adidahiya:


adidahiya:
browser-tools orb is failing to install Chrome (https://github.com/CircleCI-Public/browser-tools-orb/issues/60), which breaks our build, but the tests are passing fine for me locally. and this is just a docs change. going to merge :shipit:

PR #5792, opened 2022-12-08
[datetime2] fix(DateRangeInput2): input value when controlled

### LINKED ISSUE #5791: DateRangeInput2 in controlled mode uses previous formatted selected value
#### Environment

- __Package version(s)__: @blueprintjs/datetime2 0.9.1
- __Operating System__: macOS Monterey 12.6.1
- __Browser name and version__: Google Chrome 107.0.5304.121

#### Code Sandbox

Link to a minimal repro (fork [this code sandbox](https://codesandbox.io/s/blueprint-sandbox-et9xy)): https://codesandbox.io/s/gallant-yalow-5ml2hd?file=/src/DatetimeExample.tsx

#### Steps to reproduce

The first DateRangeInput2 is in controlled mode with no time precision with close on selection false

```
      <DateRangeInput2
        closeOnSelection={false}
        formatDate={formatDate}
        parseDate={parseDate}
        value={value1}
        onChange={setValue1}
      />
```

1. Select any date range
2. While the date range is open select another date

#### Actual behavior

The input shows the previously selected date until the input is blurred / the popover is closed. If close on selection is true then this is generally not a problem because the selected value always appears correctly.

#### Expected behavior

The input should show the currently selected date. This works correctly in the second DateRangeInput2 with a time precision set.

tzyl:
#### Checklist

- [x] Includes tests
- [ ] Update documentation

DateRangeInput2 input focus handler should read from the controlled value rather than the state value.

When a new date is selected the onChange callback is called which updates the controlled value. On the next render cycle componentDidUpdate syncs the state value with the controlled props value and also refocuses the boundary input at the same time. During this input focus the stale value from the state is read before it has been updated leading to the input display value showing the previous value.

The input blur and input change handlers already special case controlled mode could it have been intentional to have this behaviour in the input focus handler for any reason?

#### Screenshot

I tested this on the docs app locally by changing the example to be controlled:

```
@@ -120,6 +120,7 @@ export class DateRangeInput2Example extends React.PureComponent<ExampleProps, Da
                     {...spreadProps}
                     {...format}
                     onChange={this.handleRangeChange}
+                    value={range}
                     footerElement={showFooterElement ? exampleFooterElement : undefined}
                     timePickerProps={
                         enableTimePicker
```

| Before | After |
|-|-|
controlled_before | controlled_after

### DIFF
--- a/packages/datetime2/src/components/date-range-input2/dateRangeInput2.tsx
+++ b/packages/datetime2/src/components/date-range-input2/dateRangeInput2.tsx
@@ -687,7 +687,12 @@ export class DateRangeInput2 extends AbstractPureComponent2<DateRangeInput2Props
 
     private handleInputFocus = (_e: React.FormEvent<HTMLInputElement>, boundary: Boundary) => {
         const { keys, values } = this.getStateKeysAndValuesForBoundary(boundary);
-        const inputString = DatePickerUtils.getFormattedDateString(values.selectedValue, this.props, true);
+        const isValueControlled = this.isControlled();
+        const inputString = DatePickerUtils.getFormattedDateString(
+            isValueControlled ? values.controlledValue : values.selectedValue,
+            this.props,
+            true,
+        );
 
         // change the boundary only if the user explicitly focused in the field.
         // focus changes from hovering don't count; they're just temporary.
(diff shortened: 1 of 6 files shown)


adidahiya:
@tzyl it was a bit tricky to add a unit test which failed with the existing behavior... but I've added one that may catch regressions in this behavior in the future.

adidahiya (approved):

PR #5795, opened 2022-12-11
dataNode on TreeNode should not be optional Fix

### LINKED ISSUE #5779: dataNode on TreeNode should not be optional
#### Environment

- __Package version(s)__: 

#### Feature request

This line: https://github.com/palantir/blueprint/blob/develop/packages/core/src/components/tree/treeNode.tsx#L83

```
    /**
     * An optional custom user object to associate with the node.
     * This property can then be used in the `onClick`, `onContextMenu` and `onDoubleClick`
     * event handlers for doing custom logic per node.
     */
    nodeData?: T;
```

should not be optional. It prevents type checking. If a user wants that to be optional then they can pass in undefined as a union for the generic parameter

#### Examples

can't do:

```
const a: ITreeNode<{type: "a"}> = {} as ITreeNode<{type: "a"}>;

if (a.dataNode.type == "a") // can't do this as a.dataNode can be undefined
```

if user actually wants undefined they can do `ITreeNode<{type: "a"} | undefined>`

SoulPancake:
#### Checklist

- [ ] Includes tests
- [Y] Update documentation


SoulPancake:
can this be reviewed

PR #5802, opened 2022-12-13
[popover2] feat: new prop targetProps

bvandrc:
#### Checklist

- [X] Includes tests
- [X] Update documentation

Allow `targetProps` to be passed to popover2 that get passed to the `span` wrapper of the popover target.

Mutually exclusive with  `renderTarget` , since  `renderTarget` doesn't get wrapped.

With existing functionality, this could still be accomplished by specifying a custom `renderTarget` that wraps their child in a `span`, ie before:

```
 <Popover2 {...popoverProps}>
    <Button  {...buttonProps} />
  </Popover2>
```

after:

```  
<Popover2
    {...popoverProps}
    renderTarget={(props) => (
      <span {...props} {...targetProps}>
            <Button  {...buttonProps} />
      </span>
    )}
  />
```

But with this proposed change, one could do:

```
 <Popover2 {...popoverProps} targetProps={targetProps}>
    <Button  {...buttonProps} />
  </Popover2>
```

So, for us, this isn't an "essential" change, since we can accomplish the same with `renderTarget`-- but I wanted to propose it to the Blueprint team.


packages/popover2/src/popover2.tsx
```diff
@@ -356,17 +360,26 @@ export class Popover2<T> extends AbstractPureComponent2<Popover2Props<T>, IPopov
...
+            // similarly, this class is mainly useful for targets like <Button>, <InputGroup>, etc.
+            [CoreClasses.FILL]: fill,
+        };
+
         let target: JSX.Element | undefined;
 
         if (renderTarget !== undefined) {
             target = renderTarget({
                 ...targetProps,
+                className: classNames(targetProps.className, targetModifierClasses),
```
bvandrc: since we apply the fill class to the child target when renderTarget isn't defined, I figured we should apply this class to the custom target as well

packages/popover2/src/popover2SharedProps.ts
```diff
@@ -283,6 +283,13 @@ export interface IPopover2SharedProps<TProps> extends OverlayableProps, Props {
      */
     targetTagName?: keyof JSX.IntrinsicElements;
 
+    /**
+     * HTML props for the target element.
+     *
+     * This prop is mutually exclusive with the `renderTarget` API.
+     */
+    targetWrapperProps?: React.HTMLProps<HTMLSpanElement> | React.HTMLProps<HTMLDivElement>;
```
adidahiya:
```suggestion
    targetProps?: TProps;
```

adidahiya (changes requested):
Thanks for the PR @bvandercar-vt. This enhancement makes sense to me, especially after doing a bunch of Popover2 migrations of our internal code. When I initially designed Popover2, I figured most target customizations should happen via `renderTarget`, and that would be sufficient, but I now realize that it's a lot better from a developer experience perspective if we allow minor modifications to target element attributes through a new prop like this.

One use case I imagine this will be useful for is something like:

```
<Popover2 targetProps={{ style: { /* absolute positioning */ } }}>
```

On naming... to avoid potential confusion/conflation with "wrapper" terminology from Popover "v1", and because the existing prop which customizes the `.bp4-popover-target` element is named `targetTagName` (rather than `targetWrapperTagName`), I think this should be called `targetProps`.

packages/popover2/src/popover2SharedProps.ts
bvandrc: OK done, please review.

In the case of a `renderTarget`, `targetProps` props will also be appliedto the `renderTarget`, so I made the warning basically say that they will be applied, but that it doesn't really make sense to do it that way vs. just applying them in your `renderTarget`.

bvandrc: @adidahiya  I'm not sure why CI is failing, is `TProps` not the best type for `targetProps` here maybe?

adidahiya: `TProps` is the correct type to use here, but adding this constraint exposed some type issues with the way I had designed Popover2. It required adding a stricter typedef to `validateProps`, which further exposed some complex issues with the `renderTarget` API and the inferred value of `TProps`. I think I've solved those issues in the commits I just pushed to this branch.

adidahiya:
I've updated your PR to no longer inject `props.targetProps` into `props.renderTarget` -- it was causing type issues and it's not necessary to support. I've also updated the error message to reflect this change.

packages/popover2/src/popover2.tsx
```diff
@@ -327,46 +339,58 @@ export class Popover2<T> extends AbstractPureComponent2<Popover2Props<T>, IPopov
...
             // applied to the generated target wrapper element.
             className: classNames(className, Classes.POPOVER2_TARGET, {
                 [Classes.POPOVER2_OPEN]: isOpen,
                 // this class is mainly useful for button targets
-                [CoreClasses.ACTIVE]: !isControlled && isOpen && !isHoverInteractionKind,
+                [CoreClasses.ACTIVE]: isOpen && !isControlled && !isHoverInteractionKind,
             }),
             ref,
-            ...(targetEventHandlers as unknown as T),
+            ...targetEventHandlers,
```
adidahiya, line 377: note this improvement in the type definitions :) no more ugly `as unknown`!

packages/popover2/src/popover2SharedProps.ts
```diff
@@ -223,7 +240,9 @@ export interface IPopover2SharedProps<TProps> extends OverlayableProps, Props {
      *
      * Mutually exclusive with `children` and `targetTagName` props.
      */
-    renderTarget?: (props: Popover2TargetProps & TProps) => JSX.Element;
```
adidahiya, line 226: `TProps` was not getting inferred correctly before this change. You can see an example of this in `popover2MinimalExample.tsx`:

```
                <Popover2
                    {...baseProps}
                    minimal={true}
                    renderTarget={({ isOpen, ref, ...p }) => (
                        <Button {...p} active={isOpen} elementRef={ref} intent={Intent.PRIMARY} text="Minimal" />
                    )}
                />
```

here, `p` was getting inferred as an empty object `{}`, which is not what we want. that basically suppresses type checks on that object and allows anything through.

after this change, `p` correctly gets the types of the attributes injected by Popover2 here:

image

adidahiya (approved):

packages/popover2/src/popover2.tsx
```diff
@@ -327,46 +339,58 @@ export class Popover2<T> extends AbstractPureComponent2<Popover2Props<T>, IPopov
...
         // Ensure target is focusable if relevant prop enabled
         const targetTabIndex = openOnTargetFocus && isHoverInteractionKind ? 0 : undefined;
-        const targetProps = {
+        const ownTargetProps = {
             "aria-haspopup":
                 this.props.popupKind ??
-                (this.props.interactionKind === Popover2InteractionKind.HOVER_TARGET_ONLY ? undefined : "true"),
+                (this.props.interactionKind === Popover2InteractionKind.HOVER_TARGET_ONLY
+                    ? undefined
+                    : ("true" as "true")),
```
bvandrc, line 367: could just do  `"true" as const` , same result

adidahiya: nope, unfortunately I tried that -- it only works for const enums

adidahiya:
@bvandercar-vt this is nearly good to go, I just need to test it out locally to see if it breaks anything downstream... stay tuned, I'll try to get this in for the release I'm cutting today

adidahiya:
This change seems to work well downstream for the `renderTarget` API... it allows us to clean up some code, too, for example:

```diff
class MyComponent extends React.PureComponent {
    public render() {
        return (
-           <Popover2<React.HTMLProps<HTMLDivElement>>
+           <Popover2
                renderTarget={this.renderTarget}
                content="content"
            />
        );
    }

    private renderTarget = ({
        isOpen,
        ref,
        ...targetProps
-    }: Popover2TargetProps & React.HTMLProps<HTMLDivElement>) => {
+    }: Popover2TargetProps & Popover2ClickTargetHandlers) => {
        return (
            <div
                ref={ref}
                {...targetProps}
            />
        );
    };
}
```

PR #5807, opened 2022-12-15
feat(MenuItem, MenuItem2): roleStructure="listitem"

bvandrc:
#### Checklist

- [X] Includes tests
- [X] Update documentation

using `roleStructure` prop of `MenuItem2`, allow a configuration where the components have no assigned roles.

### DIFF
--- a/packages/core/src/components/menu/menuItem.tsx
+++ b/packages/core/src/components/menu/menuItem.tsx
@@ -93,7 +93,7 @@ export interface IMenuItemProps extends ActionProps, LinkProps {
      * `<li role=undefined`
      *     `<a role=undefined`
      *
-     *  which can be used if this item is within a `list` parent
+     *  which can be used if this item is within a basic `<ul/>` (or `role="list"`) parent.
      *
      * @default "menuitem"
      */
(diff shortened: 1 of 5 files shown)


adidahiya:
@bvandercar-vt the code changes seem reasonable, but can you explain what kind of UI this would be useful in? Where are you rendering a MenuItem that is neither a "menuitem" or "listoption"?

bvandrc:
> @bvandercar-vt the code changes seem reasonable, but can you explain what kind of UI this would be useful in? Where are you rendering a MenuItem that is neither a "menuitem" or "listoption"?

We have a unique case where we are adding delete buttons to the `MenuItem2`s of a `Select2`, using `Select2`'s `itemRenderer` prop. In this case, the 2 existing `MenuItem2` aria `role` configurations (the `roleStructure` prop-- "listoption" and "menuitem") **do not allow additional interactive elements to be within them**-- this causes an aXe a11y failure.

So, in our case, we are manually changing the `Select2` menu's `role` from `listbox` to `list` (using `menuProps`) , so that we can set its `MenuItem2` items to just be plain no-role `li`'s (using this PR), so that we can include buttons in each `li` without creating accessibility failures.

adidahiya (approved):

PR #5815, opened 2022-12-30
[select] feat(QueryList): add `ref` to `ItemRendererProps`

### LINKED ISSUE #3369: [Select] custom itemListRenderer breaks scrollToActiveItem
#### Environment

- __Package version(s)__: core@3.13.0, select@3.6.1
- __Browser and OS versions__: Chrome/Mac

#### Steps to reproduce

[Edit w0wmq7p76k](https://codesandbox.io/s/w0wmq7p76k)

1. Open [Sandbox](https://codesandbox.io/s/w0wmq7p76k)
2. Select `Fight Club`
3. Reopen Select

#### Actual behavior

QueryList does not scroll to `Fight Club`, but to `Schindler’s List`.

#### Expected behavior

Should scroll to `Fight Club`

#### Possible solution

As mentioned in the [docs](https://blueprintjs.com/docs/#select/select-component.custom-menu), `itemListRenderer`can be used to `group items under a common heading`. But additional `MenuItem` shift the scroll behaviour. It seems, the problem lies within [QueryList#getActiveElement](https://github.com/palantir/blueprint/blob/develop/packages/select/src/components/query-list/queryList.tsx#L262), which uses the lists children by index. This does not account for additional items.

shim-flounce:
#### Checklist

- [x] Includes tests
- [x] Update documentation

See https://github.com/palantir/blueprint/issues/3369#issuecomment-469887277

I didn't add any tests because it seems like there's no tests for scrolling, and I wasn't sure how to best write such a test:

https://github.com/palantir/blueprint/blob/b6c1518ea2f524accfba8045ce6f22cac7c5b56e/packages/select/test/queryListTests.tsx#L215-L217

`getActiveElement` being exposed as desired

### DIFF
--- a/packages/select/src/common/listItemsProps.ts
+++ b/packages/select/src/common/listItemsProps.ts
@@ -49,6 +49,7 @@ export interface ListItemsProps<T> extends Props {
     /** If provided, this function will be used in place of the default implementation. */
     getActiveElement?: (props: {
         activeItem: CreateNewItem | T | null;
+        filteredItems: T[];
         index: number;
         itemsParent: HTMLElement;
     }) => HTMLElement | undefined;
--- a/packages/select/src/common/listItemsProps.ts
+++ b/packages/select/src/common/listItemsProps.ts
@@ -46,6 +46,13 @@ export interface ListItemsProps<T> extends Props {
      */
     activeItem?: T | CreateNewItem | null;
 
+    /** If provided, this function will be used in place of the default implementation. */
+    getActiveElement?: (props: {
+        activeItem: CreateNewItem | T | null;
+        index: number;
+        itemsParent: HTMLElement;
+    }) => HTMLElement | undefined;
+
     /** Array of items in the list. */
     items: T[];
 
--- a/packages/select/src/common/listItemsProps.ts
+++ b/packages/select/src/common/listItemsProps.ts
@@ -46,14 +46,6 @@ export interface ListItemsProps<T> extends Props {
      */
     activeItem?: T | CreateNewItem | null;
 
-    /** If provided, this function will be used in place of the default implementation. */
-    getActiveElement?: (props: {
-        activeItem: CreateNewItem | T | null;
-        filteredItems: T[];
-        index: number;
-        itemsParent: HTMLElement;
-    }) => HTMLElement | undefined;
-
     /** Array of items in the list. */
     items: T[];
 
--- a/packages/select/src/components/query-list/queryList.tsx
+++ b/packages/select/src/components/query-list/queryList.tsx
@@ -34,12 +34,6 @@ import {
 export type QueryListProps<T> = IQueryListProps<T>;
 /** @deprecated use QueryListProps */
 export interface IQueryListProps<T> extends ListItemsProps<T> {
-    /** If provided, this function will be used in place of the default implementation. */
-    getActiveElement?: (props: {
-        activeItem: CreateNewItem | T | null;
-        index: number;
-        itemsParent: HTMLElement;
-    }) => HTMLElement | undefined;
     /**
      * Initial active item, useful if the parent component is controlling its selectedItem but
      * not activeItem.
--- a/packages/select/src/components/query-list/queryList.tsx
+++ b/packages/select/src/components/query-list/queryList.tsx
@@ -169,6 +169,8 @@ export class QueryList<T> extends AbstractComponent2<QueryListProps<T>, IQueryLi
 
     private itemsParentRef?: HTMLElement | null;
 
+    private itemRefs = new Map<number, HTMLElement>();
+
     private refHandlers = {
         itemsParent: (ref: HTMLElement | null) => (this.itemsParentRef = ref),
     };
@@ -394,6 +396,13 @@ export class QueryList<T> extends AbstractComponent2<QueryListProps<T>, IQueryLi
                 index,
                 modifiers,
                 query,
+                ref: node => {
+                    if (node) {
+                        this.itemRefs.set(index, node);
+                    } else {
+                        this.itemRefs.delete(index);
+                    }
+                },
             });
         }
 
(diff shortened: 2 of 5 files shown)


packages/select/src/components/query-list/queryList.tsx
```diff
@@ -34,6 +34,12 @@ import {
 export type QueryListProps<T> = IQueryListProps<T>;
 /** @deprecated use QueryListProps */
 export interface IQueryListProps<T> extends ListItemsProps<T> {
+    /** If provided, this function will be used in place of the default implementation. */
```
shim-flounce, line 37: I borrowed the wording from https://github.com/palantir/blueprint/blob/b6c1518ea2f524accfba8045ce6f22cac7c5b56e/packages/core/src/context/hotkeys/hotkeysProvider.tsx#L98

packages/select/src/components/query-list/queryList.tsx
```diff
@@ -417,11 +423,18 @@ export class QueryList<T> extends AbstractComponent2<QueryListProps<T>, IQueryLi
     private getActiveElement() {
         const { activeItem } = this.state;
         if (this.itemsParentRef != null) {
+            const { getActiveElement } = this.props;
             if (isCreateNewItem(activeItem)) {
                 const index = this.isCreateItemFirst() ? 0 : this.state.filteredItems.length;
+                if (typeof getActiveElement === "function") {
+                    return getActiveElement({ activeItem, index, itemsParent: this.itemsParentRef });
```
shim-flounce: Passing all the props that might be useful when implementing a custom `getActiveElement`

packages/select/src/common/listItemsProps.ts
```diff
@@ -46,6 +46,13 @@ export interface ListItemsProps<T> extends Props {
      */
     activeItem?: T | CreateNewItem | null;
 
+    /** If provided, this function will be used in place of the default implementation. */
```
shim-flounce: I borrowed the wording from https://github.com/palantir/blueprint/blob/b6c1518ea2f524accfba8045ce6f22cac7c5b56e/packages/core/src/context/hotkeys/hotkeysProvider.tsx#L98

packages/select/src/components/query-list/queryList.tsx
```diff
@@ -34,12 +34,6 @@ import {
 export type QueryListProps<T> = IQueryListProps<T>;
 /** @deprecated use QueryListProps */
 export interface IQueryListProps<T> extends ListItemsProps<T> {
-    /** If provided, this function will be used in place of the default implementation. */
-    getActiveElement?: (props: {
```
shim-flounce: I moved this into `ListItemProps` so that users can pass this to components like `MultiSelect2`

packages/select/src/components/query-list/queryList.tsx
```diff
@@ -417,11 +417,18 @@ export class QueryList<T> extends AbstractComponent2<QueryListProps<T>, IQueryLi
     private getActiveElement() {
         const { activeItem } = this.state;
         if (this.itemsParentRef != null) {
+            const { getActiveElement } = this.props;
```
shim-flounce: I figured that instead of delegating entirely to `props.getActiveElement()` (the user would then have to re-implement the `isCreateNewItem` check for example), it was more ergonomic to delegate to it the way I did.

shim-flounce:
@adidahiya Apologies for the direct ping, but do you have any thoughts on the proposed change?

adidahiya:
The code seems reasonable so far, can you demonstrate usage of this new prop in the docs somehow? I think you'd need to design a new example option which uses `itemListRenderer` in a meaningful way. I'd prefer to add new props / functionality only if it's been unit tested or demonstrated clearly in `docs-app`.

shim-flounce:
> The code seems reasonable so far, can you demonstrate usage of this new prop in the docs somehow? I think you'd need to design a new example option which uses `itemListRenderer` in a meaningful way. I'd prefer to add new props / functionality only if it's been unit tested or demonstrated clearly in `docs-app`.

Makes sense! To be honest I just blindly followed the suggestion in the linked issue, but I think it’s a bit awkward to use (although it does satisfy my use case). I’ll try to see if there’s a better way to solve this and update the PR (might be a couple of weeks as I’m going on vacation though). Thanks for your review anyway!

shim-flounce:
@adidahiya I've updated the Select example to include a grouped option that [uses this new prop](https://github.com/palantir/blueprint/pull/5815/commits/fcc349e14f8e1ac9fa5e8184a6ae0af9991ea93f#diff-952c30feb77b36731735281f0292ac2c7f37322b98ba9d40d4e5f351afbe5cbdR174-R191). That being said, I think it would be better to enhance `ItemRendererProps` with a `ref` and let users wire it up to the right element?

packages/select/src/components/query-list/queryList.tsx
```diff
@@ -419,10 +428,12 @@ export class QueryList<T> extends AbstractComponent2<QueryListProps<T>, IQueryLi
         if (this.itemsParentRef != null) {
             if (isCreateNewItem(activeItem)) {
                 const index = this.isCreateItemFirst() ? 0 : this.state.filteredItems.length;
-                return this.itemsParentRef.children.item(index) as HTMLElement;
+                return this.itemRefs.get(index) ?? (this.itemsParentRef.children.item(index) as HTMLElement);
```
shim-flounce: The fallback is for backwards compat (if a user didn't wire up the new `ref` to `elementRef` on the `MenuItem` for example)

shim-flounce:
@adidahiya I've changed the approach but kept the grouped example for you to verify scroll behaviour. I'm happy to drop the example though to keep the diff small.

packages/select/src/components/query-list/queryList.tsx
```diff
@@ -394,6 +396,13 @@ export class QueryList<T> extends AbstractComponent2<QueryListProps<T>, IQueryLi
                 index,
                 modifiers,
                 query,
+                ref: node => {
```
shim-flounce, line 399: Followed approach in https://beta.reactjs.org/learn/manipulating-the-dom-with-refs#how-to-manage-a-list-of-refs-using-a-ref-callback

packages/select/src/components/query-list/queryList.tsx
```diff
@@ -422,7 +431,9 @@ export class QueryList<T> extends AbstractComponent2<QueryListProps<T>, IQueryLi
                 return this.itemsParentRef.children.item(index) as HTMLElement;
```
shim-flounce, line 431: Strictly speaking, we could also wire up a ref to `createNewItemRenderer` so that we can easily look it up, but I think even without doing so `0` and `this.state.filteredItems.length` should be good enough as indices.

adidahiya (approved):
nicely done @shim-flounce, I can clearly see this fixes the linked issue in the [docs preview build](https://output.circle-artifacts.com/output/job/6c96260e-9be1-4783-ab4c-1e44f066fe1d/artifacts/0/packages/docs-app/dist/index.html#select/select2):

PR #5822, opened 2023-01-04
[popover2] feat(Tooltip2): compact appearance

adidahiya:

This PR introduces the `compact` prop for Tooltip2, which allows the components to render their content with smaller padding than the default appearance.

The new compact appearance is encouraged in data-dense UIs:

| default | compact |
| - | - |
| image | image |

I also noticed a minor typographic issue for `<Code>` spans inside text which stood out in the example for this new prop. The default `vertical-align: baseline` makes inline `<Code>` too low compared to its adjacent text, and I've found that a value of `vertical-align: text-bottom` looks better. Right now I'm conservatively applying this fix to the new `.bp4-compact` modifier class only.

#### Not in this PR (considered, but reverted)

~For Popover2, `compact={true}` applies when `popoverClassName={Popover2Classes.POPOVER2_CONTENT_SIZING}` is set. The default 20px padding for this modifier class has often felt too large in practical usage, which forced Blueprint users to apply style overrides.~

| default | compact |
| - | - |
| image | image |

### DIFF
--- a/packages/popover2/src/tooltip2.tsx
+++ b/packages/popover2/src/tooltip2.tsx
@@ -38,6 +38,14 @@ export interface ITooltip2Props<TProps = DefaultPopover2TargetHTMLProps>
     content: JSX.Element | string;
 
     /**
+     * Whether to use a compact appearance, which reduces the visual padding around
+     * tooltip content.
+     *
+     * @default false
+     */
+    compact?: boolean;
+
+    /**
      * The amount of time in milliseconds the tooltip should remain open after
      * the user hovers off the trigger. The timer is canceled if the user mouses
      * over the target before it expires.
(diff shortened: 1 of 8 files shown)


packages/popover2/src/tooltip2.tsx
```diff
@@ -103,24 +105,16 @@ export class Tooltip2<T> extends React.PureComponent<Tooltip2Props<T>> {
...
-        }
+        this.popoverRef.current?.reposition();
     }
 
     // any descendant ContextMenu2s may update this ctxState
     private renderPopover = (ctxState: Tooltip2ContextState) => {
         const { children, disabled, intent, popoverClassName, ...restProps } = this.props;
-        const classes = classNames(
-            Classes.TOOLTIP2,
-            { [CoreClasses.MINIMAL]: this.props.minimal },
```
adidahiya, line 116: this line was intentionally removed. it's not necessary to apply this class here because Popover2 will do it anyway since we're passing through the `minimal` prop

CPerinet:
Do we need it for Popover? Tooltips should only have a couple lines of text in it where Popover have more content. At the same time, compact mode make the spacing from elements to sides lower than between elements. This is "wrong" information architecture and end-up looking unbalanced.

adidahiya:
@CPerinet I found a bunch of popovers in our internal code bases with 10px padding, so I figured this compact modifier might be useful for those use cases. I did also find a handful of popovers with 5px or 15px padding (even some 12px, 8px...), so maybe this styling isn't as standard as I'd like it to be. I will have to think some more about this problem.

> compact mode make the spacing from elements to sides lower than between elements. This is "wrong" information architecture and end-up looking unbalanced.

Good point... in the docs example this is an issue because the "footer" content section uses a 15px top margin:

image

For now, I'll revert the Popover2 changes and just keep this PR scoped to Tooltip2.

adidahiya:
Iterated with @CPerinet over Slack, we've settled on a 5/7px padding for the compact tooltip. We may adjust this later if necessary.

Issue #3967, opened 2020-02-17
DateRangePicker : changing "value" (controlled) only updates Date, but not Time


eleveurdechevres:
#### Environment

- __Package version(s)__: datetime 3.15.2
- __Browser and OS versions__: Chrome 79.0.3945.130

If possible, link to a minimal repro (fork [this code sandbox](https://codesandbox.io/s/blueprint-sandbox-et9xy)): 

#### Steps to reproduce

                                <DateRangePicker
                                    value={this.selectedDateRange}
                                    onChange={(selectedDateRange: DateRange) => this.selectedDateRange = selectedDateRange}
                                    allowSingleDayRange={true}
                                    contiguousCalendarMonths={false}
                                    timePrecision={'second'}
                                    shortcuts={false}
                                    timePickerProps={{
                                        useAmPm: true
                                    }}
                                />

#### Actual behavior

In controlled mode, if i change manually this.selectedDateRange in another component, the date is modified in the DateRangePicker, but not the time

#### Expected behavior

TimePicker values must be updated when range value is changed.

adidahiya:
@eleveurdechevres it would really help if you provided a repro by forking the code sandbox so I can further investigate

eleveurdechevres:
Hello !
Sorry, I was on holidays ^^.

Here is a repo for testing :
https://github.com/eleveurdechevres/DateRangePicker_test

- The buttons permit to choose different DateRanges.
- "From date" and "To date" are displayed on the bottom (date.toIsoString(), perhaps it is not the best method)
- You see that when you click on the buttons, the date is selected in the picker, but the time is not.
I have seen also bad dates, just before sending you this message, perhaps there is another little problem due to the example ?

Have a good day,
Freddy

eleveurdechevres:
Note : the other problem I saw : by default the time is 12:00:00 AM => it indicates incorrect dates

MonkeyAndres:
Same here!

sharpcoder7:
This component uses areRangesEqual function which does not check if time was changed.
Also, this component has time property in its state which is not updated in a controlled mode, so time input boxes are not updated as well

adidahiya:
Duplicate of #5502

PR #4980, opened 2021-10-22
Fix Date Range Input - time is not updating from value

izmy:
#### Checklist

- [ ] Includes tests
- [ ] Update documentation

#### Changes proposed in this pull request:

When props value is changing, DateRangePicker does not reflect the change in time picker.

PR #5841, opened 2023-01-12
[datetime] fix(DateRangePicker): accept controlled updates for start/end time

### LINKED ISSUE #5502: Modifying the time doesn't show up on the timepicker in DateRangePicker
#### Environment

- __Package version(s)__: 4.3.4
- __Operating System__: MacOS Monterey
- __Browser name and version__: Firefox 103.0.2

#### Code Sandbox

Link to a minimal repro: https://codesandbox.io/s/blueprint-sandbox-forked-gb14b7

#### Steps to reproduce

1. Select two dates in the daterangepicker

#### Actual behavior

Modifying the endTime (from 00:00 to 23:59) doesn't show the modified end time in the time picker. (I modify the time in the onChange)

#### Expected behavior

When I change the time on the endDate value, it should update on the DateRangePicker end time

#### Possible solution

It looks like it doesn't update the startTime or endTime in the code when the props change, only on the construction. Modify it so it does do that.

tinopetros:
#### Checklist

- [x] Includes tests
- [ ] Update documentation

- `areRangesEqual` is used to check value changes when the DateRangePicker component is controlled as hinted [here](https://github.com/palantir/blueprint/issues/3967#issuecomment-668111971). This function now ensures that both the day and time in the date object match.

- Tests as proof that this now works. The current code passes updated tests, whereas older code failed:

```
DateUtils
    areRangesEqual
       returns false for
          [DATE_1, DATE_5] and [DATE_1, DATE_5_01]
             1) AssertionError: expected true to equal false
                at Context.<anonymous> (/var/folders/bf/99sjtr2d4_s2jzr5ct9nfb5x086lgg/T/_karma_webpack_575875/commons.js:169997:89)
```

#### Screenshot

N/A

### DIFF
--- a/packages/datetime/src/common/dateUtils.ts
+++ b/packages/datetime/src/common/dateUtils.ts
@@ -39,8 +39,9 @@ export function areRangesEqual(dateRange1: DateRange, dateRange2: DateRange) {
     } else {
         const [start1, end1] = dateRange1;
         const [start2, end2] = dateRange2;
-        const areStartsEqual = (start1 == null && start2 == null) || areSameDay(start1, start2);
-        const areEndsEqual = (end1 == null && end2 == null) || areSameDay(end1, end2);
+        const areStartsEqual =
+            (start1 == null && start2 == null) || (areSameDay(start1, start2) && areSameTime(start1, start2));
+        const areEndsEqual = (end1 == null && end2 == null) || (areSameDay(end1, end2) && areSameTime(start1, start2));
         return areStartsEqual && areEndsEqual;
     }
 }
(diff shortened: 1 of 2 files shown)


adidahiya:
This change seems fine in theory. But I don't see how this is functionally getting used in DateRangePicker correctly (at least not yet)... the `getStateChange()` function never updates `state.time`, so the time pickers will not get the updated start/end times:

https://github.com/palantir/blueprint/blob/38a25b882f1295b334ff67e9f6e7b511d04cbd16/packages/datetime/src/dateRangePicker.tsx#L712

I think you need to also add a unit test for `<DateRangePicker>` controlled usage, to really verify that https://github.com/palantir/blueprint/issues/5502 is fixed after this PR. Something that updates the `value` prop and then verifies that the time pickers get updated to render the new time values.

adidahiya:
It actually looks like someone else tried to make a similar fix, you probably just need to apply the same change in this PR: https://github.com/palantir/blueprint/pull/4980

adidahiya:
Closing due to inactivity

PR #5842, opened 2023-01-13
Fix#5825 DateRangeInput add defaultTimeValue to set start and end Time

### LINKED ISSUE #5825: DateRangeInput2: timePickerProps defaultValue is ignored
#### Environment

- __Package version(s)__: 0.9.6
- __Operating System__: IOS
- __Browser name and version__: Google latest

#### Code Sandbox

https://codesandbox.io/s/little-dawn-x128py?file=/src/DatetimeExample.tsx

#### Steps to reproduce

1. Set the DateTime picker in controlled or uncontrolled mode
2. set a default time in the timePickerProps
3. Select dates

#### Actual behaviour
The default time is set correctly in the time input, but when selecting the dates, it gets ignored.

#### Expected behaviour

Selecting the date should consider the default time being used. 

#### Possible Solution
On selection, get the time values and apply them to the date selected.

Happy to look into it and open a PR. 

Also, I would like to allow the independent setting of timePickerProps. A typical scenario where you want to set a different start and end time. but it is a different topic

danicase:
#### Checklist

- [x] Includes tests
- [x] Update documentation

Adding a `defaultTimeValue` prop to set the state of the TimePickers without a `defaultvalue` or `value`. 
It overrides `timePickerProps.defaultValue` and allows the start and end time to be set independently.

I tried reusing `timePickerProps.defaultValue`, but the type gets messy.

### DIFF
--- a/packages/datetime/src/dateRangePicker.tsx
+++ b/packages/datetime/src/dateRangePicker.tsx
@@ -70,6 +70,13 @@ export interface IDateRangePickerProps extends DatePickerBaseProps, Props {
     contiguousCalendarMonths?: boolean;
 
     /**
+     * Initial time the `TimePicker` will display.
+     * This should not be set if `defaultValue` or `value` is set.
+     * This overrides `timePickerProps.defaultValue`
+     */
+    defaultTimeValue?: DateRange;
+
+    /**
      * Initial `DateRange` the calendar will display as selected.
      * This should not be set if `value` is set.
      */
@@ -196,7 +203,7 @@ export class DateRangePicker extends AbstractPureComponent2<DateRangePickerProps
     public constructor(props: DateRangePickerProps, context?: any) {
         super(props, context);
         const value = getInitialValue(props);
-        const time: DateRange = value;
+        const time: DateRange = getInitialTime(props);
         const initialMonth = getInitialMonth(props, value);
 
         // if the initial month is the last month of the picker's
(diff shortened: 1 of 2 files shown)


palantirtech:
Thanks for your interest in [palantir/blueprint](https://github.com/palantir/blueprint), @danicase! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

danicase:
Hi, I have already signed the contributor license agreement.
@adidahiya, thanks so much for the hard work; I love blueprint. 
Is there anything else I need to do before this PR can even be looked at?

adidahiya:
@danicase sorry for the delay, I'm taking a look now

adidahiya:
It seems like this PR steps around the issue of the component not handling `timePickerProps.defaultValue`. Adding the new prop `defaultTimeValue` as you've done here seems like a separate request, a convenience that is somewhat unrelated to the specific issue in #5825. I'm not sure if we need a new prop.

I think a better approach would be to update the state change handlers to account for `timePickerProps.defaultValue`, I've started a PR here: https://github.com/palantir/blueprint/pull/5940

danicase:
Thanks for the reply, @adidahiya.

Yes, I was hoping to kill two birds with one stone 😄 . With this setting, it does force the time picker to add the defaultValue and shouldn't be ignored by the handler as it is set in the state. 
`defaultTimeValue` was not my first option, 
but the idea to possibly change the type of'timePickerProps.defaultValue`  to be an array and account for start and end Time was much scarier.
It would introduce the idea that time pickers can be treated independently, which is false.

So I checked how some components override some props of his child to fit the need of the component. So I thought to try the same approach. 

I see what you did in https://github.com/palantir/blueprint/pull/5940, 

It's an interesting approach, but why don't we trust the time picker in the first place? This would avoid the addition of a `hasModifiedTimePickers`, which is yet another state to handle. 

In any case, I'm happy to close this if there are better ways to address the issue. 💪

adidahiya:
closing in favor of #5940. I will try to fix that up and merge it this week.

PR #5909, opened 2023-01-31
[popover2] fix(Popover2): get correct click target in shadow DOM

### LINKED ISSUE #5908: Clicking Popover button reopens it when in Shadow DOM
#### Environment

- __Package version(s)__:  4.10.1
- __Operating System__: Win 11

#### Sandbox example
https://codesandbox.io/s/vibrant-solomon-7xvetc

#### Steps to reproduce

1. Put Popover inside Shadow DOM
1. Click on Popover button (it is opened)
1. Click on Popover button once again

#### Actual behavior

Popover closed and immediately opened

#### Expected behavior

Popover should be closed

#### Possible solution

In Popover2.tsx in function "handleOverlayClose" we should check not for "e.target", instead we should find the actual target from "composedPath", the same way it was done in "overlay.tsx" in function "handleDocumentClick"

kutensky:
#### Checklist

- [ ] Includes tests
- [ ] Update documentation

Get the actual click target. The same way it was done in "overlay.tsx" in function "handleDocumentClick"

### DIFF
--- a/packages/popover2/src/popover2.tsx
+++ b/packages/popover2/src/popover2.tsx
@@ -685,7 +685,8 @@ export class Popover2<T extends DefaultPopover2TargetHTMLProps> extends Abstract
             return;
         }
 
-        const eventTarget = e.target as HTMLElement;
+        const event = (e.nativeEvent ?? e) as Event;
+        const eventTarget = (event.composed ? event.composedPath()[0] : event.target) as HTMLElement;
         // if click was in target, target event listener will handle things, so don't close
         if (!Utils.elementIsOrContains(this.targetElement, eventTarget) || e.nativeEvent instanceof KeyboardEvent) {
             this.setOpenState(false, e);


packages/popover2/src/popover2.tsx
```diff
@@ -685,7 +685,7 @@ export class Popover2<T extends DefaultPopover2TargetHTMLProps> extends Abstract
             return;
         }
 
-        const eventTarget = e.target as HTMLElement;
+        const eventTarget = (e.composed ? e.composedPath()[0] : e.target) as HTMLElement;
```
adidahiya: this approach seems reasonable, but I think you'll need to access `e.nativeEvent.composedPath` ([docs](https://beta.reactjs.org/reference/react-dom/components/common#react-event-object))

adidahiya (changes requested):

packages/popover2/src/popover2.tsx
kutensky: It makes sense according to e is "React.SyntheticEvent". But actually the caller of this function passes native event instead of React.SyntheticEvent.
I guess in that case I can check that "e" has "nativeEvent" and use it, otherwise use "e"

kutensky: So I updated the code

adidahiya (approved):

adidahiya:
thanks @kutensky

kutensky:
> thanks @kutensky

Thank you for such a quick response. That fix will help to close a bug in our application

PR #5917, opened 2023-02-02
[core] fix(Tag): always set tabIndex on remove button

bvandrc:
#### Checklist

- [N/A] Includes tests
- [N/A] Update documentation

Even when the overall Tag has `interactive=false`, the `onRemove` remove button should always be tabbable. Cannot imagine a scenario where a user would want to disable tabbing to this button.


packages/core/src/components/tag/tag.tsx
```diff
@@ -165,13 +165,7 @@ export class Tag extends AbstractPureComponent2<TagProps> {
...
         const isLarge = large || tagClasses.indexOf(Classes.LARGE) >= 0;
         const removeButton = isRemovable ? (
-            <button
-                aria-label="Remove Tag"
-                type="button"
-                className={Classes.TAG_REMOVE}
-                onClick={this.onRemoveClick}
-                tabIndex={interactive ? tabIndex : undefined}
-            >
+            <button aria-label="Remove Tag" type="button" className={Classes.TAG_REMOVE} onClick={this.onRemoveClick}>
```
adidahiya: shouldn't this have `tabIndex={tabIndex}`?

bvandrc: that works, changed

adidahiya (approved):

PR #5903, opened 2023-01-30
Multiselect2, TagInput: undefined remove, change handlers should not render tag-remove button

nurseiit:
#### Checklist

- [x] Includes tests
- [x] Update documentation

#### Changes proposed in this pull request:

Updating `<MultiSelect2 />` and `<TagInput />` components to handle cases where their respective `onRemove` or `onChange` handlers not provided.

1. For `TagInput` when passing `onRemove` handler to `<Tag />`, we check if either of `onRemove` or `onChange` props is supplied to `TagInput`. If not, we pass `undefined` – this makes sure we don't render tag-remove button when no handlers are defined.

2. For `MultiSelect2` when passing `onRemove` prop to `<TagInput />`, we check if any of `onRemove` or `tagInputProps.onRemove` or `tagInputProps.onChange` is defined. If not, we do not pass any remove handlers down to `TagInput`.

#### Reviewers should focus on:

Making sure this does not break any components that depend on `TagInput`.

#### Screenshot

Screenshot 2023-01-30 at 14 51 47

PR #5925, opened 2023-02-06
[DRAFT] feat(TagInput, MultiSelect2): readOnly prop

adidahiya:
Currently based on top of https://github.com/palantir/blueprint/pull/5903

Semantically building off @CPerinet's work to add `<InpuGroup readOnly={true} />` in https://github.com/palantir/blueprint/pull/5839

- Add support for a new `readOnly` prop to both `TagInput` and `MultiSelect2` components.
- Demo usage of the new prop in the interactive docs examples.

#### Screenshot

image

What should be done about @nurseiit's use case where they want to make a multi-select interactive only on click/focus, like EditableText?

I've demonstrated one approach to this in this PR, where we use `tagInputProps.onClick` and `popoverProps.onClosed` to set the value of the `readOnly` prop, but the UX is not very smooth - it requires two clicks to open the MultiSelect2:

Note that the visual affordance of EditableText is missing here, too:

#### Checklist

- [ ] Includes tests
- [x] Update documentation

### DIFF
--- a/packages/core/src/components/forms/_common.scss
+++ b/packages/core/src/components/forms/_common.scss
@@ -113,7 +113,8 @@ $control-group-stack: (
     padding-left: $pt-grid-size;
   }
 
-  &[readonly] {
+  &[readonly],
+  &.#{$ns}-read-only {
     box-shadow: inset 0 0 0 1px $pt-divider-black;
   }
 
--- a/packages/core/src/components/forms/inputGroup.tsx
+++ b/packages/core/src/components/forms/inputGroup.tsx
@@ -206,6 +206,7 @@ export class InputGroup extends AbstractPureComponent2<InputGroupProps2, IInputG
             Classes.intentClass(intent),
             {
                 [Classes.DISABLED]: disabled,
+                // N.B. this 'read-only' class does not apply any styles, but may be useful as a CSS selector
                 [Classes.READ_ONLY]: readOnly,
                 [Classes.FILL]: fill,
                 [Classes.LARGE]: large,
--- a/packages/core/src/components/tag-input/tagInput.tsx
+++ b/packages/core/src/components/tag-input/tagInput.tsx
@@ -110,7 +110,8 @@ export interface ITagInputProps extends IntentProps, Props {
      *
      * This callback essentially implements basic `onAdd` and `onRemove` functionality and merges
      * the two handlers into one to simplify controlled usage.
-     * ```
+     *
+     * If both `onRemove` and `onChange` are not provided, rendered tag items will not have remove icons.
      */
     onChange?: (values: React.ReactNode[]) => boolean | void;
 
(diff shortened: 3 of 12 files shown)


adidahiya:
<h3>Adjust popover placement</h3>
Build artifact links for this commit: <strong><a href="https://output.circle-artifacts.com/output/job/e0c742c7-54eb-42c3-8a77-5820f92878f8/artifacts/0/packages/docs-app/dist/index.html">documentation</a> | <a href="https://output.circle-artifacts.com/output/job/e0c742c7-54eb-42c3-8a77-5820f92878f8/artifacts/0/packages/landing-app/dist/index.html">landing</a> | <a href="https://output.circle-artifacts.com/output/job/e0c742c7-54eb-42c3-8a77-5820f92878f8/artifacts/0/packages/table-dev-app/dist/index.html">table</a> | <a href="https://output.circle-artifacts.com/output/job/e0c742c7-54eb-42c3-8a77-5820f92878f8/artifacts/0/packages/demo-app/dist/index.html">demo</a></strong>

CPerinet:
I believe this is slightly different from the read-only PR. We have 3 class of config UI:

- Form: not meant to be seen / edited after saved. It should bias towards showing all interactions by default.
- Editable: is meant to be read and edited at the same time. We want to find a good balance of reducing visual weight but still showing interactions (hover &c.)
- Edit mode: is meant to either be read or edited. Both "mode" can either use different components or use the `read-only` prop of form elements.

In our case, we are looking at "Editable". To extend the logic we have for editable text, we could consider doing the following:
- When not hovering, only show the tags without cross
- When hovering, show a hint of the input border (same style as Editable text). and show the tag crosses
- When focusing, show as current

It would reduce the visual weight of the element while not being hovered.
It would not add any additional clicks for interactions.
It may push elements as users hover the page and we display crosses, pushing items down. For this, we could add a slight delay to debounce un-intentional hovers.

nurseiit:
Hi @adidahiya, thanks for taking this further! I like these changes a lot.

> I've demonstrated one approach to this in this PR, where we use tagInputProps.onClick and popoverProps.onClosed to set the value of the readOnly prop, but the UX is not very smooth - it requires two clicks to open the MultiSelect2:

Following up on the UX part and the need to double-click to open - we ended up handling it manually through `tagInputProps` via:
```
            {
                inputProps: {
                    // prevent clicks on input from closing the popover
                    onClick: (e: React.MouseEvent<HTMLInputElement, MouseEvent>) => e.stopPropagation(),
                    // open popover on first click
                    autoFocus: true,
                },
                tagProps: {
                    // prevent clicks on remove tag button from closing the popover
                    onClick: (e: React.MouseEvent<HTMLDivElement, MouseEvent>) => e.stopPropagation(),
                },
            },
```

Also more on the UX, we noticed that clicking anywhere in the TagInput (which includes removing items via X - tag-remove button) would close the popover of MultiSelect2 (as it would lose focus). As such we manually added handling of click events' propagation via `stopPropagation`.

EDIT: playing around with the newest example, I can see that the transition between readonly and edit modes is a bit sluggish – apart from requiring multiple clicks to open the popup, it requires some specific steps to "lose" the focus and become readonly again. We run into this same behaviour as well and ended up handling the popover in a "controlled" way by updating `popoverProps`' `isOpen` manually via focus events and its `onClose` prop.

EDIT2: (nit) the new example does not allow selecting more than one item at a time.

PR #5950, opened 2023-02-15
[core] feat(Tab): support icons and tags inside tab titles

CPerinet:


adidahiya:
I updated the API to use two props `tagContent` and `tagProps`, I think this is a little cleaner (I don't like the `{ children: "4" }` syntax as much):

packages/core/src/components/tabs/_tabs.scss
```diff
@@ -40,7 +40,9 @@ $tab-indicator-width: 3px !default;
     flex-direction: column;
 
     .#{$ns}-tab {
+      align-items: center;
       border-radius: $pt-border-radius;
+      display: flex;
```
adidahiya, line 45: it seems like we are not demonstrating the behavior of this styling change in the docs example

adidahiya: nevermind, I see it now. looks good

packages/core/src/components/tabs/tabTitle.tsx
```diff
@@ -57,8 +73,14 @@ export class TabTitle extends AbstractPureComponent2<TabTitleProps> {
                 role="tab"
                 tabIndex={disabled ? undefined : selected ? 0 : -1}
             >
+                {icon != null && <Icon icon={icon} intent={intent} className={Classes.TAB_ICON} />}
                 {title}
                 {children}
+                {tagContent != null && (
+                    <Tag minimal={true} intent={tagProps?.intent ?? intent} className={Classes.TAB_TAG} {...tagProps}>
```
adidahiya: we should combine `tagProps.className` with our custom class name here

packages/core/src/components/tabs/tab.tsx
```diff
@@ -61,6 +64,18 @@ export interface ITabProps extends Props, Omit<HTMLDivProps, "id" | "title" | "o
...
     title?: React.ReactNode;
+
+    /** Name of a Blueprint UI icon (or an icon element) to render before the children. */
+    icon?: IconName | MaybeElement;
+
+    /** Content to render inside a `<Tag>` after the children. */
+    tagContent?: TagProps["children"];
+
+    /**
+     * Props to customize the `<Tag>` rendered after the children.
```
adidahiya, line 78: we should note that these tags are minimal by default, so `{ minimal: true }` is unnecessary

adidahiya (changes requested):
self-review

adidahiya (approved):

PR #5951, opened 2023-02-15
Set default icon colour

CPerinet:
Currently, the icon default colour is not applied so it can inherit the text colour. It optimises for the "edge" case where text is "primary" or some other colour. On the flip side, most icons end-up not having their intended colour `$pt-icon-color` or `$pt-dark-icon-color` 

This is a breaking change.

### DIFF
--- a/packages/core/src/components/icon/_icon.scss
+++ b/packages/core/src/components/icon/_icon.scss
@@ -8,6 +8,8 @@
 
 // the icon class which will contain an SVG icon
 .#{$ns}-icon {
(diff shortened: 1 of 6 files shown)


packages/core/src/components/icon/_icon.scss
```diff
@@ -8,6 +8,8 @@
 
 // the icon class which will contain an SVG icon
 .#{$ns}-icon {
+  @include pt-icon-colors();
```
CPerinet, line 11: This is the breaking change.

PR #5953, opened 2023-02-16
feat(tag): always make remove button appear interactive on hover, even when tag itself not interactive

bvandrc:
…n when tag itself not interactive

#### Checklist

- [ ] Includes tests
- [ ] Update documentation


adidahiya (changes requested):
This class has no effect on the remove button. It doesn't apply any styles.

bvandrc:
> This class has no effect on the remove button. It doesn't apply any styles.

OK, will cancel the ticket. May put in a feature request for what I'm looking for here.

PR #5961, opened 2023-02-20
[core] feat(Tabs): new 'fill' prop

CPerinet:


packages/core/src/components/tabs/tabs.tsx
```diff
@@ -285,9 +293,8 @@ export class Tabs extends AbstractPureComponent2<TabsProps, ITabsState> {
 
         let indicatorWrapperStyle: React.CSSProperties = { display: "none" };
         if (selectedTabElement != null) {
-            const { clientHeight, clientWidth, offsetLeft, offsetTop } = selectedTabElement;
```
CPerinet: `clientHeight` is not needed thanks to `align-self: stretch;` on the tab

packages/docs-app/src/styles/_examples.scss
```diff
@@ -482,10 +482,6 @@
   margin-top: $pt-grid-size * 0.5;
 }
 
-.docs-tabs-example .#{$ns}-navbar .#{$ns}-tab {
```
CPerinet, line 550: Example of override that wont be necessary after this PR

packages/core/src/components/tabs/_tabs.scss
```diff
@@ -103,6 +111,10 @@ $tab-indicator-width: 3px !default;
   position: relative;
   vertical-align: top;
 
+  display: flex;
```
CPerinet: This is the dangerous bit. If the tab is used with unwrapped content, flex will apply to it. It will align content on a row and vertically centred.
It removes the need for using `line-height` making the `fill` implementation much nicer and will allow for adding `icon` and `tag` prop in [this pr](https://github.com/palantir/blueprint/pull/5950)

packages/core/src/components/tabs/_tabs.scss
```diff
@@ -94,8 +102,11 @@ $tab-indicator-width: 3px !default;
 
 .#{$ns}-tab {
   @include overflow-ellipsis();
+  align-items: center;
+  align-self: stretch;
   color: $pt-text-color;
   cursor: pointer;
+  display: flex;
```
adidahiya, line 111: ~It looks like this does cause some issues for existing code which is inserting tags into tab titles (red border added for extra clarity):~

Edit: nevermind, we're good. the `align-items` style seems to fix things

| Before | After (without `align-items: center`) | After (with `align-items: center`) |
| - | - | - |
| image | image | image |


the code for this is something like:

```ts
<Tab
  title={
    <>
      Pull requests
      {" "}
      <span className={classNames(Classes.TAG, Classes.MINIMAL, Classes.ROUND)}>0</span>
    </>
  }
/>
```

packages/core/src/components/tabs/tabs.tsx
adidahiya: it looks like we still need this for vertical tabs.

adidahiya (approved):

PR #5963, opened 2023-02-20
New vertical tab style

CPerinet:
Modernisation of the select style for vertical tabs.

Screenshot 2023-02-20 at 19 09 26
Screenshot 2023-02-20 at 19 09 33


adidahiya:
@CPerinet it looks like the line height has increased, the tabs are less compact now. Is that intentional?

Also is it possible to preserve some sort of transition animation like we used to have:

PR #5966, opened 2023-02-22
[core] feat(TagInput): resizable input that scales as you type

JosephChotard:
#### Checklist

- [N/A] Includes tests
- [N/A] Update documentation

Currently the `TagInput` component (used by `MultiSelect2`) does not expand to accomodate the text the user is typing. This is because `input` elements don't scale with their input. This causes annoying situations where the user cannot see what they are typing when there is space to display it. 
To address this issue, I created a small `ResizeableInput` component (could be included in `TagInput` but this makes the intent clearer). This new component acts like a regular `input` element for every other component but it has some coding magic that causes it to fit the size of it's content. It will never grow bigger than it's closest parent with a `position: relative`. 

This applies to both `TagInput` and `MultiSelect`
I checked that:
* The input never stretches wider than the parent
* The input scales with it's content
* The input is never smaller than 80px
* The existing props still work (large, disabled, add on blur, ...)

This PR does not introduce changes to the API and is fully backwards compatible.

#### Screenshot

Old Input scaling:
numbers - old
New Input scaling:
numbers - new

Old input text being cut off
wrap old

New input text wrapping
wrap new


packages/core/src/components/tag-input/resizeableInput.tsx
```diff
@@ -0,0 +1,50 @@
...
+ */
+
+import React, { forwardRef, useEffect, useRef, useState } from "react";
+
+import { HTMLInputProps } from "../../common";
+
+type ResizeableInputProps = HTMLInputProps;
+export type Ref = HTMLInputElement;
+
+export const ResizeableInput = forwardRef<Ref, ResizeableInputProps>(function ResizeableInput(props, ref) {
```
adidahiya: I believe "resizable" is more correct, at least for US English. We use that elsewhere in Blueprint, so please rename all references to "resizeable" -> "resizable" (including the file name).

packages/core/src/components/tag-input/resizeableInput.tsx
```diff
@@ -0,0 +1,50 @@
...
+    const [content, setContent] = useState("");
+    const [width, setWidth] = useState(0);
+    const span = useRef<HTMLSpanElement>(null);
+
+    useEffect(() => {
+        if (span.current) {
```
adidahiya: nit: please use explicit boolean coercion

```suggestion
        if (span.current != null) {
```

packages/core/src/components/tag-input/resizeableInput.tsx
```diff
@@ -0,0 +1,50 @@
...
+            setWidth(span.current.offsetWidth);
+        }
+    }, [content]);
+
+    const { onChange, ...otherProps } = props;
+
+    const changeHandler: React.ChangeEventHandler<HTMLInputElement> = evt => {
```
adidahiya: nit: use `handle...` naming for event callbacks

```suggestion
    const handleInputChange: React.ChangeEventHandler<HTMLInputElement> = evt => {
```

packages/core/src/components/tag-input/resizeableInput.tsx
```diff
@@ -0,0 +1,50 @@
...
+        setContent(evt.target.value);
+        onChange?.(evt);
+    };
+
+    return (
+        <span>
+            <span
+                id="hide"
+                ref={span}
+                style={{
```
adidahiya: prefer styling with CSS rather than inline styles. you should create a `_resizableInput.scss` file next to this one and create a new class following our existing naming conventions...

```scss
.#{$ns}-resizable-input-span {
  max-height: 0;
  max-width: 100%;
  min-width: 80px;
  position: absolute;
  white-space: nowrap;
  z-index: -100;
}
```

packages/core/src/components/tag-input/resizeableInput.tsx
```diff
@@ -0,0 +1,50 @@
...
+                    maxHeight: "0",
+                    maxWidth: "100%",
+                    minWidth: "80px",
+                    position: "absolute",
+                    whiteSpace: "nowrap",
+                    zIndex: -100,
```
adidahiya: why is this necessary? can it be simply `-1`? can you leave a code comment?

packages/core/src/components/tag-input/resizeableInput.tsx
```diff
@@ -0,0 +1,50 @@
...
+                }}
+            >
+                {/* Need to replace spaces with the html character for them to be preserved */}
```
adidahiya: :+1:

adidahiya (changes requested):
Nice, this seems to work really well in the [docs preview build](https://output.circle-artifacts.com/output/job/5d65e360-4486-435a-998d-09b9bccb2b61/artifacts/0/packages/docs-app/dist/index.html#core/components/tag-input)!

Looks like the tests are failing because we're using shallow rendering in many of the TagInput tests, and the `<ResizeableInput>` children don't get rendered. We can fix that by switching to `mount()` instead of `shallow()`.

JosephChotard:
https://github.com/enzymejs/enzyme/issues/1473
https://github.com/enzymejs/enzyme/issues/1081
Changed the tests to use `mount` and modified a few `.simulate` calls as they caused the tests to hang (not sure why), instead I call the prop directly, I don't believe we need to test the event passing as that's React's responsibility.

packages/core/src/components/tag-input/resizeableInput.tsx
JosephChotard: ack

packages/core/src/components/tag-input/resizeableInput.tsx
JosephChotard: whoops, fixing it

JosephChotard:
@adidahiya done all the changes, lmk if there's anything else

packages/core/src/common/classes.ts
```diff
@@ -163,6 +163,8 @@ export const INPUT_GROUP = `${INPUT}-group`;
 export const INPUT_LEFT_CONTAINER = `${INPUT}-left-container`;
 export const INPUT_ACTION = `${INPUT}-action`;
 
+export const INPUT_RESIZABLE_INPUT_SPAN = `${NS}-resizable-input-span`;
```
adidahiya: nit: the variable name should match the class name more closely:

```suggestion
export const RESIZABLE_INPUT_SPAN = `${NS}-resizable-input-span`;
```

packages/core/src/components/_index.scss
```diff
@@ -37,6 +37,7 @@
 @import "tabs/tabs";
 @import "tag/tag";
 @import "tag-input/tag-input";
+@import "tag-input/resizeable-input";
```
adidahiya: missed a file name here:

```suggestion
@import "tag-input/resizable-input";
```

packages/core/src/components/tag-input/resizableInput.tsx
```diff
@@ -0,0 +1,39 @@
+/* !
+ * (c) Copyright 2022 Palantir Technologies Inc. All rights reserved.
+ */
+
+import React, { forwardRef, useEffect, useRef, useState } from "react";
+
+import { Classes, HTMLInputProps } from "../../common";
+
+type ResizableInputProps = HTMLInputProps;
+export type Ref = HTMLInputElement;
```
adidahiya, line 9: I don't think this typedef needs to exist. Just inline the type: `forwardRef<HTMLInputElement, ResizableInputProps>`

packages/core/src/components/tag-input/resizableInput.tsx
```diff
@@ -0,0 +1,39 @@
+/* !
+ * (c) Copyright 2022 Palantir Technologies Inc. All rights reserved.
+ */
+
+import React, { forwardRef, useEffect, useRef, useState } from "react";
+
+import { Classes, HTMLInputProps } from "../../common";
+
+type ResizableInputProps = HTMLInputProps;
```
adidahiya: nit: export for consistency

```suggestion
export type ResizableInputProps = HTMLInputProps;
```

packages/core/test/tag-input/tagInputTests.tsx
```diff
@@ -32,19 +26,17 @@ import { Button, Classes, Intent, Keys, Tag, TagInput, TagInputProps } from "../
...
 describe("<TagInput>", () => {
     it("passes inputProps to input element", () => {
         const onBlur = sinon.spy();
-        const input = shallow(<TagInput values={VALUES} inputProps={{ autoFocus: true, onBlur }} />).find("input");
+        const input = mount(<TagInput values={VALUES} inputProps={{ autoFocus: true, onBlur }} />).find("input");
         assert.isTrue(input.prop("autoFocus"));
         // check that event handler is proxied
         const fakeEvent = { flag: "yes" };
-        input.simulate("blur", fakeEvent);
+        input.prop("onBlur")?.(fakeEvent as any);
```
adidahiya, line 39: I don't love this change but it appears to be the best path forward given the Enzyme issues you just linked. I'm surprised to learn that `.simulate()` is no longer recommended. And at some point I will move away from Enzyme anyway so these tests will get updated...

adidahiya (changes requested):
looking good, just a few more minor things

packages/core/test/tag-input/tagInputTests.tsx
JosephChotard: agreed

JosephChotard:
@adidahiya thanks for all the feedback, should be addressed now

adidahiya (approved):
LGTM, thanks for the thorough PR @JosephChotard :+1:

bvandercar-vt:
@adidahiya this change has created a bug where the `placeholder` is no longer present for a `MultiSelect2`

nerdstep:
Looks like there might be a regression with the input width not expanding to the full width of the component when using the `fill` prop:

image

packages/core/src/components/tag-input/resizableInput.tsx
```diff
@@ -0,0 +1,38 @@
...
+    const { onChange, ...otherProps } = props;
+
+    const handleInputChange: React.ChangeEventHandler<HTMLInputElement> = evt => {
+        onChange?.(evt);
+        setContent(evt?.target?.value ?? "");
+    };
+
+    return (
+        <span>
+            <span id="hide" ref={span} className={Classes.RESIZABLE_INPUT_SPAN}>
```
bvandrc, line 31: Not sure why this `id` prop is needed / where it's used, but it's causing aXe a11y failures because all `id`s must be **unique**.

Must either remove `id` prop, or use `uniqueId` for it.

bvandrc: PR: https://github.com/palantir/blueprint/pull/5984

adidahiya:
@nerdstep thanks for pointing out the regression. Fix here: https://github.com/palantir/blueprint/pull/5985

PR #5967, opened 2023-02-22
[core] feat(MenuDivider): add titleId prop

bvandrc:
add `id` prop so can use `aria-labelledby` on menuDivider title text

#### Checklist

- [NA] Includes tests
- [X] Update documentation

### DIFF
--- a/packages/core/src/components/menu/menuDivider.tsx
+++ b/packages/core/src/components/menu/menuDivider.tsx
@@ -32,7 +32,7 @@ export interface IMenuDividerProps extends Props {
     title?: React.ReactNode;
 
     /** Optional `id` prop for the header title. */
-    id?: string;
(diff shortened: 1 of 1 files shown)


packages/core/src/components/menu/menuDivider.tsx
```diff
@@ -30,6 +30,9 @@ export interface IMenuDividerProps extends Props {
 
     /** Optional header title. */
     title?: React.ReactNode;
+
+    /** Optional `id` prop for the header title. */
+    id?: string;
```
adidahiya:
```suggestion
    titleId?: string;
```

adidahiya (changes requested):

adidahiya (approved):

PR #5975, opened 2023-02-27
[core] fix(TextArea): resize vertically in controlled mode

### LINKED ISSUE #5828: TextArea Does Not Resize Vertically When Props Change After Initial Mount
#### Environment

This bug seems to be present in all environments, but sharing my details for completeness:

- __Package version(s)__: 4.12.0
- __Operating System__: macOS
- __Browser name and version__: Chrome 108

#### Code Sandbox

https://codesandbox.io/s/text-area-resize-bug-g0iz78?file=/src/index.tsx

#### Steps to reproduce

See linked code sandbox; but code is short, so will share it here.

```
const App = () => {
  const [text, setText] = useState("A");
  const longText = "AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA";
  useEffect(() => {
    const timer = setTimeout(() => {
      setText(longText);
    }, 5000);
    return () => clearTimeout(timer);
  }, []);
  return (
    <div>
      <TextArea value={text} growVertically={true} />
    </div>
  );
};
```

#### Actual behavior

The component does not grow vertically when props change post initial mount.

#### Expected behavior

The component should grow vertically when props change post initial mount.

#### Possible solution

As a quick fix, copying the logic from https://github.com/palantir/blueprint/blob/develop/packages/core/src/components/forms/textArea.tsx#L78 into componentDidUpdate should do the job. Something like the below:

```
public componentDidUpdate(prevProps: ITextAreaProps) {
    if (this.props.growVertically && this.textareaElement !== null && prevProps.value !== this.props.value) {
        this.setState({
            height: this.textareaElement?.scrollHeight,
        });
    }
}
```

AaronJRubin:
#### Checklist

- [ ] Includes tests
- [ ] Update documentation

Resize text area when its value changes due to a new string being passed in through props

How we can test this - at first, I tried adding a test like the below:

```ts
    it("resizes when props change if growVertically is true", () => {
        const initialText = "A";
        const longText = "AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA";
        const wrapper = mount(<TextArea growVertically={true} value={initialText} style={{ marginTop: 10 }} />);
        const textarea = wrapper.find("textarea");
        const initialHeight = (textarea.getDOMNode() as HTMLElement).style.height;
        wrapper.setProps({ value: longText });
        const newHeight = (textarea.getDOMNode() as HTMLElement).style.height;
        assert.notEqual(newHeight, initialHeight);
    });
```

But both newHeight and initialHeight are always 0, so this test fails. The existing test for the resize behavior (`can fit large initial content`) appears to be spurious; the comparison at https://github.com/palantir/blueprint/blob/develop/packages/core/test/forms/textAreaTests.tsx#L67 is also just comparing the strings `0px` and `0px`.

I manually tested this by building another app against the locally-compiled blueprint-core package and observing that the buggy behavior was fixed, but I'm not sure if that meets our standards for testing.

### DIFF
--- a/packages/core/src/components/forms/textArea.tsx
+++ b/packages/core/src/components/forms/textArea.tsx
@@ -54,10 +54,6 @@ export interface TextAreaState {
     height?: number;
 }
 
-export interface TextAreaSnapshot {
-    scrollHeight: number;
-}
-
 // this component is simple enough that tests would be purely tautological.
 /* istanbul ignore next */
 /**
@@ -65,7 +61,7 @@ export interface TextAreaSnapshot {
  *
  * @see https://blueprintjs.com/docs/#core/components/text-inputs.text-area
  */
-export class TextArea extends AbstractPureComponent2<TextAreaProps, TextAreaState, TextAreaSnapshot> {
+export class TextArea extends AbstractPureComponent2<TextAreaProps, TextAreaState> {
     public static displayName = `${DISPLAYNAME_PREFIX}.TextArea`;
 
     public state: TextAreaState = {};
@@ -92,7 +88,7 @@ export class TextArea extends AbstractPureComponent2<TextAreaProps, TextAreaStat
         this.maybeSyncHeightToScrollHeight();
     }
 
-    public componentDidUpdate(prevProps: TextAreaProps, _prevState: TextAreaState) {
+    public componentDidUpdate(prevProps: TextAreaProps) {
         if (prevProps.inputRef !== this.props.inputRef) {
             setRef(prevProps.inputRef, null);
             this.handleRef = refHandler(this, "textareaElement", this.props.inputRef);
--- a/packages/core/src/components/forms/textArea.tsx
+++ b/packages/core/src/components/forms/textArea.tsx
@@ -50,8 +50,12 @@ export interface ITextAreaProps extends IntentProps, Props, React.TextareaHTMLAt
     inputRef?: React.Ref<HTMLTextAreaElement>;
 }
 
-export interface ITextAreaState {
+export interface TextAreaState {
     height?: number;
+}
+
+export interface TextAreaSnapshot {
+    scrollHeight: number;
 }
 
 // this component is simple enough that tests would be purely tautological.
@@ -61,10 +65,10 @@ export interface ITextAreaState {
  *
  * @see https://blueprintjs.com/docs/#core/components/text-inputs.text-area
  */
-export class TextArea extends AbstractPureComponent2<TextAreaProps, ITextAreaState> {
+export class TextArea extends AbstractPureComponent2<TextAreaProps, TextAreaState, TextAreaSnapshot> {
     public static displayName = `${DISPLAYNAME_PREFIX}.TextArea`;
 
-    public state: ITextAreaState = {};
+    public state: TextAreaState = {};
 
     // used to measure and set the height of the component on first mount
     public textareaElement: HTMLTextAreaElement | null = null;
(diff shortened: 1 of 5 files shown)


adidahiya:
@AaronJRubin I think the tests are not working because we're not mounting into an `attachTo` container element. I'll poke at the test suite in this PR

adidahiya:
@AaronJRubin I wasn't able to get the tests working, but I've added a little example to the docs which demonstrates controlled usage w/ `growVertically`. it seems to work with your change. I went ahead and implemented some code style fixes / improvements on top of your commit.



Note that it does not _shrink vertically_ automatically, but this matches the existing behavior with the uncontrolled API.

packages/core/src/components/forms/textArea.tsx
```diff
@@ -50,21 +50,25 @@ export interface ITextAreaProps extends IntentProps, Props, React.TextareaHTMLAt
     inputRef?: React.Ref<HTMLTextAreaElement>;
 }
 
-export interface ITextAreaState {
+export interface TextAreaState {
     height?: number;
 }
 
+export interface TextAreaSnapshot {
```
adidahiya: oops, I'm not using this snapshot anymore, that was WIP... will remove

packages/core/src/components/forms/textArea.tsx
```diff
@@ -75,22 +79,29 @@ export class TextArea extends AbstractPureComponent2<TextAreaProps, ITextAreaSta
         this.props.inputRef,
     );
 
-    public componentDidMount() {
-        if (this.props.growVertically && this.textareaElement !== null) {
-            // HACKHACK: this should probably be done in getSnapshotBeforeUpdate
```
adidahiya, line 80: this doesn't really help us. removed this comment

packages/core/src/components/forms/textArea.tsx
```diff
@@ -75,22 +79,29 @@ export class TextArea extends AbstractPureComponent2<TextAreaProps, ITextAreaSta
...
-            /* eslint-disable-next-line react/no-did-mount-set-state */
```
adidahiya, line 81: I don't see a significantly better way to do this, and I think the potential for layout thrashing is minimal. removed this comment.

packages/core/src/components/forms/textArea.tsx
```diff
@@ -75,22 +79,29 @@ export class TextArea extends AbstractPureComponent2<TextAreaProps, ITextAreaSta
...
+            }
         }
+    };
+
+    public componentDidMount() {
+        this.maybeSyncHeightToScrollHeight();
     }
 
-    public componentDidUpdate(prevProps: TextAreaProps) {
+    public componentDidUpdate(prevProps: TextAreaProps, _prevState: TextAreaState) {
```
adidahiya: can revert this line change

packages/core/src/components/forms/textArea.tsx
```diff
@@ -75,22 +79,29 @@ export class TextArea extends AbstractPureComponent2<TextAreaProps, ITextAreaSta
...
         if (prevProps.inputRef !== this.props.inputRef) {
             setRef(prevProps.inputRef, null);
             this.handleRef = refHandler(this, "textareaElement", this.props.inputRef);
             setRef(this.props.inputRef, this.textareaElement);
         }
+
+        if (prevProps.value !== this.props.value || prevProps.style !== this.props.style) {
```
adidahiya, line 98: note that `props.style` changes can cause layout changes which affect `scrollHeight`

adidahiya (changes requested):
self-review

adidahiya (approved):

PR #5978, opened 2023-02-28
[select] fix: set aria-disabled attribute on combobox elements

bvandrc:
Disabled prop should be applied to the element that has `role="combobox"`, not just the `input`.

### DIFF
--- a/packages/select/src/components/multi-select/multiSelect2.tsx
+++ b/packages/select/src/components/multi-select/multiSelect2.tsx
@@ -288,8 +288,8 @@ export class MultiSelect2<T> extends AbstractPureComponent2<MultiSelect2Props<T>
                     "aria-controls": this.listboxId,
                     ...popoverTargetProps,
                     ...targetProps,
-                    "aria-expanded": isOpen,
                     "aria-disabled": disabled,
+                    "aria-expanded": isOpen,
                     // Note that we must set FILL here in addition to TagInput to get the wrapper element to full width
                     className: classNames(targetProps.className, popoverTargetProps.className, {
                         [CoreClasses.FILL]: fill,
--- a/packages/select/src/components/multi-select/multiSelect2.tsx
+++ b/packages/select/src/components/multi-select/multiSelect2.tsx
@@ -289,11 +289,11 @@ export class MultiSelect2<T> extends AbstractPureComponent2<MultiSelect2Props<T>
                     ...popoverTargetProps,
                     ...targetProps,
                     "aria-expanded": isOpen,
+                    "aria-disabled": disabled,
                     // Note that we must set FILL here in addition to TagInput to get the wrapper element to full width
                     className: classNames(targetProps.className, popoverTargetProps.className, {
                         [CoreClasses.FILL]: fill,
                     }),
-                    disabled,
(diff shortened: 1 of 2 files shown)


adidahiya:
According to [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/disabled#overview):

> The disabled attribute is supported by [`<button>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button), [`<fieldset>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/fieldset), [`<keygen>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/keygen), [`<optgroup>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/optgroup), [`<option>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/option), [`<select>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/select), [`<textarea>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/textarea) and [`<input>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input).

So it sounds like it's not valid on `<div>` or `<span>` elements... do you have any documentation to support this change?

bvandrc:
> According to [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/disabled#overview):
> 
> > The disabled attribute is supported by [`<button>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button), [`<fieldset>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/fieldset), [`<keygen>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/keygen), [`<optgroup>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/optgroup), [`<option>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/option), [`<select>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/select), [`<textarea>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/textarea) and [`<input>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input).
> 
> So it sounds like it's not valid on `<div>` or `<span>` elements... do you have any documentation to support this change?

Ah, you're right. I thought changing via `role="combobox"` would allow `disabled`. But this documentation says:

image

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-disabled

Switched to `aria-disabled`.

packages/select/src/components/multi-select/multiSelect2.tsx
```diff
@@ -289,6 +289,7 @@ export class MultiSelect2<T> extends AbstractPureComponent2<MultiSelect2Props<T>
                     ...popoverTargetProps,
                     ...targetProps,
                     "aria-expanded": isOpen,
+                    "aria-disabled": disabled,
```
adidahiya: you need to swap these lines (yes, I know the lint rule is kind of annoying)

bvandrc: lint rules are there for good reason! Fixed.

adidahiya (approved):

PR #5997, opened 2023-03-06
[core] fix(TagInput): make sure ResizableInput span is invisible

JosephChotard:
#### Checklist

- [ N/A] Includes tests
- [ N/A] Update documentation

In some scenarios, other styles might clash making the supposedly hidden span not invisible...

#### Screenshot

Example of it broken:
Screenshot 2023-03-06 at 17 53 53

### DIFF
--- a/packages/core/src/components/tag-input/_resizable-input.scss
+++ b/packages/core/src/components/tag-input/_resizable-input.scss
@@ -5,9 +5,9 @@
   max-height: 0;
   max-width: 100%;
   min-width: $pt-grid-size * 8;
+  opacity: 0;
+  overflow: hidden;
   position: absolute;
   white-space: nowrap;
   z-index: -1;
-  opacity: 0;
-  overflow: hidden;
 }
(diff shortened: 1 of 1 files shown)


JosephChotard:
@adidahiya I also just noticed that when there are no tags selected the input doesn't scale with the content (it only does it when at least one tag is selected). I've found some workarounds but they're quite ugly so will refine and submit another PR for it

adidahiya:
> when there are no tags selected the input doesn't scale with the content

@JosephChotard by "content" here, do you mean the placeholder text?

packages/core/src/components/tag-input/_resizable-input.scss
```diff
@@ -8,4 +8,6 @@
   position: absolute;
   white-space: nowrap;
   z-index: -1;
+  opacity: 0;
+  overflow: hidden;
```
adidahiya: need to alphabetize CSS properties

JosephChotard: done

JosephChotard:
> > when there are no tags selected the input doesn't scale with the content
> 
> @JosephChotard by "content" here, do you mean the placeholder text?

No the actual text typed by the user. The input only expands larger when at least one tag is present. Seems to be a styling bug and not a logic one.
Screenshot 2023-03-06 at 21 24 13
Screenshot 2023-03-06 at 21 23 59

adidahiya (approved):

PR #6007, opened 2023-03-08
[datetime] fix(TimeInput): text color when disabled in dark mode

bvandrc:
Bug fix:
TimeInput component currently does not visually look disabled when in dark mode.
Have confirmed this change is effective.

### DIFF
--- a/packages/datetime/src/_timepicker.scss
+++ b/packages/datetime/src/_timepicker.scss
@@ -100,7 +100,7 @@ $timepicker-control-width: $pt-grid-size * 3.3 !default;
     .#{$ns}-timepicker-divider-text {
       // disabled state always overrides over styles
       /* stylelint-disable-next-line declaration-no-important */
-      color: $input-color-disabled  !important;
+      color: $input-color-disabled !important;
       cursor: not-allowed;
     }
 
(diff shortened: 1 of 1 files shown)


bvandrc:
@adidahiya I have finally got docs app running on windows-- have confirmed this change is effective.

packages/datetime/src/_timepicker.scss
```diff
@@ -98,13 +98,17 @@ $timepicker-control-width: $pt-grid-size * 3.3 !default;
 
     .#{$ns}-timepicker-input,
     .#{$ns}-timepicker-divider-text {
-      color: $input-color-disabled;
+      // disabled state always overrides over styles
+      /* stylelint-disable-next-line declaration-no-important */
+      color: $input-color-disabled !important;
```
adidahiya: don't use `!important`. instead, add style rules inside the `.#{$ns}-dark` block to achieve higher specificity. this is also using the wrong color. there is a different color for disabled input text in dark theme.

you should add near the bottom of this file:

```scss
.#{$ns}-dark .#{$ns}-timepicker {
  // ...

  &.#{$ns}-disabled {
    .#{$ns}-timepicker-input,
    .#{$ns}-timepicker-divider-text,
    .#{$ns}-timepicker-arrow-button,
    .#{$ns}-timepicker-arrow-button:hover {
      color: $dark-input-color-disabled;
    }
  }
}
```

adidahiya (changes requested):

adidahiya:
Btw for reference, here is the bug in @blueprintjs/datetime v4.4.25:

packages/datetime/src/_timepicker.scss
bvandrc: done and tested!

adidahiya (approved):

PR #5935, opened 2023-02-10
[datetime2] feat: DatePicker3 using react-day-picker v8

adidahiya:
### Checklist

- [x] Includes tests
- [x] Update documentation

### Screenshot

image

### Changes proposed in this pull request:

Add react-day-picker (a.k.a. "RDP") v8.x as a dependency of `@blueprintjs/datetime2` and create new "v3" variant of DatePicker which uses this new version of the day picker library.

<details>
<summary>Details on upgrade complications</summary>

When I first looked at the [v8 upgrade guide](https://react-day-picker.js.org/guides/upgrading), I thought we might be able to mitigate the breaking changes and handle them all within Blueprint code. It would be nice if we could support _both_ v7 and v8 at the same time. Reviewing the breaking changes suggests that this might be possible, as they fall into a few manageable categories:

* :ballot_box_with_check: breaking changes that require no mitigation, or a simple rename within Blueprint code
  * `todayButton` removed
  * `canChangeMonth` → `disableNavigation`
  * `selectedDays` → `selected`
  * DayPickerInput removed
* ✅ breaking name changes that can (in theory) be mitigated / managed by Blueprint and not exposed to consumers (although this would generate code cruft)
    * `initialMonth` → `defaultMonth`
    * `disabledDays` → `disabled`
    * `showWeekNumbers` → `showWeekNumber`
    * `daysOfWeek` → `dayOfWeek`
    * `firstDayOfWeek` → `weekStartsOn`
* ❓not sure about these changes
    * v8 use the locale setting to calculate the week days and numbers. The previous versions were using [ISO week dates](https://en.wikipedia.org/wiki/ISO_week_date). To toggle the ISO week dates, use the `ISOWeek` prop.

However, after attempting the upgrade and compiling with the new `react-day-picker` types, it became clear that it was going to be difficult to implement `DatePicker` and `DateRangePicker` in a way that worked with both v7 and v8. Some of those changes include:

- how caption and navbar elements are rendered
- how modifier styles (both built-in and custom) are applied

So, instead of making the `@blueprintjs/datetime` components compatible with react-day-picker v8, I've decided to create "v2" variants in `@blueprintjs/datetime2`.

</details>

### Questions for reviewers

Edit: we decided to go with Option 1, using the "v3" name.

[...]

PR #6025, opened 2023-03-15
a11y(DatePicker): move year and month selects from caption to navbar (fixes tab order and a11y failure)

### LINKED ISSUE #6058: incorrect tab order in DatePicker header, DatePicker also failing aXe a11y checks
#### Environment

#### Steps to reproduce

1. Run aXe devtools on any DatePicker
2. See failure:
image

Failure: a grid caption should not contain interactive elements. The caption is within the role="grid" element, and the grid's caption should not contain the selects, because it's supposed to be only that, a caption for the grid.

3. Tab through DatePicker. Element order goes left arrow -> right arrow -> month select -> year select. Should go from left to right visibily, left arrow ->  month select -> year select ->

#### Expected behavior

- Have correct tab order based on visual order of elements
- Don't fail aXe rules

#### Possible solution

Move month year selects to `DayPicker`'s `navbarElement` and out of its `captionElement`. Place in between the prev/next arrows. Modify CSS to keep visually same.

bvandrc:
Fixes #6058

Reason:
- Fix tab order of datepicker navbar
- Fix A11y failure issue-- grid caption should not contain interactive elements

#### Checklist

- [X] Includes tests
- [N/A] Update documentation

Move Year and Month selects out of React-Day-Picker caption (`renderCaption`) to navbar (`renderNavbar`). Fixes tab order of these elements with the navigation arrows, and fixes accessibility issue-- a `grid` caption should not contain interactive elements. The caption is within the role="grid" element, and the grid's `caption` should not contain the `select`s, because it's supposed to be only that, a caption for the grid. 

A11y failures aside, these components generally make more sense as a "navigation" component than a "caption".

#### Screenshot
**Tab Order Before and After:**
ScreenRecorderProject1     ScreenRecorderProject2

**Proof that is visually and functionally same as before:**
ScreenRecorderProject3_2

**aXe failure:**

image


adidahiya:
> grid caption should not contain interactive elements

@bvandercar-vt do you have documentation you can reference for this claim?

bvandercar-vt:
> > grid caption should not contain interactive elements
> 
> @bvandercar-vt do you have documentation you can reference for this claim?

aXe failure:

image

Problem is that the `caption` is within the `role="grid"` element. Not only does this break the above a11y rule, it probably shouldn't be part of the calendar days grid anyways.

Plus, this fixes the tab order.

bvandercar-vt:
@adidahiya made a new update. Realized that before, the Month Year text was in the `grid` caption which could be beneficial to screenreaders. So, added a `screenreader-only` invisible caption that includes that text in the caption. That element is now the new `datePickerCaption`.

bvandrc:
> > grid caption should not contain interactive elements
> 
> @bvandercar-vt do you have documentation you can reference for this claim?

Updated the PR description with the aXe failure and more detailed info.

bvandrc:
@adidahiya I added GIFs to PR description

bvandrc:
@adidahiya refactored this

adidahiya:
thanks for your effort so far on this PR @bvandercar-vt. I feel like this change has the chance to be breaking in some very rare cases since we try to guarantee no DOM layout breaks in Blueprint components. There are also some major changes to date picker caption / navbar rendering coming in react-day-picker v8 (see https://github.com/palantir/blueprint/pull/5935). For these reasons, I feel like it's probably best to wait to make this change until we upgrade to react-day-picker v8, where we expect to make some DOM breaking changes anyway. This should be possible in the next few weeks / month.

adidahiya:
@bvandercar-vt we are now using react-day-picker v8 in the "V3" components in `@blueprintjs/datetime2`. All a11y fixes should be made there.

Most of the code paths now use react-day-picker's caption rendering, so if there are a11y issues there, you should bring it up with that library maintainer.

One code path goes through a custom caption renderer that lives inside Blueprint code, though: non-contiguous date range pickers with multiple months. Those [utilize](https://github.com/palantir/blueprint/blob/94a1963648bc5d820a66c8e64ee8eeae22327369/packages/datetime2/src/components/date-range-picker3/nonContiguousDayRangePicker.tsx#L81) a custom [`<DatePicker3Caption>` component](https://github.com/palantir/blueprint/blob/94a1963648bc5d820a66c8e64ee8eeae22327369/packages/datetime2/src/components/react-day-picker/datePicker3Caption.tsx).

I'm going to close this PR for now, and if you want to pick up the a11y work, you may do so in a new PR. Thanks!

PR #6044, opened 2023-03-28
fix(MenuItem,MenuItem2): Enter keypress should trigger onClick

### LINKED ISSUE #6056: MenuItem / MenuItem2 enter button doesn't have same effect as click
#### Environment

#### Code Sandbox

Link to a minimal repro (fork [this code sandbox](https://codesandbox.io/p/sandbox/blueprint-sandbox-2023-fjo3z4)): 

#### Steps to reproduce

1. Use tab key to focus to a MenuItem that does not have the `listoption` role (in other words, a MenuItem that is not part of `Select`/`MultiSelect`)-- one good example of this is the left shortcuts in https://blueprintjs.com/docs/#datetime/daterangepicker :

image

2. Hit enter key or space bar. Nothing happens.

#### Actual behavior

Has no effect.

#### Expected behavior

Same behavior as clicking the menu item.

#### Possible solution

PR: https://github.com/palantir/blueprint/pull/6044

bvandrc:
#### Checklist

- [X] Includes tests
- [N/A] Update documentation

Enable pressing enter on `menuitem` components. React does not apply `onClick` when enter is pressed on a `role="menuitem"` component like it does for a button, apply fix for this.

### DIFF
--- a/packages/core/src/common/utils/domUtils.ts
+++ b/packages/core/src/common/utils/domUtils.ts
@@ -142,3 +142,8 @@ function throttleImpl<T extends Function>(
     };
     return func as any as T;
 }
+
+export function clickElementOnKeyPress(keys: string[]) {
+    return (e: React.KeyboardEvent) =>
+        keys.some(key => e.key === key) && e.target.dispatchEvent(new MouseEvent("click", { ...e, view: undefined }));
+}
--- a/packages/core/src/common/utils/index.ts
+++ b/packages/core/src/common/utils/index.ts
@@ -16,6 +16,7 @@
 
 export * from "./compareUtils";
 export {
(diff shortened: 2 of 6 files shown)


adidahiya:
I can confirm this works in the [docs preview build](https://output.circle-artifacts.com/output/job/ac7cb896-2771-4337-86d5-24eddf4c54a1/artifacts/0/packages/docs-app/dist/index.html#datetime/datepicker) (this behavior was not possible before):



Not sure how useful the spacebar triggering will be, as that often causes the page to scroll, but it seems reasonable to leave in for now.

adidahiya (approved):
thanks for adding unit tests :+1:

PR #6045, opened 2023-03-28
fix(tabTitle): incorrect tabIndex

bvandrc:
#### Checklist

- [N/A] Includes tests
- [N/A] Update documentation

Fix tabIndex for tabTitle-- available tabs SHOULD be tabbable.


adidahiya (dismissed):

packages/core/src/components/tabs/tabTitle.tsx
```diff
@@ -71,7 +71,7 @@ export class TabTitle extends AbstractPureComponent2<TabTitleProps> {
                 id={generateTabTitleId(parentId, id)}
                 onClick={disabled ? undefined : this.handleClick}
                 role="tab"
-                tabIndex={disabled ? undefined : selected ? 0 : -1}
+                tabIndex={disabled ? undefined : selected ? -1 : 0}
```
adidahiya, line 74: do you have any documentation to justify this change?

bvandercar-vt: If the tab is _not_ selected, you'd want to be able to use the tab button to select it, hence setting `tabIndex=0`.

If the tab _is_ currently selected, clicking does nothing, so `tabIndex=-1` since `onClick` is also undefined.

bvandrc: Nevermind, we can close this PR. After reading this https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/tab_role , it does currently work as intended.

I did not realize that tabs should be navigated via the arrow keys, not via the tab button

PR #6092, opened 2023-04-27
[popover2] fix(Popover2): don't react twice to simulated keyboard clicks

### LINKED ISSUE #5775: Select2 does not open properly with space or enter when focused
#### Environment

- __Package version(s)__: 4.12.0
- __Operating System__: OSX
- __Browser name and version__: Chrome 107.0.5304.121

#### Code Sandbox

Link to a minimal repro (fork [this code sandbox](https://codesandbox.io/s/blueprint-sandbox-et9xy)): https://blueprintjs.com/docs/#select/select2

#### Steps to reproduce

1. Get focus on the select in the docs page
2. Press space or enter

#### Actual behavior

Popover opens and closes instantly.

#### Expected behavior

Popover stays open.

#### Possible solution

Probably combination of a key listener and the DOMs general handling of space/enter on buttons, since `ArrowDown` seems to work fine.

justinbhopper:
#### Checklist

- [ ] Includes tests
- [ ] Update documentation

Adds check in `Popover2` to expect and ignore a simulated click that follows any keyboard click (enter/space) event.

The root of the bug is caused by the following in single keystroke (up and down) when pressing Enter on a button that is wrapped within a popover:

1. onKeyDown "Enter" occurs on AbstractButton, which bubbles to the Popover2. Popover2 [handles this and opens as a result](https://github.com/palantir/blueprint/blob/develop/packages/popover2/src/popover2.tsx#L356-L358).
2. onKeyUp "Enter" occurs on AbstractButton, which it is [programmed to dispatch a click event](https://github.com/palantir/blueprint/blob/develop/packages/core/src/components/button/abstractButton.tsx#L198).
3. The click event [reaches Popover2](https://github.com/palantir/blueprint/blob/develop/packages/popover2/src/popover2.tsx#L354), which causes it to immediately toggle back to close.

This PR adds code to track when enter/space happens prior to a click event, and then to ignore any subsequent simulated click event (detected via `isTrusted` flag).  If any real click or other keyboard events occur, the tracking property is reset.


justinbhopper:
@adidahiya An alternative fix is to remove the onKeyDown handler that was added to Popover2.  Popover (v1) does not support this, and it is ultimately the root cause of the issue.

justinbhopper:
Another alternative: Keep the onKeyDown handler, but detect if the target of the event has a `bp4-button` classname.  If it does, we do not open from it, because we expect that a simulated click event will come later.

adidahiya:
@justinbhopper thanks for the PR. I like the idea of checking `e.isTrusted`, this seems like a good use of that property. I would suggest going even further and _also_ checking that the keyDown event came from a known Blueprint component that triggers "fake" click events (in this case, `e.target.matches(Classes.BUTTON)`. This approach will change Popover2 behavior in the least number of cases, which minimizes the possibility of unintended consequences from this PR.

I think we should add unit tests for this change. I poked at the code a bit and tried to add a test but I wasn't able to come up with a failing test ([see my work in progress commit here](https://github.com/palantir/blueprint/commit/a40fff5bf145f3ab437224539cd7809ea4085949)). Can you try adding a unit test or two?

I'm not 100% sure but it seems like the behavior is different based on whether `autoFocus` is enabled (the [Popover2 example](https://blueprintjs.com/docs/#popover2-package/popover2) does not have the bug, while the [Select2 example](https://blueprintjs.com/docs/#select/select2) does), so we should test both cases when `autoFocus` is and isn't enabled.

justinbhopper:
@adidahiya You were right, `autoFocus=true` was the reason that the Popover2 example did not have the same issue.

This is because autoFocus steals focus as soon as `keydown` occurs.  This means the subsequent `keyup` does not happen on the `Button` anymore, but instead on some element within the overlay/popover.

I've added tests based on your WIP example.  I had to use a real `focus()` call instead of a simulated one, because it seems simulated ones don't actually update the `document.activeElement` in tests.  

I've also narrowed the simulated-click check to only be for bp4-button as you requested.

adidahiya (changes requested):
I'm still getting double reactions in the Select2 example in the latest [docs preview build](https://output.circle-artifacts.com/output/job/a2899fa9-19a1-4b09-a3e6-c880a6eadbe0/artifacts/0/packages/docs-app/dist/index.html#select/select2) (pressing both Enter and Space keys):

justinbhopper:
@adidahiya Turns out that was caused by #6149.  The popover for Select2 was being dismissed because the focus was immediately going into the Filterable input field, and the old logic was interpreting incorrectly as a click on a menu item.

While investigating this, I found out that Select2 completely overrides the onKeyDown that is normally handled by Popover2.  This means its not possible for the Popover2 to know whether the last keydown event was a keyboard click.  

Therefore, I coded it to just make the assumption that it should ignore simulated click events from Button if it's current state is already open.  If the user presses Enter/Space while the popover is open, the keydown event will already close it, so the simulated click doesn't serve any purpose in any realistic scenario.

adidahiya (changes requested):
code changes look fine, and I can now open the select popover with enter/space keys. but there is a regression where selecting an item in the list with the Enter key no longer closes the popover in the latest [docs preview build](https://output.circle-artifacts.com/output/job/3f7cc9e6-3bfa-42c2-b9f4-8fc38a453d68/artifacts/0/packages/docs-app/dist/index.html#select/select2):



here is the expected behavior on ~develop~ __edit: I was actually testing on the last docs release, core v4.18.0__ (requires clicking to open):

justinbhopper:
@adidahiya I just pulled latest from develop and it seems like this regression bug exists there too.  I am guessing #6149 had this unintended side effect and its not due to the changes in this PR.

The target of the event passed in `onItemSelect` is apparently the target button, and not the MenuItem element like the code seems to think.

adidahiya:
@justinbhopper ah, sorry about that, you're right. There is a regression on develop, I can see it on this commit preview build: https://github.com/palantir/blueprint/commit/4ee638c40b5613f139155214fd28875def271508#comments. I'll fix that after merging this PR.

adidahiya (approved):

adidahiya:
Note for posterity: there was apparently some code in QueryList which attempted to work around this problem already. But it seems like the `event.preventDefault()` was insufficient or buggy: https://github.com/palantir/blueprint/blob/db161b4aa566bd523edc83954d910f629a0d6439/packages/select/src/components/query-list/queryList.tsx#L549-L553

PR #6093, opened 2023-04-27
[core] feat(Overlay, Portal): allow portal to stop event propagation

### LINKED ISSUE #6124: Events propagate through Portals in Overlay elements
#### Environment

- __Package version(s)__: core v4.18
- __Operating System__: macOS

#### Code Sandbox

https://codesandbox.io/p/sandbox/awesome-mountain-njbt9k?file=%2FREADME.md

#### Steps to reproduce

- Render a `<Dialog>` on top of an interactive element.
- Interact with the background element through the Dialog.

#### Actual behavior

Events bubble through React Portals as per the [docs](https://legacy.reactjs.org/docs/portals.html#event-bubbling-through-portals)

See https://github.com/facebook/react/issues/11387

#### Expected behavior

There should be a way to capture all events in Portals, preventing them from propagating to the rest of the DOM.

#### Possible solution

1. add a boolean prop to `PortalProps` and `OverlayableProps` to capture all events at the Portal and prevent them from bubbling up the tree
2. add a prop which supports a _list_ of event types to capture at the Portal, like https://github.com/palantir/blueprint/pull/6093

dylanrcooke:
#### Checklist

- [ ] Includes tests
- [x] Update documentation

Currently react have a [longstanding issue](https://github.com/facebook/react/issues/11387) where portals propagate events up the react tree. This PR allows a list of events to be passed in to prevent propagation through the portal on.

Whether or not this should be configurable, or if we should always stopPropagation on events on the portal element.

#### Screenshot

No user facing changes.

### DIFF
--- a/packages/core/src/components/overlay/overlay.tsx
+++ b/packages/core/src/components/overlay/overlay.tsx
@@ -110,11 +110,13 @@ export interface IOverlayableProps extends IOverlayLifecycleProps {
     portalContainer?: HTMLElement;
 
     /**
-     * A list of DOM events to pass to the portal to stop propagation on.
+     * A list of DOM events which should be stopped from propagating through the Portal.
      * This prop is ignored if `usePortal` is `false`.
-     * Stopgap resolution for https://github.com/facebook/react/issues/11387
+     *
+     * @see https://legacy.reactjs.org/docs/portals.html#event-bubbling-through-portals
+     * @see https://github.com/palantir/blueprint/issues/6124
      */
-    stopPropagationEvents?: Array<keyof HTMLElementEventMap>;
+    portalStopPropagationEvents?: Array<keyof HTMLElementEventMap>;
 
     /**
      * A callback that is invoked when user interaction causes the overlay to close, such as
@@ -328,7 +330,7 @@ export class Overlay extends AbstractPureComponent2<OverlayProps, IOverlayState>
                 <Portal
                     className={this.props.portalClassName}
                     container={this.props.portalContainer}
-                    stopPropagationEvents={this.props.stopPropagationEvents}
+                    stopPropagationEvents={this.props.portalStopPropagationEvents}
                 >
                     {transitionGroup}
                 </Portal>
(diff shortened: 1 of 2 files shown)


packages/core/src/components/overlay/overlay.tsx
```diff
@@ -109,6 +109,15 @@ export interface IOverlayableProps extends IOverlayLifecycleProps {
...
     portalContainer?: HTMLElement;
 
+    /**
+     * A list of DOM events which should be stopped from propagating through the Portal.
+     * This prop is ignored if `usePortal` is `false`.
+     *
+     * @see https://legacy.reactjs.org/docs/portals.html#event-bubbling-through-portals
+     * @see https://github.com/palantir/blueprint/issues/6124
+     */
+    portalStopPropagationEvents?: Array<keyof HTMLElementEventMap>;
```
adidahiya, line 119: adjusted the name of this prop based on the other "portal" props on this interface. I admit it's getting a bit unwieldy and a `portalProps` prop might be better, but this is OK for now...

packages/core/src/components/portal/portal.tsx
```diff
@@ -128,6 +143,14 @@ export class Portal extends React.Component<PortalProps, IPortalState> {
         }
         return container;
     }
+
+    private addStopPropagationListeners(eventNames?: Array<keyof HTMLElementEventMap>) {
```
adidahiya, line 147: renamed these functions to be more specific

adidahiya:
> Whether or not this should be configurable, or if we should always stopPropagation on events on the portal element.

This should definitely be configurable; I don't want to enable new behavior by default that might break existing applications.

There's another question of whether the prop should just be a simple `boolean` which makes Portal capture _all_ events (rather than a specific denylist). Again, I'm opting to be a little conservative here and going with the way you've initially implemented it, which requires users to know the list of events which they might be having problems with on the underlying content. I think this will be the least surprising for users. For example, we don't want to introduce unnecessary issues with drag & drop listeners that might get blocked if we start capturing _all_ events.

adidahiya (approved):

PR #6146, opened 2023-05-11
[core] deprecate Toaster in favor of OverlayToaster

adidahiya:

Properly deprecate `Toaster` API in favor of `OverlayToaster`, for forwards-compatibility with v5.0. See https://github.com/palantir/blueprint/wiki/Toaster-5.0-changes

This is most commonly referenced in user code like this:

```diff
- const myToaster = Toaster.create();
+ const myToaster = OverlayToaster.create();
```

Also mark `IToasterProps` as deprecated, to be replaced by `OverlayToasterProps`. We can auto-fix this _type_ with the `no-deprecated-type-references` lint rule, but we don't have the no-deprecated-components rule set up to auto-fix _components_ right now.

### DIFF
--- a/packages/core/src/components/toast/toaster.tsx
+++ b/packages/core/src/components/toast/toaster.tsx
@@ -38,7 +38,7 @@ export type ToasterPosition =
 /** @deprecated use ToasterInstance */
 export type IToaster = ToasterInstance;
 
-/** Public API methods available on a `<Toaster>` component instance. */
+/** Public API methods available on a `<OverlayToaster>` component instance. */
 export interface ToasterInstance {
     /**
      * Shows a new toast to the user, or updates an existing toast corresponding to the provided key (optional).
@@ -58,10 +58,11 @@ export interface ToasterInstance {
 }
 
 /**
- * Props supported by the `<Toaster>` component.
- * These props can be passed as an argument to the static `Toaster.create(props?, container?)` method.
+ * Props supported by the `OverlayToaster` component.
+ *
+ * These props can be passed as an argument to the static `OverlayToaster.create(props?, container?)` method.
  */
-export interface IToasterProps extends Props {
+export interface OverlayToasterProps extends Props {
     /**
      * Whether a toast should acquire application focus when it first opens.
      * This is disabled by default so that toasts do not interrupt the user's flow.
@@ -85,7 +86,7 @@ export interface IToasterProps extends Props {
      * Whether the toaster should be rendered into a new element attached to `document.body`.
      * If `false`, then positioning will be relative to the parent element.
      *
-     * This prop is ignored by `Toaster.create()` as that method always appends a new element
+     * This prop is ignored by `OverlayToaster.create()` as that method always appends a new element
      * to the container.
      *
      * @default true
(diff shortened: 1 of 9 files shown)


adidahiya:
<h3>[core] deprecate Toaster in favor of OverlayToaster</h3>
Build artifact links for this commit: <strong><a href="https://output.circle-artifacts.com/output/job/6c7ade7e-66cb-4487-b27b-9ee0745c4639/artifacts/0/packages/docs-app/dist/index.html">documentation</a> | <a href="https://output.circle-artifacts.com/output/job/6c7ade7e-66cb-4487-b27b-9ee0745c4639/artifacts/0/packages/landing-app/dist/index.html">landing</a> | <a href="https://output.circle-artifacts.com/output/job/6c7ade7e-66cb-4487-b27b-9ee0745c4639/artifacts/0/packages/table-dev-app/dist/index.html">table</a> | <a href="https://output.circle-artifacts.com/output/job/6c7ade7e-66cb-4487-b27b-9ee0745c4639/artifacts/0/packages/demo-app/dist/index.html">demo</a></strong>

packages/core/src/components/toast/toaster.tsx
```diff
@@ -268,5 +272,12 @@ export class Toaster extends AbstractPureComponent2<IToasterProps, IToasterState
     };
 }
 
-export const OverlayToaster = Toaster;
-export type OverlayToasterProps = IToasterProps;
+/** @deprecated use the new, more specific component name `OverlayToaster` instead (forwards-compatible with v5) */
+export const Toaster = OverlayToaster;
+/** @deprecated use the new, more specific type `ToasterInstance` instead (forwards-compatible with v5) */
+// eslint-disable-next-line @typescript-eslint/no-redeclare
+export type Toaster = ToasterInstance;
```
dlech, line 279: FYI, this is a breaking change, e.g. https://github.com/pybricks/pybricks-code/actions/runs/4989185014/jobs/8932767904?pr=1741#step:5:6

adidahiya: ah, sorry about that, thanks for flagging. I suppose I could fix this by changing it to:

```suggestion
// eslint-disable-next-line @typescript-eslint/no-redeclare
export type Toaster = typeof OverlayToaster;
```

adidahiya: Opened a PR: https://github.com/palantir/blueprint/pull/6165

dlech: It should probably just be deleted until v5 to avoid complications.

adidahiya: I don't think removing this line will be backwards-compatible.

Before, `Toaster` was a class, so it could be referenced as a type or a value.

After this change, it's re-exported as a value with `const Toaster = OverlayToaster`. But that will break existing usage of `Toaster` as a type (you could work around this with `typeof Toaster`).

To restore `Toaster` as a _type_ in the public API, I think we need an `export type Toaster = ...` statement.

dlech: :+1:  makes sense.

PR #6231, opened 2023-06-21
fix(menuDivider): no tabIndex on role="separator"

bvandrc:
#### Checklist

- [N/A] Includes tests
- [N/A] Update documentation

We must remove `tabIndex={-1}`, is causing aXe a11y testing failure. Read discussion here:
https://github.com/dequelabs/axe-core/issues/4062

### DIFF
--- a/packages/core/src/components/menu/menuDivider.tsx
+++ b/packages/core/src/components/menu/menuDivider.tsx
@@ -45,10 +45,14 @@ export class MenuDivider extends React.Component<MenuDividerProps> {
 
     public render() {
         const { className, title, titleId } = this.props;
+        const hasTitle = title !== null;
 
         return (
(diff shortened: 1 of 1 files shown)


packages/core/src/components/menu/menuDivider.tsx
```diff
@@ -45,16 +45,11 @@ export class MenuDivider extends React.Component<MenuDividerProps> {
 
     public render() {
         const { className, title, titleId } = this.props;
-        if (title == null) {
-            // simple divider
```
adidahiya, line 49: I liked the previous code style, even though it was more verbose. I like that it explicitly lists out the two types of dividers. Can you refactor your PR to make the most minimal change necessary?

adidahiya (changes requested):

PR #6234, opened 2023-06-22
[v5] [core] feat: remove @juggle/resize-observer in favor of native API

### LINKED ISSUE #6233: Remove ResizeObserver polyfill for Blueprint v5
Various polyfills have been removed for Blueprint v5 as per https://github.com/palantir/blueprint/issues/5143. MDN reports [that all modern browsers have support for `ResizeObserver`](https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver#browser_compatibility), so I believe it should be fine to remove the `@juggler/resize-observer` ponyfill which BP core currently depends on:

https://github.com/palantir/blueprint/blob/974cbe1aa8ada21e178fa16703d2b189501be8d5/packages/core/package.json#L48

This should reduce bundle size a bit and likely provide a faster `ResizeObserver` implementation for users as I imagine the native implementation is more efficient than the ponyfill. (In addition, I may have traced a memory leak in our application to the use of `@juggle/resize-observer`. I'm still verifying that aspect of things though and it should make sense to stop using it in BP regardless.)

If for some reason removing the ponyfill isn't possible, it'd be great to make BP use the native `ResizeObserver` class instead of the ponyfill when available. This change could be applied to current BP v4 since it shouldn't break IE 11 compatibility.

jkillian:
#### Checklist

- [x] Includes tests (just the existing ones)
- [x] Update documentation

Removes the use of the `@juggle/resize-observer` polyfill in favor of directly using the native API.

Is the code to support isomoprhic rendering okay?

#### Screenshot

N/A


packages/core/src/components/resize-sensor/resizeSensor.tsx
```diff
@@ -72,7 +71,8 @@ export class ResizeSensor extends AbstractPureComponent<ResizeSensorProps> {
 
     private prevElement: HTMLElement | undefined = undefined;
 
-    private observer = new ResizeObserver(entries => this.props.onResize?.(entries));
+    private observer =
+        globalThis.ResizeObserver != null ? new ResizeObserver(entries => this.props.onResize?.(entries)) : undefined;
```
jkillian, line 78: Unsure if this is the best way to make things work correctly in an isomorphic setting, but it seemed reasonable to me and got the tests to pass locally

adidahiya:
Nice, this sounds like a good idea. Just in time for v5.0, I'm about to cut the stable release this week.

I'm working on the `next` branch -- I'll try to update your PR to target that one.

packages/core/src/components/resize-sensor/resizeSensor.tsx
adidahiya: seems fine

adidahiya:
Additional note: it sounds like the caveats listed here in ["Switching between native and polyfilled versions"](https://github.com/juggle/resize-observer#switching-between-native-and-polyfilled-versions) don't apply to us since we don't actually use any of the `ResizeEntry` data in Blueprint components.

adidahiya (approved):

jkillian:
Oops, sorry for targeting the wrong branch, thanks for taking care of that and merging!

PR #6245, opened 2023-06-26
[core] feat: Section component

CPerinet:
Screenshot 2023-06-27 at 10 42 13 PM

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1204971662887378

### DIFF
--- a/packages/core/src/common/classes.ts
+++ b/packages/core/src/common/classes.ts
@@ -217,6 +217,7 @@ export const MULTISTEP_DIALOG_NAV_TOP = `${MULTISTEP_DIALOG}-nav-top`;
 export const MULTISTEP_DIALOG_NAV_RIGHT = `${MULTISTEP_DIALOG}-nav-right`;
 
 export const SECTION = `${NS}-section`;
+export const SECTION_COLLAPSED = `${SECTION}-collapsed`;
 export const SECTION_HEADER = `${SECTION}-header`;
 export const SECTION_HEADER_LEFT = `${SECTION_HEADER}-left`;
 export const SECTION_HEADER_TITLE = `${SECTION_HEADER}-title`;
@@ -224,7 +225,7 @@ export const SECTION_HEADER_SUB_TITLE = `${SECTION_HEADER}-sub-title`;
 export const SECTION_HEADER_DIVIDER = `${SECTION_HEADER}-divider`;
 export const SECTION_HEADER_TABS = `${SECTION_HEADER}-tabs`;
 export const SECTION_HEADER_RIGHT = `${SECTION_HEADER}-right`;
-export const SECTION_CONTENT = `${SECTION}-content`;
+export const SECTION_PANEL = `${SECTION}-panel`;
 
 export const NAVBAR = `${NS}-navbar`;
 export const NAVBAR_GROUP = `${NAVBAR}-group`;
--- a/packages/core/src/common/classes.ts
+++ b/packages/core/src/common/classes.ts
@@ -221,6 +221,7 @@ export const SECTION_HEADER = `${SECTION}-header`;
 export const SECTION_HEADER_LEFT = `${SECTION_HEADER}-left`;
 export const SECTION_HEADER_TITLE = `${SECTION_HEADER}-title`;
 export const SECTION_HEADER_SUB_TITLE = `${SECTION_HEADER}-sub-title`;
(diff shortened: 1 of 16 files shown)


adidahiya:
@CPerinet I cleaned up some of the code based on my review.
- Instead of "small" appearance, I've switched to using the newer "compact" modifier (this exists for HTMLTable and Tooltip components as well, and is better suited to reduced padding visuals).
- I understand that `<Card>` is being used inside `<Section>`, but I feel like it doesn't really make sense to use `CardProps` here since we're not allowing any of the props which are unique/specific to that component. Better to define `SectionProps` by inheriting from the same underlying interfaces which `CardProps` does.
- We might want to consider further breaking down the component to also define a new `<SectionHeader>` component. This might help to disambiguate some of the prop names (for example, `rightElement`)
- There are still some edge cases for some combinations of props that need to be ironed out:
  - tabs defined & `compact={true}` 
image
  - tabs defined & `collapsible={true}`

adidahiya:
Re: edge case where tabs & collapsible are both enabled. We can work around this by making these props mutually exclusive. `tabDefinitions={...} collapsible={true}` would be unsupported.

Also, I think the collapse feature should actually use the `<Collapse>` component so that we get the nice transition animation.

adidahiya:
Note that I updated the docs page a bit and fixed the console error related to duplicate docs headings (you can't have two instances of the same heading in a `.md` page, like `## Props`).

[Latest docs preview build here](https://output.circle-artifacts.com/output/job/49d93774-0b19-440a-8ff8-7cdc8eb5d424/artifacts/0/packages/docs-app/dist/index.html#core/components/section)

image

adidahiya:
I've removed Tabs support for now. It has some complications which we need to address, and we don't need to implement all the features in one PR. We can bring back that feature in a separate PR.

adidahiya:
> * We might want to consider further breaking down the component to also define a new `<SectionHeader>` component. This might help to disambiguate some of the prop names (for example, `rightElement`)

I tried this out and it turns out to be kind of cumbersome to implement. The SectionHeader would need to know a lot of the state that Section has access to, specifically `collapsible` and `isCollapsed`. So I think the current approach is fine, at least for now.

packages/core/src/common/classes.ts
```diff
@@ -213,6 +214,17 @@ export const MULTISTEP_DIALOG_RIGHT_PANEL = `${MULTISTEP_DIALOG}-right-panel`;
 export const MULTISTEP_DIALOG_NAV_TOP = `${MULTISTEP_DIALOG}-nav-top`;
 export const MULTISTEP_DIALOG_NAV_RIGHT = `${MULTISTEP_DIALOG}-nav-right`;
 
+export const SECTION = `${NS}-section`;
+export const SECTION_COLLAPSED = `${SECTION}-collapsed`;
```
adidahiya, line 218: I prefer to add a "section"-namespaced "collapsed" class name for now rather than a `.bp5-collapsed` name.

packages/core/src/components/section/sectionPanel.tsx
```diff
@@ -0,0 +1,49 @@
...
+     *
+     * @default true
+     */
+    padded?: boolean;
+}
+
+/**
+ * Section panel component.
+ *
+ * @see https://blueprintjs.com/docs/#core/components/section-panel
```
adidahiya: this link needs to be fixed

packages/docs-app/src/examples/core-examples/sectionExample.tsx
```diff
@@ -0,0 +1,147 @@
+/*
+ * Copyright 2017 Palantir Technologies, Inc. All rights reserved.
```
adidahiya: copyright year needs to be updated

packages/docs-app/src/styles/_examples.scss
```diff
@@ -533,6 +533,23 @@
   }
 }
 
+#{example("Section")} {
+  .docs-example {
+    align-items: start;
+  }
+
+  .metadata-panel {
```
adidahiya, line 541: @CPerinet for future reference: styling of example content should be done in CSS like this

adidahiya (changes requested):

adidahiya (approved):

PR #6246, opened 2023-06-26
[core] feat: new CardList component

CPerinet:
### Demo

---

### Archive

Based on https://github.com/palantir/blueprint/pull/6245 (edit: that PR has merged ✅)

Original screenshot:

Screenshot 2023-06-26 at 4 21 09 PM

PR #6250, opened 2023-06-27
[core] feat: SwitchCard component

CPerinet:
~Based on https://github.com/palantir/blueprint/pull/6246~ edit: that PR has merged

Screenshot 2023-06-27 at 12 57 28 PM

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1204971665339519


adidahiya:
I fixed the double click issue by using the switch `onChange` handler instead of a new `onClick` on the containing Card. This required moving the card padding down to the label element.

adidahiya:
I fixed the issue where the right side of the card was non-interactive:

adidahiya:
[latest docs preview build](https://output.circle-artifacts.com/output/job/f5d34f6a-eae5-42ec-be40-3904e5f7cce4/artifacts/0/packages/docs-app/dist/index.html#core/components/card.switch-card)

CPerinet:
- The padding on card list cards does not get overriden correctly.
- Do we want to also add radio and checkbox cards?

adidahiya:
> The padding on card list cards does not get overriden correctly.

Ah, I see the issue. I'll push a fix:




> Do we want to also add radio and checkbox cards?

Yes, in a separate PR

packages/docs-app/src/examples/core-examples/switchCardExample.tsx
```diff
@@ -0,0 +1,94 @@
...
+ * limitations under the License.
+ */
+
+import * as React from "react";
+
+import { CardList, H5, Switch, SwitchCard } from "@blueprintjs/core";
+import { Example, ExampleProps, handleBooleanChange } from "@blueprintjs/docs-theme";
+
+interface SwitchCardExampleState {
+    // TODO: add compact option
```
adidahiya, line 23: @CPerinet we can add this option to this example once we implement `<Card compact={true} />` in a separate PR. The padding style variable has already been added in `_card-variables.scss` :+1:

adidahiya:
I added an example showing composition with CardList:

image

adidahiya (approved):

PR #6260, opened 2023-07-03
[core] feat(Portal): Add `portalContainer` context option

### LINKED ISSUE #6210: Adding `portalContainer` to `PortalContext`/`PortalProvider`
#### Environment

- __Package version(s)__: 

#### Feature request

The ask is to add a React-context solution which forces all (blueprint) children that who render below a given `PortalProvider` (like `<Dialog/>`, `<Popover2/>`) to have a common `React.createPortal` container target. 

This allows for the app to have more fine-grained control over the bounds of portal background overlays & visibility when components may stay mounted in the background. 

Currently the `PortalContextOptions` interface has the following field:
```ts
interface PortalContextOptions {
     portalClassName: string | undefined
}
```

We propose expanding it with

```ts
interface PortalContextOptions {
     portalClassName: string | undefined,
     portalContainer?: HTMLElement
}
```

And subsequently drilling the result of `useContext(PortalContext).portalContainer` into `portalContainer` props of `Portal` based blueprint elements. 

But still allowing explicit overrides from individual usages , like `portalContainer={props.portalContainer ?? portalContext.portalConainter ?? document.body}`.
#### Examples

```tsx

    const ref = useRef<HTMLDivElement>(null);

    <PortalProvider portalContainer={ref}>
        <Dialog isOpen={true}>
            <span>This dialog will have a custom portal container </span>
        </Dialog>
        <div ref={ref}></div>
    </PortalProvider>,
```

braeden:
#### Checklist

- [ ] Includes tests
- [ ] Update documentation

This changes forces all (blueprint) children that who render below a given `PortalProvider` (like `<Dialog/>`, `<Popover2/>`) to have a common element where `React.createPortal` usages become direct children of. 

To be more precise here, `portalContainer` acts ONLY as a parent element of a new `div` and that new `div` becomes the target container in the `React.createPortal` syntax -- so `children` are rendered into that element. 

We also memoize the `PortalProvider` value to avoid thrashing on every re-render.

- I changed the way that `addStopPropagationListeners`,  `removeStopPropagationListeners` (and `maybeAddClass`) work -- using `React.useEffect`'s cleanup function to avoid leaving, what I believe would have been, dangling listeners when the portal unmounted. 
- I removed the `hasMounted` state, since it was purely duplicative with `portalElement != null`
- I removed `Portal.defaultProps` and replaces it with an internal `?? document?.body` fallback.

### DIFF
--- a/packages/core/src/components/portal/portal.tsx
+++ b/packages/core/src/components/portal/portal.tsx
@@ -80,8 +80,8 @@ export function Portal(props: PortalProps, legacyContext: PortalLegacyContext = 
 
     const container =  props.container ?? context.portalContainer ?? document?.body
 
-    const [hasMounted, setHasMounted] = React.useState(false);
     const [portalElement, setPortalElement] = React.useState<HTMLElement>();
+    const hasMounted = portalElement != null
 
     const createContainerElement = React.useCallback(() => {
         const newContainer = document.createElement("div");
@@ -108,12 +108,10 @@ export function Portal(props: PortalProps, legacyContext: PortalLegacyContext = 
         const newPortalElement = createContainerElement();
         container.appendChild(newPortalElement);
         setPortalElement(newPortalElement);
-        setHasMounted(true);
 
         return () => {
             removeStopPropagationListeners(newPortalElement, props.stopPropagationEvents);
             newPortalElement.remove();
-            setHasMounted(false);
             setPortalElement(undefined);
         };
     }, [container, createContainerElement]);
--- a/packages/core/src/components/portal/portal.tsx
+++ b/packages/core/src/components/portal/portal.tsx
@@ -35,7 +35,7 @@ export interface PortalProps extends Props {
     /**
      * The HTML element that children will be mounted to.
      *
-     * @default document.body
+     * @default PortalContext.portalContainer ?? document.body
      */
     container?: HTMLElement;
 
@@ -78,7 +78,7 @@ const PORTAL_LEGACY_CONTEXT_TYPES: ValidationMap<PortalLegacyContext> = {
 export function Portal(props: PortalProps, legacyContext: PortalLegacyContext = {}) {
     const context = React.useContext(PortalContext);
 
-    const container =  props.container ?? context.portalContainer ?? document?.body
+    const portalContainer =  props.container ?? context.portalContainer ?? document?.body
 
     const [portalElement, setPortalElement] = React.useState<HTMLElement>();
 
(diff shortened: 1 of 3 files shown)


packages/core/src/components/portal/portal.tsx
```diff
@@ -75,84 +74,82 @@ const PORTAL_LEGACY_CONTEXT_TYPES: ValidationMap<PortalLegacyContext> = {
...
         if (blueprintPortalClassName != null && blueprintPortalClassName !== "") {
             console.error(Errors.PORTAL_LEGACY_CONTEXT_API);
-            maybeAddClass(container.classList, blueprintPortalClassName); // added via legacy context
+            maybeAddClass(newPortalElement.classList, blueprintPortalClassName); // added via legacy context
         }
 
-        return container;
-    }, [props.className, context.portalClassName]);
+        return newPortalElement;
+    }, [className, context.portalClassName, legacyContext.blueprintPortalClassName, stopPropagationEvents]);
```
adidahiya, line 102: nice, thanks for fixing this dependency list

packages/core/src/components/portal/portal.tsx
```diff
@@ -75,84 +74,82 @@ const PORTAL_LEGACY_CONTEXT_TYPES: ValidationMap<PortalLegacyContext> = {
...
  */
-export function Portal(props: PortalProps, legacyContext: PortalLegacyContext = {}) {
+export function Portal(
+    { className, stopPropagationEvents, container, onChildrenMount, children }: PortalProps,
+    legacyContext: PortalLegacyContext = {},
+) {
     const context = React.useContext(PortalContext);
 
-    const [hasMounted, setHasMounted] = React.useState(false);
+    const portalContainer = container ?? context.portalContainer ?? document?.body;
```
adidahiya, line 83: can you please update the Portal docs to reflect this behavior? Specifically, these lines should change: https://github.com/palantir/blueprint/blob/8623da5404bc1be186f696aca0a92a41bd1db5d9/packages/core/src/components/portal/portal.md#L10-L11

- Let's remove outdated references to JQuery
- Let's mention that the new portal element is added as a child of either (in order of preference): a `container` element specified as a `<Portal>` prop, a `portalContainer` element specified via `<PortalProvider>` context, or `document.body`

Also, I noticed a broken link on the Portal docs page. This line should be updated: https://github.com/palantir/blueprint/blob/8623da5404bc1be186f696aca0a92a41bd1db5d9/packages/core/src/components/portal/portal.md#L34

The correct link is `#core/context/portal-provider`

adidahiya (approved):
looks good! just need a few documentation fixes

adidahiya:
Another docs suggestion: let's update the link to React context docs. This line: https://github.com/palantir/blueprint/blob/8623da5404bc1be186f696aca0a92a41bd1db5d9/packages/core/src/components/portal/portal.md?plain=1#L31

Should instead point to: https://react.dev/learn/passing-data-deeply-with-context

adidahiya (approved):
thanks @braeden  👍🏽

PR #6268, opened 2023-07-10
[core] fix(Icon): avoid breaking rules against conditional React hooks

### LINKED ISSUE #6267: [Icon] Internal React error: Expected static flag was missing
#### Environment

- __Package version(s)__: 5.0.0
- __Operating System__: All
- __Browser name and version__: All

#### Code Sandbox

N/A

#### Steps to reproduce

Render an Icon component

#### Actual behavior

2023-07-10-211204_1025x314_scrot

#### Expected behavior

That error / warning should not be appearing, it doesn't break anything, but having the error during development is quite annoying

#### Possible solution

Looks like the [icon](https://github.com/palantir/blueprint/blob/develop/packages/core/src/components/icon/icon.tsx) component has a conditional return before usage of hooks. And it is not recommended, as per hook rules, or this comment https://github.com/facebook/react/issues/24391#issuecomment-1106410214

I think the part after the return can be extracted to a private component, something like:

```
export const Icon: React.FC<IconProps> = React.forwardRef<any, IconProps>((props, ref) => {
    const { icon } = props;
    if (icon == null || typeof icon === "boolean") {
        return null;
    } else if (typeof icon !== "string") {
        return icon;
    }
    
    return <IconInner {...props} />
}

// and then the rest of the code can go into the `<IconInner>`

```

fellmann:
#### Checklist

- [ ] ~~Includes tests~~
- [ ] ~~Update documentation~~

Fix conditional hook in Icon component. As the useEffect hook checks for `typeof icon === "string", it won't run anyway in the other cases.

It would be great to use a linter rule to check for conditional hooks, but I did not succeed to configure it. Maybe other components are affected as well.

### DIFF
--- a/packages/core/src/components/icon/icon.tsx
+++ b/packages/core/src/components/icon/icon.tsx
@@ -70,11 +70,6 @@ export interface IconProps extends IntentProps, Props, SVGIconProps, IconHTMLAtt
  */
 export const Icon: React.FC<IconProps> = React.forwardRef<any, IconProps>((props, ref) => {
     const { icon } = props;
-    if (icon == null || typeof icon === "boolean") {
-        return null;
-    } else if (typeof icon !== "string") {
-        return icon;
-    }
 
     const {
         autoLoad,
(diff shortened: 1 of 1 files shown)


palantirtech:
Thanks for your interest in [palantir/blueprint](https://github.com/palantir/blueprint), @fellmann! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

adidahiya (approved):
thanks @fellmann!

PR #6272, opened 2023-07-12
[core] fix(SectionCard): rename component, fix Sass variable names

adidahiya:
#### Checklist

- [ ] Includes tests
- [x] Update documentation

- Rename __SectionPanel__ to __SectionCard__
  - ⚠️ this is _not_ a breaking change since we haven't released #6245 yet
- Rename Sass variables in `_section.scss` to match naming used in other parts of the Sass codebase (`_tooltip.scss`, `button/_common.scss`)
- feat(`Section`): add support for `elevation` prop which accepts a subset of the possible `Elevation` values (ZERO and ONE)

Naming for the __SectionCard__ component. It's a bit tricky to name container elements like this, but I feel like this is probably the best option for now. It's not:
- a _panel_, since those can be more complex components with different kinds of content inside (we expect Section contents to be a little simpler). also, this creates some ambiguity with PanelStack.
- "content" (as was the initial name in #6245), since "content" is provided by Blueprint users, not Blueprint itself. the component is a container for content

references:
- https://ux.stackexchange.com/questions/78798/whats-the-difference-between-cards-panels-and-tiles

#### Screenshot

image

### DIFF
--- a/packages/core/src/components/section/sectionPanel.tsx
+++ b/packages/core/src/components/section/sectionPanel.tsx
@@ -20,7 +20,7 @@ import * as React from "react";
 import { Classes } from "../../common";
 import { DISPLAYNAME_PREFIX, HTMLDivProps, Props } from "../../common/props";
 
-export interface SectionPanelProps extends Props, HTMLDivProps, React.RefAttributes<HTMLDivElement> {
+export interface SectionCardProps extends Props, HTMLDivProps, React.RefAttributes<HTMLDivElement> {
     /**
      * Whether to apply visual padding inside the content container element.
      *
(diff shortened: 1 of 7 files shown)


packages/core/src/components/section/sectionPanel.tsx
```diff
@@ -30,20 +30,20 @@ export interface SectionPanelProps extends Props, HTMLDivProps, React.RefAttribu
...
 
 /**
- * Section panel component.
+ * Section card component.
  *
- * @see https://blueprintjs.com/docs/#core/components/section.section-panel
+ * @see https://blueprintjs.com/docs/#core/components/section.section-card
  */
-export const SectionPanel: React.FC<SectionPanelProps> = React.forwardRef((props, ref) => {
+export const SectionCard: React.FC<SectionCardProps> = React.forwardRef((props, ref) => {
```
qcz, line 37: The file itself was not renamed.

adidahiya: good catch! thankfully that does not affect the public API, so I can fix that without a breaking change.

PR #6278, opened 2023-07-13
[Draft] Entity title

CPerinet:


packages/core/src/common/classes.ts
```diff
@@ -153,6 +153,15 @@ export const EDITABLE_TEXT_EDITING = `${EDITABLE_TEXT}-editing`;
 export const EDITABLE_TEXT_INPUT = `${EDITABLE_TEXT}-input`;
 export const EDITABLE_TEXT_PLACEHOLDER = `${EDITABLE_TEXT}-placeholder`;
 
+export const ENTITY_TITLE = `${NS}-entity-title`;
+export const ENTITY_TITLE_ICON_CONTAINER = `${ENTITY_TITLE}-icon-container`;
+export const ENTITY_TITLE_RIGHT = `${ENTITY_TITLE}-right`;
+export const ENTITY_TITLE_TOP_LINE = `${ENTITY_TITLE}-top-line`;
+export const ENTITY_TITLE_TAGS_CONTAINER = `${ENTITY_TITLE}-tags-container`;
+export const ENTITY_TITLE_TITLE = `${ENTITY_TITLE}-title`;
+export const ENTITY_TITLE_HAS_SUBTITLE = `${ENTITY_TITLE}-has-subtitle`;
```
adidahiya, line 162: it looks like this class is unused

packages/core/src/components/entity-title/entityTitle.tsx
```diff
@@ -0,0 +1,148 @@
...
+            {icon != null && (
+                <div
+                    className={classNames(Classes.ENTITY_TITLE_ICON_CONTAINER, {
+                        [Classes.ENTITY_TITLE_HAS_SUBTITLE]: subtitleComponent != null,
+                    })}
+                >
+                    <Icon icon={icon} aria-hidden={true} tabIndex={-1} className={Classes.TEXT_MUTED} />
+                </div>
+            )}
+            <div className={Classes.ENTITY_TITLE_RIGHT}>
```
adidahiya, line 116: "right" doesn't seem like the correct terminology here, it's more of the main content. I would call it "ENTITY_TITLE_TEXT" instead

packages/core/src/components/entity-title/entityTitle.tsx
```diff
@@ -0,0 +1,148 @@
...
+export const EntityTitle: React.FC<IEntityTitleProps> = ({
+    className,
+    headingSize,
+    title,
+    icon,
+    tags,
+    subtitle,
+    titleURL,
+}) => {
+    const titleComponent = React.useMemo(() => {
```
adidahiya, line 57: this is of type `JSX.Element`, so it should be called:

```suggestion
    const titleElement = React.useMemo(() => {
```

packages/core/src/components/entity-title/entityTitle.tsx
```diff
@@ -0,0 +1,148 @@
...
+                <a target="_blank" href={titleURL} rel="noreferrer">
+                    {title}
+                </a>
+            ) : (
+                title
+            );
+
+        const titleClassName = classNames(Classes.ENTITY_TITLE_TITLE);
+
+        switch (headingSize) {
```
adidahiya, line 69: instead of this switch case syntax and exposing a new `HeadingSize` enum in the public API, I think it might be nicer to have a prop that accepts a function component:

```tsx
<EntityTitle title="Foo" heading={H4} />

interface EntityTitleProps {
    /**
     * React component to render the main title heading. This defaults to Blueprint's `<Text>` component,
     * which inherits font size from its containing element(s).
     *
     * To render larger, more prominent titles, Use Blueprint's heading components instead
     * (e.g. `{ H1 } from "@blueprintjs/core"`).
     *
     * @default Text
     */
    heading?: React.FC<any>;
}
```

then you can simply render:

```ts
React.createElement(heading, { className: Classes.ENTITY_TITLE_TITLE }, maybeTitleWithURL);
```

packages/core/src/components/entity-title/entityTitle.tsx
```diff
@@ -0,0 +1,148 @@
...
+            case HeadingSize.h5:
+                return <H5 className={titleClassName}>{maybeTitleWithURL}</H5>;
+            case HeadingSize.h6:
+                return <H6 className={titleClassName}>{maybeTitleWithURL}</H6>;
+            default:
+                return <Text className={titleClassName}>{maybeTitleWithURL}</Text>;
+        }
+    }, [title, titleURL, headingSize]);
+
+    const subtitleComponent = React.useMemo(() => {
```
adidahiya, line 87: again, this is an _element_ rather than a component value. actually you can drop the suffix altogether. it's more important to use the "maybe" prefix in this case:

```suggestion
    const maybeSubtitle = React.useMemo(() => {
```

packages/core/src/components/entity-title/entityTitle.tsx
```diff
@@ -0,0 +1,148 @@
...
+        if (!subtitle) {
```
adidahiya, line 88: don't use implicit boolean conversion

```suggestion
        if (subtitle == null || subtitle === "") {
```

packages/core/src/components/entity-title/entityTitle.tsx
```diff
@@ -0,0 +1,148 @@
...
+            return null;
+        }
+
+        return (
+            <Text
+                className={classNames(
+                    Classes.TEXT_MUTED,
+                    headingSize == null && Classes.TEXT_SMALL,
+                    Classes.ENTITY_TITLE_SUBTITLE,
```
adidahiya, line 97: use `classNames` object syntax for this logic instead of `&&`. also put the conditionally-applied classes last in the liast:

```suggestion
                    Classes.ENTITY_TITLE_SUBTITLE,
                    Classes.TEXT_MUTED,
                    { [Classes.TEXT_SMALL]: headingSize == null },
```

packages/core/src/components/entity-title/entityTitle.tsx
```diff
@@ -0,0 +1,148 @@
...
+                )}
+            >
+                {subtitle}
+            </Text>
+        );
+    }, [headingSize, subtitle]);
+
+    return (
+        <div className={classNames(className, Classes.ENTITY_TITLE, headingSize != null && css[headingSize])}>
```
adidahiya, line 106: we'll need to refactor this to work with my `heading?: React.FC` proposal above...

packages/core/src/components/entity-title/entityTitle.tsx
```diff
@@ -0,0 +1,148 @@
...
+                <div
+                    className={classNames(Classes.ENTITY_TITLE_ICON_CONTAINER, {
+                        [Classes.ENTITY_TITLE_HAS_SUBTITLE]: subtitleComponent != null,
+                    })}
+                >
+                    <Icon icon={icon} aria-hidden={true} tabIndex={-1} className={Classes.TEXT_MUTED} />
+                </div>
+            )}
+            <div className={Classes.ENTITY_TITLE_RIGHT}>
+                <div className={Classes.ENTITY_TITLE_TOP_LINE}>
```
adidahiya, line 117: "top line" doesn't sound great, how about "ENTITY_TITLE_TITLE_AND_TAGS"

packages/core/src/components/entity-title/entityTitle.tsx
```diff
@@ -0,0 +1,148 @@
...
+                {subtitleComponent}
+            </div>
+        </div>
+    );
+};
+
+EntityTitle.defaultProps = {};
+EntityTitle.displayName = `${DISPLAYNAME_PREFIX}.EntityTitle`;
+
+type ISkeletonEntityTitleProps = Props &
```
adidahiya, line 130: we don't use `I` prefixes in types or interfaces names anymore. also we typically use `interface` instead of `type` to define props interfaces:


```suggestion
interface SkeletonEntityTitleProps extends Props {
```

packages/core/src/components/entity-title/entityTitle.tsx
```diff
@@ -0,0 +1,148 @@
...
+    Pick<IEntityTitleProps, "headingSize"> & {
+        hasIcon?: boolean;
+        hasSubtitle?: boolean;
+    };
+
+export const SkeletonEntityTitle: React.FC<ISkeletonEntityTitleProps> = ({ hasSubtitle, hasIcon, ...props }) => {
```
adidahiya, line 136: __EntityTitleSkeleton__ would be a better name so that it gets grouped with EntityTitle in the imports list when we alphabetize imports.

Also, let's move this to its own file (typically keep one component per file)

adidahiya (changes requested):

adidahiya:
superseded by #6590

PR #6282, opened 2023-07-17
[core] fix(Section): apply collapsed style when empty children

CPerinet:
#### Checklist
 - [ ] Includes tests
 - [ ] Update documentation
 
Apply `Classes.SECTION_COLLAPSED` on `<Section />` when children is empty. This will remove the border bottom of the section header which is unnecessary in this case.

Section header do not have a double border anymore when children is empty. Should not have other effects.

#### Screenshot

|Before|After|
|--|--|
|Screenshot 2023-07-17 at 16 18 50|Screenshot 2023-07-17 at 16 18 56|

### DIFF
--- a/packages/core/src/components/section/section.tsx
+++ b/packages/core/src/components/section/section.tsx
@@ -104,7 +104,8 @@ export const Section: React.FC<SectionProps> = React.forwardRef((props, ref) => 
             elevation={Elevation.ZERO}
             className={classNames(className, Classes.SECTION, {
                 [Classes.COMPACT]: compact,
-                [Classes.SECTION_COLLAPSED]: collapsible && isCollapsed,
+                [Classes.SECTION_COLLAPSED]:
+                    (collapsible && isCollapsed) || React.Children.toArray(children).length === 0,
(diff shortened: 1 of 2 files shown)


packages/core/src/components/section/section.tsx
```diff
@@ -104,7 +104,8 @@ export const Section: React.FC<SectionProps> = React.forwardRef((props, ref) =>
             elevation={Elevation.ZERO}
             className={classNames(className, Classes.SECTION, {
                 [Classes.COMPACT]: compact,
-                [Classes.SECTION_COLLAPSED]: collapsible && isCollapsed,
+                [Classes.SECTION_COLLAPSED]:
+                    (collapsible && isCollapsed) || React.Children.toArray(children).length === 0,
```
adidahiya: we have a utility function for this. I'll push a commit. I realized it should be used in the ~Collapse~ Callout implementation as well.

adidahiya (changes requested):

adidahiya (approved):

Issue #3943, opened 2020-01-27
Expose onChange function of NumericInput


strusoftsawen:
`<NumericInput>` does not expose `onChange(e:ChangeEvent<HTMLInputElement>)` the same way as `<InputGroup>`. Instead, `onValueChange(n:number, s:string)` is exposed, with no information about which input is changing. This means that, annoyingly, one change handler is needed for each numeric input. Should be trivial to expose the change event of the `<input>` within.

adidahiya:
> with no information about which input is changing

I'm a little confused by this; there's only one input in a `NumericInput` component.

Can you demonstrate your issue with a code sandbox? I don't see why you need the change event of the underlying `<input>`. Note that you can attach your own DOM event handler using the `inputRef` prop, but that's a little different from a React event handler like you suggested.

strusoftsawen:
Thanks for the response. Here is a sandbox illustrating what I wish to achieve: https://codesandbox.io/s/react-typescript-xuly5

As you can see, one handleChange() method is used to deal with the onChange events of the various InputGroups. However, if I wish to use a NumericInput, there is no onChange event so I have to define one event handler for each of them. Even if there may be a way to work around this as you say, it would be neat if I could use the same logic as the InputGroups.

Let me know if I have misunderstood anything about the use of NumericInput.

strusoftsawen:
@adidahiya Any update on this? My groups is looking to generate InputGroups and NumericInputs extensively based on data and share the same onChange event handler, but this is currently not possible.

adidahiya:
@strusoftsawen I pushed a PR which should accomplish what you want, please check it out: https://github.com/palantir/blueprint/pull/3982

strusoftsawen:
@adidahiya awesome, looks exactly like what I need. Apparently it wasn't included in the 3.24.0 package release, but I'm looking forward to using it once it's released.

rkulinski:
@adidahiya I think this issue should be reopened. To be frank it seems insane that two very similar components `InputGroup` and `NumericInput` (naming is also wierd) have very different API. I think proper `onChange` method is required here.

adidahiya:
@rkulinski I don't think you'll get very far in open source conversations with that kind of language in your comment. Most of the API decisions in Blueprint come with tradeoffs and the names have been chosen for a reason. If you have a concrete criticism of or proposal for what you'd like to see in the NumericInput component, please file a new issue and use the issue template.

chetanism:
[...]

PR #6290, opened 2023-07-19
[core] feat(InputGroup): new prop `onValueChange`

thasner:
#### Checklist

- [x] Includes tests
- [x] Update documentation

Add a new `onValueChange` callback prop to `<InputGroup>` which behaves similarly to the same prop for `<NumericInput>`.

Motivation: this helps prevent common potential footguns with React 16 Event Pooling where developers sometimes forget to save `event.target.value` as a variable or use `event.persist()`, which can lead to null-pointer exceptions at runtime.

Various uses of `<InputGroup onChange>` within the Blueprint monorepo have also been updated to use the new prop to demonstrate its efficacy.

docs & tests for the new prop

#### Screenshot

N/A

### DIFF
--- a/packages/core/src/components/forms/inputGroup.tsx
+++ b/packages/core/src/components/forms/inputGroup.tsx
@@ -40,7 +40,7 @@ export interface InputGroupProps
     /** Whether this input should use large styles. */
     large?: boolean;
 
-    /** The callback invoked when the value changes.
+    /** The callback invoked when the value changes. */
     onValueChange?(valueAsString: string/*, inputElement: HTMLInputElement | null*/): void;
 
     /** Whether this input should use small styles. */
--- a/packages/core/src/components/forms/inputGroup.tsx
+++ b/packages/core/src/components/forms/inputGroup.tsx
@@ -40,8 +40,15 @@ export interface InputGroupProps
     /** Whether this input should use large styles. */
     large?: boolean;
 
-    /** The callback invoked when the value changes. */
-    onValueChange?(valueAsString: string/*, inputElement: HTMLInputElement | null*/): void;
+    /**
+     * Callback invoked when the input value changes, typically via keyboard interactions.
+     *
+     * Using this prop instead of `onChange` can help avoid common bugs in React 16 related to Event Pooling
+     * where developers forget to save the text value from a change event or call `event.persist()`.
+     *
+     * @see https://legacy.reactjs.org/docs/legacy-event-pooling.html
+     */
+    onValueChange?(value: string, targetElement: HTMLInputElement | null): void;
 
     /** Whether this input should use small styles. */
     small?: boolean;
@@ -69,9 +76,7 @@ export interface InputGroupState {
     rightElementWidth?: number;
 }
 
-const NON_HTML_PROPS: Array<keyof InputGroupProps> = [
-    "onValueChange",
-];
+const NON_HTML_PROPS: Array<keyof InputGroupProps> = ["onValueChange"];
 
 /**
  * Input group component.
@@ -167,7 +172,7 @@ export class InputGroup extends AbstractPureComponent<InputGroupProps, InputGrou
     private handleInputChange = (event: React.ChangeEvent<HTMLInputElement>) => {
         const value = event.currentTarget.value;
         this.props.onChange?.(event);
-        this.props.onValueChange?.(value/*, this.inputElement*/);
+        this.props.onValueChange?.(value, event.currentTarget);
     };
 
     private maybeRenderLeftElement() {
--- a/packages/core/src/components/forms/inputGroup.tsx
+++ b/packages/core/src/components/forms/inputGroup.tsx
@@ -170,9 +170,9 @@ export class InputGroup extends AbstractPureComponent<InputGroupProps, InputGrou
     }
 
     private handleInputChange = (event: React.ChangeEvent<HTMLInputElement>) => {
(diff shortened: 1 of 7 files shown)


packages/core/src/components/forms/inputGroup.tsx
```diff
@@ -156,6 +164,12 @@ export class InputGroup extends AbstractPureComponent<InputGroupProps, InputGrou
         }
     }
 
+    private handleInputChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+        const value = event.currentTarget.value;
+        this.props.onChange?.(event);
+        this.props.onValueChange?.(value/*, this.inputElement*/);
```
thasner: `NumericInput` also passes back reference to the `inputElement`.
Wasn't sure if I should copy that pattern, not sure what it's for.
https://github.com/palantir/blueprint/blob/95b7fea60e08c86a7d406441f268bbd1a7a6a96e/packages/core/src/components/forms/numericInput.tsx#L642

packages/core/src/components/forms/inputGroup.tsx
```diff
@@ -156,6 +164,12 @@ export class InputGroup extends AbstractPureComponent<InputGroupProps, InputGrou
         }
     }
 
+    private handleInputChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+        const value = event.currentTarget.value;
+        this.props.onChange?.(event);
```
thasner, line 174: I wasn't sure if it'd be reasonable to ban native `onChange` usage, let alone make such a breaking change.

packages/core/src/components/forms/inputGroup.tsx
```diff
@@ -40,6 +40,9 @@ export interface InputGroupProps
     /** Whether this input should use large styles. */
     large?: boolean;
 
+    /** The callback invoked when the value changes.
+    onValueChange?(valueAsString: string/*, inputElement: HTMLInputElement | null*/): void;
```
ash211: I think there are some commenting issues here

packages/core/src/components/forms/inputGroup.tsx
adidahiya: No, we won't ban it, and we can't make a breaking change like that right now. We can just update the docs and maybe create a lint rule.

packages/core/src/components/forms/inputGroup.tsx
thasner: @adidahiya do you know why `inputElement` is an arg in `NumericInput`'s `onValueChange` callback?

Would you like me to do the same here?

packages/core/src/components/forms/inputGroup.tsx
thasner: thanks Andrew. left this here as a TODO. https://github.com/palantir/blueprint/pull/6290/files#r1268590322

packages/core/src/components/forms/inputGroup.tsx
```diff
@@ -40,6 +40,9 @@ export interface InputGroupProps
     /** Whether this input should use large styles. */
     large?: boolean;
 
+    /** The callback invoked when the value changes.
```
thasner:
```suggestion
    /** The callback invoked when the value changes. */
```

packages/core/src/components/forms/inputGroup.tsx
adidahiya: @thasner it was requested here: https://github.com/palantir/blueprint/issues/3943. I think it's a good idea to add it for your new prop as well.

packages/core/src/components/forms/inputGroup.tsx
adidahiya: Event pooling is going away in React 17, so I think this will become less of a problem: https://legacy.reactjs.org/blog/2020/08/10/react-v17-rc.html#no-event-pooling

I added a note in the JSDoc for the new prop about why users might prefer this handler instead of `onChange`.

packages/core/src/components/forms/inputGroup.tsx
adidahiya: done in [86881c0](https://github.com/palantir/blueprint/pull/6290/commits/86881c0b1b9ca7722e10016572e6d77cb09620dd)

adidahiya:
[docs preview](https://output.circle-artifacts.com/output/job/f737addc-f337-4085-bfc1-e13511f7e22f/artifacts/0/packages/docs-app/dist/index.html#core/components/text-inputs.input-group) looks fine 👍🏽

adidahiya (approved):

packages/core/src/components/forms/inputGroup.tsx
thasner: sounds good.
imo the `onValueChange` signature is also an ergonomic improvement, so i anticipate it'll still have usage

packages/docs-app/src/components/icons.tsx
```diff
@@ -100,10 +100,7 @@ export class Icons extends React.PureComponent<IconsProps, IconsState> {
         return icons.filter(icon => iconFilter(this.state.filter, icon));
     }
 
-    private handleFilterChange = (e: React.SyntheticEvent<HTMLInputElement>) => {
-        const filter = (e.target as HTMLInputElement).value;
-        this.setState({ filter });
-    };
+    private handleFilterChange = (filter: string) => this.setState({ filter });
```
thasner, line 103: nice

PR #6312, opened 2023-08-01
[core] add AsyncControllableTextArea based on AsyncControllableInput & add async control option for TextArea component

nurseiit:
Related to https://github.com/palantir/blueprint/issues/4678.

#### Checklist

- [x] Includes tests
- [ ] Update documentation

a. (~~Extends the existing `AsyncControllableInput` with an option to render a `textarea` element. (the type of Props is quite complicated now, and I am open to suggestions to further simplify it)~~) -- initial approach. Ended up with a `useAsyncControllableValue` hook to use in `AsyncControllableTextArea` component.

b. Adds `asyncControl` option to `TextArea` component that ~~reuses `AsyncControllableInput`~~ uses `AsyncControllableTextArea` when set to `true`.

Basically the only thing I was afraid of and tried to avoid is any external incompatibilities with the API and component props types. But since `AsyncControllableInput` is not exported externally and only used as a part of other components I think we can get away with the introduced complex types.

Also I generalised the tests for `AsyncControllableInput` instead of copying them, one can argue this was unnecessary, but I am also open for any suggestions on this.

Thanks!


adidahiya:
<h3>add correct types & copy tests from default input</h3>
Build artifact links for this commit: <strong><a href="https://output.circle-artifacts.com/output/job/de8128e9-c34f-480e-96a8-9f981311cd6c/artifacts/0/packages/docs-app/dist/index.html">documentation</a> | <a href="https://output.circle-artifacts.com/output/job/de8128e9-c34f-480e-96a8-9f981311cd6c/artifacts/0/packages/landing-app/dist/index.html">landing</a> | <a href="https://output.circle-artifacts.com/output/job/de8128e9-c34f-480e-96a8-9f981311cd6c/artifacts/0/packages/table-dev-app/dist/index.html">table</a> | <a href="https://output.circle-artifacts.com/output/job/de8128e9-c34f-480e-96a8-9f981311cd6c/artifacts/0/packages/demo-app/dist/index.html">demo</a></strong>

adidahiya:
<h3>fix lint, object order</h3>
Build artifact links for this commit: <strong><a href="https://output.circle-artifacts.com/output/job/23cfa394-2aa5-4650-9e6b-e1ad44fee8e2/artifacts/0/packages/docs-app/dist/index.html">documentation</a> | <a href="https://output.circle-artifacts.com/output/job/23cfa394-2aa5-4650-9e6b-e1ad44fee8e2/artifacts/0/packages/landing-app/dist/index.html">landing</a> | <a href="https://output.circle-artifacts.com/output/job/23cfa394-2aa5-4650-9e6b-e1ad44fee8e2/artifacts/0/packages/table-dev-app/dist/index.html">table</a> | <a href="https://output.circle-artifacts.com/output/job/23cfa394-2aa5-4650-9e6b-e1ad44fee8e2/artifacts/0/packages/demo-app/dist/index.html">demo</a></strong>

packages/core/src/components/forms/asyncControllableInputProps.ts
```diff
@@ -0,0 +1,38 @@
...
+     * @type {"input" | "textarea"}
+     * @default "input"
+     */
+    tagName?: T;
+    inputRef?: React.Ref<AsyncControllableElement<T>>;
+
+    // NOTE: these are copied from the React.HTMLAttributes interface definition.
+    onChange?: React.ChangeEventHandler<AsyncControllableElement<T>> | undefined;
+    onCompositionStart?: React.CompositionEventHandler<AsyncControllableElement<T>> | undefined;
+    onCompositionEnd?: React.CompositionEventHandler<AsyncControllableElement<T>> | undefined;
```
nurseiit: not sure if this is the best way to do it, but I would argue this is fine since this component is not exported directly.

adidahiya:
<h3>refactor tests</h3>
Build artifact links for this commit: <strong><a href="https://output.circle-artifacts.com/output/job/6ba787b6-ed7d-4c35-90b7-627a21efcd3b/artifacts/0/packages/docs-app/dist/index.html">documentation</a> | <a href="https://output.circle-artifacts.com/output/job/6ba787b6-ed7d-4c35-90b7-627a21efcd3b/artifacts/0/packages/landing-app/dist/index.html">landing</a> | <a href="https://output.circle-artifacts.com/output/job/6ba787b6-ed7d-4c35-90b7-627a21efcd3b/artifacts/0/packages/table-dev-app/dist/index.html">table</a> | <a href="https://output.circle-artifacts.com/output/job/6ba787b6-ed7d-4c35-90b7-627a21efcd3b/artifacts/0/packages/demo-app/dist/index.html">demo</a></strong>

packages/core/src/components/forms/asyncControllableInputProps.ts
nurseiit: I did not want to further complicate the types by excluding and extending the native interfaces above.

nurseiit:
Hi @gluxon, any chance you could take a look a this? Thanks!

packages/core/test/forms/asyncControllableInputTests.tsx
```diff
@@ -25,148 +25,192 @@ import { spy } from "sinon";
...
+                );
+                return;
+            });
         });
     });
+};
+
+describe("<AsyncControllableInput>", () => {
+    runAsyncControllableInputTests({ tagName: "input", inputType: "text" });
+    runAsyncControllableInputTests({ tagName: "textarea" });
```
gluxon: I like the test deduplication here. Thanks for doing that.

What are your thoughts on `describe.each`? That would be another way to write this in more idiomatic Jest. https://jestjs.io/docs/api#describeeachtablename-fn-timeout

packages/core/src/components/forms/textArea.tsx
```diff
@@ -170,13 +190,22 @@ export class TextArea extends AbstractPureComponent<TextAreaProps, TextAreaState
             };
         }
 
-        return (
+        return asyncControl ? (
```
gluxon: Does this seem to work when testing live? The code changes look correct. Just wondering if `<TextArea asyncControl />` works as expected.

gluxon (commented):
Apologies for the late response — It took me a bit of time to come around to it, but I like how this takes advantage of generic types to ensure type safety of the input props.

---

A discussion of a long-term approach may make sense. Zooming out a bit, more recent React code tends to lean on hooks for component logic sharing. I think the need to add additional functionality to this existing `AsyncControllableInput` class component comes from the fact that it was written when component composition was a more common logic reuse pattern. In that prior world, this PR makes sense.

If we were to imagine an alternative hooks-based approach, I think adding support for new HTML textual input elements becomes significantly easier. This example isn't complete and likely fails to compile, but here's the rough direction I was thinking:

```tsx
function AsyncControllableInput(props: P) {
    const { value, onChange, ...rest } = props;
    const { textInputProps } = useAsyncControlledValue({ value, onChange });

    return <input {...rest} {...textInputProps} />
}

function AsyncControllableTextarea(props: P) {
    const { value, onChange, ...rest } = props;
    const { textInputProps } = useAsyncControlledValue({ value, onChange });

    return <textarea {...rest} {...textInputProps} />
}

function useAsyncControllableValue(opts) {
    const [isComposing, setIsComposing] = React.useState(false);
    const [nextValue, setNextValue ] = React.useState(/* ... */);
    const [hasPendingUpdate, setHasPendingUpdate ] = React.useState(/* ... */);

    function onCompositionStart() {
        setIsComposing(true)
    }

    function onCompositionEnd() {
        setIsComposing(false)
    }

    const value = isComposing || hasPendingUpdate ? nextValue : value

    /* ... */

    return {
        value,
        onChange,
        onCompositionEnd,
        onCompositionStart
    };
}
```

---

Since this PR adds a new functionality, I do wonder if it makes sense to create a new `AsyncControllableTextarea` that's based upon the hook approach. We would then battle-test that purely new functionality before switching `AsyncControllableInput` to be implemented from the same hook in a future Blueprint release.

I'm also okay extending the existing `AsyncControllableInput` to be generic since it's private anyway. I do think the hook approach is the preferable long-term approach though. What do you think?

> Basically the only thing I was afraid of and tried to avoid is any external incompatibilities with the API and component props types. But since `AsyncControllableInput` is not exported externally and only used as a part of other components I think we can get away with the introduced complex types.

I appreciate that! I scanned through the PR looking for changes to external APIs and don't see any other than the new `asyncControl` prop that was intended.

nurseiit:
> Apologies for the late response — It took me a bit of time to come around to it, but I like how this takes advantage of generic types to ensure type safety of the input props.
> 
> A discussion of a long-term approach may make sense. Zooming out a bit, more recent React code tends to lean on hooks for component logic sharing. I think the need to add additional functionality to this existing `AsyncControllableInput` class component comes from the fact that it was written when component composition was a more common logic reuse pattern. In that prior world, this PR makes sense.
> 
> If we were to imagine an alternative hooks-based approach, I think adding support for new HTML textual input elements becomes significantly easier. This example isn't complete and likely fails to compile, but here's the rough direction I was thinking:
> 
> ```tsx
> function AsyncControllableInput(props: P) {
>     const { value, onChange, ...rest } = props;
>     const { textInputProps } = useAsyncControlledValue({ value, onChange });
> 
>     return <input {...rest} {...textInputProps} />
> }
> 
> function AsyncControllableTextarea(props: P) {
>     const { value, onChange, ...rest } = props;
>     const { textInputProps } = useAsyncControlledValue({ value, onChange });
> 
>     return <textarea {...rest} {...textInputProps} />
> }
> 
> function useAsyncControllableValue(opts) {
>     const [isComposing, setIsComposing] = React.useState(false);
>     const [nextValue, setNextValue ] = React.useState(/* ... */);
>     const [hasPendingUpdate, setHasPendingUpdate ] = React.useState(/* ... */);
> 
>     function onCompositionStart() {
>         setIsComposing(true)
>     }
> 
>     function onCompositionEnd() {
>         setIsComposing(false)
>     }
> 
>     const value = isComposing || hasPendingUpdate ? nextValue : value
> 
>     /* ... */
> 
>     return {
>         value,
>         onChange,
>         onCompositionEnd,
>         onCompositionStart
>     };
> }
> ```
> 
> Since this PR adds a new functionality, I do wonder if it makes sense to create a new `AsyncControllableTextarea` that's based upon the hook approach. We would then battle-test that purely new functionality before switching `AsyncControllableInput` to be implemented from the same hook in a future Blueprint release.
> 
> I'm also okay extending the existing `AsyncControllableInput` to be generic since it's private anyway. I do think the hook approach is the preferable long-term approach though. What do you think?
> 
> > Basically the only thing I was afraid of and tried to avoid is any external incompatibilities with the API and component props types. But since `AsyncControllableInput` is not exported externally and only used as a part of other components I think we can get away with the introduced complex types.
> 
> I appreciate that! I scanned through the PR looking for changes to external APIs and don't see any other than the new `asyncControl` prop that was intended.

I really do like the hooks approach! That's what I was aiming for at first, but then gave up unfortunately. In any case, I will revisit this again with hooks. Thanks for reviewing!

gluxon:
> I really do like the hooks approach! That's what I was aiming for at first, but then gave up unfortunately. In any case, I will revisit this again with hooks. Thanks for reviewing!

Please reach out if you want a second pair of eyes while looking at this! 🙂

nurseiit:
> > I really do like the hooks approach! That's what I was aiming for at first, but then gave up unfortunately. In any case, I will revisit this again with hooks. Thanks for reviewing!
> 
> Please reach out if you want a second pair of eyes while looking at this! 🙂

Thanks! I updated the PR with hooks alternative based on the `AsyncControllableInput` component and your comment above. Let me know what do you think 😌

adidahiya:
<h3>add more comments</h3>
Build artifact links for this commit: <strong><a href="https://output.circle-artifacts.com/output/job/54aa7b6d-d87a-4d51-bb6e-f036003d386f/artifacts/0/packages/docs-app/dist/index.html">documentation</a> | <a href="https://output.circle-artifacts.com/output/job/54aa7b6d-d87a-4d51-bb6e-f036003d386f/artifacts/0/packages/landing-app/dist/index.html">landing</a> | <a href="https://output.circle-artifacts.com/output/job/54aa7b6d-d87a-4d51-bb6e-f036003d386f/artifacts/0/packages/table-dev-app/dist/index.html">table</a> | <a href="https://output.circle-artifacts.com/output/job/54aa7b6d-d87a-4d51-bb6e-f036003d386f/artifacts/0/packages/demo-app/dist/index.html">demo</a></strong>

packages/core/src/components/forms/asyncControllableTextArea.tsx
```diff
@@ -0,0 +1,46 @@
+/* !
+ * (c) Copyright 2023 Palantir Technologies Inc. All rights reserved.
+ */
+
+import * as React from "react";
+
+import { useAsyncControllableValue } from "./useAsyncControllableValue";
+
+export type IAsyncControllableTextAreaProps = React.TextareaHTMLAttributes<HTMLTextAreaElement> & {
+    inputRef?: React.Ref<HTMLTextAreaElement>;
```
gluxon: Does `React.forwardRef` make sense to use? I think we didn't have that option with the class-based component, but it's open to us now that this is a functional component.

packages/core/src/components/forms/textArea.tsx
```diff
@@ -170,13 +190,21 @@ export class TextArea extends AbstractPureComponent<TextAreaProps, TextAreaState
...
+            <AsyncControllableTextArea
```
gluxon: Assuming it makes sense to use `React.forwardRef`, another way we could write this would be:

```tsx
const TextAreaComponent = asyncControl ? AsyncControllableTextArea : textarea;
return (
  <TextAreaComponent
    {...htmlProps}
    className={rootClasses}
    onChange={this.handleChange}
    style={style}
    ref={this.handleRef}
  />
);
```

Feel free to accept that edit if you prefer it. Otherwise feel free to ignore.

packages/core/src/components/forms/asyncControllableTextArea.tsx
```diff
@@ -0,0 +1,46 @@
...
+};
+
+/**
+ * A wrapper around the low-level <textarea> component which works around a React bug
+ * the same way <AsyncControllableInput> does.
+ */
+export const AsyncControllableTextArea = React.memo<IAsyncControllableTextAreaProps>(
```
gluxon: I have a guess that `React.memo` may not be needed on this component since it returns a small `<textarea>`. Does that sound right?

packages/core/src/components/forms/useAsyncControllableValue.ts
```diff
@@ -0,0 +1,138 @@
...
+    // The latest input value, which updates during IME composition.
+    const [nextValue, setNextValue] = React.useState(propValue);
+
+    // Whether we are in the middle of a composition event.
+    const [isComposing, setIsComposing] = React.useState(false);
+
+    // Whether there is a pending update we are expecting from a parent component.
+    const [hasPendingUpdate, setHasPendingUpdate] = React.useState(false);
+
+    const cancelPendingCompositionEnd = React.useRef<(() => void) | null>(null);
```
gluxon: Does the suggestion below also work? I saw that the original version used `null` instead of `undefined` to be more clear, but I think `useRef`'s typings make `undefined` a bit more ergonomic. Feel free to ignore this suggestion if you disagree.

```suggestion
    const cancelPendingCompositionEnd = React.useRef<() => void>();
```

packages/core/src/components/forms/useAsyncControllableValue.ts
```diff
@@ -0,0 +1,138 @@
...
+
+        if (userTriggeredUpdate) {
+            if (propValue === nextValue) {
+                // parent has processed and accepted our update
+                if (hasPendingUpdate) {
+                    setValue(propValue);
+                    setHasPendingUpdate(false);
+                } else if (value !== nextValue) {
+                    setValue(nextValue);
+                }
```
gluxon: There might be opportunity to simplify this block. I know it was copied from the original version, but since we're rewriting it anyway, it might be cool to revisit.

Staring at this part, I think I can convince myself that these blocks are equivalent:

```ts
if (propValue === nextValue) {
    // parent has processed and accepted our update
    if (hasPendingUpdate) {
        setValue(propValue);
        setHasPendingUpdate(false);
    } else if (value !== nextValue) {
        setValue(nextValue);
    }
} else {
```

```ts
if (propValue === nextValue) {
    // parent has processed and accepted our update
    setValue(propValue);
    setHasPendingUpdate(false);
} else {
```

The theory being that

1. `propValue` and `nextValue` are the same, so either can be used to `setValue`. 
2. React should no-op the `setHasPendingUpdate(false)` call if it's already `false`.

packages/core/src/components/forms/useAsyncControllableValue.ts
```diff
@@ -0,0 +1,138 @@
...
+                } else {
+                    // accept controlled update overriding user action
+                    setValue(propValue);
+                    setNextValue(propValue);
+                    setHasPendingUpdate(false);
+                }
+            }
+        } else {
+            // accept controlled update, could be confirming or denying user action
+            if (value !== propValue || hasPendingUpdate) {
```
gluxon, line 115: This was added to reduce infinite renders, is that right? Could we add a comment for that? Hoping to make that clear for the next reader of this code.

packages/core/src/components/forms/useAsyncControllableValue.ts
```diff
@@ -0,0 +1,138 @@
...
+            onChange?.(event);
+        },
+        [onChange],
+    );
+
+    if (isComposing || propValue === undefined) {
+        // don't derive anything from props if:
+        // - in uncontrolled mode, OR
+        // - currently composing, since we'll do that after composition ends
+    } else {
```
gluxon: How do you feel about this suggestion?

```suggestion
    // don't derive anything from props if:
    // - in uncontrolled mode, OR
    // - currently composing, since we'll do that after composition ends
    const isStateDerivationPaused =  isComposing || propValue === undefined
    if (!isStateDerivationPaused) {
```

It feels like the empty if block could be replaced somehow, but not too opinionated on that. I do see it came from the early return in the `getDerivedStateFromProps` version of this, so I think this is okay to keep around for the hook version.

packages/core/src/components/forms/useAsyncControllableValue.ts
gluxon: Taking this a bit further, I think the nested logic can be much flatter.

```tsx
if (!isStateDerivationPaused) {
    const userTriggeredUpdate = nextValue !== value;

    if (userTriggeredUpdate && propValue === nextValue) {
        // parent has processed and accepted our update
        setValue(propValue);
        setHasPendingUpdate(false);
    } else if (userTriggeredUpdate && propValue === value) {
        // we have sent the update to our parent, but it has not been processed yet. just wait.
        // DO NOT set nextValue here, since that will temporarily render a potentially stale controlled value,
        // causing the cursor to jump once the new value is accepted
        if (!hasPendingUpdate) {
            setHasPendingUpdate(true);
        }
    } else if (userTriggeredUpdate || value !== propValue || hasPendingUpdate) {
        // accept controlled update, could be confirming or denying user action
        setValue(propValue);
        setNextValue(propValue);
        setHasPendingUpdate(false);
    }
}
```

But I could definitely use help making sure I refactored that correctly. The general strategy was to propagate the `userTriggeredUpdate` condition down and then remove layers of nesting.

If you would rather not think about further refactors and prefer to keep the existing structure as-is to match the current `getDerivedStateFromProps` approach, I think that's okay too!

gluxon (commented):
Thanks for working on this! I personally like how this came out. Just have a few more questions and I think we'll be ready to merge soon.

Played around with the latest hook-based version live. Seeing all the expected expected state updates.

Screenshot 2023-08-19 at 1 22 50 AM

packages/core/src/components/forms/useAsyncControllableValue.ts
nurseiit: Awesome, thanks!

packages/core/src/components/forms/textArea.tsx
nurseiit: Yep, `forwardRef` fits best in this case, thanks!

nurseiit:
Thanks for the review @gluxon! Updated based on the latest comments / suggestions 👍

nurseiit:
Thanks for the review @gluxon! Updated based on the latest comments / suggestions 👍

packages/core/src/components/forms/useAsyncControllableValue.ts
nurseiit: I refactored this bit to be flatter and (arguably) more readable than the initial `getDerivedStateFromProps` version in the latest commit. Thanks for the ideas above!

packages/core/src/components/forms/useAsyncControllableValue.ts
nurseiit: Added, in both places

packages/core/test/forms/asyncControllableInputTests.tsx
```diff
@@ -170,3 +172,149 @@ describe("<AsyncControllableInput>", () => {
         });
     });
 });
+
```
nurseiit, line 35: I would probably need to move this to its own file, but maybe would be better to reuse them, as suggested with `describe.each`?

adidahiya:
<h3>flatten & simplify state derival + comments</h3>
Build artifact links for this commit: <strong><a href="https://output.circle-artifacts.com/output/job/9e9660dc-cd00-45b6-97e5-1a60ea11dfba/artifacts/0/packages/docs-app/dist/index.html">documentation</a> | <a href="https://output.circle-artifacts.com/output/job/9e9660dc-cd00-45b6-97e5-1a60ea11dfba/artifacts/0/packages/landing-app/dist/index.html">landing</a> | <a href="https://output.circle-artifacts.com/output/job/9e9660dc-cd00-45b6-97e5-1a60ea11dfba/artifacts/0/packages/table-dev-app/dist/index.html">table</a> | <a href="https://output.circle-artifacts.com/output/job/9e9660dc-cd00-45b6-97e5-1a60ea11dfba/artifacts/0/packages/demo-app/dist/index.html">demo</a></strong>

packages/core/test/forms/asyncControllableInputTests.tsx
gluxon: You're right that this feels a bit strange since `AsyncControllableTextArea` has deviated from `AsyncControllableInput`. I think it would be fair to use `describe.each` (assuming that part still makes sense), but add an inline source code comment that the plan is to have `AsyncControllableInput` use the hook in a future change? Therefore it makes sense for the test to stay together?

gluxon: Would we still be able to share the tests between the `textarea` and `input` with the `describe.each` approach? I did like the initial version of the PR that did that.

gluxon (commented):
Thanks for the updates! I think everything looks good. Expecting to approve and merge once the discussion about tests https://github.com/palantir/blueprint/pull/6312/files/bdc07b6e4a19ef154d84cbc233943985c6f91cbd#r1300068035 wraps up.

packages/core/src/components/forms/useAsyncControllableValue.ts
gluxon: Looks good!

packages/core/test/forms/asyncControllableInputTests.tsx
nurseiit: Updated the tests. Though turns out `mocha` does not have an 1-to-1 alternative to `jest`'s `describe.each` – so I had to work around that with `.forEach` over testcases.

nurseiit:
Updated the tests & added a clarifying note at the top of the test file ☺️ Thanks for reviewing this @gluxon!

adidahiya:
<h3>dedup tests & add note</h3>
Build artifact links for this commit: <strong><a href="https://output.circle-artifacts.com/output/job/131b6626-e9c5-4601-94ab-2170609230aa/artifacts/0/packages/docs-app/dist/index.html">documentation</a> | <a href="https://output.circle-artifacts.com/output/job/131b6626-e9c5-4601-94ab-2170609230aa/artifacts/0/packages/landing-app/dist/index.html">landing</a> | <a href="https://output.circle-artifacts.com/output/job/131b6626-e9c5-4601-94ab-2170609230aa/artifacts/0/packages/table-dev-app/dist/index.html">table</a> | <a href="https://output.circle-artifacts.com/output/job/131b6626-e9c5-4601-94ab-2170609230aa/artifacts/0/packages/demo-app/dist/index.html">demo</a></strong>

gluxon (approved):
Thank you for the improvements and new feature @nurseiit! 🎉

packages/core/src/components/forms/asyncControllableTextArea.tsx
```diff
@@ -0,0 +1,43 @@
+/* !
+ * (c) Copyright 2023 Palantir Technologies Inc. All rights reserved.
+ */
+
+import * as React from "react";
+
+import { useAsyncControllableValue } from "./useAsyncControllableValue";
+
+export type IAsyncControllableTextAreaProps = React.TextareaHTMLAttributes<HTMLTextAreaElement>;
```
adidahiya, line 9: nit: we are no longer naming interfaces with the `I` prefix in this repo

packages/core/src/components/forms/asyncControllableTextArea.tsx
```diff
@@ -0,0 +1,43 @@
...
+                {...restProps}
+                value={value}
+                onChange={onChange}
+                onCompositionStart={onCompositionStart}
+                onCompositionEnd={onCompositionEnd}
+                ref={ref}
+            />
+        );
+    },
+);
```
adidahiya, line 43: minor: we should add a component display name for debugging purposes, e.g.

```ts
AsyncControllableTextArea.displayName = `${DISPLAYNAME_PREFIX}.AsyncControllableTextArea`;
```

packages/core/src/components/forms/useAsyncControllableValue.ts
```diff
@@ -0,0 +1,133 @@
+/* !
+ * (c) Copyright 2023 Palantir Technologies Inc. All rights reserved.
+ */
+
+import * as React from "react";
+
+interface IUseAsyncControllableValueProps<E extends HTMLInputElement | HTMLTextAreaElement> {
```
adidahiya, line 7: nit: same comment about avoiding the `I` prefix

adidahiya (commented):
Thanks @nurseiit & @gluxon! I like the new approach of using the hook to implement this behavior. We should be able to migrate AsyncControllableInput to use it as well soon.

I can fix my minor review comments in a follow-up PR.

PR #6328, opened 2023-08-21
feat(Section): Add controlled mode support

### LINKED ISSUE #6323: Section component controlled state
#### Environment

- __Package version(s)__: 5.0
- __Browser and OS versions__: Windows 11

#### Feature request

The `Section` component should allow the ability to fully control the `isOpen` state.

Allow providing the `isOpen` value to the underlying `Collapse` component.
Additionally, an `onToggle/onClick` callback would need to be registered with the `Section` header.

nerdstep:
#### Checklist

- [x] Includes tests
- [x] Update documentation

Implementation & documentation.

#### Screenshot

image


packages/docs-app/src/examples/core-examples/sectionExample.tsx
```diff
@@ -91,8 +95,26 @@ export class SectionExample extends React.PureComponent<ExampleProps, SectionExa
...
+                        />
+                        <Switch
+                            checked={this.state.isOpen}
+                            disabled={!this.state.isControlled}
+                            label="Open"
+                            onChange={this.toggleIsOpen}
+                        />
                     </>
                 )}
                 <Label>
```
adidahiya, line 95: the ordering/hierarchy of these controls is a bit confusing now. I think this "Elevation" label & slider needs to go above the "Collapse Props" section:

image

packages/docs-app/src/examples/core-examples/sectionExample.tsx
```diff
@@ -91,8 +95,26 @@ export class SectionExample extends React.PureComponent<ExampleProps, SectionExa
...
-                        <Switch checked={defaultIsOpen} label="Default is open" onChange={this.toggleDefaultIsOpen} />
+                        <H5>Collapse Props</H5>
+                        <Switch
+                            checked={defaultIsOpen}
+                            disabled={this.state.isControlled}
+                            label="Default is open"
+                            onChange={this.toggleDefaultIsOpen}
+                        />
+
+                        <H6>Control</H6>
```
adidahiya: Nit: I don't think this extra heading adds much, we can just omit it so it looks like this:

image

packages/core/src/components/section/section.tsx
```diff
@@ -34,6 +34,31 @@ import { Icon } from "../icon/icon";
...
+     * Whether the component is open or closed.
+     *
+     * Passing a boolean value to `isOpen` will enabled controlled mode for the component.
+     */
+    isOpen?: boolean;
+
+    /**
+     * Callback invoked in controlled mode when the collapse element is clicked.
+     */
+    onToggleCollapse?: () => void;
```
adidahiya: minor: I think we can abbreviate this to simply `onToggle`, as it's already inside a `collapseProps` key in `SectionProps`

```suggestion
    onToggle?: () => void;
```

packages/core/src/components/section/section.tsx
```diff
@@ -34,6 +34,31 @@ import { Icon } from "../icon/icon";
...
+
+    /**
+     * Whether the component is open or closed.
+     *
+     * Passing a boolean value to `isOpen` will enabled controlled mode for the component.
+     */
+    isOpen?: boolean;
+
+    /**
+     * Callback invoked in controlled mode when the collapse element is clicked.
```
adidahiya:
```suggestion
     * Callback invoked in controlled mode when the collapse toggle element is clicked.
```

adidahiya (changes requested):
nice thorough PR, thanks for adding unit tests!

just a few minor points of feedback. we should be able to merge this soon.

packages/docs-app/src/examples/core-examples/sectionExample.tsx
nerdstep: Yeah I was struggling a bit to understand what the design standards for this should be. I do agree that `Elevation` feels out of place on its own down there at the bottom.

My rationale for placing `CollapseProps` below `Collapsible`, is that when `collapsible=false` then the `Collapse Props` & `Control` sections are hidden. It also felt more natural that all of the collapse related items would be grouped together.

So here's the display with "Collapsible" disabled:

image

That said, would you prefer those options to always be visible and instead disable them?

Example:

image

packages/docs-app/src/examples/core-examples/sectionExample.tsx
nerdstep: Ok sounds good -- I was using the [Popover](https://blueprintjs.com/docs/#core/components/popover) docs for inspiration on this one.

packages/docs-app/src/examples/core-examples/sectionExample.tsx
adidahiya: > would you prefer those options to always be visible and instead disable them?

yeah, that sounds like a good solution too. let's go with that for now

nerdstep: done! changes pushed.

adidahiya (approved):
looks good in [docs preview build](https://output.circle-artifacts.com/output/job/ebc9ec0f-b956-4f92-b3f4-b0762cf07b80/artifacts/0/packages/docs-app/dist/index.html#core/components/section), thanks @nerdstep

PR #6331, opened 2023-08-25
[core] feat(Text): forward ref to DOM element

### LINKED ISSUE #6320: Text component doesn't correctly attach ref passed via props
The [`Text`](https://github.com/palantir/blueprint/blob/cba5612d28e0ba3d13e7ff21cf94b2e25e01ebb1/packages/core/src/components/text/text.tsx) component allows for a `ref` for the underlying HTML element via including `React.HTMLAttributes<HTMLElement>` in the props. However, the component passes its own ref in [here](https://github.com/palantir/blueprint/blob/cba5612d28e0ba3d13e7ff21cf94b2e25e01ebb1/packages/core/src/components/text/text.tsx#L73-L87). This results in a ref passed to `Text` never being invoked/set.

Ideally instead this ref should compose with whatever ref is passed

maclockard:
#### Checklist

- [ ] Includes tests
- [x] Update documentation

Makes it so that the `Text` component forwards a ref to the underlying HTML element.

Correctly following patterns of the rest of the code base

### DIFF
--- a/packages/core/src/components/text/text.tsx
+++ b/packages/core/src/components/text/text.tsx
@@ -17,10 +17,13 @@
 import classNames from "classnames";
 import * as React from "react";
 
-import { Classes } from "../../common";
+import { Classes, mergeRefs } from "../../common";
 import { DISPLAYNAME_PREFIX, Props } from "../../common/props";
 
-export interface TextProps extends Props {
+export interface TextProps
+    extends Props,
+        React.RefAttributes<HTMLElement>,
+        Omit<React.HTMLAttributes<HTMLElement>, "title"> {
     children?: React.ReactNode;
 
     /**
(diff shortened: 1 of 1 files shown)


packages/core/src/components/text/text.tsx
```diff
@@ -17,10 +17,13 @@
...
 
-import { Classes } from "../../common";
+import { Classes, mergeRefs } from "../../common";
 import { DISPLAYNAME_PREFIX, Props } from "../../common/props";
 
-export interface TextProps extends Props {
+export interface TextProps
+    extends Props,
+        React.RefAttributes<HTMLElement>,
+        Omit<React.HTMLAttributes<HTMLElement>, "title"> {
```
maclockard, line 26: Changed this to better match other `Prop` interfaces elsewhere in the code base.

adidahiya (approved):
[docs preview](https://output.circle-artifacts.com/output/job/07ccad27-6c18-4fd1-a130-1ce11ee021fe/artifacts/0/packages/docs-app/dist/index.html#core/components/text)

LGTM, thanks @maclockard!

PR #6382, opened 2023-09-13
[core] fix(NumericInput): increment/decrement with very small step size

### LINKED ISSUE #4497: NumericInput up/down keypress/button click breaks on stepsize smaller than 0.001
#### Environment

- __Package version(s)__: @blueprintjs/core: "3.38.1"
- __Operating System__: Ubuntu 20.04 LTS
- __Browser name and version__: Chrome 87.0.4280.141 

#### Code Sandbox

https://codesandbox.io/s/blueprint-sandbox-forked-98i26

#### Steps to reproduce

Provide a stepSize and minorStepSize such as 0.0001, and try to modify the input using the up/down button or the keypress. It will lock up and won't fire onValueChange correctly.

#### Actual behavior

Freezes

#### Expected behavior

Increments/Decrements value correctly

#### Possible solution

Tried to investigate a bit but couldn't figure it out. The count decimals util function seems to be working correctly however.

Rafael-Martins:
#### Checklist

- [X] Includes tests
- [ ] Update documentation


packages/core/src/components/forms/numericInputUtils.ts
```diff
@@ -30,7 +30,7 @@ function getDecimalSeparator(locale: string) {
 }
 
 export function toLocaleString(num: number, locale: string = "en-US") {
-    return sanitizeNumericInput(num.toLocaleString(locale), locale);
+    return sanitizeNumericInput(num.toLocaleString(locale, { maximumSignificantDigits: 21 }), locale);
```
adidahiya: why `21`? according to the [NumberFormat docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/NumberFormat#options), this should already be the default. maybe you need to adjust a different option like `roundingMode`?

packages/core/test/forms/numericInputTests.tsx
```diff
@@ -0,0 +1,48 @@
...
+ */
+
+import { assert } from "chai";
+import { mount } from "enzyme";
+import { spy } from "sinon";
+import * as React from "react";
+
+import { NumericInput } from "../../src";
+
+describe("<NumericInput>", () => {
```
adidahiya: there are already tests for this component in `packages/core/test/controls/numericInputTests.tsx`. can you please move your new tests into that test suite?

adidahiya (changes requested):

packages/core/src/components/forms/numericInputUtils.ts
```diff
@@ -30,7 +30,7 @@ function getDecimalSeparator(locale: string) {
 }
 
 export function toLocaleString(num: number, locale: string = "en-US") {
-    return sanitizeNumericInput(num.toLocaleString(locale), locale);
+    return sanitizeNumericInput(num.toLocaleString(locale, { roundingPriority: "morePrecision" } as any), locale);
```
Rafael-Martins, line 34: I used `as any` here because this property has not included in the type, I was wondering if we have a better way to do it?

Reference: https://github.com/microsoft/TypeScript/issues/52072

packages/core/src/components/forms/numericInputUtils.ts
Rafael-Martins: I was misunderstanding how it works, I opened a task to enhance the docs in this part https://github.com/mdn/mdn/issues/444 😄

packages/core/src/components/forms/numericInputUtils.ts
adidahiya, line 34: @Rafael-Martins why do you need to cast `as any`? I'd prefer to avoid that cast if we can. If it's _absolutely necessary_, then it should be accompanied with a `// HACKHACK: explanation` comment on the previous line.

adidahiya (changes requested):

packages/core/src/components/forms/numericInputUtils.ts
Rafael-Martins: Hey, sorry, I forgot to send my commentary about it, but the reason is: 

> I used `as any` here because this property has not included in the type, I was wondering if we have a better way to do it?
> 
> Reference: [microsoft/TypeScript#52072](https://github.com/microsoft/TypeScript/issues/52072)

I will add the HACKHACK or even try a better approach!

adidahiya (approved):
thanks @Rafael-Martins!

PR #6416, opened 2023-09-30
docs: view source link is broken for htmlSelectExample

### LINKED ISSUE #6296: docs: view source link is broken for htmlSelectExample
[View source on GitHub](https://github.com/palantir/blueprint/blob/develop/packages/docs-app/src/examples/core-examples/hTMLSelectExample.tsx) link for [HTMLSelectExample](https://blueprintjs.com/docs/#core/components/html-select) page is broken:

leads to `hTMLSelectExample.tsx` (with only the first letter in lowercase) instead of `htmlSelectExample.tsx`.

Looks like this is due to https://github.com/palantir/blueprint/blob/d73a4055f2b5f45b1042fd838a1ce2127d5088c6/packages/docs-app/src/tags/reactExamples.ts#L37

#### Steps to reproduce

1. go to https://blueprintjs.com/docs/#core/components/html-select
1. click `"View source on GitHub"`
1. observe that it leads to a non-existent file

#### Possible solution

Not sure if we should make an exception for this example only, and I figure it would break for other uppercase example files that could be added in the future.

electron97:
## Summary

- Currently the view source [link](https://github.com/palantir/blueprint/blob/develop/packages/docs-app/src/examples/core-examples/hTMLSelectExample.tsx) is not working for htmlSelectExmaple.
- This is due to [logic](https://github.com/palantir/blueprint/blob/develop/packages/docs-app/src/tags/reactExamples.ts#L38) written to generate the view source link. It takes the first character of the example name and converts it to lowerCase and append with the remaining characters. 
- Since the name of the example is `HTMLSelectExample` it converts it `hTMLSelectExample.tsx` for which we don't have any file in our codebase which results in file not found.
- To fix the issue we can either remain the file to `hTMLSelectExample.tsx` so that after the view source link is generated it finds the file in our codebase or else we can rename the example name to `HtmlSelectExample` instead of `HTMLSelectExample`.
- Since the later required changes in many places in our codebase we went ahead with renaming the file instead.

#### Fixes #6296


palantirtech:
Thanks for your interest in [palantir/blueprint](https://github.com/palantir/blueprint), @electron97! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

packages/docs-app/src/examples/core-examples/index.ts
```diff
@@ -40,7 +40,7 @@ export * from "./formGroupExample";
 export { HotkeyPiano } from "./hotkeyPiano";
 export { HotkeyTesterExample } from "./hotkeyTesterExample";
 export { HotkeysTarget2Example } from "./hotkeysTarget2Example";
-export { HTMLSelectExample } from "./htmlSelectExample";
+export { HTMLSelectExample } from "./hTMLSelectExample";
```
adidahiya, line 43: I'd rather not change the file name and break our file naming convention.

Instead, the file name matching code should be fixed to deal with component names which start with multiple uppercase characters: https://github.com/palantir/blueprint/blob/d73a4055f2b5f45b1042fd838a1ce2127d5088c6/packages/docs-app/src/tags/reactExamples.ts#L37

adidahiya (changes requested):

PR #6421, opened 2023-10-02
[datetime2] Do not use `export * as ns` for webpack 4 compatibility

### LINKED ISSUE #6420: [datetime2] is not compatible with webpack 4
#### Environment

- __Package version(s)__: datetime2: 2.0.0

#### Steps to reproduce

Build project that imports `@blueprintjs/datetime2` with webpack 4 

#### Actual behavior

```
ERROR in ./node_modules/.pnpm/@blueprintjs+datetime2@2.0.0_@types+react@17.0.67_react-dom@17.0.2_react@17.0.2/no
de_modules/@blueprintjs/datetime2/lib/esm/index.js 20:9
Module parse failed: Unexpected token (20:9)
You may need an appropriate loader to handle this file type, currently no loaders are configured to process this
 file. See https://webpack.js.org/concepts#loaders
| export { DateRangeInput3 } from './components/date-range-input3/dateRangeInput3';
| export { DateRangePicker3 } from './components/date-range-picker3/dateRangePicker3';
> export * as DateInput2MigrationUtils from './dateInput2MigrationUtils';
| export { Classes as Datetime2Classes, ReactDayPickerClasses } from './classes';

```

#### Possible solution

Do not use `export * as namespace`, looks like support for it was added in webpack 5 only https://github.com/webpack/webpack/pull/11117

Webpack 4 is quite old but it can load the rest of the bp libs so it would be great to keep it supported until blueprintjs v6.

kalekseev:
#### Fixes #6420

### DIFF
--- a/packages/datetime2/src/index.ts
+++ b/packages/datetime2/src/index.ts
@@ -18,8 +18,9 @@ export { DatePicker3, DatePicker3Props } from "./components/date-picker3/datePic
 export { DateInput3, DateInput3Props } from "./components/date-input3/dateInput3";
 export { DateRangeInput3, DateRangeInput3Props } from "./components/date-range-input3/dateRangeInput3";
 export { DateRangePicker3, DateRangePicker3Props } from "./components/date-range-picker3/dateRangePicker3";
-export * as DateInput2MigrationUtils from "./dateInput2MigrationUtils";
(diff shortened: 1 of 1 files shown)


adidahiya (approved):
thanks @kalekseev!

PR #6062, opened 2023-04-07
[core] fix(MenuDivider): handle escape key to close popover

mshende:
#### Changes proposed in this pull request:

Previously MenuDivider did not handle escape key to close popover. This should fix that.

Issue #6054, opened 2023-04-03
MenuDivider does not handle escape key to close popover


adidahiya:
#### Environment

- __Package version(s)__: core v4.17.8, popover2 v1.13.12

#### Code Sandbox

https://codesandbox.io/p/sandbox/recursing-wave-ldgezp

#### Steps to reproduce

1. Render a Menu with MenuItem2 and MenuDivider items inside a popover
2. Make the MenuItem2 components inactive (`disabled={true}` or `shouldDismissPopover={true}`)
3. Open the popover
4. Click on a MenuItem2, then press the escape key
5. Open the popover
6. Click on a MenuDivider, then press the escape key

#### Actual behavior

![2023-04-03 19 18 34](https://user-images.githubusercontent.com/723999/229647703-b8f14bd5-7c03-4e74-b074-2539b4f516c2.gif)

Popover closed after step 4
Popover stayed open after step 6

#### Expected behavior

Popover should close after step 4, and also after step 6

#### Possible solution

Seems related to https://github.com/palantir/blueprint/issues/5947, where the suggestion of adding a `tabIndex={0}` to an element which should be focusable (and therefore handle the "escape key to close overlay" behavior) seems to work. Should be an easy fix to `MenuDivider.tsx`, probably just need to add `tabIndex={0}` here: https://github.com/palantir/blueprint/blob/6d86a19b9e858c4cde66a253b1172dcf896872d8/packages/core/src/components/menu/menuDivider.tsx#L54

bvandrc:
> Should be an easy fix to MenuDivider.tsx, probably just need to add tabIndex={0} here

`tabIndex` > -1 should only be applied to interactive elements, which the divider and its text are not. Definitely should not be tabb-able.

A better solution that accomplishes what you want would be to add `onKeyDown={(event) => event.key === 'Escape' && closePopover()}`

adidahiya:
@bvandercar-vt that's a good point, I don't think we should use `tabIndex={0}`.

But the alternative solution gets tricky when you want to implement `closePopover()` as you suggested. We'd have to either:
- create new custom DOM event semantics, like a Blueprint-specific "popover close" event which is bubbled up through the DOM
- or create a new Popover2Context (this would require creating MenuDivider2 which knows about Popover2) which allows open / close context "actions" on the parent Popover2

Both of those require a bit of work which I'm not sure is worth the cost for this small feature.

I think a reasonable compromise would be to use `tabIndex={-1}`, which still solves the problem:

[...]

PR #6457, opened 2023-10-11
[core] fix(MenuDivider): remove tabIndex attr to resolve a11y failure

bvandrc:
Does not need `tabIndex={-1}`, `role="separator"` already makes it non-focusable. Having any `tabIndex` (even -1) causes this to fail aXe a11y tests, so we must remove it.

### DIFF
--- a/packages/core/src/components/menu/menuDivider.tsx
+++ b/packages/core/src/components/menu/menuDivider.tsx
@@ -47,7 +47,7 @@ export class MenuDivider extends React.Component<MenuDividerProps> {
         } else {
             // section header with title
             return (
-                <li className={classNames(Classes.MENU_HEADER, className)} role="separator" tabIndex={-1}>
+                <li className={classNames(Classes.MENU_HEADER, className)} role="separator">
                     <H6 id={titleId}>{title}</H6>
                 </li>
             );


adidahiya:
This `tabIndex` was added to fix a bug, you can see that via git blame: https://github.com/palantir/blueprint/pull/6062

Can you fix the a11y issue without causing a regression re: https://github.com/palantir/blueprint/issues/6054?

edit: in fact we even discussed this here: https://github.com/palantir/blueprint/issues/6054#issuecomment-1513501190

bvandrc:
> This `tabIndex` was added to fix a bug, you can see that via git blame: #6062
> 
> Can you fix the a11y issue without causing a regression re: #6054?
> 
> edit: in fact we even discussed this here: [#6054 (comment)](https://github.com/palantir/blueprint/issues/6054#issuecomment-1513501190) 

Lol, I totally forgot I got involved in the discussion over this change... just to revert my decision later.

I do see that setting `tabIndex={0}` was the solution used for solving the #6054 bug, but I'm not sure _why_ that worked, and I'm not sure if it was the ideal approach to solving that problem.

Regardless, it appears something has changed in the overall code between bug #6054 and now that has fixed the core of the problem— I cannot reproduce that bug, with or without my changes in this PR, and even if I remove the `tabIndex`. So, maybe that fix occurred between bp4 and bp5? Maybe it was fixed in `Popover` at some point? Either way, it's good news— it means that the core of the popover escape problem was solved, and we don't need the `tabIndex` anymore.

Here is a code sandbox from the latest blueprint codebase to show you: https://codesandbox.io/p/sandbox/jovial-matan-4w98xm?file=%2Fsrc%2FCoreExample.tsx%3A10%2C48 . As you can see, there is no `tabIndex` on the `separator`, and the bug no longer occurs.

adidahiya (approved):
ok, great, thanks for that code sandbox @bvandercar-vt. It looks like it is safe to remove this attribute.

PR #6504, opened 2023-11-01
fix(MenuItem): a11y fix - disabled has aria-disabled

bvandrc:
Add `aria-disabled` prop to disabled `MenuItem`. Improves a11y and fixes a11y aXe failure.


bvandrc:

PR #6505, opened 2023-11-01
fix(MultiStepDialog): a11y fix - disabled step tab has aria-disabled

bvandrc:
Add aria-disabled prop to disabled `Tab`. Improves a11y and fixes a11y aXe failure.


bvandrc:

PR #6506, opened 2023-11-03
[core] feat(Section): a11y improvements, titleRenderer prop

bvandrc:
Improve a11y on the `Section` component by adding various `aria` props.

- [X] Includes tests


packages/core/src/components/section/section.tsx
```diff
@@ -158,38 +161,36 @@ export const Section: React.FC<SectionProps> = React.forwardRef((props, ref) =>
...
                     aria-pressed={collapsible ? isCollapsed : undefined}
+                    aria-expanded={collapsible ? isCollapsed : undefined}
+                    aria-controls={sectionId}
                     className={classNames(Classes.SECTION_HEADER, {
                         [Classes.INTERACTIVE]: collapsible,
                     })}
-                    onClick={collapsible != null ? toggleIsCollapsed : undefined}
+                    onClick={collapsible ? toggleIsCollapsed : undefined}
                 >
-                    {isHeaderLeftContainerVisible && (
```
bvandrc, line 172: `isHeaderLeftContainerVisible` is not needed, does nothing, because of line 169 `title &&` , `isHeaderLeftContainerVisible` is always true

packages/core/src/components/section/section.tsx
```diff
@@ -110,6 +111,11 @@ export interface SectionProps extends Props, Omit<HTMLDivProps, "title">, React.
      * Note that the header will only be rendered if `title` is provided.
      */
     title?: JSX.Element | string;
+
+    /**
+     * @default H6
+     */
+    titleTagName?: keyof JSX.IntrinsicElements | React.FC<React.AllHTMLAttributes<HTMLElement>>;
```
adidahiya: can we limit this to only `HTMLHeadingElement` function components, so that users are encouraged to use Blueprint's heading components? also I think the JSDoc needs to be improved

```suggestion
    /**
     * The heading component to use when rendering the title, typically one of `H1`, `H2,
     * `H3`, `H4`, `H5`, or `H6` from `@blueprintjs/core`.
     *
     * @default H6
     */
    titleHeadingComponent?: React.FC<React.AllHTMLAttributes<HTMLHeadingElement>>;
```

packages/core/src/components/section/section.tsx
```diff
@@ -158,38 +167,37 @@ export const Section: React.FC<SectionProps> = React.forwardRef((props, ref) =>
             })}
             elevation={elevation}
             ref={ref}
+            role="region"
```
adidahiya: I appreciate the attempt to provide better markup out-of-the-box, but after looking at the [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/region_role) for this aria role, I don't think it's appropriate to add it by default to _every_ Section component. This component is too generic for that, its purpose is largely for styling rather than functionality. So we should leave it up to users to specify `role="region"` via `{...htmlProps}` if they want.

```suggestion
```

packages/core/src/components/section/section.tsx
```diff
@@ -147,9 +154,11 @@ export const Section: React.FC<SectionProps> = React.forwardRef((props, ref) =>
         }
     }, [collapseProps, isCollapsed, isControlled]);
 
-    const isHeaderLeftContainerVisible = title != null || icon != null || subtitle != null;
     const isHeaderRightContainerVisible = rightElement != null || collapsible;
 
+    const sectionId = uniqueId("section");
+    const sectionTitleId = title ? uniqueId("section-title") : undefined;
```
adidahiya, line 164:
```suggestion
    const sectionTitleId = title != null ? uniqueId("section-title") : undefined;
```

packages/core/src/components/section/section.tsx
```diff
@@ -158,38 +167,37 @@ export const Section: React.FC<SectionProps> = React.forwardRef((props, ref) =>
...
             {title && (
                 <div
                     role={collapsible ? "button" : undefined}
                     aria-pressed={collapsible ? isCollapsed : undefined}
+                    aria-expanded={collapsible ? isCollapsed : undefined}
+                    aria-controls={collapsible ? sectionId : undefined}
                     className={classNames(Classes.SECTION_HEADER, {
                         [Classes.INTERACTIVE]: collapsible,
                     })}
-                    onClick={collapsible != null ? toggleIsCollapsed : undefined}
```
adidahiya, line 170: good catch 😅

adidahiya (changes requested):

packages/core/src/components/section/section.tsx
bvandrc: Actually, the purpose for this change, for our application at least, is specifically to _not_ have it be inside of a heading element. This is so that we can pass an element to the `title` prop that has a heading element _within_ it. In our case, here is our goal:
```javascript
title={
      <div>
        <H2 id={titleId}>{title}</H2>
        {titleRightElement}
      </div>
    }
```
in which case we'd want to set `titleTagName` to `div`, since it will _contain_ the H2.

So I see 3 options for us to achieve our goal above:

1. We implement `titleHeadingComponent` the way you said, and then add a prop `titleRightElement`.
2. We implement `titleHeadingComponent` the way you said, and on our end we just add a ts ignore so that we can pass a `div` component instead of a heading component.
3. We scrap this change and add a `titleProps` prop which gets spread to the H6 so that we can pass our `div` component to `title` and set `titleProps={{ role: 'presentation'}}`

packages/core/src/components/section/section.tsx
bvandrc: done

packages/core/src/components/section/section.tsx
bvandrc: Line 174 has logic
```javascript
 {title && (
                <div
```
so we would want this logic to be the same.

bvandrc: We could change that to `title != null` too though. Depends whether you want empty string to create the title element.

packages/core/src/components/section/section.tsx
bvandercar-vt: Third option: we scrap this change and add a `titleProps` prop which gets spread to the `H6`

adidahiya: Ok, how about a render prop API instead?


```ts
type SectionTitleProps = Pick<React.HTMLAttributes<HTMLElement>, "className" | "id">;
```

```ts
/**
 * Optional title renderer function. It is recommended to render a Blueprint `<H6>` element as part of the title.
 *
 * @default H6
 */
renderTitle?: React.FC<SectionTitleProps>;
```

Note: this type is informed by a recent fix we had to make for a similar render prop API in https://github.com/palantir/blueprint/pull/6521

bvandrc: Alright I take back my last comment, your first solution would work. Kept that solution. Renamed to `renderTitle` instead of `titleHeadingComponent` though. Up for whatever you want.

packages/core/src/components/section/section.tsx
adidahiya: ok, good point, let's keep what you have

packages/core/src/components/section/section.tsx
adidahiya: upon further review, I decided that `titleRenderer` is a better name. I pushed a commit with that update

adidahiya (approved):

adidahiya:
Sorry about the unrelated `lint` job error in CI; I will fix that on develop.

Your [docs preview build](https://output.circle-artifacts.com/output/job/f6aa80f7-2914-4d9d-88ea-b21e964621bc/artifacts/0/packages/docs-app/dist/index.html#core/components/section) looks good

image

PR #6507, opened 2023-11-03
fix(Section): remove rightElement from within section header to fix a11y

bvandrc:
Remove `rightElement` from within section header in order to fix a11y failure of having nested interactive elements (when the `rightElement` contains interactive elements, such as with the docs example).


packages/core/src/components/section/section.tsx
```diff
@@ -190,17 +189,29 @@ export const Section: React.FC<SectionProps> = React.forwardRef((props, ref) =>
...
+                            <ChevronDown className={Classes.TEXT_MUTED} />
+                        ) : (
+                            <ChevronUp className={Classes.TEXT_MUTED} />
+                        ))}
+                </div>
+            )}
+
+            {rightElement && (
+                // Needs to be outside of the section header because section header may be a button (when collapsable), and
+                // interactive elements (that may be rightElement) can't be inside other buttons.
```
adidahiya, line 214: > interactive elements (that may be rightElement) can't be inside other buttons

do you have a documentation reference for this?

adidahiya (commented):
how important is this fix? Is it really worth the added complexity of all this markup and styling?

bvandrc:
@adidahiya Rebased.

packages/core/src/components/section/section.tsx
```diff
@@ -201,17 +199,30 @@ export const Section: React.FC<SectionProps> = React.forwardRef((props, ref) =>
...
+                // Use an overlay to position rightElement absolutely in the header.
+                <div className={Classes.SECTION_HEADER_OVERLAY}>
```
nerdstep, line 216: I'm not sure this change is the right solution to the problem. 

Let's say you have a `collapsible` component with an interactive `rightElement` -- with this change the `rightElement` would be rendered to the right of the collapsible chevron icon, which visually would look a little strange, IMO.

packages/core/src/components/section/section.tsx
adidahiya: Still waiting on an answer to this question, and my overall question in my last review @bvandercar-vt

packages/core/src/components/section/section.tsx
evansjohnson: I'm not sure this is the case because of the right margin and absolute positioning (perhaps the diff has changed since this comment)

packages/core/src/components/section/section.tsx
evansjohnson: @bvandercar-vt do you have a good reference for this? a quick search turned up https://accessibilityinsights.io/info-examples/web/nested-interactive/ but not sure how reliable that is/I have not tested myself

packages/core/src/components/section/section.tsx
```diff
@@ -201,17 +199,30 @@ export const Section: React.FC<SectionProps> = React.forwardRef((props, ref) =>
...
+                    <div
+                        className={Classes.SECTION_HEADER_RIGHT}
+                        style={{
+                            // 25px accounts for the collapse chevron
+                            marginRight: collapsible ? 25 : undefined,
```
evansjohnson, line 221: Could there be any issues with this element overlapping other elements (I think just title/subtitle)? It's possible we'd need to do something to visually take up the space.

packages/core/src/components/section/section.tsx
evansjohnson: I'm also not sure this applies here because the header is just a div with an `onClick` event and a `button` role.

I opened a related issue yesterday: https://github.com/palantir/blueprint/issues/6872

evansjohnson (changes requested):
Note that `title` and `subtitle` could also be any arbitrary element, possibly interactive.

I think we may want to align this PR with a fix for https://github.com/palantir/blueprint/issues/6872. I imagine we leave the full header area only interactive through the div `onClick`, and provide tab access/action through only the chevron. This should have full support for keyboard users while avoiding the nested interactive problem, but allowing the nested interactive elements for mouse users as it seems is the intent of the design.

PR #6565, opened 2023-11-26
fix(a11y): popover target aria-expanded

bvandrc:
Add `aria-expanded` prop to popover target. Improve a11y.


adidahiya (approved):

adidahiya:
[`aria-expanded` MDN docs](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-expanded) for reference. It looks like this is compatible with "popup" elements and `aria-haspopup` 👍🏽

PR #6568, opened 2023-11-26
fix(a11y): TimePicker, DateInput a11y fixes

bvandrc:
Basing a11y structure off of this example: https://www.w3.org/WAI/ARIA/apg/patterns/combobox/examples/combobox-datepicker/

### DIFF
--- a/packages/core/src/components/section/section.tsx
+++ b/packages/core/src/components/section/section.tsx
@@ -115,7 +115,7 @@ export interface SectionProps extends Props, Omit<HTMLDivProps, "title">, React.
     /**
      * Optional title renderer function. If provided, it is recommended to include a Blueprint `<H6>` element
      * as part of the title. The render function is supplied with `className` and `id` attributes which you must
-     * forward to the DOM.
+     * forward to the DOM, as well as the `children` which is the passed `title`.
      *
      * @default H6
      */
--- a/packages/core/src/components/section/section.tsx
+++ b/packages/core/src/components/section/section.tsx
@@ -115,7 +115,7 @@ export interface SectionProps extends Props, Omit<HTMLDivProps, "title">, React.
     /**
      * Optional title renderer function. If provided, it is recommended to include a Blueprint `<H6>` element
      * as part of the title. The render function is supplied with `className` and `id` attributes which you must
-     * forward to the DOM, as well as the `children` which is the passed `title`.
+     * forward to the DOM.
      *
      * @default H6
      */
--- a/packages/datetime/src/components/date-input/dateInput.tsx
+++ b/packages/datetime/src/components/date-input/dateInput.tsx
@@ -23,7 +23,6 @@
 /* eslint-disable deprecation/deprecation, @blueprintjs/no-deprecated-components */
 
 import classNames from "classnames";
-import { uniqueId } from "lodash";
 import * as React from "react";
 import type { DayPickerProps } from "react-day-picker";
 
@@ -246,7 +245,7 @@ export const DateInput: React.FC<DateInputProps> = React.memo(function _DateInpu
 
     const inputRef = React.useRef<HTMLInputElement | null>(null);
     const popoverContentRef = React.useRef<HTMLDivElement | null>(null);
-    const popoverId = uniqueId("date-picker");
+    const popoverId = Utils.uniqueId("date-picker");
 
     // State
     // ------------------------------------------------------------------------
--- a/packages/datetime/src/components/date-input/dateInput.tsx
+++ b/packages/datetime/src/components/date-input/dateInput.tsx
@@ -23,6 +23,7 @@
 /* eslint-disable deprecation/deprecation, @blueprintjs/no-deprecated-components */
 
 import classNames from "classnames";
+import { uniqueId } from "lodash";
 import * as React from "react";
 import type { DayPickerProps } from "react-day-picker";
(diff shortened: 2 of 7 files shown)


packages/datetime/src/components/date-input/dateInput.tsx
```diff
@@ -245,6 +246,7 @@ export const DateInput: React.FC<DateInputProps> = React.memo(function _DateInpu
 
     const inputRef = React.useRef<HTMLInputElement | null>(null);
     const popoverContentRef = React.useRef<HTMLDivElement | null>(null);
+    const popoverId = uniqueId("date-picker");
```
adidahiya: can you use `Utils.uniqueId()` instead? `import { Utils} from "@blueprintjs/core"`

packages/datetime/src/components/time-picker/timePicker.tsx
```diff
@@ -197,6 +195,7 @@ export class TimePicker extends React.Component<TimePickerProps, TimePickerState
                 onKeyDown={this.getInputKeyDownHandler(unit)}
                 onKeyUp={this.getInputKeyUpHandler(unit)}
                 role={this.props.showArrowButtons ? "spinbutton" : undefined}
+                type="number"
```
adidahiya, line 197: there's a reason why we did not use this attribute, as you can see this change broke the test suite. you'll need to revert this part

adidahiya (changes requested):

packages/datetime/src/components/time-picker/timePicker.tsx
bvandrc: I think that test can just be modified with this change. The test is essentially "if entry is not a number, display error", however setting `type=number` disallows non-number entry in the first place and solves this case / problem from the start.

Is there a reason to not have it be `type=number` when that's exactly what it is?

Tests updated: https://github.com/palantir/blueprint/commit/abe73d10ddb38d485631059a69a63bb6da102f21

bvandercar-vt: Tests updated!  https://github.com/palantir/blueprint/pull/6568/commits/abe73d10ddb38d485631059a69a63bb6da102f21

bvandrc: In fact, if we use `type=number`,  we can just use the `min` and `max` props instead of `aria-valuemin` and `aria-valuemax`

packages/core/src/components/section/section.tsx
```diff
@@ -115,7 +115,7 @@ export interface SectionProps extends Props, Omit<HTMLDivProps, "title">, React.
     /**
      * Optional title renderer function. If provided, it is recommended to include a Blueprint `<H6>` element
      * as part of the title. The render function is supplied with `className` and `id` attributes which you must
-     * forward to the DOM.
+     * forward to the DOM, as well as the `children` which is the passed `title`.
```
adidahiya: please remove this unrelated change from this PR. You can fix this JSDoc in a separate PR.

also the wording is a bit awkward, I would say this instead:

```suggestion
     * forward to the DOM. The `title` prop is also passed along to this renderer via `props.children`.
```

packages/datetime/src/components/time-picker/timePicker.tsx
adidahiya: I tried to dig up the original reason why we didn't use `type="number"` when the component was originally written 8 years ago I and couldn't find a compelling one. Perhaps the browser support was incomplete, or there wasn't good support for doing the style overrides you've added in your recent commits. I suppose we can switch to `type="number"` now, as it does seem like the right markup to use in the long run. I tested it out in the [docs preview build](https://output.circle-artifacts.com/output/job/2d566d4e-bcfd-4966-b768-af49e578f300/artifacts/0/packages/docs-app/dist/index.html#datetime/timepicker) and things seem to be working fine :+1:

packages/datetime/test/components/timePickerTests.tsx
```diff
@@ -144,16 +144,27 @@ describe("<TimePicker>", () => {
         hourInput.value = "2";
         TestUtils.Simulate.change(hourInput);
         assert.strictEqual(hourInput.value, "2");
+        assert.isFalse(hourInput.classList.contains(CoreClasses.intentClass(Intent.DANGER)));
     });
 
-    it("allows invalid text entry, but shows visual indicator", () => {
+    it("disallows allows non-number text entry", () => {
```
adidahiya: I think there's a typo:

```suggestion
    it("disallows non-number text entry", () => {
```

adidahiya (changes requested):

packages/core/src/components/section/section.tsx
bvandrc: Done

packages/datetime/test/components/timePickerTests.tsx
bvandrc: Fixed.

adidahiya (approved):
changes lgtm, can you please run `yarn lint-fix` inside `packages/datetime/`?

bvandrc:
> changes lgtm, can you please run `yarn lint-fix` inside `packages/datetime/`?

@adidahiya  Done.

adidahiya (approved):

PR #6578, opened 2023-11-30
Update default button background color to white

CPerinet:
Before
Screenshot 2023-11-30 at 18 08 56

After
Screenshot 2023-11-30 at 18 08 52

### DIFF
--- a/packages/core/src/components/button/_common.scss
+++ b/packages/core/src/components/button/_common.scss
@@ -37,8 +37,8 @@ $dark-button-box-shadow-active:
   0 1px 2px rgba($black, 0.4) !default;
 
 $button-color-disabled: $pt-text-color-disabled !default;
-$button-background-color: $light-gray5 !default;
-$button-background-color-hover: $light-gray4 !default;
+$button-background-color: $white !default;
+$button-background-color-hover: $light-gray5 !default;
 $button-background-color-active: $light-gray2 !default;
 $button-background-color-disabled: rgba($light-gray1, 0.5) !default;
 $button-background-color-active-disabled: rgba($light-gray1, 0.7) !default;


adidahiya:
<h3>Update default button background color to white</h3>
Build artifact links for this commit: <strong><a href="https://output.circle-artifacts.com/output/job/20266ac5-1d84-4f0f-aba5-2d89279ab651/artifacts/0/packages/docs-app/dist/index.html">documentation</a> | <a href="https://output.circle-artifacts.com/output/job/20266ac5-1d84-4f0f-aba5-2d89279ab651/artifacts/0/packages/landing-app/dist/index.html">landing</a> | <a href="https://output.circle-artifacts.com/output/job/20266ac5-1d84-4f0f-aba5-2d89279ab651/artifacts/0/packages/table-dev-app/dist/index.html">table</a> | <a href="https://output.circle-artifacts.com/output/job/20266ac5-1d84-4f0f-aba5-2d89279ab651/artifacts/0/packages/demo-app/dist/index.html">demo</a></strong>

adidahiya:
Hmm, I'm not sure I agree with this change. The default button now looks _very_ similar to outlined buttons:

| Default | Outlined |
| - | - |
| image | image |

When two options look very similar, it begs the question of _why are these both available?_ And IMO it would be better if the design system was more opinionated and only allowed one of them, or kept them separate as more visually distinct options. If we do move in this direction, then I think `outlined` would have to be deprecated. But personally, I like the light gray background of the default button...

CPerinet:
On a white background, default and outlined when no intent will look quite similar indeed. That said, on any other background they would look more different — we could remove the outlined but I'm unsure if that'd be worth the effort given all other intent would remain different

nerdstep:
As an alternate idea, what if buttons (and maybe other components) could support a "theme color" instead of, or in addition to, the intent colors. In other words, why can't we have a violet button? :)

FWIW, we've implemented this for our project on some components:

image

image

The default is still gray, but you can also use a white or black version if you want, or any of the other colors.

image

PR #6599, opened 2023-12-11
Add `OverlayToaster.createAsync` method to support React 18

### LINKED ISSUE #6239: Update OverlayToaster component to use ReactDOM.createRoot() for React 18 compatibility
#### Environment

- __Package version(s)__: 5.0.0
- __Operating System__: MacOS
- __Browser name and version__: Chrome 114.0.5735.133

#### Steps to reproduce

Use the OverlayToaster component in a React 18 application.
Observe the console warnings about ReactDOM.render() being deprecated.

#### Actual behavior

The OverlayToaster component in BlueprintJS uses the ReactDOM.render() method, which is deprecated in React 18. This causes the application to behave as if it's running React 17, according to the warning message from React.

#### Expected behavior

The OverlayToaster component should use the new ReactDOM.createRoot() method introduced in React 18, so that it can take advantage of the new features and improvements in React 18.

#### Possible solution

Update the OverlayToaster component to use ReactDOM.createRoot() instead of ReactDOM.render(). This might involve refactoring the component to accommodate the new root API, which involves creating a root and then calling render on it, as opposed to the legacy API where render is called directly with the component and container.

gluxon:
## PR Stack

- https://github.com/palantir/blueprint/pull/6601
- https://github.com/palantir/blueprint/pull/6599

## Checklist

- [x] Includes tests
- [x] Update documentation

## Changes

Fixes #6239. A new `OverlayToaster.createAsync()` utility is added alongside `OverlayToaster.create()`.

This asynchronous function allows a custom `domRenderer` utility to be passed to better support codebases on React 18.

```tsx
import { OverlayToaster } from "@blueprintjs/core";
import { createRoot } from "react-dom/client";

const toaster = await OverlayToaster.createAsync(/* props */ {}, {
  domRenderer: (toaster, containerElement) => createRoot(containerElement).render(toaster),
});

toaster.show({ message: "Hello React 18!" })
```

The existing `OverlayToaster.create()` method uses `ReactDOM.render` without any ability to customize this call, which [results in a console warning when upgrading to React 18](https://react.dev/blog/2022/03/08/react-18-upgrade-guide#updates-to-client-rendering-apis).

Screenshot 2023-12-11 at 5 38 23 PM

#### Why is the new API asynchronous?

The change to Blueprint's API reflects a change in React's API. The new `createRoot` function from `react-dom/client` no longer renders components synchronously.

```ts
import { createRoot } from "react-dom/client";

const toaster = React.createRef<OverlayToaster>();

createRoot(containerElement)
  .render(<OverlayToaster {...props} ref={toaster} usePortal={false} />)
  
// OverlayToaster render function is not yet executed.

// ‼️ This is null ‼️
ref.current

setTimeout(() => {
  // Okay, now the OverlayToaster render() function has ran.
  ref.current // now available
}, 0)
```

This is different than `ReactDOM.render`, which does synchronously populate the `ref`.

```tsx
import * as ReactDOM from "react-dom";

const toaster = React.createRef<OverlayToaster>();

ReactDOM.render(
  <OverlayToaster {...props} ref={toaster} usePortal={false} />,
  containerElement
);

// OverlayToaster render function has ran by this point.

ref.current // instance of <OverlayToaster />
```

This seems to be an intentional change in React 18. See the “_What about the render callback?_” section under [Replacing render with createRoot](https://github.com/reactwg/react-18/discussions/5#top).

#### Why do I need to pass in a custom `domRenderer` value for React 18?

The `createRoot` function is an import on `react-dom/client`. Blueprint is an NPM library that needs to support multiple bundlers (e.g. Webpack, Vite). Most bundlers will fail if it encounters a non-existent submodule import. Since the current major version of Blueprint needs to support React 16 to 18, imports into 18-specific code paths can't be made directly in Blueprint.

As a workaround, consumers of Blueprint can provide the `createRoot` function to Blueprint. This is an application of [dependency injection](https://en.wikipedia.org/wiki/Dependency_injection).

When Blueprint drops support for React 16 in a future major version, the `domRenderer` option will change its current default from `ReactDOM.render` to a function using the new `createRoot` API. This breaking change will make `OverlayToaster.createAsync` easier to use in the future.

#### How do I migrate?

The most one-to-one conversion would be from:

```tsx
OverlayToaster.create().show({ message: "Hello!" });
```

to:

```tsx
OverlayToaster.createAsync().then(toaster => toaster.show({ message: "Hello!" }));
```

You may notice:

1. We're swallowing errors that happen when rendering the component. There's no rejection handler on the promise object.
2. The `OverlayToaster` component is never unmounted or removed from the DOM after the message is dismissed.

Both of these are true of the original synchronous `OverlayToaster.create()` API. The new API makes these existing problems more apparent. We recommend [mounting the toaster directly to your React application tree](https://blueprintjs.com/docs/#core/components/toast.react-component-usage) if the first problem is a concern, and [sharing the Toaster to avoid the second problem](https://blueprintjs.com/docs/#core/components/toast.example).

- Whether the new API makes sense.
- Reviewing each commit individually should be easier than reviewing all changes at once. Hiding whitespace only changes should also make the diff easier to look at since a few tests were nested.
- Several refactors had to be made to the `OverlayToaster` tests to make them reusable for both `OverlayToaster.create` and `OverlayToaster.createAsync`.
- I briefly spent some time attempting to add a React 18 dependency for testing, but hit issues with NPM aliases discovering `@types/react-dom@18`. I can take another stab if we think a React 18 test is valuable to commit, but it's a bit hard to have multiple version of React in the same codebase, even for testing.

### DIFF
--- a/packages/core/src/common/errors.ts
+++ b/packages/core/src/common/errors.ts
@@ -93,6 +93,10 @@ export const TOASTER_CREATE_NULL =
     ns +
     ` OverlayToaster.create() is not supported inside React lifecycle methods in React 16.` +
     ` See usage example on the docs site.`;
+export const TOASTER_CREATE_ASYNC_NULL =
+    ns +
+    ` OverlayToaster.createAsync() received a null component ref. This can happen if called inside React lifecycle ` +
+    `methods in React 16. See usage example on the docs site.`;
 export const TOASTER_MAX_TOASTS_INVALID =
     ns + ` <OverlayToaster> maxToasts is set to an invalid number, must be greater than 0`;
 export const TOASTER_WARN_INLINE =
--- a/packages/core/src/common/errors.ts
+++ b/packages/core/src/common/errors.ts
@@ -92,11 +92,12 @@ export const SPINNER_WARN_CLASSES_SIZE = ns + ` <Spinner> Classes.SMALL/LARGE ar
 export const TOASTER_CREATE_NULL =
     ns +
     ` OverlayToaster.create() is not supported inside React lifecycle methods in React 16.` +
-    ` See usage example on the docs site.`;
+    ` See usage example on the docs site. https://blueprintjs.com/docs/#core/components/toast.example`;
 export const TOASTER_CREATE_ASYNC_NULL =
     ns +
     ` OverlayToaster.createAsync() received a null component ref. This can happen if called inside React lifecycle ` +
-    `methods in React 16. See usage example on the docs site.`;
+    `methods in React 16. See usage example on the docs site. ` +
+    `https://blueprintjs.com/docs/#core/components/toast.example`;
 export const TOASTER_MAX_TOASTS_INVALID =
     ns + ` <OverlayToaster> maxToasts is set to an invalid number, must be greater than 0`;
 export const TOASTER_WARN_INLINE =
--- a/packages/core/src/components/toast/toast.md
+++ b/packages/core/src/components/toast/toast.md
@@ -36,6 +36,8 @@ There are three ways to use __OverlayToaster__:
     myToaster.show({ ...toastOptions });
     ```
 
+    We recommend calling `OverlayToaster.createAsync` once in your application and [sharing the created instance](#core/components/toast.example) throughout your application.
+
     A synchronous `OverlayToaster.create()` static method is also available, but will be phased out since React 18+ no longer synchronously renders components to the DOM.
 
     ```ts
@@ -115,6 +117,19 @@ function synchronousFn() {
 
 Note that `OverlayToaster.createAsync()` will throw an error if invoked inside a component lifecycle method, as
 `ReactDOM.render()` will return `null` resulting in an inaccessible toaster instance.
+
+<div class="@ns-callout @ns-intent-primary @ns-icon-info-sign @ns-callout-has-body-content">
+    <h5 class="@ns-heading">Beware of memory leaks</h5>
+
+The static `createAsync` and `create` methods create a new `OverlayToaster` instance for the full lifetime of your
+application. Since there's no React parent component, these methods create a new DOM node as a container for the
+rendered `<OverlayToaster>` component. Every `createAsync` call will add a new DOM node. We do not recommend creating a
+new `Toaster` every time a toast needs to be shown. To minimize leaking:
+
+1. Call `OverlayToaster.createAsync` once in an application and [share the instance](#core/components/toast.example).
+2. Consider one of the alternative APIs that mount the `<OverlayToaster>` somewhere in the application's React component tree. This provides component lifecycle management out of the box. See [_React component usage_](#core/components/toast.react-component-usage) for an example.
+
+</div>
 
 @interface Toaster
 
--- a/packages/core/src/components/toast/toast.md
+++ b/packages/core/src/components/toast/toast.md
@@ -175,6 +175,26 @@ The example below uses the `OverlayToaster.createAsync()` static method. Clickin
 
 @reactExample ToastCreateAsyncExample
 
+#### React 18
+
+To maintain backwards compatibility with React 16 and 17, `OverlayToaster.createAsync` uses `ReactDOM.render` out of the box. This triggers a [console warning on React 18](https://react.dev/blog/2022/03/08/react-18-upgrade-guide#updates-to-client-rendering-apis). A future major version of Blueprint will drop support for React versions before 18 and switch the default rendering function from `ReactDOM.render` to `createRoot`.
+
+If you're using React 18, we recommend passing in a custom `domRenderer` function.
+
+```tsx
+import { OverlayToaster } from "@blueprintjs/core";
+import { createRoot } from "react-dom/client";
+
+const toaster = await OverlayToaster.createAsync(toasterProps, {
+    // Use createRoot() instead of ReactDOM.render(). This can be deleted after
+    // a future Blueprint version uses createRoot() for Toasters by default.
+    domRenderer: (toaster, containerElement) => createRoot(containerElement).render(toaster),
+});
+
+toaster.show({ message: "Hello React 18!" })
+```
+
+
 @## React component usage
 
 Render the `<OverlayToaster>` component like any other element and supply `<Toast>` elements as `children`. You can
--- a/packages/core/src/components/toast/toast.md
+++ b/packages/core/src/components/toast/toast.md
@@ -30,7 +30,14 @@ horizontally aligned along the left edge, center, or right edge of its container
 
 There are three ways to use __OverlayToaster__:
 
-1. __Recommended__: use the `OverlayToaster.create()` static method to create a new `Toaster` instance:
+1. __Recommended__: use the `OverlayToaster.createAsync()` static method to create a new `Toaster` instance:
+    ```ts
+    const myToaster: Toaster = await OverlayToaster.createAsync({ position: "bottom" });
+    myToaster.show({ ...toastOptions });
+    ```
+
+    A synchronous `OverlayToaster.create()` static method is also available, but will be phased out since React 18+ no longer synchronously renders components to the DOM.
+
     ```ts
     const myToaster: Toaster = OverlayToaster.create({ position: "bottom" });
     myToaster.show({ ...toastOptions });
--- a/packages/core/src/components/toast/toast.md
+++ b/packages/core/src/components/toast/toast.md
@@ -143,6 +143,10 @@ export class App extends React.PureComponent {
 }
 ```
 
+The example below uses the `OverlayToaster.createAsync()` static method. Clicking the button will create a new toaster mounted to `<body>`, show a message, and unmount the toaster from the DOM once the message is dismissed.
+
+@reactExample ToastCreateAsyncExample
+
 @## React component usage
 
 Render the `<OverlayToaster>` component like any other element and supply `<Toast>` elements as `children`. You can
--- a/packages/core/src/components/toast/toast.md
+++ b/packages/core/src/components/toast/toast.md
@@ -86,21 +86,34 @@ enable `autoFocus` for an individual `OverlayToaster` via a prop, if desired.
 
 @## Static usage
 
-__OverlayToaster__ provides the static `create` method that returns a new `Toaster`, rendered into an
+__OverlayToaster__ provides the static `createAsync` method that returns a new `Toaster`, rendered into an
 element attached to `<body>`. A toaster instance has a collection of methods to show and hide toasts in its given container.
 
 ```ts
-OverlayToaster.create(props?: ToasterProps, container = document.body): Toaster
+OverlayToaster.createAsync(props?: OverlayToasterProps, options?: OverlayToasterCreateOptions): Promise<Toaster>;
 ```
+
+@interface OverlayToasterCreateOptions
 
 The toaster will be rendered into a new element appended to the given `container`.
 The `container` determines which element toasts are positioned relative to; the default value of `<body>` allows them to use the entire viewport.
 
-Note that the return type is `Toaster`, which is a minimal interface that exposes only the instance
-methods detailed below. It can be thought of as `OverlayToaster` minus the `React.Component` methods,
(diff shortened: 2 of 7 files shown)


packages/core/test/toast/overlayToasterTests.tsx
```diff
@@ -29,122 +29,132 @@ describe("OverlayToaster", () => {
...
+    describe("create with default props", () => {
+        before(() => {
+            testsContainerElement = document.createElement("div");
+            document.documentElement.appendChild(testsContainerElement);
+            toaster = OverlayToaster.create({}, testsContainerElement);
+        });
 
-    afterEach(() => {
-        toaster.clear();
-        ReactDOM.unmountComponentAtNode(testsContainerElement);
```
gluxon: We were previously calling `ReactDOM.unmountComponentAtNode` on the wrong DOM element. The `OverlayToaster.create()` method creates a new `<div>` within `testsContainerElement` to render on.

packages/core/test/toast/overlayToasterTests.tsx
```diff
@@ -174,9 +174,21 @@ describe("OverlayToaster", () => {
...
 
-            it("does not exceed the maximum toast limit set", () => {
+        describe("with maxToasts set to finite value", () => {
+            before(() => {
+                testsContainerElement = document.createElement("div");
+                document.documentElement.appendChild(testsContainerElement);
                 toaster = spec.create({ maxToasts: 3 }, testsContainerElement);
+            });
+
+            after(() => {
```
gluxon, line 204: Without the additional cleanup in a few tests that call `OverlayToaster.create()` directly (instead of using the existing `Toaster` instance), I was seeing the "_does not attach toast container to body on script load_" test fail.

packages/core/src/components/toast/toast.md
```diff
@@ -30,7 +30,14 @@ horizontally aligned along the left edge, center, or right edge of its container
 
 There are three ways to use __OverlayToaster__:
 
-1. __Recommended__: use the `OverlayToaster.create()` static method to create a new `Toaster` instance:
+1. __Recommended__: use the `OverlayToaster.createAsync()` static method to create a new `Toaster` instance:
```
gluxon, line 33: I would actually swap the recommended approach here since `OverlayToaster.create()` and `OverlayToaster.createAsync()` both leak memory and add a DOM node to the page _forever_. Thoughts?

adidahiya:
<h3>Add OverlayToaster.createAsync test</h3>
Build artifact links for this commit: <strong><a href="https://output.circle-artifacts.com/output/job/fdcd8dda-2196-4156-968a-da4a1b4e31e1/artifacts/0/packages/docs-app/dist/index.html">documentation</a> | <a href="https://output.circle-artifacts.com/output/job/fdcd8dda-2196-4156-968a-da4a1b4e31e1/artifacts/0/packages/landing-app/dist/index.html">landing</a> | <a href="https://output.circle-artifacts.com/output/job/fdcd8dda-2196-4156-968a-da4a1b4e31e1/artifacts/0/packages/table-dev-app/dist/index.html">table</a> | <a href="https://output.circle-artifacts.com/output/job/fdcd8dda-2196-4156-968a-da4a1b4e31e1/artifacts/0/packages/demo-app/dist/index.html">demo</a></strong>

adidahiya:
<h3>Add OverlayToaster.createAsync test</h3>
Build artifact links for this commit: <strong><a href="https://output.circle-artifacts.com/output/job/ee93a99a-9f26-4e99-8227-9fa7579a4f54/artifacts/0/packages/docs-app/dist/index.html">documentation</a> | <a href="https://output.circle-artifacts.com/output/job/ee93a99a-9f26-4e99-8227-9fa7579a4f54/artifacts/0/packages/landing-app/dist/index.html">landing</a> | <a href="https://output.circle-artifacts.com/output/job/ee93a99a-9f26-4e99-8227-9fa7579a4f54/artifacts/0/packages/table-dev-app/dist/index.html">table</a> | <a href="https://output.circle-artifacts.com/output/job/ee93a99a-9f26-4e99-8227-9fa7579a4f54/artifacts/0/packages/demo-app/dist/index.html">demo</a></strong>

adidahiya:
<h3>Add OverlayToaster.createAsync test</h3>
Build artifact links for this commit: <strong><a href="https://output.circle-artifacts.com/output/job/7ba58d77-e938-41cd-8732-d0e6efc5bc80/artifacts/0/packages/docs-app/dist/index.html">documentation</a> | <a href="https://output.circle-artifacts.com/output/job/7ba58d77-e938-41cd-8732-d0e6efc5bc80/artifacts/0/packages/landing-app/dist/index.html">landing</a> | <a href="https://output.circle-artifacts.com/output/job/7ba58d77-e938-41cd-8732-d0e6efc5bc80/artifacts/0/packages/table-dev-app/dist/index.html">table</a> | <a href="https://output.circle-artifacts.com/output/job/7ba58d77-e938-41cd-8732-d0e6efc5bc80/artifacts/0/packages/demo-app/dist/index.html">demo</a></strong>

packages/core/src/components/toast/toast.md
adidahiya: Ok, sure, I'm down with that, we can change the recommended approach. So we'd reverse the order of approaches: 3 becomes 1 (recommended), and 1 becomes 3. Since it seems like most people prefer having an imperative API in their application to trigger toasts rather than storing a list of Toasts in app state and rendering `<Toast>` themselves.


It's worth adding more docs under the "Static usage" section around L105 with your note about memory leaks and leaving a DOM node on the page forever. Also perhaps update the code snippets there to use `createAsync` instead of `create`?

packages/core/src/common/errors.ts
```diff
@@ -93,6 +93,10 @@ export const TOASTER_CREATE_NULL =
     ns +
     ` OverlayToaster.create() is not supported inside React lifecycle methods in React 16.` +
     ` See usage example on the docs site.`;
+export const TOASTER_CREATE_ASYNC_NULL =
+    ns +
+    ` OverlayToaster.createAsync() received a null component ref. This can happen if called inside React lifecycle ` +
+    `methods in React 16. See usage example on the docs site.`;
```
adidahiya: nit: which documentation site? Blueprint or React? Might be useful to just add a link

adidahiya (changes requested):
thanks for improving the test suite, nice to see the amount of code re-use there 👍🏽 

code changes & behavior look good, I'm requesting some docs changes in this review

would you mind adding a new (small) docs example using `OverlayToaster.createAsync()` which gets shown in this part of the docs? it doesn't need to have options:

image

this will help us to have a quick & easy way to test the imperative API interactively in the docs preview. Currently, `toastExample.tsx` only uses the declarative API.

gluxon:
Thanks for all the feedback! Agree that the docs need to have some improvements. I'm still working through a few more clarifications and will move the PR to draft mode until I wrap those up.

adidahiya:
<h3>Move "React component usage example" further up</h3>
Build artifact links for this commit: <strong><a href="https://output.circle-artifacts.com/output/job/a08d7c43-0d25-40df-9a06-69dd2fcbfcf4/artifacts/0/packages/docs-app/dist/index.html">documentation</a> | <a href="https://output.circle-artifacts.com/output/job/a08d7c43-0d25-40df-9a06-69dd2fcbfcf4/artifacts/0/packages/landing-app/dist/index.html">landing</a> | <a href="https://output.circle-artifacts.com/output/job/a08d7c43-0d25-40df-9a06-69dd2fcbfcf4/artifacts/0/packages/table-dev-app/dist/index.html">table</a> | <a href="https://output.circle-artifacts.com/output/job/a08d7c43-0d25-40df-9a06-69dd2fcbfcf4/artifacts/0/packages/demo-app/dist/index.html">demo</a></strong>

packages/core/src/common/errors.ts
gluxon: Agree with it being useful to just add a link. This message was copied from the `TOASTER_CREATE_NULL`. I'll update the original message as well. https://github.com/palantir/blueprint/pull/6599/commits/4779ba20a8115ac9cea8cc19c49cba263e74b1ea

packages/core/src/components/toast/toast.md
gluxon: > Ok, sure, I'm down with that, we can change the recommended approach. So we'd reverse the order of approaches: 3 becomes 1 (recommended), and 1 becomes 3. Since it seems like most people prefer having an imperative API in their application to trigger toasts rather than storing a list of Toasts in app state and rendering `<Toast>` themselves.

I pushed up a few commits that switched 1 and 3, but decided to walk this back. The primary factor that changed my mind was that it becomes harder to pass the `OverlayToaster` ref down the React component tree. I think that will be a bigger point of confusion than I originally thought when I suggested switching the recommendations. Thanks for being open to my suggestion.

The latest push keeps the current recommendation since it may be the best of all options, but adds more notes around only creating one `Toaster` instance.

> It's worth adding more docs under the "Static usage" section around L105 with your note about memory leaks and leaving a DOM node on the page forever.

Can do! Added in https://github.com/palantir/blueprint/pull/6599/commits/d375696c1a25cbfa9db2cc772beb4816fce788a2.

> Also perhaps update the code snippets there to use `createAsync` instead of `create`?

Good suggestion. Pushed up a commit that switches the docs and code snippets to `createAsync` here: https://github.com/palantir/blueprint/pull/6599/commits/9b1fb9334162d6726ed0403e676de0f14afd398a

gluxon (commented):
> would you mind adding a new (small) docs example using OverlayToaster.createAsync() which gets shown in this part of the docs? it doesn't need to have options

Can do! Done in commit https://github.com/palantir/blueprint/pull/6599/commits/9bdf637917c0df5e9fbec5b00d8f2f4760d152ef.

I ended up moving this to the _Example_ section since the code snippets were very similar to what the interactive example implements. If that's not a good place, I can move it back to where you suggested.

gluxon:
This should be ready for another review after the holidays. I wanted to finish my in-progress changes before I lost context over the next few weeks. Please feel free to ignore this until next year. Thanks!

adidahiya:
<h3>Add React 18 domRenderer example</h3>
Build artifact links for this commit: <strong><a href="https://output.circle-artifacts.com/output/job/34bd249a-5e35-41d2-b32a-9aac17370717/artifacts/0/packages/docs-app/dist/index.html">documentation</a> | <a href="https://output.circle-artifacts.com/output/job/34bd249a-5e35-41d2-b32a-9aac17370717/artifacts/0/packages/landing-app/dist/index.html">landing</a> | <a href="https://output.circle-artifacts.com/output/job/34bd249a-5e35-41d2-b32a-9aac17370717/artifacts/0/packages/table-dev-app/dist/index.html">table</a> | <a href="https://output.circle-artifacts.com/output/job/34bd249a-5e35-41d2-b32a-9aac17370717/artifacts/0/packages/demo-app/dist/index.html">demo</a></strong>

packages/core/src/components/toast/toast.md
```diff
@@ -79,23 +88,49 @@ enable `autoFocus` for an individual `OverlayToaster` via a prop, if desired.
...
+
+The static `createAsync` and `create` methods create a new `OverlayToaster` instance for the full lifetime of your
+application. Since there's no React parent component, these methods create a new DOM node as a container for the
+rendered `<OverlayToaster>` component. Every `createAsync` call will add a new DOM node. We do not recommend creating a
+new `Toaster` every time a toast needs to be shown. To minimize leaking:
+
+1. Call `OverlayToaster.createAsync` once in an application and [share the instance](#core/components/toast.example).
+2. Consider one of the alternative APIs that mount the `<OverlayToaster>` somewhere in the application's React component tree. This provides component lifecycle management out of the box. See [_React component usage_](#core/components/toast.react-component-usage) for an example.
+
+</div>
```
adidahiya, line 132: these docs look good 👍🏽

adidahiya (approved):
new example works well in docs preview

packages/core/src/components/toast/toast.md
gluxon: Thanks!

PR #6603, opened 2023-12-15
style: props objects satisfies React.HtmlProps

bvandrc:
Just a style change, adds `satisfies` for objects that are html props, that previously were un-typed.


adidahiya (approved):
nice one, TIL about [`satisfies`](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-9.html#the-satisfies-operator) 💯

PR #6604, opened 2023-12-15
fix(Popover): a11y aria props applied to child target, not wrapper

bvandrc:
`aria-expanded` and `aria-haspopover`  should be applied to the _interactive_ element, which is the passed _child target_, not the span _wrapper_. Compared to the span wrapper, the child element is the interactive element because A) it is likely to be an interactive element such as a `<Button>` and B) it's what gets the `tabIndex` prop.


bvandercar-vt:
@adidahiya I do not think the current test failure is related to the changes in this PR.

bvandrc:
@adidahiya  passing now

adidahiya (approved):
there is some risk to this change as it touches a very fundamental component, but I think it's the right direction to move in. We had to make a similar attribute change in one of our apps due to a11y checks failing on the popover target element, so it would be great if Popover fixed this across the board for everyone by applying these `aria-` attributes on the correct DOM element.

Issue #6241, opened 2023-06-25
Unable to run dev environment on Windows


UsmanMuhammad:
#### Environment

- __Package version(s)__: 5.0.0-alpha.0
- __Operating System__: Windows 11
- __Browser name and version__: Chrome V114.0.5735.134

#### Steps to reproduce

1. Clone the repo
2. Run Yarn
3. Run Yarn Verify

While trying to install `windows-build-tools`, it says "Node.js now includes build tools for Windows. You probably no longer need this tool." and the command fails.
#### Actual behavior

Doesn't compile one package.
Error:
 1/15 targets failed, including the following:
         - @blueprintjs/docs-data:compile

[docs-data] ERROR when generating JSON docs data:
file:///C:/Users/path/blueprint/packages/docs-data/compile-docs-data.mjs:32

#### Expected behavior

Should compile all packages and run the app

#### Possible solution

No Idea

adidahiya:
@UsmanMuhammad there may be issues developing Blueprint from source in a Windows environment. Unfortunately we don't have the bandwidth to test & maintain all our development tools in Windows, and we rely on community support to get it working when there are issues. If you are able to fix this dev task in your environment, please share your solution and/or send a PR :)

PR #6611, opened 2023-12-18
Feature: Add Prop to NonIdealState to Disable Muted Icon Style

AlexKempen:
#### Introduction

This PR adds a prop `mutedIcon` to `NonIdealState` which can be used to explicitly opt-out of the muted style for the `icon`. This makes it possible to, for example, use an `<Icon>` component with an `intent` specified alongside `NonIdealState`, which is currently impossible since `Classes.NON_IDEAL_STATE_VISUAL` always overrides the `intent` (even when passing an explicit `<Icon>` JSX element). 

#### Checklist

- [x] Includes tests
- [x] Update documentation

This PR adds a `mutedIcon` prop to `NonIdealState`. `mutedIcon` defaults to `true` to preserve backwards-comparability. When set to `false`, `Classes.NON_IDEAL_STATE_VISUAL` is no longer applied to the `icon` prop, which can prevent issues with certain props getting overridden in an undesirable (and unexpected) way. 

#### Screenshot
Example of a `NonIdealState` with a custom `<Icon>` and `intent`:
fix

#### First Time Contributor Notes

As a first time contributor on this repo, I'll briefly discuss my experience getting the code to run and build. I first tried on my Windows 10 machine, which I ultimately gave up on. The primary issues I encountered there were:

- `windows-build-tools` refused to install for me. The primary suggestion I saw while troubleshooting was that it is now natively available with Node as an option during install. I'm not sure if that worked for me or not.
- Issues with `C:` paths being disallowed. The suggestion I saw while troubleshooting was to use a later version of Node (the post said they saw it resolved going from Node 12 to Node 16), but it was nevertheless occurring with the project's Node 18.

Since I couldn't find any fixes for the above issues, I eventually gave up and switched to Ubuntu in WSL, which worked much better. My main feedback there is that ChromeHeadless is needed to run `yarn test`, which isn't mentioned in the documentation (as far as I can tell). These commands resolved it for me:
```
wget https://dl.google.com/linux/direct/google-chrome-stable_current_amd64.deb
sudo apt install ./google-chrome-stable_current_amd64.deb
```

### DIFF
--- a/packages/core/src/components/non-ideal-state/nonIdealState.tsx
+++ b/packages/core/src/components/non-ideal-state/nonIdealState.tsx
@@ -46,7 +46,7 @@ export interface NonIdealStateProps extends Props {
      */
     description?: React.ReactChild;
 
-    /** The name of a Blueprint icon or a JSX Element (such as `<Spinner/>`) to render above the title. */
+    /** The name of a Blueprint icon or a JSX element (such as `<Spinner/>`) to render above the title. */
     icon?: IconName | MaybeElement;
 
     /**
@@ -55,6 +55,13 @@ export interface NonIdealStateProps extends Props {
      * @default NonIdealStateIconSize.STANDARD
      */
     iconSize?: NonIdealStateIconSize;
+
+    /**
+     * Whether the `icon` should attempt to use a muted appearence.
+     *
+     * @default true
+     */
+    mutedIcon?: boolean;
 
     /**
      * Component layout, either vertical or horizontal.
@@ -78,6 +85,7 @@ export class NonIdealState extends AbstractPureComponent<NonIdealStateProps> {
     public static defaultProps: Partial<NonIdealStateProps> = {
         iconSize: NonIdealStateIconSize.STANDARD,
         layout: "vertical",
+        mutedIcon: true,
     };
 
     public render() {
@@ -94,13 +102,13 @@ export class NonIdealState extends AbstractPureComponent<NonIdealStateProps> {
     }
 
     private maybeRenderVisual() {
-        const { icon, iconSize } = this.props;
+        const { icon, iconSize, mutedIcon } = this.props;
         if (icon == null) {
             return undefined;
         } else {
             return (
                 <div
-                    className={Classes.NON_IDEAL_STATE_VISUAL}
+                    className={classNames(mutedIcon && Classes.NON_IDEAL_STATE_VISUAL)}
                     style={{ fontSize: `${iconSize}px`, lineHeight: `${iconSize}px` }}
                 >
                     <Icon icon={icon} size={iconSize} aria-hidden={true} tabIndex={-1} />
--- a/packages/core/src/components/non-ideal-state/nonIdealState.tsx
+++ b/packages/core/src/components/non-ideal-state/nonIdealState.tsx
@@ -111,7 +111,13 @@ export class NonIdealState extends AbstractPureComponent<NonIdealStateProps> {
                     className={Classes.NON_IDEAL_STATE_VISUAL}
                     style={{ fontSize: `${iconSize}px`, lineHeight: `${iconSize}px` }}
                 >
-                    <Icon className={classNames({ [Classes.ICON_MUTED]: iconMuted })} icon={icon} size={iconSize} aria-hidden={true} tabIndex={-1} />
+                    <Icon
+                        className={classNames({ [Classes.ICON_MUTED]: iconMuted })}
+                        icon={icon}
+                        size={iconSize}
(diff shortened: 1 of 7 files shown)


palantirtech:
Thanks for your interest in [palantir/blueprint](https://github.com/palantir/blueprint), @AlexKempen! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

packages/core/src/components/non-ideal-state/nonIdealState.tsx
```diff
@@ -94,13 +102,13 @@ export class NonIdealState extends AbstractPureComponent<NonIdealStateProps> {
...
     private maybeRenderVisual() {
-        const { icon, iconSize } = this.props;
+        const { icon, iconSize, mutedIcon } = this.props;
         if (icon == null) {
             return undefined;
         } else {
             return (
                 <div
-                    className={Classes.NON_IDEAL_STATE_VISUAL}
+                    className={classNames(mutedIcon && Classes.NON_IDEAL_STATE_VISUAL)}
```
adidahiya: Let's not remove the class in this PR, as I think it is still semantically valuable to be able to target the "visual" element in the DOM here. Instead, let's adjust the styles to target a new class applied to the `<Icon>` element below.

packages/core/src/components/non-ideal-state/nonIdealState.tsx
```diff
@@ -94,13 +102,13 @@ export class NonIdealState extends AbstractPureComponent<NonIdealStateProps> {
...
                     style={{ fontSize: `${iconSize}px`, lineHeight: `${iconSize}px` }}
                 >
                     <Icon icon={icon} size={iconSize} aria-hidden={true} tabIndex={-1} />
```
adidahiya:
```suggestion
                    <Icon className={classNames({ [Classes.ICON_MUTED]: iconMuted })} icon={icon} size={iconSize} aria-hidden={true} tabIndex={-1} />
```

packages/core/src/components/non-ideal-state/nonIdealState.tsx
```diff
@@ -56,6 +56,13 @@ export interface NonIdealStateProps extends Props {
      */
     iconSize?: NonIdealStateIconSize;
 
+    /**
+     * Whether the `icon` should attempt to use a muted style.
+     *
+     * @default true
+     */
+    mutedIcon?: boolean;
```
adidahiya: nit: let's call this `iconMuted` to get the prop visually grouped with the other icon-related props in alphabetical sorting

```suggestion
    iconMuted?: boolean;
```

packages/core/src/components/non-ideal-state/nonIdealState.tsx
adidahiya:
```suggestion
                    className={Classes.NON_IDEAL_STATE_VISUAL}
```

adidahiya (changes requested):
Thanks for the PR and your notes about the contribution process @AlexKempen. I'm aware of development issues in Windows environments, we haven't had the bandwidth to fix them but we're open to PRs (see #6241)

As for the contents of this PR: I'd suggest moving the muted styling to an "icon modifier" class - see comments below. That would mean that `_non-ideal-state.scss` changes to:

```scss

.#{$ns}-non-ideal-state-visual {
  color: $gray3;
}
```

and the removed styles move to `_icon.scss`:

```scss
#{$ns}-icon {
  // ... existing styles

  &.#{$ns}-icon-muted svg {
    fill-opacity: 15%;
    // need to show overflow for some strokes on paths that reach the edge of the icon bounding box
    overflow: visible;

    path {
      stroke: $gray3;
      stroke-opacity: 50%;
      stroke-width: 0.5px;
    }
  }

  .#{$ns}-dark & {
    .#{$ns}-icon-muted svg {
      fill-opacity: 20%;
    }
  }
}
```

and we get a new string constant for the modifier class in `classes.ts`:

```ts
export const ICON_MUTED = `${ICON}-muted`;
```

AlexKempen:
@adidahiya ping since it's been a little while - is there anything else you'd like me to do with this?

adidahiya (approved):
looks great, thanks for making all those adjustments @AlexKempen!

the new prop looks good in [docs preview](https://output.circle-artifacts.com/output/job/2b8fbb9c-e753-4ce5-b428-0e0ccf1c86dd/artifacts/0/packages/docs-app/dist/index.html#core/components/non-ideal-state), and I was able to toggle it to `true` using React dev tools:

image

PR #6621, opened 2024-01-03
[core] feat(Tabs): allow panel prop to be a renderer

bvandrc:
feat(Tabs): allow `panel` prop to be a renderer, pass tabTitleId and tabPanelId to the renderer.

Reason: a developer may want to use these IDs for accessibility, such as `aria-controls`.

#### Checklist

- [x] Includes tests
- [x] Update documentation


packages/core/test/tabs/tabsTests.tsx
```diff
@@ -128,6 +129,21 @@ describe("<Tabs>", () => {
...
     });
 
+    it("passes correct tabTitleId and tabPanelId to panel renderer", () => {
+        mount(
+            <Tabs id={ID}>
+                <Tab
+                    id="first"
+                    panel={({ tabTitleId, tabPanelId }) => {
+                        assert.equal(tabTitleId, generateTabTitleId(ID, "first"));
+                        assert.equal(tabPanelId, generateTabPanelId(ID, "first"));
```
adidahiya, line 139: are you sure these assertions are getting run in this test? I think you need to use the Mocha async `done` syntax.

adidahiya (commented):
seems like a reasonable change. thanks for adding a unit test. just one question about the test...

packages/core/test/tabs/tabsTests.tsx
bvandrc: Yeah I was a little confused on how to write this test, was curious how you would do it.

> are you sure these assertions are getting run in this test?

Yep, it does work. If I change the expected value, I see a failure.

adidahiya (approved):

PR #6644, opened 2024-01-10
a11y(Tooltip): "tooltip" role

bvandrc:
#### Checklist

- [N/A] Includes tests
- [N/A] Update documentation

Improve tooltip a11y by adding the following:
- apply `role="tooltip"` to the Tooltip's `content` - https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/tooltip_role
- apply an `id` to the tooltip content, and `aria-describedby={id}` to the target


packages/core/src/components/tooltip/tooltip.tsx
```diff
@@ -134,6 +152,13 @@ export class Tooltip<
                     },
                 }}
                 {...restProps}
+                renderTarget={renderTarget ? props => renderTarget({ ...props, tooltipId }) : undefined}
+                content={
+                    // want popover validation warnings to appear
+                    isContentEmpty(content)
+                        ? content
+                        : Utils.ensureElement(content, "div", { role: "tooltip", id: tooltipId })
```
bvandercar-vt, line 160: or instead of this, could add a `contentProps` prop to `Popover`?

bvandrc:
Closing so that can create other PRS

PR #6649, opened 2024-01-11
feat(Popover): spread child props

bvandrc:
Spread child props to allow overriding of `aria-expanded`

### DIFF
--- a/packages/core/src/components/popover/popover.tsx
+++ b/packages/core/src/components/popover/popover.tsx
@@ -386,7 +386,6 @@ export class Popover<
             ...targetEventHandlers,
         } satisfies React.HTMLProps<HTMLElement>;
         const childTargetProps = {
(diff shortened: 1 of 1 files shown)


adidahiya:
This change sounds reasonable, but I'm curious why you need to override `aria-expanded`? When does the automatic value applied by the component not work?

bvandrc:
> This change sounds reasonable, but I'm curious why you need to override `aria-expanded`? When does the automatic value applied by the component not work?

`aria-expanded` causes a11y failure if it's applied to an item that doesn't have an interactive `role`. For example, if you apply `aria-expanded` to a `span` that doesn't have `role` of `button`, `input`, `menuitem`, etc.

However I think we can delete this PR because the item _is_ interactive so someone _should_ be applying a role to pass a11y.

PR #6670, opened 2024-01-23
Sanitize docs content to prevent XSS

gtsp233:
## Fix for Cross-Site Scripting (XSS) Vulnerability

Hi, I've found a Cross-Site Scripting (XSS) vulnerability in the package @icedesign/richtext-renderer.

**Vulnerability Details:**
- **Severity**: High/Critical
- **Description**: There's a risk of malicious content is passed to docs

#### POC(Proof of Concept)
```
import React, { useEffect } from "react";
import { Documentation } from "@blueprintjs/docs-theme";

const App = () => {
  return (
    <Documentation
      docs={{
        nav: [],
        pages: {
          test: {
            reference: "foo",
            route: "foo",
            sourcePath: "foo",
            title: "foo",
            contents: ["<img src='' onerror=alert(1)>"],
          },
        },
        docs: {
          pages: [],
        },
      }}
      defaultPageId="foo"
    />
  );
};

export default App;

```

Sanitize the HTML before rendering it using dangerouslySetInnerHtml

I've already fixed this issue, and have submitted a pull request with the necessary changes. Please review and merge my pull request to resolve this vulnerability. Thanks!

### DIFF
--- a/packages/docs-theme/package.json
+++ b/packages/docs-theme/package.json
@@ -41,7 +41,8 @@
         "@documentalist/client": "^5.0.0",
         "classnames": "^2.3.1",
         "fuzzaldrin-plus": "^0.6.0",
-        "tslib": "~2.6.2"
+        "tslib": "~2.6.2",
+        "dompurify": "^3.0.8"
     },
     "peerDependencies": {
         "@types/react": "^16.14.41 || 17 || 18",
--- a/packages/docs-theme/package.json
+++ b/packages/docs-theme/package.json
@@ -40,9 +40,9 @@
         "@blueprintjs/select": "workspace:^",
         "@documentalist/client": "^5.0.0",
         "classnames": "^2.3.1",
+        "dompurify": "^3.0.8",
         "fuzzaldrin-plus": "^0.6.0",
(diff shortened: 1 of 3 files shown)


palantirtech:
Thanks for your interest in [palantir/blueprint](https://github.com/palantir/blueprint), @gtsp233! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

adidahiya:
This is a developer build tool that is not exposed directly to end users (only developers building documentation sites). Why does this sanitization need to happen in the Blueprint package? If it's a concern for a downstream consumer, then perhaps @icedesign/richtext-renderer should do the sanitization?

ericjeney (commented):
Thanks for filing, @gtsp233!

However, I agree with @adidahiya's inclination that the Blueprint docs theme is only intended to be used with trusted input, and this sanitization should be performed by the consumer if the input cannot be trusted.

If we did want to fix this within Blueprint, I think we'd have to fix a few other places as well (such as the `dangerouslySetInnerHTML`'s in `css.tsx`).

PR #6691, opened 2024-02-01
feat(SegmentedControl): a11y - set role to radiogroup

bvandrc:
Improves a11y of new SegmentedControl item, giving it the `radiogroup` role and also allowing it to be controlled with arrow keys.

- [X] Added Tests


adidahiya:
Various open source UI toolkits have different approaches for Segmented Control a11y:

- [Primer](https://primer.style/components/segmented-control#accessibility) _explicitly_ advises against using `radiogroup`, `tablist`, or `toolbar` ARIA roles
- [MUI's Toggle Button Group](https://mui.com/material-ui/react-toggle-button/) uses `role="group"`
- [Mantine UI](https://mantine.dev/core/segmented-control/) uses `role="radiogroup"`

Can you explain the motivations for your change and make an argument for setting the default as you've done in this PR?

I think a safer, better approach might be to add good documentation to the component docs suggesting that users should set their own ARIA role appropriately by specifying the `role` attribute.

bvandrc:
@adidahiya  that is tricky. After reading through those, I agree my original choice of `listbox` was incorrect.

It sounds like `radiogroup` is the best option, I changed the code to that. Despite what that Primer link says, I see no evidence why it shouldn't be `radiogroup` when that's exactly how it behaves.

If you would like me to add a `role` prop-- `role?="group" | "list" | "radiogroup"`-- and then we set the children roles and aria props based on that as well, I'm up for it. But IMO, I don't think anyone would complain about it being hard set to `radiogroup`

bvandercar-vt:
@adidahiya seems like this has gone stale? Please see my last comment.

evansjohnson (changes requested):
> why it shouldn't be radiogroup when that's exactly how it behaves

Right now the individual buttons are focusable, not the whole element with focus moved between with arrow keys as in a radio group. The non-checked buttons should have `tabIndex={-1}` and we should hook up arrow key listeners to move between the options if we want this to behave like a radio group. 

I too do not see the argument against treating this like a radio group [for the reason of](https://primer.style/components/segmented-control#accessibility) "which would require a save button to apply the changes" - I see no reference to this in [this w3 example](https://www.w3.org/WAI/ARIA/apg/patterns/radio/examples/radio-rating/) or the [mdn docs](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/radiogroup_role).

---

The [SegmentedControl docs](https://blueprintjs.com/docs/#core/components/segmented-control) do say "similar to a [Radio](https://blueprintjs.com/docs/#core/components/radio) group" so this seems worth doing to me, as long as we hook up [the expected interaction behavior](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/radiogroup_role#keyboard_interactions) as well.

bvandrc:
@evansjohnson Done, added arrow key control and the proper `tabIndex` setting! Tested, works great.

packages/core/src/common/utils/keyboardUtils.ts
```diff
@@ -25,3 +25,21 @@ export function isKeyboardClick(event: React.KeyboardEvent<HTMLElement>) {
...
+
+/**
+ * Direction multiplier for component such as radiogroup, tablist
+ *
+ * @param upMovesLeft
+ * If true, up arrow returns same as left arrow, down arrow returns same as right arrow.
+ * If false, down arrow returns same as left arrow, up arrow returns same as right arrow.
+ * @returns -1 for left, 1 for right, undefined if not an arrow keypress.
+ */
+export function getArrowKeyDirection(event: React.KeyboardEvent<HTMLElement>, upMovesLeft: boolean = true) {
```
evansjohnson: `upMovesLeft` is interesting - are we currently following expected behavior correctly? it definitely makes sense in the slider case for up to move right but I'm wondering if up should always be to the right

this feels like something that may change based on locale, possibly in addition to component usage

not sure we need to figure this out now since this keeps existing behavior but leaving this comment for posterity at least

packages/core/src/components/segmented-control/segmentedControl.tsx
```diff
@@ -92,34 +94,72 @@ export const SegmentedControl: React.FC<SegmentedControlProps> = React.forwardRe
         value: controlledValue,
         ...htmlProps
     } = props;
+    const [selectedIndex, setSelectedIndex] = React.useState<number>(
```
evansjohnson: we should keep this code more like it was before - where we use this local state index only if using the `defaultValue` option to avoid duplicating state, then on every render reconcile the index of the user provided `controlledValue` with this local index as a fallback

evansjohnson (commented):
this is looking good - some last comments and then will leave to @ggdouglas for approval since he'll be owning the next BP release

packages/core/src/common/utils/keyboardUtils.ts
bvandrc: - `Slider`: Definitely makes sense for `up = right`.

- `Tabs`: the [official aria tabs example](https://www.w3.org/WAI/ARIA/apg/patterns/tabs/examples/tabs-manual/) doesn't have any effect from the up/down arrow keys, only left/right. So if anything, we should remove the effect of an up/down keypress at some point. But with this PR, the existing implementation goes unchanged.

- `SegmentedControl`: since we are giving this the `radiogroup` role, [here in the official example](https://www.w3.org/WAI/ARIA/apg/patterns/radio/examples/radio-rating/) of this role you can see that `up = left`. The action description is given as well:
image

packages/core/src/components/segmented-control/segmentedControl.tsx
bvandrc: Hmm it's tricky because we want to add keypress control with this PR, which is much easier to change to the next index vs. the next value. I can take a shot at it though.

bvandrc: Done! Reduced the diff a ton, mostly the same code as was before, but arrow keys work as indended.

bvandrc:
@ggdouglas this should be ready for final review!

packages/core/src/components/segmented-control/segmentedControl.tsx
```diff
@@ -104,14 +108,48 @@ export const SegmentedControl: React.FC<SegmentedControlProps> = React.forwardRe
...
+        (e: React.KeyboardEvent<HTMLDivElement>) => {
+            const direction = getArrowKeyDirection(e, true);
+            const { current: outerElement } = outerRef;
+            if (direction == undefined || !outerElement) return;
+
+            const focusedElement = Utils.getActiveElement(outerElement)?.closest<HTMLButtonElement>("button");
+            if (!focusedElement) return;
+
+            // must rely on DOM state because we have no way of mapping `focusedElement` to a React.JSX.Element
+            const enabledOptionElements = Array.from(outerElement.querySelectorAll("button")).filter(
```
ggdouglas: Could we leverage the selector query to filter out disabled buttons? e.g.
```suggestion
            const enabledOptionElements = Array.from(outerElement.querySelectorAll("button:not(:disabled)"));
```

ggdouglas (commented):
I think this looks good! Will approve after linter errors are fixed.

packages/core/src/components/segmented-control/segmentedControl.tsx
bvandrc: Good idea, done!

bvandrc:
@ggdouglas made some additional changes actually-- added a `role` prop so consumers can use a different aria role for the segmented control if they wish, since there seems to be some general debate on what the proper role is.

bvandrc:
@ggdouglas the lint error is actually unrelated to this PR, it's happening in my other PRs as well. Here's the fix: https://github.com/palantir/blueprint/pull/6972

ggdouglas (approved):

PR #6696, opened 2024-02-02
[icons] feat: "binary-number", "curly-braces"

waynecai1213:

Add "binary number" and "curly braces" icons.

#### Screenshot

image image


packages/icons/icons.json
```diff
@@ -4184,5 +4184,20 @@
...
+        "codepoint": 62295
+    },
+    {
+        "displayName": "Curly braces",
+        "iconName": "curly-braces",
+        "tags": "data, type, format，punctuation, grammar",
+        "group": "miscellaneous",
+        "codepoint": 62296
     }
+
```
adidahiya: this empty line is failing prettier formatting checks

I recommend enabling the Prettier extension for VSCode, that will prevent issues like this from being committed if you format your files on save.

```suggestion
```

packages/icons/icons.json
```diff
@@ -4184,5 +4184,20 @@
         "tags": "footnote, notebook, note, reference, text, citation",
         "group": "editor",
         "codepoint": 62294
+    },
+    {
+        "displayName": "Binary",
+        "iconName": "binary",
```
adidahiya: I think the names should be more specific here, it's not immediately obvious that we're dealing with numbers. Unlike the `"floating-point"` icon, in this case there is potential ambiguity when dealing with things like binary trees, etc.

```suggestion
        "displayName": "Binary number",
        "iconName": "binary-number",
```

packages/icons/icons.json
```diff
@@ -4184,5 +4184,20 @@
...
+        "tags": "data, type, number, numeric, format",
+        "group": "miscellaneous",
```
adidahiya: I think this new icon belongs in the "data" group.

```suggestion
        "group": "data",
```

existing icons like `"floating-point"` also should move to the "data" group, but that can happen in a separate PR

packages/icons/icons.json
```diff
@@ -4184,5 +4184,20 @@
...
+        "codepoint": 62295
+    },
+    {
+        "displayName": "Curly braces",
+        "iconName": "curly-braces",
+        "tags": "data, type, format，punctuation, grammar",
```
adidahiya, line 4198: can you explain the connection with "data" for this icon?

What is this icon going to be used for?

does it belong in the "editor" group of icons?

adidahiya (changes requested):

packages/icons/icons.json
waynecai1213: We want to use this icon for the "Struct" data type in Pipeline Builder.

resources/icons/20px/binary.svg
adidahiya, line 1: you need to update the icon file names to match the updated icon name

waynecai1213: Updated! thank you!!

adidahiya:
both sizes look good in [docs preview](https://output.circle-artifacts.com/output/job/c0786c51-465c-441b-a8b8-88abb404a297/artifacts/0/packages/docs-app/dist/index.html#icons/icons-list)

image

image

adidahiya (approved):

PR #6697, opened 2024-02-03
Prevent scrolling when a MenuItem is selected with the spacebar

ericjeney:
This fixes a small bug that caused the page to scroll when a user selected a `<MenuItem>` using the spacebar, which is visible on the Menu [docs page](https://blueprintjs.com/docs/#core/components/menu).

### DIFF
--- a/packages/core/src/common/utils/domUtils.ts
+++ b/packages/core/src/common/utils/domUtils.ts
@@ -152,7 +152,11 @@ function throttleImpl<T extends Function>(
 export function clickElementOnKeyPress(keys: string[]) {
     return (e: React.KeyboardEvent) => {
         if (keys.some(key => e.key === key)) {
-            e.preventDefault(); // Prevent spacebar from scrolling the page
+            // Prevent spacebar from scrolling the page unless we're in a text field
+            if (!elementIsTextInput(e.target as HTMLElement)) {
+                e.preventDefault();
+            }
+
             e.target.dispatchEvent(new MouseEvent("click", { ...e, view: undefined }));
         }
     };
--- a/packages/core/src/common/utils/domUtils.ts
+++ b/packages/core/src/common/utils/domUtils.ts
@@ -150,6 +150,10 @@ function throttleImpl<T extends Function>(
 }
 
 export function clickElementOnKeyPress(keys: string[]) {
-    return (e: React.KeyboardEvent) =>
-        keys.some(key => e.key === key) && e.target.dispatchEvent(new MouseEvent("click", { ...e, view: undefined }));
+    return (e: React.KeyboardEvent) => {
+        if (keys.some(key => e.key === key)) {
+            e.preventDefault(); // Prevent spacebar from scrolling the page
+            e.target.dispatchEvent(new MouseEvent("click", { ...e, view: undefined }));
+        }
+    };
 }
--- a/packages/docs-app/src/examples/core-examples/menuExample.tsx
+++ b/packages/docs-app/src/examples/core-examples/menuExample.tsx
@@ -16,13 +16,14 @@
 
 import * as React from "react";
 
-import { Classes, H5, Icon, Menu, MenuDivider, MenuItem } from "@blueprintjs/core";
+import { Classes, H5, Icon, InputGroup, Menu, MenuDivider, MenuItem } from "@blueprintjs/core";
 import { Example, type ExampleProps } from "@blueprintjs/docs-theme";
 
 import { getSizeProp, type Size, SizeSelect } from "./common/sizeSelect";
 
 export function MenuExample(props: ExampleProps) {
     const [size, setSize] = React.useState<Size>("regular");
+    const [count, setCount] = React.useState<number>(0);
 
     const options = (
         <>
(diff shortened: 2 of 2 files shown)


adidahiya:
<h3>Prevent scroll in clickElementOnKeyPress</h3>
Build artifact links for this commit: <strong><a href="https://output.circle-artifacts.com/output/job/d09b38fc-4e53-4719-aa1c-bfed01892cc2/artifacts/0/packages/docs-app/dist/index.html">documentation</a> | <a href="https://output.circle-artifacts.com/output/job/d09b38fc-4e53-4719-aa1c-bfed01892cc2/artifacts/0/packages/landing-app/dist/index.html">landing</a> | <a href="https://output.circle-artifacts.com/output/job/d09b38fc-4e53-4719-aa1c-bfed01892cc2/artifacts/0/packages/table-dev-app/dist/index.html">table</a> | <a href="https://output.circle-artifacts.com/output/job/d09b38fc-4e53-4719-aa1c-bfed01892cc2/artifacts/0/packages/demo-app/dist/index.html">demo</a></strong>

packages/core/src/common/utils/domUtils.ts
```diff
@@ -150,6 +150,10 @@ function throttleImpl<T extends Function>(
 }
 
 export function clickElementOnKeyPress(keys: string[]) {
-    return (e: React.KeyboardEvent) =>
-        keys.some(key => e.key === key) && e.target.dispatchEvent(new MouseEvent("click", { ...e, view: undefined }));
+    return (e: React.KeyboardEvent) => {
+        if (keys.some(key => e.key === key)) {
+            e.preventDefault(); // Prevent spacebar from scrolling the page
```
adidahiya: this could possibly cause issues if a MenuItem has a text input inside it, I think. it's not really recommended, but I wouldn't be surprised if something like this exists in the wild:

https://codesandbox.io/p/devbox/distracted-cookies-sv5mlx

image


we should test / verify this behavior in the docs by adding an item which contains an `<InputGroup>` to the [Menu docs example](https://blueprintjs.com/docs/#core/components/menu.menu-item)

one idea for preventing a regression here is to check if the keypress occurred in a text input, using another utility fn in this same module:

```ts
if (elementIsTextInput(e.target as HTMLElement)) {
  // don't prevent default
}
```

adidahiya (changes requested):

packages/core/src/common/utils/domUtils.ts
ericjeney: Ah, I didn't consider someone could be nesting a text element in a `MenuItem`. I can add a check and docs example for that.

Though if that were the case today, it would be issuing a click on the `MenuItem` any time they hit the spacebar. I assume we'd want to keep that behavior to avoid any breaks?

packages/core/src/common/utils/domUtils.ts
```diff
@@ -150,6 +150,14 @@ function throttleImpl<T extends Function>(
 }
 
 export function clickElementOnKeyPress(keys: string[]) {
-    return (e: React.KeyboardEvent) =>
-        keys.some(key => e.key === key) && e.target.dispatchEvent(new MouseEvent("click", { ...e, view: undefined }));
+    return (e: React.KeyboardEvent) => {
+        if (keys.some(key => e.key === key)) {
+            // Prevent spacebar from scrolling the page unless we're in a text field
+            if (!elementIsTextInput(e.target as HTMLElement)) {
```
ericjeney, line 156: This will allow the space character to enter the `TextInput`, but will still click the element. That's a bit odd, but technically prevents a break here.

packages/docs-app/src/examples/core-examples/menuExample.tsx
```diff
@@ -39,6 +40,12 @@ export function MenuExample(props: ExampleProps) {
                 <MenuItem icon="new-object" text="New object" />
                 <MenuItem icon="new-link" text="New link" />
                 <MenuDivider />
+                <MenuItem
```
ericjeney, line 43: We didn't have any menu examples that performed an action on click previously.

adidahiya:
<h3>Add more Menu docs examples</h3>
Build artifact links for this commit: <strong><a href="https://output.circle-artifacts.com/output/job/6d38e246-0333-4fa7-a84f-2a21690f71b9/artifacts/0/packages/docs-app/dist/index.html">documentation</a> | <a href="https://output.circle-artifacts.com/output/job/6d38e246-0333-4fa7-a84f-2a21690f71b9/artifacts/0/packages/landing-app/dist/index.html">landing</a> | <a href="https://output.circle-artifacts.com/output/job/6d38e246-0333-4fa7-a84f-2a21690f71b9/artifacts/0/packages/table-dev-app/dist/index.html">table</a> | <a href="https://output.circle-artifacts.com/output/job/6d38e246-0333-4fa7-a84f-2a21690f71b9/artifacts/0/packages/demo-app/dist/index.html">demo</a></strong>

packages/core/src/common/utils/domUtils.ts
adidahiya: > it would be issuing a click on the MenuItem any time they hit the spacebar

Yeah I suppose so, but they might be doing their own `preventDefault` to work around this. Agree that we should preserve the existing behavior as much as possible to avoid surprises

adidahiya (approved):
nice, I like the improved docs example, it works nicely when I hit spacebar

image

PR #6713, opened 2024-02-16
PanelStack2: Prevent extra render due to updated previous panel

YoruNoHikage:
Hi,

I had a `TextArea` component with a jumping cursor from the moment I updated to Blueprint 5 from 4.

This component is located in a second panel within a `PanelStack2`  but the first one had changing props at every render.
When digging through the differences between v4 and v5, I found out that `handleClose` has been added to the `useMemo` dependencies of `PanelWrapper` triggering a render every time `previousPanel` changes.

So for now, to go around this, I wrap the previous panel with a `useMemo` as well.

#### Checklist

- [ ] Includes tests
- [ ] Update documentation

Extracting `previousPanel !== undefined` outside the callback prevent extra renders by using a boolean instead.

### DIFF
--- a/packages/core/src/components/panel-stack2/panelView2.tsx
+++ b/packages/core/src/components/panel-stack2/panelView2.tsx
@@ -60,12 +60,13 @@ export const PanelView2: PanelView2Component = <T extends Panel<object>>({
     previousPanel,
     showHeader,
 }: PanelView2Props<T>) => {
+    const hasPreviousPanel = previousPanel !== undefined;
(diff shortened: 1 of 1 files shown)


palantirtech:
Thanks for your interest in [palantir/blueprint](https://github.com/palantir/blueprint), @YoruNoHikage! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

adidahiya (approved):

PR #6747, opened 2024-03-18
link to select component from multiselect docs

bryan-mclellan:
The previous link wasn't working for me and I believe it should redirect to the select component instead of the generic select docs. 

#### Checklist

- [ ] Includes tests
- [ ] Update documentation


ericjeney (approved):
Thank you!

PR #6751, opened 2024-03-26
chore: remove datetime package dependencies to lodash

PetrusAsikainen:
#### Checklist

- [x] Includes tests (well, seems to pass existing ones)
- [ ] Update documentation

I noticed that `@blueprintjs/datetime` depends on `lodash` for two extremely trivial use cases. I've changed those as such:

- `getCurrentTimezone`, a function without arguments, used `memoize`.

  I've replaced it with a global variable - while this is a new global `let`, the previous one was a mutable object stored in a `const`, so it's essentially the same.

  The function was also declared as `() => string`, meaning there was no official way to access its lodash cache property either.
- `timezoneUtils` used `isEmpty` on a variable declared as `string | null | undefined`, out of which `null` and `undefined` were already checked for. I replaced this with a simple manual check

Additionally, `@blueprintjs/datetime2` depended on `@types/lodash` without using it for anything, so I've removed it as well.

This effectively means you can build a Blueprint app without depending transitively on lodash.

Whether there are edge cases that might break from these changes.


packages/datetime/src/common/timezoneUtils.ts
```diff
@@ -69,7 +68,7 @@ export function getDateObjectFromIsoString(
     if (value === undefined) {
         return undefined;
     }
-    if (value === null || isEmpty(value)) {
+    if (value === null || value === "") {
```
ericjeney:
```suggestion
    if (value === null || value === "" || typeof value !== "string") {
```

Can you add a `typeof` check here to be extra defensive in the event a non-TypeScript consumer is passing in an incorrect value here (such as a number) and expecting to get `null` returned?

`isEmpty` evaluates to `true` on numbers, so I'd like to preserve that behavior to avoid an accidental break.

ericjeney (commented):
Thanks! I agree with removing the dep on `lodash`, especially since it was so light.

packages/datetime/src/common/timezoneUtils.ts
PetrusAsikainen: Seems reasonable. Interesting that here lodash.isEmpty very much *doesn't* do what it [says on the tin](https://lodash.com/docs/4.17.15#isEmpty):

>Checks if `value` is an empty object, collection, map, or set.

...which a number clearly isn't - handling for primitives isn't specified.

ericjeney (approved):

PR #6753, opened 2024-03-29
fix cleanup for overlay2 component

### LINKED ISSUE #6752: Dialog component does not cleanup the overlay's side effects on redirects without onClose
#### Environment

- __Package version(s)__: 5.7.2
- __Operating System__: MacOS 14.4 
- __Browser name and version__: Google Chrome Version 123.0.6312.59 (Official Build) (arm64)

#### Code Sandbox

Link to a minimal repro: [repro](https://codesandbox.io/p/devbox/loving-ioana-sxd3py?file=%2Fsrc%2Fpages%2FPageWithDialog.tsx%3A12%2C24)

#### Steps to reproduce

1. Open dialog
2. Click on the "Main page" href inside the dialog

#### Actual behavior

Main page has bp5-overlay-open class that prevents scroll across the app

#### Expected behavior

Scroll is still working

mkorabelskyi:
#### Checklist

- [x ] Includes tests
- [ ] Update documentation

TLDR `isOpen` is almost never `true` inside the useEffect callback.

Removed the if statement which was using an outdated state variable inside of the `useEffect` and prevented from cleanup on component unmount, which could happen if user forced the modal to close (e.g. navigating from the parent view). In this case `useEffect` doesn't have any dependencies and only returns a callback that will run on the component unmount, but since the `useEffect` doesn't have any dependencies, body of the useEffect won't be re-evaluated on the subsequent component re-renders. So the callback function will only keep in its scope the initial component's values (`isOpen` and `overlayWillClose` in this case). It might be okay to keep the initial value of the overlayWillClose function, but if the component was initialized with `isOpen === false` then the `overlayWillClose` function will never be called.  

Double checking for potential side-effects of the if statement removal

### DIFF
--- a/packages/core/src/components/overlay2/overlay2.tsx
+++ b/packages/core/src/components/overlay2/overlay2.tsx
@@ -328,11 +328,10 @@ export const Overlay2 = React.forwardRef<OverlayInstance, Overlay2Props>((props,
     // run once on unmount
     React.useEffect(() => {
         return () => {
-            if (isOpen) {
-                // if the overlay is still open, we need to run cleanup code to remove some event handlers
-                overlayWillClose();
-            }
+            // we need to run cleanup code to remove some event handlers from the overlay element
+            overlayWillClose();
         };
+
         // eslint-disable-next-line react-hooks/exhaustive-deps
     }, []);
 
--- a/packages/core/test/overlay2/overlay2Tests.tsx
+++ b/packages/core/test/overlay2/overlay2Tests.tsx
@@ -628,6 +628,47 @@ describe("<Overlay2>", () => {
             });
         });
 
+        describe("after closing by navigating out of the view", () => {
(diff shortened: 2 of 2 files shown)


palantirtech:
Thanks for your interest in [palantir/blueprint](https://github.com/palantir/blueprint), @maxpolski! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

ericjeney (approved):
Thank you for fixing, and for adding the test!

Your explanation makes sense, and running this cleanup code on every unmount mirrors the behavior of the original overlay component better as well:
https://github.com/palantir/blueprint/blob/bca2e0560effa78681bed1ae491f701996cf1737/packages/core/src/components/overlay/overlay.tsx#L180-L182

PR #6758, opened 2024-04-05
fix #6734 checkbox icon opacity issue

### LINKED ISSUE #6734: Disabled checkbox check icon has confusing fill opacity
#### Environment

- __Package version(s)__: 
- __Operating System__: MacOS Sonoma 14.3 
- __Browser name and version__: Chrome 122.0.6261.94 

#### Steps to reproduce

1. Blueprint Checkbox/Disabled/Check icon in code/web has a fill of White [#FFFFFF], 100% opacity
2. This is the same fill as the Checkbox/Default/Check icon
3. The checkbox does not appear disabled as a result

#### Actual behavior

The Checkbox/Disabled/Check icon has a fill of White [#FFFFFF], 100% opacity

#### Expected behavior

The Checkbox/Disabled/Check icon should have a fill of White [#FFFFFF], 50% opacity. This is consistent with the disabled icon styles for buttons/other components and will clearly communicate that the checkbox is not interactive.

#### Possible solution

The Checkbox/Disabled/Check icon should have a fill of White [#FFFFFF], 50% opacity. This is consistent with the disabled icon styles for buttons/other components and will clearly communicate that the checkbox is not interactive.

image

JamesHighsmith:
#### Fixes #6734

This pull request addresses an issue with the opacity of the checkbox icon in its disabled state. Previously, the opacity was set to 0.6, which did not provide enough contrast against the background, making it hard to distinguish for some users.

Changes proposed in this pull request:
Adjust the opacity of the checkbox icon in the disabled state from 0.6 to 0.5, improving visibility and design consistency.
Reviewers should focus on:
The change in _controls.scss for the checkbox icon opacity. Ensure this adjustment does not affect other components or create visual regressions in different states.
Screenshot:
No visual change significant enough to require a screenshot, but the difference can be observed in disabled checkboxes across the application where this change applies.

This minor adjustment ensures better consistency with the design language and improves the user experience by making the disabled state more distinguishable.

### DIFF
--- a/packages/core/src/components/forms/_controls.scss
+++ b/packages/core/src/components/forms/_controls.scss
@@ -58,7 +58,7 @@ $control-indicator-spacing: $pt-grid-size !default;
   input:disabled#{$selector} ~ .#{$ns}-control-indicator {
     background: $control-checked-background-color-disabled;
     box-shadow: none;
-    color: rgba($white, 0.6);
+    color: rgba($white, 0.5);
 
     @media (forced-colors: active) and (prefers-color-scheme: dark) {
       background-color: $pt-high-contrast-mode-disabled-border-color;


palantirtech:
Thanks for your interest in [palantir/blueprint](https://github.com/palantir/blueprint), @JamesHighsmith! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

ggdouglas:
On inspection of this, I don't see the proposed changes going into effect. This is what I'm seeing, where the checked and indeterminate states are still at 100% opacity. 

Screenshot 2024-08-12 at 21 26 39@2x

Since the indicator icon is a embedded SVG icon, it doesn't inherit the `color` of the `.bp5-control-indicator` class.

Instead, could we set the opacity directly on the `::before` element? It would look something like:
```scss
&.#{$ns}-checkbox {
  // ...
  input:disabled ~ .#{$ns}-control-indicator::before {
    opacity: 0.5;
  }
  // ...
}
```
This would be similar to what we already do for reducing the opacity on disabled radio button indicators https://github.com/palantir/blueprint/blob/4767947f098c4604e2e38d212a8973d7d39f6df5/packages/core/src/components/forms/_controls.scss#L320-L321

Screenshot 2024-08-12 at 22 01 56@2x

ggdouglas (changes requested):
Supportive of the change here, but we need some slightly more targeted styles to get the desired effect.

ggdouglas:
Fixed in #6962

PR #6762, opened 2024-04-10
Fix hotkeys dialog having 2 scrollbars

okaynik:
#### Checklist

- [ ] Includes tests
- [ ] Update documentation

Fix hotkey dialog having 2 scrollbars.

Both `hotkey-column` and `dialog-body-scroll-container` have `overflow: auto` property. Add `DialogBody` to `HotkeysDialogLegacy` and remove `overflow: scroll` from `hotkey-column`.

#### Screenshot
| Before | After    |
| ------  | ------  |
| image | image |

### DIFF
--- a/packages/core/src/components/hotkeys/_hotkeys.scss
+++ b/packages/core/src/components/hotkeys/_hotkeys.scss
@@ -24,8 +24,6 @@
 
 .#{$ns}-hotkey-column {
   margin: auto;
-  max-height: 80vh;
-  overflow-y: auto;
   padding: $pt-grid-size * 3;
 
   .#{$ns}-heading {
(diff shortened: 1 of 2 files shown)


michael-yx-wu (commented):
I cannot quite repro the before state on the current version of the docs with the scrollbars displayed side-by-side, although I do see two scrollbars overlapped on top of each other briefly and can confirm that scrolling doesn't work well because of the two scrollbars.

Instead of disabling the scroll container on `DialogBody`, can you instead add a `DialogBody` to the legacy implementation. Replacing this line with the actual component: https://github.com/palantir/blueprint/blob/0f8d6d9689e42ca09d36b3f63c564e19f485b91d/packages/core/src/legacy/hotkeysDialogLegacy.tsx#L128

Then we could probably just remove these styles from the `HOTKEY_COLUMN` class
https://github.com/palantir/blueprint/blob/0f8d6d9689e42ca09d36b3f63c564e19f485b91d/packages/core/src/components/hotkeys/_hotkeys.scss#L27-L28

ericjeney (approved):
Thanks! I was able to confirm this by setting "Show scroll bars" to "Always" in macOS's System Settings.

PR #6767, opened 2024-04-12
Fix `TagInput` adding a tag when `Enter` is pressed while composing

jrafidi:
#### Fixes

The `TagInput` component has a built-in `Enter` handler for adding a tag based on the current input value. This key is hit when composing languages such as Korean or Japanese, but not for the reason of adding a tag.

The current implementation will both add a tag of the current value and keep the current value, which is unexpected behavior.

This implementation should ignore `Enter` keystrokes when the user is in the middle of composing.

#### Checklist

- [x] Includes tests
- [ ] Update documentation

See above description.

Whether or not we need to fix this in other places in Blueprint.

#### Screenshot

Screenshot 2024-04-12 at 11 30 15 AM
Screenshot 2024-04-12 at 11 30 19 AM

### DIFF
--- a/packages/core/src/components/tag-input/tagInput.tsx
+++ b/packages/core/src/components/tag-input/tagInput.tsx
@@ -421,7 +421,8 @@ export class TagInput extends AbstractPureComponent<TagInputProps, TagInputState
 
         let activeIndexToEmit = activeIndex;
 
-        if (event.key === "Enter" && value.length > 0) {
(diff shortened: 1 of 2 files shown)


ericjeney (approved):
Thanks for fixing!

There are a few other places where we fix this today, such as in https://github.com/palantir/blueprint/pull/4266, but I think that's partially because the `isComposing` property on the native keyboard event wasn't rolled out broadly at the time.

jeremydorne (approved):

PR #6776, opened 2024-04-22
[Callout] Add rightElement prop

ljlinsc:
#### Checklist

- [X] Includes tests
- [X] Update documentation

Add `rightElement` prop to `Callout` component.

#### Screenshot

|Before|After|
|-|-|
|image|image|
|image|image|
|image|image|
|image|image|

**`rightElement` as a button wrapped in a tooltip:**

image

**`rightElement` as long text:**

image

### DIFF
--- a/packages/core/changelog/@unreleased/pr-6776.v2.yml
+++ b/packages/core/changelog/@unreleased/pr-6776.v2.yml
@@ -1,0 +1,5 @@
+type: improvement
+improvement:
+  description: '[Callout] Add rightElement prop'
+  links:
+  - https://github.com/palantir/blueprint/pull/6776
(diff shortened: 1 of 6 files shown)


evansjohnson (changes requested):
This could cause some weird interactions if the rightElement content is taller than the other content, and possibly some issues with these having the same flex basis from the wrapping `div`s.

You could get pretty close to this by defining your own child container with `display: flex; justify-content: space-between; align-items: center;`, it would just be centered on the other child content not including the header. Or you could get pretty much on par with this PR by defining your own `H5` inside the `children` you provide.

We could consider adding a docs example where the `title` is not provided to be rendered in the `children` with a totally custom layout, to show that something like this is possible.

ljlinsc:
Closing due to above reasons + having any action button moved to the bottom instead of the right!

PR #6779, opened 2024-04-30
Remove round corner for search inputs

CPerinet:
#### Checklist

- [ ] Includes tests
- [x] Update documentation

Remove the rounded corners of search inputs.


CPerinet:
@adidahiya

There are instances when the round corner looks off — typically when placed next to a square-er component such as Button. As a result, the default input is used in many instances instead of the `type="search"`. Thus, we lose the semantic and esc behaviour.

I appreciate that, with this change, there is not much in the way of telling a search input from a default input. The way I see it being done is through `icon="IconNames.SEARCH"` and `placeholder="Search..."`. Maybe we could set those as default for the search input?

PR #6782, opened 2024-05-02
Fix DateRangeInput3 closing on time change in Examples

moon1ock:
`this.props.timePrecision` which is supposed to be an alias for `this.props.timePickerProps.precision` seems to be never set and is always null.

I tried going down the rabbit hole to figure out how to update it properly, but it's just a can of worms.
Contributing a solution to the docs: it's not elegant, but it works

### DIFF
--- a/packages/datetime2/changelog/@unreleased/pr-6782.v2.yml
+++ b/packages/datetime2/changelog/@unreleased/pr-6782.v2.yml
@@ -1,0 +1,5 @@
+type: fix
+fix:
+  description: Fix DateRangeInput3 closing on time change
+  links:
+  - https://github.com/palantir/blueprint/pull/6782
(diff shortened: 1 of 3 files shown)


ericjeney (changes requested):
Thanks for spotting this and digging in to where the bug is, @moon1ock!

I don't think we should merge this docs change, though. I'd like the docs to continue to represent the "good" / expected ways that people should use components, rather than make it seem like this bug is an expected part of our API.

I recognize that's not a very satisfying answer since we haven't fixed the bug itself yet, but I'm glad you found a workaround that should work for you in your codebase at least. 🙂

evansjohnson:
closing in favor of https://github.com/palantir/blueprint/pull/6881 to fix the underlying issue - thank you for the lead on this!

PR #6785, opened 2024-05-03
[core] refactor: Use `isTimeoutEnabled` constant in other parts of `<Toast2>`

gluxon:
Optional change on top of https://github.com/palantir/blueprint/pull/6783. This should be a no-op.


pgoldberg (dismissed):

pgoldberg (approved):
I just noticed my approval was invalidated!

Issue #4026, opened 2020-03-25
Outlined button should not gain extra width and height because of the border


CPerinet:
#### Environment

- __Package version(s)__: 
- __Browser and OS versions__: 

If possible, link to a minimal repro (fork [this code sandbox](https://codesandbox.io/s/blueprint-sandbox-et9xy)): 

#### Steps to reproduce

1. 
1. 
1. 

#### Actual behavior

Outlined buttons are 2px wider and 2px higher than not outlined ones.

#### Expected behavior

Outlined buttons should have the same width and height as not outlined ones.

#### Possible solution

Use box shadow or reduce padding.

sonovawolf:
+1, thanks for responding this rapidly 🤗

PR #6788, opened 2024-05-06
[Button Group] Add outlined prop

ljlinsc:
#### Checklist

- [ ] Includes tests
- [X] Update documentation

Add `outlined` prop to `ButtonGroup` component.

#### Screenshot

|No outline|Outlined (new changes)|Regular Button|
|-|-|-|
|image|image|image|
|image|image||
|image|image|image|

### DIFF
--- a/packages/core/src/components/button/_button-group.scss
+++ b/packages/core/src/components/button/_button-group.scss
@@ -35,6 +35,7 @@ Markup:
 .#{$ns}-fill - Buttons expand equally to fill container
 .#{$ns}-large - Use large buttons
 .#{$ns}-minimal - Use minimal buttons
+.#{$ns}-outlined - Use outlined buttons
 .#{$ns}-vertical - Vertical layout
 
 Styleguide button-group
@@ -96,7 +97,7 @@ Styleguide button-group
   }
 
   // support wrapping buttons in a tooltip, which adds a wrapper element
-  &:not(.#{$ns}-minimal) {
+  &:not(.#{$ns}-minimal), &.#{$ns}-outlined  {
     > .#{$ns}-popover-wrapper:not(:first-child) .#{$ns}-button,
     > .#{$ns}-button:not(:first-child) {
       border-bottom-left-radius: 0;
@@ -111,7 +112,7 @@ Styleguide button-group
     }
   }
 
-  &.#{$ns}-minimal {
+  &.#{$ns}-minimal, &.#{$ns}-outlined {
     .#{$ns}-button {
       @include pt-button-minimal();
     }
@@ -131,6 +132,12 @@ Styleguide button-group
         border-top-right-radius: 0;
         margin-right: -$button-border-width;
       }
+    }
+  }
+
+  &.#{$ns}-outlined {
+    .#{$ns}-button {
+      @include pt-button-outlined();
     }
   }
 
@@ -213,7 +220,7 @@ Styleguide button-group
       width: 100%;
     }
 
-    &:not(.#{$ns}-minimal) {
+    &:not(.#{$ns}-minimal), &.#{$ns}-outlined {
       > .#{$ns}-popover-wrapper:first-child .#{$ns}-button,
       > .#{$ns}-button:first-child {
         border-radius: $pt-border-radius $pt-border-radius 0 0;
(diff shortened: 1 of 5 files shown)


michael-yx-wu (dismissed):

adidahiya:
Nit: I think the border colors might need adjustments, the vertical borders between buttons are _slightly_ more opaque (brighter in dark theme, darker in light theme) than the outer border of the group:

CleanShot 2024-05-07 at 08 28 08
CleanShot 2024-05-07 at 08 29 07

packages/core/src/components/button/_button-group.scss
```diff
@@ -211,9 +230,13 @@ Styleguide button-group
       margin-right: 0 !important; /* stylelint-disable-line declaration-no-important */
       // needed to ensure buttons take up the full width when wrapped in a Popover:
       width: 100%;
+
+      &:not(:last-child) {
+        border-bottom: none;
+      }
```
evansjohnson: could we only apply this rule inside the `outlined` wrapper?

packages/core/src/components/button/buttonGroup.tsx
```diff
@@ -70,13 +77,14 @@ export interface ButtonGroupProps extends Props, HTMLDivProps, React.RefAttribut
...
     (props, ref) => {
-        const { alignText, className, fill, minimal, large, vertical, ...htmlProps } = props;
+        const { alignText, className, fill, minimal, outlined, large, vertical, ...htmlProps } = props;
         const buttonGroupClasses = classNames(
             Classes.BUTTON_GROUP,
             {
                 [Classes.FILL]: fill,
                 [Classes.LARGE]: large,
                 [Classes.MINIMAL]: minimal,
+                [Classes.OUTLINED]: outlined,
```
evansjohnson, line 87: 🎉

evansjohnson (commented):
just one small comment - could you push a commit so the PR preview builds again?

packages/core/src/components/button/_button-group.scss
```diff
@@ -107,11 +108,17 @@ Styleguide button-group
     > .#{$ns}-button:not(:last-child) {
       border-bottom-right-radius: 0;
       border-top-right-radius: 0;
+    }
+  }
+
+  &:not(.#{$ns}-minimal):not(.#{$ns}-outlined) {
```
evansjohnson, line 114: is this needed? I'm seeing the button group grow in size when `outlined` which I don't think is intentional

(added some extra buttons locally to test)

vertical (extra 1px per button, expect for last is what I'd assume):
Screenshot 2024-07-03 at 11 31 45 AM

Screenshot 2024-07-03 at 11 31 41 AM

horizontal (2px per button):
Screenshot 2024-07-03 at 11 33 55 AM

Screenshot 2024-07-03 at 11 34 00 AM

packages/core/src/components/button/_button-group.scss
```diff
@@ -223,12 +242,20 @@ Styleguide button-group
...
+    &:not(.#{$ns}-minimal):not(.#{$ns}-outlined) {
       > .#{$ns}-popover-wrapper:not(:last-child) .#{$ns}-button,
       > .#{$ns}-button:not(:last-child) {
         margin-bottom: -$button-border-width;
       }
     }
+
+    &.#{$ns}-outlined {
+      > .#{$ns}-button:not(:last-child) {
+        border-bottom: none;
```
evansjohnson, line 256: could we do a transparent color border instead of none here? `border-bottom-color: transparent` - it looks like that should fix any small shifts when toggling this prop

packages/core/src/components/button/_button-group.scss
```diff
@@ -134,6 +141,18 @@ Styleguide button-group
...
   }
 
+  &.#{$ns}-outlined {
+    > .#{$ns}-button {
+      @include pt-button-outlined();
+    }
+
+    &:not(.#{$ns}-vertical) {
+      > .#{$ns}-button:not(:last-child) {
+        border-right: none;
```
evansjohnson, line 151: here I think `none` instead of transparent color actually helps, since the `Button` component itself currently grows 2px when outlined. this brings that down to 1px at least

evansjohnson (commented):
leaving this review just in case it's helpful in the future - @ljlinsc highlighted https://github.com/palantir/blueprint/issues/4026 so I don't think we should worry about the slight px difference for now since it's also an issue in the underlying button component

evansjohnson (approved):
Thank you for the PR!

We may need to revisit some of this CSS if addressing https://github.com/palantir/blueprint/issues/4026 but this seems fine for now. The underlying button also displays some px shift when toggling this prop, so not worrying about it here either for now.

In practice if users use outline consistently (don't toggle for the same button group) this shouldn't be a big issue.

PR #6802, opened 2024-05-15
Add new Tabs option to enable tab navigation to non-active tabs

### LINKED ISSUE #6759: Cannot use tab key to navigate through Blueprint tabs
#### Environment

- __Package version(s)__: 5.8.2
- __Operating System__: MacOS Sonoma 14.4
- __Browser name and version__: Edge, Chrome
- 
#### Steps to reproduce

Try to tab through the tabs in the example docs [here](https://blueprintjs.com/docs/#core/components/tabs)
 
#### Actual behavior

For some reason only the currently selected tab can be focused by tabbing through. Non-selected tabs do not get focus. 

https://github.com/palantir/blueprint/assets/35851484/9eea76cd-b885-48f4-b440-032af5b0fc88

#### Expected behavior

All tabs unless disabled should be able to be focused on by tabbing through

evansjohnson:
#### Fixes https://github.com/palantir/blueprint/issues/6759

Note that current behavior follows the [W3 guidance for the tabs pattern](https://www.w3.org/WAI/ARIA/apg/patterns/tabs/), so we should consider closing out the filed issue without making any fixes or adding any features. At a minimum, I think the default behavior should follow the guidance, possibly with an opt-in option to add this additional tab navigation pattern to use as consumers of the Tabs component feels is appropriate.

Ex one use case for this is if the usage of the `Tabs` component, for whatever reason, feels or looks more like a button group, and users expect to be able to tab to all tabs.

#### Checklist

Can consider tests pending decision to support or not

- [ ] Includes tests
- [x] Update documentation

- Add a new prop `enableTabbedNavigation` to the `Tabs` component, which allows using the tab key to access non-active tabs

- Do we want to add this option?
- Prop naming - also considered `insertNonActiveTabsInTabOrder` and `addNonActiveTabsToTabOrder`

#### Screenshot

Screenshot 2024-05-15 at 2 04 05 PM

### DIFF
--- a/packages/core/src/components/tabs/tabTitle.tsx
+++ b/packages/core/src/components/tabs/tabTitle.tsx
@@ -36,6 +36,9 @@ export interface TabTitleProps extends TabProps {
 
     /** Whether the tab is currently selected. */
     selected: boolean;
+
+    /** Whether to enable tab navigation to non-disabled, non-active, tabs. */
+    enableTabbedNavigation: boolean;
 }
 
 export class TabTitle extends AbstractPureComponent<TabTitleProps> {
@@ -53,6 +56,7 @@ export class TabTitle extends AbstractPureComponent<TabTitleProps> {
             icon,
             tagContent,
             tagProps,
+            enableTabbedNavigation,
             ...htmlProps
         } = this.props;
         const intent = selected ? Intent.PRIMARY : Intent.NONE;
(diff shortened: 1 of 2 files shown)


invliD:
Ah, I was not aware of that guidance when I commented on the issue. Given how explicit the guidance is on this, I'd say we should close it out as wontfix since we currently support that guidance exactly.

evansjohnson:
I agree with leaning towards wontfix, and mostly opened this since it was easy enough to show for consideration

I was stretching a bit with
> one use case for this is if the usage of the Tabs component, for whatever reason, feels or looks more like a button group, and users expect to be able to tab to all tabs.

In this case if this isn't the tab navigation pattern you want, then it seems you simply don't want the tabs component

ericjeney:
+1 to closing this as "won't fix" given the explicitness of the guidance. Thanks, @evansjohnson!

PR #6803, opened 2024-05-15
Avoid trapping focus in Table

### LINKED ISSUE #3834: Blueprint table is a keyboard trap
#### Environment

- __Package version(s)__:  Blueprint table 3.8.1
- __Browser and OS versions__:  Chrome

If possible, link to a minimal repro (fork [this code sandbox](https://codesandbox.io/s/blueprint-sandbox-et9xy)): https://codesandbox.io/s/blueprint-sandbox-hiwri

#### Steps to reproduce

1. Use tab key to move focus into blueprint table

#### Actual behavior

It's not possible to move focus out of the table using the keyboard - as demonstrated in repo above, once focus is in the table you cannot tab to either button on the page

#### Expected behavior

You should be able to tab back to the previous or onto the next element on the page.

evansjohnson:
#### Checklist

- [ ] Includes tests
- [x] Update documentation
  - testable in existing table docs

Immediately preventing default on tab key presses seems problematic since sometimes we return early and do nothing on the tab key press. This happens when the selection is already at the edge of the table and will not move further. This PR proposes that in these cases we should allow focus to freely leave the table.

I considered other solutions including:
- Making this a new optional option.
    - In the next major version, this option could be defaulted to true, and possibly removed in the future if we see no valid cases where trapping focus is desired.
- Introducing a new pattern to move focus into the table, such as Enter when the table is focused to switch to current behavior, and otherwise have the table be a single element in the tab order.
    - Esc would "exit" this interaction mode
    - Big downside of not being very discoverable
- Are there other best practices for table interaction to consider?

- Is this the interaction pattern we want?
- Should this be applied as a fix, or a new option?

#### Screenshot

n/a

### DIFF
--- a/packages/table/src/tableHotkeys.ts
+++ b/packages/table/src/tableHotkeys.ts
@@ -144,9 +144,6 @@ export class TableHotkeys {
     // no good way to call arrow-key keyboard events from tests
     /* istanbul ignore next */
     private handleFocusMove = (e: KeyboardEvent, direction: "up" | "down" | "left" | "right") => {
-        e.preventDefault();
-        e.stopPropagation();
-
         const { focusedCell } = this.state;
         if (focusedCell == null) {
             // halt early if we have a selectedRegionTransform or something else in play that nixes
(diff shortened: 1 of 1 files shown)


gluxon:
Thanks for this fix!

I'm personally leaning towards not optionalizing this and treating it as a fix. I can't imagine any users that would want the Blueprint table to trap focus within it. 🤔

gluxon (approved):
Approving based on my thoughts above. I think this is the right fix. Let me gut check this with another person to make sure this sounds reasonable.

pgoldberg (approved):
I agree that this should just be treated as a fix – I imagine no one wants the behavior of tables trapping focus

PR #6786, opened 2024-05-03
fix(menu): remove negative margin to prevent horizontal scroll

chaejunlee:
#### Checklist

- [ ] Includes tests
- [ ] Update documentation

#### Changes proposed in this pull request:

Previously in this PR https://github.com/palantir/blueprint/pull/6269, `menu-header` (or `menu-divider`) had negative horizontal margin to make "dividers take up full menu width".

However this caused horizontal overflow and scroll. Removing the horizontal margin solves the scroll behavior.

#### Reviewers should focus on:

Previous PR doesn't really give the reason why it should have negative margin to achieve full width, while it already is taking full width.

If the negative margin was there just to make the header take the full width, negative margin is not needed and should be removed.

AS-IS

Screenshot 2024-05-03 at 15 18 42

TO-BE

Screenshot 2024-05-03 at 15 19 30

#### Screenshot

`<Select />` where negative margin header is removed.

Screenshot 2024-05-03 at 15 28 57

PR #6269, opened 2023-07-10
[core] feat: add MenuItem small modifier, extend MenuDivider width

adidahiya:
#### Checklist

- [ ] Includes tests
- [x] Update documentation

#### Changes proposed in this pull request:

- feat(`MenuItem`): add support for `small` size modifier prop
- feat(`MenuItem`): remove constraint that prevents both selected "tick" icon and `icon` prop from being shown at the same time
- fix(`MenuItem`): "selected" styles no longer extend "active" styles (this is mostly an internal change)

#### Reviewers should focus on:

No regressions in selected styling and possible CSS overrides consumers may be applying on `.bp5-menu-item.bp5-selected` elements.

PR #6804, opened 2024-05-15
No scrollbar in Select component when menu has dividers

### LINKED ISSUE #6780: Select has a horizontal scrollbar
#### Steps to reproduce

Turn on grouped in the docs here. https://blueprintjs.com/docs/#select/select-component

Hard to say how long this has been an issue because not all users have the "always show scrollbars" browser/OS setting enabled.

image (2)

Need to look into whether it's a docs-app bug or a Select component styles bug.

evansjohnson:
#### Checklist

- [ ] Includes tests
- [ ] Update documentation

I don't think these are relevant for this PR

For a menu divider inside of a select-popover, we already have applied a 5px padding inside the popover. This means to get to full popover width we need to apply the -5px margin to the outer list element, and therefor also undo the -5px margin on the menu divider.

This PR is a variation on https://github.com/palantir/blueprint/pull/6786, which correctly identified that the -5px margin on the menu divider is responsible for the scrollbar being added, but this PR additionally applies this -5px margin on the list element so that the divider continues to take up the full width.

- Any other styling options I'm not thinking of that may be negatively impacted
- Any concerns of this fix itself being a breaking change, since this style issue has likely existed since https://github.com/palantir/blueprint/pull/6269 was merged almost a year ago?

#### Screenshot

Before
Screenshot 2024-05-15 at 2 20 22 PM

After
Screenshot 2024-05-15 at 2 19 48 PM
*note right and bottom padding outside of list are now also removed. this seems to match what we do in the menu popover

### DIFF
--- a/packages/select/src/components/select/_select.scss
+++ b/packages/select/src/components/select/_select.scss
@@ -3,11 +3,13 @@
 
 @import "../../common/variables";
 
+$select-padding: $pt-grid-size * 0.5;
+
 .#{$ns}-select-popover {
   .#{$ns}-popover-content {
     // use padding on container rather than margin on input group
     // because top margin leaves some empty space with no background color.
-    padding: $pt-grid-size * 0.5;
+    padding: $select-padding;
   }
 
   .#{$ns}-input-group {
@@ -19,10 +21,13 @@
     max-width: $select-popover-max-width;
     overflow: auto;
     padding: 0;
+    margin-left: -$select-padding;
+    margin-right: -$select-padding;
+    margin-bottom: -$select-padding;
 
     &:not(:first-child) {
       // adjust padding to account for that on .#{$ns}-popover-content above
-      padding-top: $pt-grid-size * 0.5;
+      padding-top: $select-padding;
     }
   }
 
--- a/packages/select/src/components/select/_select.scss
+++ b/packages/select/src/components/select/_select.scss
@@ -17,25 +17,19 @@ $select-padding: $pt-grid-size * 0.5;
   }
 
   .#{$ns}-menu {
-    // offset the padding added by the select componenent so menu dividers can remain full width
+    // offset the padding added by the select componenent so menu can remain full width with padding on either side
+    // for menu items
     margin-left: -$select-padding;
     margin-right: -$select-padding;
     max-height: $select-popover-max-height;
     max-width: $select-popover-max-width;
     overflow: auto;
-    padding: 0;
+    padding: 0 5px;
 
     &:not(:first-child) {
       // adjust padding to account for that on .#{$ns}-popover-content above
       padding-top: $select-padding;
     }
-  }
-
-  .#{$ns}-menu-item {
-    // re-apply margin unset on menu to the actual menu items. unlike menu dividers, menu items are not intended
-    // to be full width
-    margin-left: $select-padding;
-    margin-right: $select-padding;
   }
 
   &.#{$ns}-popover-match-target-width {
--- a/packages/select/src/components/select/_select.scss
+++ b/packages/select/src/components/select/_select.scss
@@ -24,7 +24,7 @@ $select-padding: $pt-grid-size * 0.5;
     max-height: $select-popover-max-height;
     max-width: $select-popover-max-width;
     overflow: auto;
-    padding: 0 5px;
+    padding: 0 $select-padding;
 
     &:not(:first-child) {
       // adjust padding to account for that on .#{$ns}-popover-content above
--- a/packages/select/src/components/select/_select.scss
+++ b/packages/select/src/components/select/_select.scss
@@ -17,13 +17,13 @@ $select-padding: $pt-grid-size * 0.5;
   }
 
   .#{$ns}-menu {
+    margin-bottom: -$select-padding;
+    margin-left: -$select-padding;
+    margin-right: -$select-padding;
     max-height: $select-popover-max-height;
     max-width: $select-popover-max-width;
     overflow: auto;
     padding: 0;
-    margin-left: -$select-padding;
-    margin-right: -$select-padding;
-    margin-bottom: -$select-padding;
 
     &:not(:first-child) {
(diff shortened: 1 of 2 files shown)


chaejunlee:
I think going over the `<Menu />` is necessary. Can you post the screenshot of it?

evansjohnson:
There's no changes to the menu in this PR, [except when inside a select popover](https://github.com/palantir/blueprint/pull/6804/files#diff-f0f5c40bacc20327ba1f185471a11a4fe4790e1c22b66b8a38e727640de649cfR313)

I checked for good measure and in the generated docs preview you can see the styling menu is unaffected:
Screenshot 2024-05-16 at 5 28 33 PM

Note in https://github.com/palantir/blueprint/pull/6786 since the -5px margin is removed there is a slight visual regression in the screenshot where the border no longer goes to the edge of the menu, which seems to be what the intent of that style is.

gluxon (commented):
## Menu Items Flush

Is it intended that the menu items themselves are flush with the border of the popover after these changes? I see from your description that we wanted to make the horizontal dividers flush, which I agree with.

Screenshot 2024-05-30 at 1 15 34 PM

Here's the before picture again for reference. I think the intended visual design was to have a bit of spacing around them menu item since the light blue hover background has a slight border radius. That border radius doesn't look as great if the background is how flush with the border.

Screenshot 2024-05-30 at 1 15 23 PM

## Suggestion

I'm wondering if we should tweak `@mixin menu-divider()`'s implementation to instead be a `::before` pseudo-element that is "[out of flow](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_flow_layout/In_flow_and_out_of_flow)".

The negative margin would then only be on the pseudo-element driving the horizontal border. Since the pseudo-element is "out of flow", we would eliminate unintended CSS layout calculations (like this one) entirely in the future.

evansjohnson:
Thanks for the close look and good catch @gluxon!

I wrapped up this approach since it was easy to add back the margin one more time for the items, but given how much back and forth this is I may try the psuedo element approach quickly as well.

As a nice to have - it looks like there was a small stying weirdness where there is empty space padding to the left but not right of the selection. With fixing the scrollbar overflow this also sizes the menu items correctly to have padding on both sides.

## develop
Screenshot 2024-06-18 at 8 55 35 AM

Screenshot 2024-06-18 at 8 55 29 AM

## this PR
Screenshot 2024-06-18 at 8 56 00 AM

Screenshot 2024-06-18 at 8 56 07 AM

with scrollbars showing only when scrolling still looks good
Screenshot 2024-06-18 at 9 04 10 AM
Screenshot 2024-06-18 at 9 04 01 AM

packages/select/src/components/select/_select.scss
```diff
@@ -3,29 +3,41 @@
...
     padding: 0;
 
     &:not(:first-child) {
       // adjust padding to account for that on .#{$ns}-popover-content above
-      padding-top: $pt-grid-size * 0.5;
+      padding-top: $select-padding;
     }
   }
 
+  .#{$ns}-menu-item {
```
evansjohnson: I wonder if there may be other classes we need to target with adding back the margin we take away above...

From what I could tell from the docs it seems like only elements that get this `bp-menu-item` class are intended to go into a menu, but I suppose it may be allowed by the API to develop custom menu items that may now shift around with this change?

evansjohnson: ahh, I think this was a good note... pushing to make the divider and menu items not be a special case got me to bring back the padding on the actual menu, which allows the divider to keep it's normal negative margin

packages/select/src/components/select/_select.scss
```diff
@@ -3,26 +3,32 @@
...
   .#{$ns}-menu {
+    // offset the padding added by the select componenent so menu can remain full width with padding on either side
+    // for menu items
+    margin-left: -$select-padding;
+    margin-right: -$select-padding;
     max-height: $select-popover-max-height;
     max-width: $select-popover-max-width;
     overflow: auto;
-    padding: 0;
+    padding: 0 5px;
```
gluxon: nitpick: Does it make sense for this to also be `$select-padding`?

```suggestion
    padding: 0 $select-padding;
```

packages/select/src/components/select/_select.scss
```diff
@@ -3,26 +3,32 @@
...
+    padding: $select-padding;
   }
 
   .#{$ns}-input-group {
     margin-bottom: 0;
   }
 
   .#{$ns}-menu {
+    // offset the padding added by the select componenent so menu can remain full width with padding on either side
+    // for menu items
```
gluxon: nit: Is it worth a comment that this is specifically to allow dividers to be flush with the popover's edge?

packages/select/src/components/select/_select.scss
gluxon: Nice!

gluxon:
> As a nice to have - it looks like there was a small stying weirdness where there is empty space padding to the left but not right of the selection. With fixing the scrollbar overflow this also sizes the menu items correctly to have padding on both sides.

Very nice! 😁

gluxon (dismissed):

packages/select/src/components/select/_select.scss
evansjohnson: indeed!

packages/select/src/components/select/_select.scss
```diff
@@ -3,26 +3,32 @@
...
-    padding: $pt-grid-size * 0.5;
+    padding: $select-padding;
   }
 
   .#{$ns}-input-group {
     margin-bottom: 0;
   }
 
   .#{$ns}-menu {
+    // offset the padding added by the select componenent so menu can remain full width with padding on either side
```
evansjohnson, line 20:
```suggestion
    // offset the padding added by the select componenent so menu can remain full width with padding on either side
    // this is important so menu dividers can be full width, and so that menu items can have padding to the left and right
```

gluxon (approved):

Issue #6171, opened 2023-05-23
[V5] ContextMenu does not lazy render popover anymoe


ptgamr:
#### Environment

- __Package version(s)__: 5.0.0-alpha.3
- __Operating System__: MacOS
- __Browser name and version__: Chrome

#### Steps to reproduce

- We're migrating `ContextMenu2` --> `ContextMenu`, but see some degrade in performance for some of our table.

#### Actual behavior

- Lots of portal divs are created in advance, this is not happening with `ContextMenu2`, it used to be lazy,
![2023-05-24-001917_611x460_scrot](https://github.com/palantir/blueprint/assets/1619740/3156bf65-5a49-4b3f-bfb5-da5fc483f830)

- I am trying to pass `lazy: true` to `popoverProps` but TypeScript complaining, apparently because we only pick a handful of props.

#### Expected behavior
- lazy should be the default, like it was with `ContextMenu2`

#### Possible solution

I'm not too sure this time...

adidahiya:
good catch, thanks for flagging the issue @ptgamr

adidahiya:
@ptgamr actually, this is not the bug I thought it was. ContextMenu2 in v4 and ContextMenu in v5 _are_ lazy; they use the default prop value of `lazy: true` in `<Overlay>`. Being lazy, however, does still mean that a portal is created for each context menu on mount. You can see this in the v4 docs:

![image](https://github.com/palantir/blueprint/assets/723999/28a947a4-989d-4340-bbd6-51fd0c35fb5e)

It sounds like you might either:

1. be requesting a feature for the __Portal__ component to be lazy about mounting its children
2. looking for an equivalent to `ContextMenu.show()` from v4, which doesn't do anything until you call it since it's an imperative API

For (2), you can use `showContextMenu()` in v4 or v5. This will avoid adding elements to the document when your table is first mounted.

ptgamr:
Oh, yes, I'm sorry, my bad. We're actually migrating from `@ContextMenuTarget`. 

However, I'm not quite sure what you said is align with the docs:

> Being lazy, however, does still mean that a portal is created for each context menu on mount. You can see this in the v4 docs:

The docs say:

> If lazy=true and usePortal={true}, the Portal containing the children is created and attached to the DOM when the overlay is opened for the first time; otherwise this happens when the component mounts

https://blueprintjs.com/docs/versions/5/#core/components/overlay

ptgamr:
[...]

PR #6805, opened 2024-05-15
Remove `componentDidMount` and `componentDidUpdate` hooks from `<Popover>`

ab-pm:
While using `<Popover>` in controlled mode to work around #6171, I noticed that it always renders twice when `onInteraction` changes my `isOpen` state. On first render, `renderTarget` is called with outdated arguments (closed instead of open), only immediately after in the second render it gets passed the correct values.
You can repro with something like
```jsx

<Popover
	…
	isOpen={isOpen}
	onInteraction={(nextOpenState, event) => {
		console.log('interaction', nextOpenState, event);
		setIsOpen(nextOpenState);
	}}
	renderTarget={props => {
		console.log('Rendering popover target', isOpen, props);
		return …;
	}}
/>
```
I traced the problem down to `setState` being called from `componentDidUpdate`. Instead the component should simply take the `isOpen` value from the props and ignore its local state altogether in controlled mode.

#### Checklist

- [x] Includes tests
- [x] Update documentation

* avoid calling `setState` from `componentDidUpdate`, which leads to the component rendering twice in controlled mode
* instead, just ignore local state when the component is in controlled mode
* remove `componentDidMount` and `componentDidUpdate` hooks from `<Popover>` altogether
* assert that `onInteraction` is always called with an event
* also remove `hasDarkParent` and `isClosingViaEscapeKeypress` from react component state, put them in class properties instead and update them whenever the overlay is opening/closing

* dark mode and esc key behavior
* coding style (I'm unfamiliar with Bluprint guidelines)
* testing (I've only done this in the web editor, haven't actually installed the dev environment and ran any tooling)
* whether no longer calling `onInteraction` due to props change is a bugfix or a compatibility issue

### DIFF
--- a/packages/core/src/components/popover/popover.tsx
+++ b/packages/core/src/components/popover/popover.tsx
@@ -132,9 +132,7 @@ export interface PopoverProps<TProps extends DefaultPopoverTargetHTMLProps = Def
 }
 
 export interface PopoverState {
-    hasDarkParent: boolean;
-    // when an ESC keypress interaction closes the overlay, we want to force-enable `shouldReturnFocusOnClose` behavior
-    isClosingViaEscapeKeypress: boolean;
+    /** Whether the popover is open. Ignored in controlled mode. */
     isOpen: boolean;
 }
 
@@ -176,9 +174,7 @@ export class Popover<
     };
 
     public state: PopoverState = {
-        hasDarkParent: false,
-        isClosingViaEscapeKeypress: false,
-        isOpen: this.getIsOpen(this.props),
+        isOpen: this.props.defaultIsOpen!,
     };
 
     /**
@@ -203,12 +199,18 @@ export class Popover<
      */
     public targetRef = React.createRef<HTMLElement>();
 
+    // a flag that tells us to whether the target is in dark mode, so that also the overlay gets rendered in dark mode if inside a portal
+    private targetHasDarkParent: boolean | undefined = false;
+
     /**
      * Overlay2 transition container element ref.
      */
     private transitionContainerElement = React.createRef<HTMLDivElement>();
 
     private cancelOpenTimeout?: () => void;
+
+    // when an ESC keypress interaction closes the overlay, we want to force-enable `shouldReturnFocusOnClose` behavior
+    private isClosingViaEscapeKeypress: boolean = false;
 
     // a flag that lets us detect mouse movement between the target and popover,
     // now that mouseleave is triggered when you cross the gap between the two.
@@ -220,8 +222,6 @@ export class Popover<
 
     // Reference to the Poppper.scheduleUpdate() function, this changes every time the popper is mounted
     private popperScheduleUpdate?: () => Promise<Partial<PopperState> | null>;
-
-    private isControlled = () => this.props.isOpen !== undefined;
 
     // arrow is disabled if minimal, or if the arrow modifier was explicitly disabled
     private isArrowEnabled = () => !this.props.minimal && this.props.modifiers?.arrow?.enabled !== false;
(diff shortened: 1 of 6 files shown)


palantirtech:
Thanks for your interest in [palantir/blueprint](https://github.com/palantir/blueprint), @ab-pm! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

ab-pm:
@palantir do you need anything else? Do I need to provide a changelog entry before the PR even is reviewed?

ericjeney:
Nothing else necessary, and you can ignore the changelog check! We should be able to give this a review next week.

gluxon (commented):
Thank you! This fix is very appreciated.  I was also looking at https://github.com/palantir/blueprint/pull/6771 recently, which was another case where we should have simply done a simple nullish coalesce like you did instead of complicated state handling.

https://github.com/palantir/blueprint/blob/22355c2c6aa16d704493427b64ab7893a93f8999/packages/core/src/components/popover/popover.tsx#L246

## Pure Rendering

I'm a bit nervous about the other changes in this PR that move the `hasDarkParent` and `isClosingViaEscapeKeypress` state booleans to be instance variables. I'm not sure if that change will work as expected with React 18's concurrent renderer, which expects the render output of a component to be a pure function of its state and props.

https://react.dev/reference/react/StrictMode#fixing-bugs-found-by-double-rendering-in-development

> [React assumes that every component you write is a pure function.](https://react.dev/learn/keeping-components-pure) This means that React components you write must always return the same JSX given the same inputs (props, state, and context).

I acknowledge that you'll probably find existing Blueprint code that violates this best practice since it only became a stronger recommendation as of React 18. We shouldn't actively move away from this best practice in new code though.

## Request

Could we revert the changes to `hasDarkParent` and `isClosingViaEscapeKeypress` for now? If there's factors I'm missing that make this safe to do, I'd be open to hear those explanations. Thanks.

ab-pm:
> React 18's concurrent renderer expects the render output of a component to be a pure function of its state and props.

I'm under the impression that guidance is a bit simplified and pertains only to function components. For class components, it would need to talk about the `render` *method*, which gets the whole instance as its input `this` argument and is still pure in terms of that. Rendering a `Popover` does neither mutate the instance (or its `.state` or `.props`) and does not have any side effects either. Notice the instance properties are only updated from event handlers. So I'm pretty sure the code is still on the safe side here.

Admittedly, the guidance does not talk about rendering of `useRef` values, which would be the function component equivalent to class component instance properties. It's probably not the best idea since changing a ref/instance property does not cause the component to re-render, but I think it works fine here. `targetHasDarkParent` and `isClosingViaEscapeKeypress` would only ever change when the popover opens or closes, not on renders in between, right?

Tbh I'm more concerned whether moving the `targetHasDarkParent` update from `componentDidMount`/`componentDidUpdate` to the open/close handler works properly (regardless whether it's a state or property update). I'm not sure how this is meant to work, and have no resources to test this on my own. In particular, I might have broken it for controlled popovers? Tbh the whole approach seems sketchy to me - https://legacy.reactjs.org/docs/strict-mode.html for example did mention that side effects (like state updates) from `componentDidMount`/`componentDidUpdate` should be avoided, but I don't have a better solution either. In a function component I'd probably have used a layout effect, or even better just relied on a context for dark mode, but I didn't want to do an entire refactoring in this PR :-)

gluxon:
Hey @ab-pm, I'm going to hand this over to @ggdouglas and @jscheiny, who are more actively maintaining Blueprint at the moment.

> > React 18's concurrent renderer expects the render output of a component to be a pure function of its state and props.
> 
> I'm under the impression that guidance is a bit simplified and pertains only to function components. For class components, it would need to talk about the `render` _method_, which gets the whole instance as its input `this` argument and is still pure in terms of that. Rendering a `Popover` does neither mutate the instance (or its `.state` or `.props`) and does not have any side effects either. Notice the instance properties are only updated from event handlers. So I'm pretty sure the code is still on the safe side here.

Do you have evidence for that claim?

This doesn't match my understanding of the concurrent renderer. If a `render()` function returns different results for the same props and state, it won't be concurrent renderer compatible. That breaks the fundamental assumption the concurrent renderer makes about being able to bail out in-progress work and try again later. Whether the component is a function or class-based component doesn't seem relevant.

I don't believe class based components get some special exception from the pure function guidance. If that were the case, they can't be used with React 18's new concurrent renderer.

PR #6811, opened 2024-05-18
[core] fix: do not colour FormGroup sub label based on intent

### LINKED ISSUE #6808: Is it correct for the FormGroup sub label colour to be determined by the intent?
#### Environment

- __Package version(s)__: @blueprintjs/core 5.10.3
- __Browser name and version__: Google Chrome 125.0.6422.60

#### Code Sandbox

Link to a minimal repro: https://blueprintjs.com/docs/#core/components/form-group

#### Steps to reproduce

1. Enable "Show sub label"
1. Set the intent to e.g. "Danger"

#### Actual behavior

The sub label inherits the colour of the intent.

image

#### Expected behavior

The sub label, like the label should remain unchanged when an intent is supplied.

image

Unlike the helper text, which I understand should be used to display information based on the supplied value (such as "This username is already taken"), the sub label is not contextual and should not be re-coloured based on the intent of the FormGroup. 

#### Possible solution

Do not colour the sub label based on the intent, and instead maintain a constant colour as is done for the label. I think that this change should be made because the sub label should be used to provide further information about the label, and the label does not change colour based on the intent.

berryde:
#### Checklist

- [ ] Includes tests
- [ ] Update documentation

No longer derive the sub label colour of FromGroup from the intent.

#### Screenshot

**Before**

image

image

**After**

image

image

### DIFF
--- a/packages/core/changelog/@unreleased/pr-6811.v2.yml
+++ b/packages/core/changelog/@unreleased/pr-6811.v2.yml
@@ -1,0 +1,5 @@
+type: improvement
+improvement:
+  description: Do not colour FormGroup sub label based on intent
+  links:
+  - https://github.com/palantir/blueprint/pull/6811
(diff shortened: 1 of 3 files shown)


berryde:
I understand that this is an opinionated change - I think it is best illustrated by the validation example provided in the linked issue:

**Before**

image

**After**

image

CPerinet:
The change makes sense to me

invliD (approved):

PR #6821, opened 2024-05-23
chore(NonIdealState): refactored non ideal state component to a functional component

ahmadJT:
#### Checklist

- [ ] Includes tests
- [ ] Update documentation

- [x] refactor non ideal state component to a functional component

- NonIdealState component

### DIFF
--- a/packages/core/src/components/non-ideal-state/nonIdealState.tsx
+++ b/packages/core/src/components/non-ideal-state/nonIdealState.tsx
@@ -19,7 +19,7 @@ import * as React from "react";
 
 import { type IconName, IconSize } from "@blueprintjs/icons";
 
-import { AbstractPureComponent, Classes, DISPLAYNAME_PREFIX, type MaybeElement, type Props } from "../../common";
+import { Classes, DISPLAYNAME_PREFIX, type MaybeElement, type Props } from "../../common";
 import { ensureElement } from "../../common/utils";
 import { H4 } from "../html/html";
 import { Icon } from "../icon/icon";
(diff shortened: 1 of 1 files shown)


packages/core/src/components/non-ideal-state/nonIdealState.tsx
```diff
@@ -79,34 +79,20 @@ export interface NonIdealStateProps extends Props {
...
+    action,
+    children,
+    className,
+    description,
+    icon,
+    iconMuted = true,
+    iconSize = NonIdealStateIconSize.STANDARD,
+    layout = "vertical",
+    title,
+}) => {
```
ggdouglas, line 92: Can we destructure these inside of the function body rather than in the arguments? I think this is more in line with the structure we use for other functional components that have multiline props in Blueprint.

```tsx
export const NonIdealState: React.FC<NonIdealStateProps> = props => {
    const {
        action,
        children,
        className,
        description,
        icon,
        iconMuted = true,
        iconSize = NonIdealStateIconSize.STANDARD,
        layout = "vertical",
        title,
    } = props;
    // ...
}
```

packages/core/src/components/non-ideal-state/nonIdealState.tsx
```diff
@@ -119,21 +105,30 @@ export class NonIdealState extends AbstractPureComponent<NonIdealStateProps> {
...
-        const { description, title } = this.props;
-        if (title == null && description == null) {
-            return undefined;
-        } else {
-            return (
+            )
+        );
+    };
+
+    const maybeRenderText = () => {
```
ggdouglas, line 112: Instead of keeping these as inline functions, can we encapsulate the conditional in a couple variables, say:
```tsx
const shouldRenderVisual = icon != null;
const shouldRenderText = title != null || description != null;
```

and then inline them below:
```tsx
<div className={classNames(Classes.NON_IDEAL_STATE, `${Classes.NON_IDEAL_STATE}-${layout}`, className)}>
    {shouldRenderVisual && (
        <div
            className={Classes.NON_IDEAL_STATE_VISUAL}
            style={{ fontSize: `${iconSize}px`, lineHeight: `${iconSize}px` }}
        >
            <Icon
                className={classNames({ [Classes.ICON_MUTED]: iconMuted })}
                icon={icon}
                size={iconSize}
                aria-hidden={true}
                tabIndex={-1}
            />
        </div>
    )}
    {shouldRenderText && (
        <div className={Classes.NON_IDEAL_STATE_TEXT}>
            {title && <H4>{title}</H4>}
            {description && ensureElement(description, "div")}
        </div>
    )}
    {action}
    {children}
</div>
```
I think this would improve the code clarity a bit, and it precludes having to make additional unnecessary function calls.

packages/core/src/components/non-ideal-state/nonIdealState.tsx
```diff
@@ -119,21 +105,30 @@ export class NonIdealState extends AbstractPureComponent<NonIdealStateProps> {
...
+        return (
+            !!title &&
+            !!description && (
```
ggdouglas, line 115: I believe the logic here should be:
```ts
!!title || !!description
```
Otherwise, the text will only render only if _both_ title and description are set. Having just a title and no description (or vice versa) is a valid configuration for this component.

Better yet, can we be a bit more explicit with our operators here? (since this is TypeScript)
```ts
title != null || description != null
```

ggdouglas (changes requested):
This is almost ready; requesting a few minor stylistic and one functional change.

PR #6834, opened 2024-05-28
use gap instead of margin for tab

JosephChotard:
By doing this, tabs can natively support RTL without breaking any LTR

Instead of using margin for tab spacing use flex gap

- [x] Checking that it looks the same in both LTR
- [x] That it's no longer broken in RTL

#### Screenshot

Looks identical

### DIFF
--- a/packages/core/src/components/tabs/_tabs.scss
+++ b/packages/core/src/components/tabs/_tabs.scss
@@ -94,7 +94,7 @@ $tab-indicator-width: 3px !default;
   margin: 0;
   padding: 0;
   position: relative;
-  gap: $pt-grid-size * 2
+  gap: $pt-grid-size * 2;
 }
 
 .#{$ns}-tab {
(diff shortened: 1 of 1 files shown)


JosephChotard:
Existing view in RTL mode (notice margins break)
Screenshot 2024-05-28 at 16 39 39
New view:
Screenshot 2024-05-28 at 16 40 58

LTR looks identical
For these screenshots I added `direction: rtl` to a wrapper around the tab. We use RTL for Arabic / Hebrew text

JosephChotard:
Example in arabic.

Now:
Screenshot 2024-05-28 at 16 48 05

Before:
Screenshot 2024-05-28 at 16 49 18

gluxon (approved):
Thank you!

Glad to see older CSS hacks cleaned up as `gap` for flexbox properties gets wider browser usage.

This might break if something overrides the `.bp5-tab-list` display property, but Blueprint consumers shouldn't be doing that anyway.

stewartfeasby:
Hi

This completely breaks vertical tabbing now, which previously had no gap but now has massive gaps.

Thanks

gluxon:
Thanks for catching that @stewartfeasby. Putting up a fix here: https://github.com/palantir/blueprint/pull/6885

PR #6846, opened 2024-06-12
a11y(Alert): give Alert alertdialog role

bvandrc:
#### Checklist

- [N/A] Includes tests
- [N/A] Update documentation

1. Make `Dialog` `role` prop configurable (default=`"dialog"`)
2. Set `Alert`'s `Dialog`'s `role` to `alertdialog`: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/alertdialog_role


packages/core/src/components/alert/alert.tsx
```diff
@@ -172,6 +172,7 @@ export class Alert extends AbstractPureComponent<AlertProps> {
         return (
             <Dialog
                 {...overlayProps}
+                role="alertdialog"
```
evansjohnson, line 175: We may also want to set [`aria-modal="true"` when setting this role](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-modal)

---

We may also want to follow this guidance if adding this role

> The alert dialog must have at least one focusable control — such as Confirm, Close, and Cancel — and focus must be moved to that control when the alert dialog appears. Alertdialogs can have additional interactive controls such as text fields and checkboxes.

From my logging it appears focus is initially on the start of the dialog focus trap
Screenshot 2024-06-17 at 11 15 56 PM

packages/core/src/components/dialog/dialog.tsx
```diff
@@ -65,6 +65,11 @@ export interface DialogProps extends OverlayableProps, BackdropProps, Props {
      */
     isCloseButtonShown?: boolean;
 
+    /**
+     * @default "dialog"
+     */
+    role?: React.AriaRole;
```
evansjohnson: Could we have an explicit list of values we expect to be valid given the setup of this component? `"dialog" | "alertdialog"` feels like a reasonable place to start that can be expanded as needed.

packages/core/src/components/alert/alert.tsx
bvandrc: Good call. Tied `aria-modal` to the `enforceFocus` prop in `Dialog`, these seemed to be one in the same. https://github.com/palantir/blueprint/pull/6846/commits/1a9709bad0885b216f89c996f5f34273054024c0

packages/core/src/components/dialog/dialog.tsx
bvandrc: https://github.com/palantir/blueprint/pull/6846/commits/97104bbe7054b687d62fd36107e2448147f12742

evansjohnson (approved):
LGTM

PR #6847, opened 2024-06-13
fix: prevent PanelStack2 from opening in the wrong direction

### LINKED ISSUE #6723: PanelStack2's animation when opening a new panel happens in the wrong direction
#### Environment

- __Package version(s)__: @blueprintjs/core ^5.9.1
- __Operating System__: Mac OS X / stackblitz
- __Browser name and version__: Microsoft Edge Version 121.0.2277.128

#### ~~Code Sandbox~~ Stackblitz

(The code sandbox cannot be forked right now, it shows a warning about insufficient credits)

https://stackblitz.com/edit/vitejs-vite-sqvhmd?file=src%2FPanelStack.tsx

#### Steps to reproduce

Only reproducible if NOT using strict mode and in development OR with a production build.

1. Click open: panel opens from the right ✅ 
2. Click close: panel closes from the left ✅ 
3. Click open: panel opens from the left 🔴 

#### Actual behavior

The first time the panel opens from the right as it should. All subsequent times it opens from the left.

I believe this is because when the current stack of the component changes, the `direction` state is "pop" when it should be "push", until the `useEffect` sets it to "push" but it is too late to affect the transition component.

#### Expected behavior

The panel should always open from the right.

#### Possible solution

I already made a POC that seems to solve the issue. I make sure the direction is not a state but a value computed immediately in the render function based on the latest and the previous stack's length.

If this is confirmed to be a bug I'm happy to submit a PR.

stropitek:
#### Fixes #6723 

As additional context that was not described in the original issue, I realized that the problem only exists when using react 18, not with react 16.

I did an example of the issue in an application running with react 18 in stackblitz, one without the fix, the other with the fix. By testing the example as described in #6723, you can verify the fix.

Before the fix: https://stackblitz.com/edit/vitejs-vite-4d5jhz
After the fix: https://stackblitz.com/edit/vitejs-vite-txizds

#### Checklist

- [ ] Includes tests
- [ ] Update documentation

The changes removes one of the state that can be derived from other values directly, preventing unnecessary re-renders.

Making sure the change does not break any existing behaviour.

### DIFF
--- a/packages/core/src/components/panel-stack2/panelStack2.tsx
+++ b/packages/core/src/components/panel-stack2/panelStack2.tsx
@@ -19,6 +19,7 @@ import * as React from "react";
 import { CSSTransition, TransitionGroup } from "react-transition-group";
 
 import { Classes, DISPLAYNAME_PREFIX, type Props } from "../../common";
+import { usePrevious } from "../../hooks";
 
 import type { Panel } from "./panelTypes";
 import { PanelView2 } from "./panelView2";
@@ -103,12 +104,8 @@ export const PanelStack2: PanelStack2Component = <T extends Panel<object>>(props
         () => (isControlled ? propsStack.slice().reverse() : localStack),
         [localStack, isControlled, propsStack],
     );
-    const stackLength = React.useRef<number>(stack.length);
-    // Adjust the direction in case the stack size has changed, controlled or uncontrolled
-    const direction = stack.length - stackLength.current < 0 ? "pop" : "push";
-    React.useEffect(() => {
-        stackLength.current = stack.length;
-    }, [stack]);
+    const prevStackLength = usePrevious(stack.length) ?? stack.length;
+    const direction = stack.length - prevStackLength < 0 ? "pop" : "push";
 
     const handlePanelOpen = React.useCallback(
         (panel: T) => {
--- a/packages/core/src/components/panel-stack2/panelStack2.tsx
+++ b/packages/core/src/components/panel-stack2/panelStack2.tsx
@@ -97,7 +97,6 @@ export const PanelStack2: PanelStack2Component = <T extends Panel<object>>(props
         stack: propsStack,
     } = props;
     const isControlled = propsStack != null;
-    const [direction, setDirection] = React.useState("push");
 
     const [localStack, setLocalStack] = React.useState<T[]>(initialPanel !== undefined ? [initialPanel] : []);
     const stack = React.useMemo(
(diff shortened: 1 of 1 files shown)


palantirtech:
Thanks for your interest in [palantir/blueprint](https://github.com/palantir/blueprint), @stropitek! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

packages/core/src/components/panel-stack2/panelStack2.tsx
```diff
@@ -97,19 +97,16 @@ export const PanelStack2: PanelStack2Component = <T extends Panel<object>>(props
...
     const isControlled = propsStack != null;
-    const [direction, setDirection] = React.useState("push");
 
     const [localStack, setLocalStack] = React.useState<T[]>(initialPanel !== undefined ? [initialPanel] : []);
     const stack = React.useMemo(
         () => (isControlled ? propsStack.slice().reverse() : localStack),
         [localStack, isControlled, propsStack],
     );
     const stackLength = React.useRef<number>(stack.length);
+    // Adjust the direction in case the stack size has changed, controlled or uncontrolled
```
evansjohnson: nit: this comment doesn't seem to apply anymore now that the state is gone - I think the logic reads pretty explanatory without a comment "if current length is greater than previous length, pop the panel, otherwise push"

packages/core/src/components/panel-stack2/panelStack2.tsx
```diff
@@ -97,19 +97,16 @@ export const PanelStack2: PanelStack2Component = <T extends Panel<object>>(props
...
     } = props;
     const isControlled = propsStack != null;
-    const [direction, setDirection] = React.useState("push");
 
     const [localStack, setLocalStack] = React.useState<T[]>(initialPanel !== undefined ? [initialPanel] : []);
     const stack = React.useMemo(
         () => (isControlled ? propsStack.slice().reverse() : localStack),
         [localStack, isControlled, propsStack],
     );
     const stackLength = React.useRef<number>(stack.length);
```
evansjohnson: nit: related to below and existing but this is really `prevStackLength`, we could clean this up with something like

```
// default prevStackLength to current length on first render to avoid handling undefined separately
const prevStackLength = usePrevious(stack.length) ?? stack.length;
```

evansjohnson (dismissed):
Thank you for the fix! Agreed regarding deriving from props/previous props rather than stored as separate state.

Feel free to merge as is and I can open a cleanup PR since my comments aren't needed for the fix, otherwise I can re-approve if you push changes.

stropitek:
Thank you for the review @evansjohnson!

I updated the PR based on it.

packages/core/src/components/panel-stack2/panelStack2.tsx
```diff
@@ -97,21 +98,14 @@ export const PanelStack2: PanelStack2Component = <T extends Panel<object>>(props
...
     );
-    const stackLength = React.useRef<number>(stack.length);
-    React.useEffect(() => {
-        if (stack.length !== stackLength.current) {
-            // Adjust the direction in case the stack size has changed, controlled or uncontrolled
-            setDirection(stack.length - stackLength.current < 0 ? "pop" : "push");
-        }
-        stackLength.current = stack.length;
-    }, [stack]);
+    const prevStackLength = usePrevious(stack.length) ?? stack.length;
```
evansjohnson, line 107: leaving comment for posterity - technically `stack.length` is not the previous stack length, but it lets us not worry about handling the `undefined` case, and there is no animation on the initial render anyways

evansjohnson (approved):
LGTM

PR #6851, opened 2024-06-18
Do not set `tabIndex` when no popover will show

benrucker:
#### Checklist

- [x] Includes tests
- [ ] Update documentation

This PR makes `Popover`s not add a `tabIndex` to the target when the `Popover` will not render due to `disabled={true}` or `content={undefined}`. This has been done by making sure that we do _not_ override `tabIndex` on the popover target in those scenarios even if `openOnTargetFocus` is also true.

This is useful when rendering shared components that have conditional `Tooltips` implemented via the `content` or `disabled` props. Before this change, those components would always be focusable.

This PR does not address `Popover`s that are rendered in control mode and not open, though they exhibit similar behavior. This could be a follow-up change if desired.

#### Popover -> Tooltip nesting

This PR is consistent with previous behavior regarding nesting a `Tooltip` in a `Popover`.

For context, default `Popover`s do not apply a `tabIndex=0` to their targets. This is because `Popover`s are by default `PopoverInteractionKind.CLICK`, which does not meet the pre-requisite for getting `tabIndex=0` applied to the target of `PopoverInteractionKind.HOVER`.

Previously, a `Popover` with a nested `Tooltip` would always have a focusable target. Now, that will only be true if the `Tooltip` is both not disabled and has content.

This behavior has been codified in the test suite.

- If the tests make sense
    - Is sharing the button reference brittle? Should I extract the props instead, or make it a `getButton` function of some sort?
- Does this need a docs update? A further comment on the `openOnTargetFocus` TSDoc?

#### After

> ⚠️ Note that I've added new lines here for demonstration purposes to exhibit the new behavior. No changes to the docs are being proposed in this PR.

https://github.com/palantir/blueprint/assets/12519846/43efdbb4-da3d-45cd-bcd9-7f7aa4237e0d

Lines that are no longer focusable: 
- Tooltip with no content
- Disabled Tooltip
- Popover with a nested disabled Tooltip
- Disabled Popover with a nested disabled Tooltip

Lines that were never focusable:
- Popover (not disabled)

Lines that are notably still focusable:
- Popover with nested Tooltip
- Disabled Popover with a nested Tooltip

### DIFF
--- a/packages/core/src/components/popover/popover.tsx
+++ b/packages/core/src/components/popover/popover.tsx
@@ -344,7 +344,7 @@ export class Popover<
     public reposition = () => this.popperScheduleUpdate?.();
 
     private renderTarget = ({ ref: popperChildRef }: ReferenceChildrenProps) => {
-        const { children, className, fill, openOnTargetFocus, renderTarget } = this.props;
+        const { children, className, content, disabled, fill, openOnTargetFocus, renderTarget } = this.props;
         const { isOpen } = this.state;
         const isControlled = this.isControlled();
         const isHoverInteractionKind = this.isHoverInteractionKind();
@@ -375,7 +375,8 @@ export class Popover<
                       onKeyDown: this.handleKeyDown,
                   };
         // Ensure target is focusable if relevant prop enabled
-        const targetTabIndex = openOnTargetFocus && isHoverInteractionKind ? 0 : undefined;
+        const targetTabIndex =
+            content != null && !disabled && openOnTargetFocus && isHoverInteractionKind ? 0 : undefined;
         const ownTargetProps = {
             // N.B. this.props.className is passed along to renderTarget even though the user would have access to it.
             // If, instead, renderTarget is undefined and the target is provided as a child, this.props.className is
--- a/packages/core/src/components/popover/popover.tsx
+++ b/packages/core/src/components/popover/popover.tsx
@@ -249,11 +249,10 @@ export class Popover<
     }
 
     public render() {
-        const { disabled, content, placement, position = "auto", positioningStrategy } = this.props;
+        const { disabled, placement, position = "auto", positioningStrategy } = this.props;
         const { isOpen } = this.state;
 
-        const isContentEmpty = content == null || (typeof content === "string" && content.trim() === "");
-        if (isContentEmpty) {
+        if (this.getIsContentEmpty()) {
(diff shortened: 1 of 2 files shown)


packages/core/src/components/popover/popover.tsx
```diff
@@ -375,7 +375,8 @@ export class Popover<
                       onKeyDown: this.handleKeyDown,
                   };
         // Ensure target is focusable if relevant prop enabled
-        const targetTabIndex = openOnTargetFocus && isHoverInteractionKind ? 0 : undefined;
+        const targetTabIndex =
+            content != null && !disabled && openOnTargetFocus && isHoverInteractionKind ? 0 : undefined;
```
evansjohnson: could you split out and reuse this check?
`const isContentEmpty = content == null || (typeof content === "string" && content.trim() === "");`

evansjohnson (changes requested):
Thank you for the PR!

This is looking good, I've encountered this issue in the wild and believe I've done annoying things like optionally wrap `Tooltip` when not disabled to avoid this. Let's just get the empty check cleaned up to support the `trim` case then I think this should be ready to merge.

---

for good measure I also tested

disabling tooltip (popover) shouldn't override already focusable target element 
```
<Tooltip
    disabled={true}
    content="tooltip!"
    intent="danger"
    placement="right"
>
    <Button onClick={() => console.log("hello world")}>I should still be focusable</Button>
</Tooltip>
```

keeps existing `tabIndex`
```
<Tooltip
    disabled={true}
    content="tooltip!"
    intent="danger"
    placement="right"
>
    <div tabIndex={0}>target text, not sure why anyone would define tabIndex here but...</div>
</Tooltip>
```
Screenshot 2024-06-26 at 3 48 34 PM

and in the `renderTarget` case the consumer is fully responsible for rendering anyways - if they accept our "recommended" `tabIndex`, they'll get this new behavior, otherwise they can always do whatever they want.

packages/core/src/components/popover/popover.tsx
benrucker: Okay! I pulled that into a private class method. Let me know if you envisioned a different way to share it between `renderTarget` and `render`. Thanks for the review!

evansjohnson (dismissed):
thank you for the fix!

evansjohnson (approved):

PR #6853, opened 2024-06-18
Don't show slider fill before value equal to initial value

### LINKED ISSUE #6844: Slider fill with negative min, positive max, initialValue of 0 has the wrong track fill at zero
#### Steps to reproduce

1. configure a Slider with a negative min and positive max (or, any case where `min < initialValue && max > initialValue && value === initialValue`)
3. move the Slider handle to zero (or if not using the default initialValue, then move the Slider handle to `initialValue`)

#### Actual behavior
The track fills from min to zero (or `initialValue`) with the primary intent. 

#### Expected behavior
No track fill, since it's only supposed to be between initialValue and value. 

#### Possible solution

the issue is that [here](https://github.com/palantir/blueprint/blob/45461a2707c01797b022b9922eb7d717da1a9fa0/packages/core/src/components/slider/slider.tsx#L73) when the slider's value equals the initialValue, the first slider handle created has an intentBefore defined. Then the track fill is created from three segments:
min to first slider handle [-100% to 0%]
first slider handle to initial value [0% to 0%]
initial value to max [0% to 100%]
Since the first segment's endpoint has an intentBefore , it gets filled in, in this case with the primary intent 

It seems like it should only set an intentBefore if value > initialValue and not if value === initialValue.

evansjohnson:
#### Checklist

- [x] Includes tests
- [n/a] Update documentation

I can't think of a reason a user would want the track to be filled from min to value when value === initial value, but then have value to initial value filled when value is just slightly less.

I imagine this case just wasn't explicitly considered, since often min may equal initial value and this issue is not apparent.

If there's any valid use case for this track fill behavior.

#### Screenshot

Issue observable on slider example if showing fill for bottom example
Screenshot 2024-06-18 at 10 19 00 AM
Screenshot 2024-06-18 at 10 18 57 AM
Screenshot 2024-06-18 at 10 18 54 AM

### DIFF
--- a/packages/core/src/components/slider/slider.tsx
+++ b/packages/core/src/components/slider/slider.tsx
@@ -70,7 +70,7 @@ export class Slider extends AbstractPureComponent<SliderProps> {
                 <MultiSlider.Handle
                     value={value!}
                     intentAfter={value! < initialValue! ? intent : undefined}
-                    intentBefore={value! >= initialValue! ? intent : undefined}
+                    intentBefore={value! > initialValue! ? intent : undefined}
                     onChange={onChange}
                     onRelease={onRelease}
                     htmlProps={handleHtmlProps}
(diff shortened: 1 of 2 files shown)


gluxon (approved):
> I can't think of a reason a user would want the track to be filled from min to value when value === initial value, but then have value to initial value filled when value is just slightly less.

Agree. Thanks for this fix!

Issue #6076, opened 2023-04-19
Consider using `:focus-visible` instead of `:focus` for styling focus states


ggdouglas:
### Summary

With [`:focus-visible`](https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-visible) now being supported across [all modern browsers](https://caniuse.com/?search=focus-visible), it is possible to replace older JS-based techniques for managing cross-browser focus styles (such as [What Input?](https://github.com/ten1seven/what-input) or Blueprint's [FocusStyleManager](https://blueprintjs.com/docs/#core/accessibility.focus-management)) with a browser-native solution. This would involve updating the global styles applied to `:focus` here: https://github.com/palantir/blueprint/blob/d7e8bc67fc1657b6943838b846381fc05e148f59/packages/core/src/accessibility/_focus-states.scss#L4-L6

Doing so would allow users of Blueprint to _opt-in_ to focus rings via `FocusStyleManager.alwaysShowFocus()` rather than always having to manually _opt-out_ of seeing focus rings on click via `FocusStyleManager.onlyShowFocusOnTabs()`.

### Context

If the `FocusStyleManager` is not initialized/included along with Blueprint, focus rings will appear around elements when clicked. These focus rings outlines, while vital to accessibility for keyboards and other non-pointing devices, have long been considered undesirable when applied as a result of a mouse/pointer interaction.

![focus-ring-click-example](https://user-images.githubusercontent.com/2968519/232974182-d6a381f7-d6cd-40f5-ad19-24a14a5c527e.gif)
> Example of a focus ring appearing, undesirably, on click

---

| <video src="https://user-images.githubusercontent.com/2968519/232971947-0f2a548b-a016-4fc6-8ab5-29ad9263a87a.mp4">  | <video src="https://user-images.githubusercontent.com/2968519/232971942-a1e01b28-9afc-4754-ac17-fe9d093e3f2a.mp4">|
| ------------- | ------------- |
| The blue focus ring outlines **_are not_** present when clicking on elements. | The blue focus ring outlines _**are**_ present when clicking on elements. |

---

Chrome, Firefox, and Safari have all replaced `:focus` with `:focus-visible` in their user agent stylesheets. These changes occurred in February 2021 for both Chrome and Firefox and December 2021 for Safari. This means that, by default, these browsers will no longer show focus rings on many elements, such as buttons, where they did previously (see the [spec](url) for `:focus-visible`)
[...]

PR #6854, opened 2024-06-18
Update focus outline color to meet WCAG 2.2 contrast standards

evansjohnson:
#### Checklist

- [n/a] Includes tests
- [x] Update documentation

Update color of focus outline to meet 3:1 contrast standard for UI component focus states

https://webaim.org/resources/contrastchecker/

- Our default focus indicator won't meet contrast standards for arbitrary colors - but should we provide any buffer so that the default focus indicator meets standards up to a certain shade of background color, rather than just `$white`/`$black`?
    - I currently went with an option that meets contrast standards against all light-gray colors in light theme, and all dark-gray colors when in dark theme
- Should be good in light and dark themes
- [Design review checklist](https://github.com/palantir/blueprint/pull/6854#issuecomment-2200510897)

#### WCAG Guidelines
From what I can tell, WCAG [did not require minimum contrast in 2.1](https://www.w3.org/WAI/WCAG21/Understanding/focus-visible), but in [2.2 has a section on focus appearance](https://www.w3.org/WAI/WCAG22/Understanding/focus-appearance.html) which specifies the 3:1 minimum ratio

#### Screenshot
Current:
Screenshot 2024-06-21 at 8 28 04 AM
Screenshot 2024-06-21 at 8 27 54 AM

This PR:
Screenshot 2024-06-21 at 8 28 04 AM
Screenshot 2024-06-21 at 8 27 54 AM


evansjohnson:
Using [RGBA -> RGB converter](https://borderleft.com/toolbox/rgba/) and putting that result into a [contrast checker](https://webaim.org/resources/contrastchecker/).

Note after aligning on an option I will look into replicating a similar decision for dark theme. For now I will talk in terms of light theme to stay explicit.

### Option 1 - only optimize for white background, bump opacity from 0.6 to 0.742
Screenshot 2024-06-18 at 10 45 19 AM
Screenshot 2024-06-18 at 10 45 25 AM

### Option 2 - contrast buffer to support first level of light gray ($light-gray5) background, opacity to 0.78
Screenshot 2024-06-18 at 11 01 51 AM
Screenshot 2024-06-18 at 11 01 47 AM

### Option 3 - contrast buffer to support third level of light gray ($light-gray3) background, opacity to 0.862
Screenshot 2024-06-18 at 11 00 34 AM
Screenshot 2024-06-18 at 11 00 30 AM

### Option 4 - contrast buffer to support all light gray colors, opacity to 0.95
Screenshot 2024-06-18 at 11 03 55 AM
Screenshot 2024-06-18 at 11 03 40 AM

packages/core/src/components/forms/_common.scss
```diff
@@ -100,6 +102,10 @@ $control-group-stack: (
   transition: $input-transition;
   vertical-align: middle;
 
+  .bp5-focus-enhanced-contrast & {
+    outline: none;
```
evansjohnson: it might be better to just save this for a breaking change in a future release, or decide it's not enough of a breaking change to hold on and just release this now.

I'm a bit worried about the class based approach since there will be specificity differences between the class based approach, and later changing the color without the class.

packages/core/src/components/forms/_controls.scss
```diff
@@ -197,7 +197,7 @@ $control-indicator-spacing: $pt-grid-size !default;
 
   input:focus ~ .#{$ns}-control-indicator {
     @include focus-outline();
-    outline: $blue3 solid 2px;
+    outline: $blue3 solid 2px; // remove this line for bp6 for a consistent focus outline style
```
evansjohnson: now that the contrast is bumped up by default, I think we can get rid of this override as part of a breaking change

evansjohnson (commented):
this feels like a borderline breaking change, since it has the potential to make UI snapshots fail. I don't think it could cause some cases to have worse contrast because the current focus color already does not have enough contrast with the extremes of the color spectrum

evansjohnson:
Screenshot 2024-06-21 at 5 28 53 PM

trying to figure out intent inputs before marking this ready to review

packages/core/src/common/_color-aliases.scss
```diff
@@ -25,6 +25,11 @@ $pt-app-elevated-background-color: $light-gray4 !default;
 $pt-dark-app-elevated-background-color: $dark-gray3 !default;
 
 $pt-outline-color: rgba($blue3, 0.6) !default;
+// 0.752 opacity to meet WCCAG 2.2 minimum contrast guidelines with all light-gray backgrounds in light theme
+// and all dark-gray backgrounds in dark theme
+$pt-focus-indicator-color: rgba($blue2, 0.752) !default;
+// between $blue4 and $blue5
+$pt-dark-focus-indicator-color: rgba(#7bb0fb, 0.752) !default;
```
invliD: Can we base this off of an actual BP color? We should avoid using any other colors.

packages/core/src/common/_mixins.scss
```diff
@@ -91,6 +91,17 @@ $pt-dark-intent-text-colors: (
...
+  // this should become default in bp6, providing new behavior as a class for now to avoid breaking changes
+  // to focus colors
+  /* stylelint-disable-next-line @blueprintjs/no-prefix-literal */
+  .bp5-focus-enhanced-contrast & {
+    outline: $pt-focus-indicator-color solid 2px;
+  }
+  /* stylelint-disable-next-line @blueprintjs/no-prefix-literal */
+  .#{$ns}-dark.bp5-focus-enhanced-contrast & {
+    outline-color: $pt-dark-focus-indicator-color;
+  }
```
invliD: I think we should just do it. I wouldn't call this a breaking change. If we do introduce this class, you should use `#{$ns}` instead of `bp5`, which is what this lint rule is trying to tell you.

packages/core/src/docs/accessibility.md
```diff
@@ -59,4 +59,14 @@ const MyComponent = () => ({
 
 Colors have been designed to be accessible to as many people as possible, even those who are
 visually impaired or experiencing any kind of colorblindness. Our colors have not only been chosen
-to go well together but to also adhere to [WCAG 2.0](https://www.w3.org/TR/WCAG20/) standards.
+to go well together, but to also adhere to [WCAG 2.0](https://www.w3.org/TR/WCAG20/) standards.
+
+@### Enhanced focus indication [WCAG 2.2 focus appearance](https://www.w3.org/WAI/WCAG22/Understanding/focus-appearance.html)
+
+All Blueprint elements descending from an element with the `bp5-focus-enhanced-contrast` class, will
+received enhanced focus indication. With this class applied at the root, the focus contrast meets the
```
invliD: The root part is technically inaccurate, as it doesn't have to be the root (as the first part of the sentence mentions).

packages/core/src/common/_mixins.scss
evansjohnson: Okay cool, I wasn't sure the history of how careful we've been with minor things like this that could be viewed as a break.

Ack on the prefix - I hard coded bp5 since this was only intended to work for the bp5 version and removed in bp6, but will just remove this.

packages/core/src/docs/accessibility.md
evansjohnson: yea, I also noticed it could be an issue if the dark theme class wasn't applied on the same element

just making the change outright with no transition class in bp5 should make this simpler

packages/core/src/common/_color-aliases.scss
evansjohnson: We use opacity so frequently it seems we don't really have a fixed color palette anyways, but happy to update to `$blue5` which should have slightly more contrast against dark theme backgrounds. I thought it was a little bright initially, but I'm seeing we were using `$blue5` with no opacity as the outline color specifically for `Switch`s in dark theme, so there's some precedent.

evansjohnson (commented):
TODO:
- [x] Audit intent color contrast, `pt-input-intent` and `pt-dark-input-intent`
    - [x] $pt-intent-colors
    - [x] $pt-dark-input-intent-box-shadow-colors
- [x] Audit focus outline of all components (planning to leave out of this PR) - [a few more here](https://github.com/palantir/blueprint/issues?q=sort%3Aupdated-desc+is%3Aissue+is%3Aopen+label%3A%22Domain%3A+accessibility%22+author%3Aevansjohnson) but highlighting some main ones:
    - [-] [`Card` no focus state](https://github.com/palantir/blueprint/issues/6870)
    - [-] [`FileInput` no visual focus state](https://github.com/palantir/blueprint/issues/6859)
    - [-] [`Tree` items cannot be focused](https://github.com/palantir/blueprint/issues/6860)
    - [-] Tag focus outline does not have offset like other components 
Screenshot 2024-06-25 at 3 06 11 PM

- [-] Check on history of why components that appear like text inputs have a shadow focus style while other components have offset outline style
    - For now, went with an option that retains `box-shadow` for text inputs, but styled in a way that makes it look closer to a solid outline. I figured this reduces chances of regressions since this will just change the color/opacity used but not the method of adding a focus indicator for these components.
- [x] Consider https://github.com/palantir/blueprint/issues/6076 along with this PR, or as a follow up in the same release
    - This was already tagged a breaking change - we could consider at the next major version bump but not needed for now.

packages/core/src/common/_color-aliases.scss
```diff
@@ -25,6 +25,10 @@ $pt-app-elevated-background-color: $light-gray4 !default;
 $pt-dark-app-elevated-background-color: $dark-gray3 !default;
 
 $pt-outline-color: rgba($blue3, 0.6) !default;
```
evansjohnson, line 27: Not changing this color because of the comment at the top of this file stating that is generally considered a breaking change. Perhaps we want to make an exception here...

packages/core/src/common/_color-aliases.scss
```diff
@@ -25,6 +25,11 @@ $pt-app-elevated-background-color: $light-gray4 !default;
...
+// 0.752 opacity to meet WCCAG 2.2 minimum contrast guidelines with all light-gray backgrounds in light theme
+// and all dark-gray backgrounds in dark theme
+$pt-focus-indicator-color: rgba($blue2, 0.752) !default;
```
evansjohnson, line 30: $blue2 on $white - [3.77:1 contrast](https://webaim.org/resources/contrastchecker/?fcolor=215DB0C0&bcolor=FFFFFF)
$blue2 on $light-gray1 - [3:1 contrast](https://webaim.org/resources/contrastchecker/?fcolor=215DB0c0&bcolor=D3D8DE)

packages/core/src/common/_color-aliases.scss
```diff
@@ -25,6 +25,11 @@ $pt-app-elevated-background-color: $light-gray4 !default;
...
+// dark theme opacity could be less and still meet minimum contrast, but keep 0.752 to stay consistent
+$pt-dark-focus-indicator-color: rgba($blue5, 0.752) !default;
```
evansjohnson, line 32: $blue5 on $black - [5.77:1 contrast](https://webaim.org/resources/contrastchecker/?fcolor=8ABBFFc0&bcolor=111418)
$blue5 on $dark-gray5 - [3.39:1 contrast](https://webaim.org/resources/contrastchecker/?fcolor=8ABBFFC0&bcolor=404854)

---

for comparison - $blue4 on $dark-gray5 - [2.24:1 contrast](https://webaim.org/resources/contrastchecker/?fcolor=4C90F0C0&bcolor=404854)

packages/core/src/components/forms/_common.scss
```diff
@@ -15,14 +15,14 @@ $input-color-disabled: $button-color-disabled !default;
 $input-placeholder-color: $pt-text-color-muted !default;
 $input-background-color: $white !default;
 $input-background-color-disabled: rgba($light-gray1, 0.5) !default;
-$input-shadow-color-focus: $pt-intent-primary !default;
+$input-shadow-color-focus: $pt-focus-indicator-color;
```
evansjohnson, line 18: if this can't be relied on externally - remove and use `$pt-focus-indicator-color` directly in the handful of places it's used. Same for dark theme version below.

evansjohnson:
Intent contrast ratios against different background colors:
Screenshot 2024-07-01 at 10 33 28 AM



Color | Hex | BP color | Contrast with $white/$black at 0.752 opacity | Contrast with $light-gray-1/$dark-gray-5 at 0.752 opacity | Contrast with $white/$black at 1 opacity | Contrast with $light-gray-1/$dark-gray-5 at 1 opacity | Contrast with $light-gray-2/$dark-gray-4 at 1 opacity | Contrast with $light-gray-4
-- | -- | -- | -- | -- | -- | -- | -- | --
Light primary | #2d72d2 | $blue3 | 3.06 | 2.42 | 4.72 | 3.29 | 3.55 |  
Light success | #238551 | $green3 | 3.01 | 2.36 | 4.62 | 3.22 | 3.48 |  
Light warning | #c87619 | $orange3 | 2.47 | 1.95 | 3.46 | 2.41 | 2.61 | 3
Light danger | #cd4246 | $red3 | 3.19 | 2.53 | 4.7 | 3.28 | 3.54 |  
Dark primary | #4c90f0 | $blue4 | 3.76 | 2.24 | 5.76 | 2.88 | 3.36 |  
Dark success | #32a467 | $green4 | 3.79 | 2.26 | 5.84 | 2.92 | 3.4 |  
Dark warning | #ec9a3c | $orange4 | 5.07 | 2.98 | 8.13 | 4.07 | 4.75 |  
Dark danger | #e76a6e | $red4 | 3.81 | 2.25 | 5.87 | 2.94 | 3.43 |

packages/core/src/components/forms/_common.scss
evansjohnson: it looks like we explicitly distribute `_colors.scss`, `_color-aliases` and `_variables`, but not this file. could consider updating to use `$pt-focus-indicator-color` directly

`"dist:variables": "generate-css-variables --retainDefault true ../../colors/src/_colors.scss common/_color-aliases.scss common/_variables.scss",`

packages/core/src/components/forms/_common.scss
```diff
@@ -56,27 +56,33 @@ $control-group-stack: (
...
+  $color: $input-shadow-color-focus,
+  $focus: false,
+  $outer-shadow-focus-width: 1px,
+) {
   @if $focus {
     @return
-      inset border-shadow(1, $color, 1px),
-      border-shadow(0.3, $color, 2px);
+      inset border-shadow(0.752, $color, 1px),
+      border-shadow(0.752, $color, $outer-shadow-focus-width);
```
evansjohnson, line 67: border-shadow width stayed the same in the case of intent shadows (since they already start with 1px in un-focused state, the focus state is a 3px shadow outline)

for non-intent inputs, the shadow width reduced by 1px. I expect this to mean that shadows that worked in previous situations will continue to work (ex there won't be clipping introduced from something like a new outline offset or wider shadow)

packages/core/src/components/forms/_controls.scss
```diff
@@ -197,7 +197,6 @@ $control-indicator-spacing: $pt-grid-size !default;
 
   input:focus ~ .#{$ns}-control-indicator {
     @include focus-outline();
-    outline: $blue3 solid 2px;
```
evansjohnson, line 200: from what I can tell this was just to provide increased contrast, no reason to not use the consistent `focus-outline` style now

packages/core/src/components/forms/_controls.scss
```diff
@@ -632,10 +631,6 @@ $control-indicator-spacing: $pt-grid-size !default;
       background-color: $dark-control-background-color-hover;
     }
 
-    input:focus ~ .#{$ns}-control-indicator {
-      outline: $blue5 solid 2px;
```
evansjohnson, line 636: same comment as line 200 in this file

packages/core/src/common/_color-aliases.scss
evansjohnson, line 1: colors have changed - but not the method of how the focus indicator is applied (2px offset 2px outline) so I expect minimal chance of breaking changes

evansjohnson:
Compare [public docs](https://blueprintjs.com/docs/#core) vs latest "documentation" link from latest @svc-palantir-github comment

Design review checklist:
- [ ] Light theme
    - [ ] Existing high contrast focus style components like `Checkbox`, `Radio`, `Switch`
    - [ ] Outline focus style on a few interactive components, like `Button`, `Menu`, `Tabs`
    - [ ] Text input styled components like `InputGroup`, `NumericInput`, `EditableText`
    - [ ] Text input styled components with `intent`
- [ ] Dark theme
    - [ ] Existing high contrast focus style components like `Checkbox`, `Radio`, `Switch`
    - [ ] Outline focus style on a few interactive components, like `Button`, `Menu`, `Tabs`
    - [ ] Text input styled components like `InputGroup`, `NumericInput`, `EditableText`
    - [ ] Text input styled components with `intent`

alex-dales:
> 
> Design review checklist:
> 
> * [x]  Light theme
>   
>   * [x]  Existing high contrast focus style components like `Checkbox`, `Radio`, `Switch`
>   * [x]  Outline focus style on a few interactive components, like `Button`, `Menu`, `Tabs`
>   * [x]  Text input styled components like `InputGroup`, `NumericInput`, `EditableText`
>   * [x]  Text input styled components with `intent`
> * [x]  Dark theme
>   
>   * [x]  Existing high contrast focus style components like `Checkbox`, `Radio`, `Switch`
>   * [x]  Outline focus style on a few interactive components, like `Button`, `Menu`, `Tabs`
>   * [x]  Text input styled components like `InputGroup`, `NumericInput`, `EditableText`
>   * [x]  Text input styled components with `intent`

**Looks good! Some comments and questions below:**

### Comments

- Unrelated to the contrast changes, but I noticed that with the `segmented control` when an option between two other options is selected and we focus the left-most option, the focus border is behind the selected option. When we focus the right-most option this isn't the case. Could we fix the the focus for the left-most option here to not be clipped? 

**Left-most option focused**
Screenshot 2024-07-01 at 3 19 20 PM

**Right-most option focused**
Screenshot 2024-07-01 at 3 19 25 PM


- For a `numeric input` with `intent` and in `light mode` I noticed the input field overlaps the focus border of the right buttons. Could we fix this to bring the focus border to the front?

Screenshot 2024-07-01 at 3 54 31 PM


### Questions

- Do `submenus` intentionally have a non-rounded focus border? I was expecting that they'd also be rounded to be consistent with the `focus` state for `menu items`

Screenshot 2024-07-01 at 3 21 04 PM

- Was just curious what was the rationale behind moving from the two-colored stroke to the single colored stroke for `input fields`? I do think it looks cleaner. Just wanted to flag that we still have the previous visual treatment for the border when the `control card` is in the `selected` state. I realize it is a different state, but should we also consider updating the `control card` here to be consistent? 

**Previous**
Screenshot 2024-07-01 at 3 40 37 PM

**New**
Screenshot 2024-07-01 at 3 40 28 PM

**Control card option selected**
Screenshot 2024-07-01 at 3 42 32 PM

evansjohnson:
Ah great find on the `ControlCard`. I'll get that updated and do another second pass myself and look for others that may not be up to date.

---

Since this is already a delicate change to get released, I'd strongly prefer to track all of the border overlap and other issues noted separately, but am happy to continue to work on them, thanks for flagging explicit examples! For this PR I want to keep a close eye on making sure we don't have any regressions, but would expect existing bugs to remain in place for now. If any of the bugs are particularly bad in your opinion, and perhaps become even more obvious with the increased contrast, I could fix as needed as pre-requisites to this PR.

---

My thinking on moving to a solid border color is that we need to get much darker and 2px width to meet minimum contrast standards, and having 0.752 opacity right next to 1 opacity looks a little off in a way the previous light shadow didn't.

alex-dales:
> Ah great find on the ControlCard. I'll get that updated and do another second pass myself and look for others that may not be up to date.

Sounds good!
<br />

> ...perhaps become even more obvious with the increased contrast, I could fix as needed as pre-requisites to this PR

I don't think these bugs are particularly blocking, so I think it is ok to tackle this in a FLUP PR so you can get these changes in.
<br />

> My thinking on moving to a solid border color is that we need to get much darker and 2px width to meet minimum contrast standards...

Makes sense, I think the solid border also looks cleaner!

evansjohnson:
From the [demo app](https://output.circle-artifacts.com/output/job/6670a2a4-76bc-40b2-aaa6-aa621a2d9373/artifacts/0/packages/demo-app/dist/index.html) it looks like there's some custom toast focus styles that look like they don't have high contrast
Screenshot 2024-07-01 at 6 24 11 PM

confirmed here:
Screenshot 2024-07-01 at 6 59 18 PM


---

current contrast bumping [$white focus opacity](https://github.com/palantir/blueprint/blob/9754f3d6eda3889bac7b086c2cedb206bd1a30d9/packages/core/src/components/toast/_toast.scss#L50) to 0.75 from 0.5:
Screenshot 2024-07-01 at 6 51 01 PM
Screenshot 2024-07-01 at 6 50 41 PM
Screenshot 2024-07-01 at 6 50 23 PM
Screenshot 2024-07-01 at 6 50 03 PM

---

updating toast focus color to match text color with 0.75 opacity
Screenshot 2024-07-01 at 6 53 55 PM

Screenshot 2024-07-01 at 6 57 20 PM
Screenshot 2024-07-01 at 6 57 11 PM

evansjohnson:
I'm wondering if we're conflating focus and selection styles and may be anchoring on the focus state because that's what this PR is changing. FWIW [that style is not reused and is unique to `ControlCard`](https://github.com/palantir/blueprint/blob/develop/packages/core/src/components/card/_card.scss#L75) so this is just limited to the relatively new `ControlCard` component.

---

There is an argument to be made that the current selection state already follows the current focus state, so if we're updating one we should update the other. But I could also see this staying the same, since from what I've read so far this is a guideline for focus state which is more transient than a selection state and needs to grab the users attention.

alex-dales:
Toast focuses look good! 

---

Yeah agreed, I don't want to over-index on the control card selection state since _it is different_. After mocking up the 2px border for the selection state on the control card, I do like visually how the solid border looks – it is almost more attention-grabbing compared to the current. I'd be curious to get some more thoughts from the design team here, but this shouldn't be blocking for this PR!

Screenshot 2024-07-02 at 11 14 43 AM

evansjohnson:
Thanks for highlighting those existing issues! I opened #6890 and #6889 to track

evansjohnson:
Let me know if you end up wanting to change selected styles and we can make sure there's at least a ticket open to track

packages/core/src/components/toast/_toast.scss
```diff
@@ -47,7 +47,7 @@ $toast-intent-colors: (
 
     &:focus {
       // blue outline color shows poorly on colored bg
-      outline-color: rgba($white, 0.5);
+      outline-color: rgba($text-color, 0.75);
```
invliD: Was this intentional? Seems odd to use the text color for the outline.

invliD (dismissed):

packages/core/src/components/toast/_toast.scss
evansjohnson: I left an exploration of this here: https://github.com/palantir/blueprint/pull/6854#issuecomment-2201197249

The text color is what allows us to easily flip to a dark outline when rendering against a light color background. Alternate option to define a separate focus color variable, but I'm not sure what we'd really choose differently - these are already against intent backgrounds so it's a bit of a weird case.

alex-dales:
> Let me know if you end up wanting to change selected styles and we can make sure there's at least a ticket open to track

Sorry, I missed this last week! Will let you know if we want to make a change here and FLUP as needed over Slack.

packages/core/src/components/toast/_toast.scss
invliD: I'm not saying the color itself is weird, and choosing the same color seems fine. It's just odd to use the variable for _text_ color for something that isn't text.

evansjohnson: ah yes - updated to `$text-and-focus-color`, didn't think of anything good to combine them generically, ex `$content-color` doesn't sound right

invliD (approved):

PR #6858, opened 2024-06-24
Fix end time selection before selecting date

### LINKED ISSUE #6856: "Allow single day range" not respected if using end time picker before selecting date
Screenshot 2024-06-24 at 8 30 34 AM

#### Steps to reproduce
- https://blueprintjs.com/docs/#datetime2/date-range-picker3
- Note "Allow single day range" is unchecked
- Set precision to minute
- Select today's date
- Modify time of end date
- See single date range selection

When fixing https://github.com/palantir/blueprint/issues/6483 this will likely become the behavior for any date, not just selecting today's date, since using the selected start date as a default for the end date when selecting a time is likely more applicable.

#### Actual behavior
If selecting today's date, modifying the time of the end date causes the date range to be a single day, even if the `allowSingleDayRange` prop is `true`.

#### Expected behavior
Should not be able to select an end date equal to the start date when `allowSingleDayRange` is not `true`.

#### Possible solution
Add one day to the default value when `allowSingleDayRange` is not `true`.

### LINKED ISSUE #6863: Picking a second time for a single day changes the date
#### Environment

- __Package version(s)__: 
- __Operating System__: macOs Venture 13.6
- __Browser name and version__: Google Chrome Version 118.0.5993.88 (Official Build) (arm64)

#### Code Sandbox

[Link to a minimal repro - can see it in the docs
](https://blueprintjs.com/docs/#datetime2/date-range-picker3)
#### Steps to reproduce

1. Enable the time picker (doesn't matter if allowSingleDayRange is on or off)
2. Click on a single day in the DateRangePicker3
3. Change its end time

#### Actual behavior

The second date in the range becomes today's date at the time you selected

#### Expected behavior

The second date in the range becomes the date you selected at the time you selected. Also, the selection highlighting sometimes looks wrong - the whole month has the pale blue selection and all of the dates in between the date you selected and the current day are dark blue. 

#### Possible solution

The second date in the range is treated as null when only one date has been clicked (or if allowSingleDayRange is on, when only one date has been clicked one time -- if you click it a second time this behavior does not occur). The TimePicker then attempts to modify a null date, so it falls back to today's date based on [this line of code](https://github.com/palantir/blueprint/blob/5ba7594729977b385770bc815f0df11514f922b2/packages/datetime2/src/components/date-range-picker3/dateRangePicker3.tsx#L266C9-L266C9). I think changing this so it does something like "grab the date at the other index if this one is null but the time isn't null" would fix it.

evansjohnson:
#### Fixes https://github.com/palantir/blueprint/issues/6856

Previously, users could end up with a reversed selection range if modifying the time input of the end date before selecting a date. Today's date was used as the default in these cases, leaving the user with a reversed range if the start date was later than today's date.

#### Checklist
- [x] Includes tests
- [-] Update documentation

- Fix behavior of default date such that selected range cannot end up reversed.
- When `allowSingleDayRange` is not `true`, ensure the default date is 1 day off from the already selected date (otherwise keep today's date as default if no dates are selected).

- There's a slight behavior break in that users could previously select a single day range by modifying the time with today's date selected as the start date. This seems like a bug that's okay to fix but would like a second opinion on this.
- Any better solutions than to add/subtract a day from the default when `allowSingleDayRange` is not `true`?
- Would you prefer to see the `allowSingleDayRange` change split out?

### DIFF
--- a/packages/datetime2/src/components/date-range-picker3/dateRangePicker3.tsx
+++ b/packages/datetime2/src/components/date-range-picker3/dateRangePicker3.tsx
@@ -15,7 +15,7 @@
  */
 
 import classNames from "classnames";
-import { format, addDays } from "date-fns";
+import { addDays, format } from "date-fns";
 import * as React from "react";
 import type { DateFormatter, DayModifiers, DayMouseEventHandler, ModifiersClassNames } from "react-day-picker";
 
--- a/packages/datetime2/src/components/date-range-picker3/dateRangePicker3.tsx
+++ b/packages/datetime2/src/components/date-range-picker3/dateRangePicker3.tsx
@@ -292,7 +292,7 @@ export class DateRangePicker3 extends DateFnsLocalizedComponent<DateRangePicker3
         }
 
         return otherDate;
-    }
+    };
 
     private handleTimeChangeLeftCalendar = (time: Date) => {
         this.handleTimeChange(time, 0);
--- a/packages/datetime2/src/components/date-range-picker3/dateRangePicker3.tsx
+++ b/packages/datetime2/src/components/date-range-picker3/dateRangePicker3.tsx
@@ -277,10 +277,9 @@ export class DateRangePicker3 extends DateFnsLocalizedComponent<DateRangePicker3
         this.setState({ value: newDateRange, time: newTimeRange });
     };
 
-    // When a user sets the time value before choosing a date, we need to pick a date for
-    // them to have something to store the time on.
+    // When a user sets the time value before choosing a date, we need to pick a date for them
     // The default depends on the value of the other date since there's an invariant
-    // that the left/0 date is always less than the right/1 date.
+    // that the left/0 date is always less than the right/1 date
     private getDefaultDate = (dateIndex: number) => {
         const { value } = this.state;
         const otherIndex = dateIndex === 0 ? 1 : 0;
--- a/packages/datetime2/src/components/date-range-picker3/dateRangePicker3.tsx
+++ b/packages/datetime2/src/components/date-range-picker3/dateRangePicker3.tsx
@@ -15,7 +15,7 @@
  */
 
 import classNames from "classnames";
-import { format } from "date-fns";
+import { format, addDays } from "date-fns";
 import * as React from "react";
 import type { DateFormatter, DayModifiers, DayMouseEventHandler, ModifiersClassNames } from "react-day-picker";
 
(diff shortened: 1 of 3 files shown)


packages/datetime2/src/components/date-range-picker3/dateRangePicker3.tsx
```diff
@@ -265,17 +265,35 @@ export class DateRangePicker3 extends DateFnsLocalizedComponent<DateRangePicker3
...
         newDateRange[dateIndex] = newValue;
         const newTimeRange: DateRange = [time[0], time[1]];
         newTimeRange[dateIndex] = newTime;
+
         this.props.onChange?.(newDateRange);
         this.setState({ value: newDateRange, time: newTimeRange });
     };
 
+    private getDefaultDate = (dateIndex: number) => {
+        const { value } = this.state;
```
gluxon, line 284: nitpick: I'm not too worried about this since this is already happening in the existing code, but I wonder if we need to grab existing state in the `setState` callback rather than reading directly off of `this.state`.

Otherwise we could end up with weird bugs if `this.state` is referring to stale values.

In general, any time "next state" is computed from some kind of prior state, I usually read the prior state in a `setState` callback.

packages/datetime2/src/components/date-range-picker3/dateRangePicker3.tsx
```diff
@@ -265,17 +265,35 @@ export class DateRangePicker3 extends DateFnsLocalizedComponent<DateRangePicker3
...
         const newDateRange: DateRange = [value[0], value[1]];
         newDateRange[dateIndex] = newValue;
         const newTimeRange: DateRange = [time[0], time[1]];
         newTimeRange[dateIndex] = newTime;
+
         this.props.onChange?.(newDateRange);
         this.setState({ value: newDateRange, time: newTimeRange });
     };
 
+    private getDefaultDate = (dateIndex: number) => {
```
gluxon, line 283: Just as a suggestion, I think we can add an inline source code comment to explain why we need to be careful about selecting the next default date.

Maybe something around:

1. When someone sets the time value before choosing a date, we need to "infer" or pick a date for them to initialize.
1. The next default date for the 0 or 1 index depends on the value of the other index since there's an invariant that the left/0 date is always less than the right/1 date.

gluxon (dismissed):
Thank you!

packages/datetime2/src/components/date-range-picker3/dateRangePicker3.tsx
evansjohnson, line 283:
```suggestion
    // When a user sets the time value before choosing a date, we need to pick a date for
    // them to have something to store the time on.
    // The default depends on the value of the other date since there's an invariant
    // that the left/0 date is always less than the right/1 date.
    private getDefaultDate = (dateIndex: number) => {
```

packages/datetime2/src/components/date-range-picker3/dateRangePicker3.tsx
evansjohnson: oof yea, I remember dealing with stuff like this back when seeing class components more often

in this case I'm thinking let's leave as is since as you call out we access `this.state` just before calling this helper method

packages/datetime2/src/components/date-range-picker3/dateRangePicker3.tsx
```diff
@@ -265,17 +265,39 @@ export class DateRangePicker3 extends DateFnsLocalizedComponent<DateRangePicker3
...
         newTimeRange[dateIndex] = newTime;
+
         this.props.onChange?.(newDateRange);
         this.setState({ value: newDateRange, time: newTimeRange });
     };
 
+    // When a user sets the time value before choosing a date, we need to pick a date for
+    // them to have something to store the time on.
+    // The default depends on the value of the other date since there's an invariant
+    // that the left/0 date is always less than the right/1 date.
```
evansjohnson:
```suggestion
    // When a user sets the time value before choosing a date, we need to pick a date for them
    // The default depends on the value of the other date since there's an invariant
    // that the left/0 date is always less than the right/1 date
```

packages/datetime2/src/components/date-range-picker3/dateRangePicker3.tsx
gluxon: Agree! Let's leave as-is. I almost didn't make this comment because it was an existing issue.

gluxon (dismissed):
Sweet! Thanks for the comment on `getDefaultDate`.

gluxon (approved):

PR #6861, opened 2024-06-24
Fix docs crash when navigating with hotkeys

### LINKED ISSUE #6381: Hotkey navigation on blueprintjs.com/docs sometimes makes the page crash
#### Environment

https://blueprintjs.com/docs/
Chrome 116.0.5845.179
Mac OS X

#### Steps to reproduce

1. Navigate to https://blueprintjs.com/docs/#core/components/breadcrumbs
1. Click somewhere in the page body. Press `[`.

#### Actual behavior

The URL changes to https://blueprintjs.com/docs/#core/components, and the page goes blank. Errors are reported in the console:

```
Uncaught TypeError: Cannot read properties of null (reading 'offsetTop')
    at t.maybeScrollToActivePageMenuItem (docs-app.js:1:5894972)
    at t.componentDidUpdate (docs-app.js:1:5894075)
    at dr (docs-app.js:1:990333)
    at Ll (docs-app.js:1:1007962)
    at t.unstable_runWithPriority (docs-app.js:1:1035731)
    at Ks (docs-app.js:1:951919)
    at Nl (docs-app.js:1:1004621)
    at bl (docs-app.js:1:1000768)
    at docs-app.js:1:952210
    at t.unstable_runWithPriority (docs-app.js:1:1035731)
```

Refreshing and using the browser back button does not get you out of the blank page.

#### Expected behavior

Navigation is either prevented between certain subdivisions or skips the headers instead of crashing the page.

evansjohnson:
#### Checklist
- [n/a] Includes tests
- [n/a] Update documentation

Ignore non-navigational headers so that navigating with hotkeys does not crash the page, and users cannot get to a page with no useful info.

- Unintended changes
- Any ideas on improving the targeting of non-navigational headers

#### Also considered
Allowing these headers to be navigated to, but you end up on a useless page
Screenshot 2024-06-24 at 2 40 15 PM

#### Screenshot
Current:
Screenshot 2024-06-24 at 2 38 05 PM

This PR:
Screenshot 2024-06-24 at 2 38 15 PM

### DIFF
--- a/packages/docs-theme/src/components/documentation.tsx
+++ b/packages/docs-theme/src/components/documentation.tsx
@@ -122,6 +122,18 @@ export interface DocumentationProps extends Props {
     tagRenderers: TagRendererMap;
 }
 
+// HACKHACK: this is brittle
+// detect Components page and subheadings
+const COMPONENTS_PATTERN = /\/components(\.[\w-]+)?$/;
+const CONTEXT_PATTERN = /\/context(\.[\w-]+)?$/;
+const HOOKS_PATTERN = /\/hooks(\.[\w-]+)?$/;
+const LEGACY_PATTERN = /\/legacy(\.[\w-]+)?$/;
+export const isNavSection = ({ route }: HeadingNode) =>
+    COMPONENTS_PATTERN.test(route) ||
+    CONTEXT_PATTERN.test(route) ||
+    HOOKS_PATTERN.test(route) ||
+    LEGACY_PATTERN.test(route);
+
 export interface DocumentationState {
     activeApiMember: string;
     activePageId: string;
@@ -156,8 +168,13 @@ export class Documentation extends React.PureComponent<DocumentationProps, Docum
         // build up static map of all references to their page, for navigation / routing
         this.routeToPage = {};
         eachLayoutNode(this.props.docs.nav, (node, parents) => {
-            const { reference } = isPageNode(node) ? node : parents[0]!;
-            this.routeToPage[node.route] = reference;
+            if (isNavSection(node)) {
+                return;
+            } else if (isPageNode(node)) {
+                this.routeToPage[node.route] = node.reference;
+            } else if (parents[0] != null) {
+                this.routeToPage[node.route] = parents[0].reference;
+            }
         });
     }
 
@@ -349,7 +366,11 @@ export class Documentation extends React.PureComponent<DocumentationProps, Docum
         const { activeSectionId } = this.state;
         // only scroll nav menu if active item is not visible in viewport.
         // using activeSectionId so you can see the page title in nav (may not be visible in document).
-        const navItemElement = this.navElement.querySelector<HTMLElement>(`a[href="#${activeSectionId}"]`)!;
+        const navItemElement = this.navElement.querySelector<HTMLElement>(`a[href="#${activeSectionId}"]`);
+        if (navItemElement == null) {
+            return;
+        }
+
         const scrollOffset = navItemElement.offsetTop - this.navElement.scrollTop;
         if (scrollOffset < 0 || scrollOffset > this.navElement.offsetHeight) {
             // reveal two items above this item in list
--- a/packages/docs-theme/src/components/documentation.tsx
+++ b/packages/docs-theme/src/components/documentation.tsx
@@ -158,6 +158,8 @@ export class Documentation extends React.PureComponent<DocumentationProps, Docum
         eachLayoutNode(this.props.docs.nav, (node, parents) => {
             if (isPageNode(node)) {
                 if (this.props.navigatorExclude?.(node)) {
+                    // if node is excluded from navigation, don't store it in the route to page map
+                    // to ensure the user cannnot navigate to it with hotkeys or through the URL
                     return;
                 }
                 this.routeToPage[node.route] = node.reference;
--- a/packages/docs-theme/src/components/documentation.tsx
+++ b/packages/docs-theme/src/components/documentation.tsx
@@ -122,18 +122,6 @@ export interface DocumentationProps extends Props {
     tagRenderers: TagRendererMap;
 }
 
-// HACKHACK: this is brittle
-// detect Components page and subheadings
-const COMPONENTS_PATTERN = /\/components(\.[\w-]+)?$/;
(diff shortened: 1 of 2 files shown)


packages/docs-theme/src/components/documentation.tsx
```diff
@@ -156,8 +168,13 @@ export class Documentation extends React.PureComponent<DocumentationProps, Docum
...
         this.routeToPage = {};
         eachLayoutNode(this.props.docs.nav, (node, parents) => {
-            const { reference } = isPageNode(node) ? node : parents[0]!;
-            this.routeToPage[node.route] = reference;
+            if (isNavSection(node)) {
+                return;
+            } else if (isPageNode(node)) {
+                this.routeToPage[node.route] = node.reference;
+            } else if (parents[0] != null) {
+                this.routeToPage[node.route] = parents[0].reference;
```
evansjohnson, line 167: the `!` was probably fine but no reason to not gracefully fail here I think rather than error out if we end up with something unexpected

packages/docs-theme/src/components/documentation.tsx
```diff
@@ -156,8 +168,13 @@ export class Documentation extends React.PureComponent<DocumentationProps, Docum
         // build up static map of all references to their page, for navigation / routing
         this.routeToPage = {};
         eachLayoutNode(this.props.docs.nav, (node, parents) => {
-            const { reference } = isPageNode(node) ? node : parents[0]!;
-            this.routeToPage[node.route] = reference;
+            if (isNavSection(node)) {
+                return;
```
gluxon: nit: Might be worth a comment that we never want to put the user on a "nav section" since these aren't click-able and don't have a real page. They're just presentational groupings of pages.

packages/docs-theme/src/components/documentation.tsx
```diff
@@ -122,6 +122,18 @@ export interface DocumentationProps extends Props {
...
+// detect Components page and subheadings
+const COMPONENTS_PATTERN = /\/components(\.[\w-]+)?$/;
+const CONTEXT_PATTERN = /\/context(\.[\w-]+)?$/;
+const HOOKS_PATTERN = /\/hooks(\.[\w-]+)?$/;
+const LEGACY_PATTERN = /\/legacy(\.[\w-]+)?$/;
+export const isNavSection = ({ route }: HeadingNode) =>
+    COMPONENTS_PATTERN.test(route) ||
+    CONTEXT_PATTERN.test(route) ||
+    HOOKS_PATTERN.test(route) ||
+    LEGACY_PATTERN.test(route);
```
gluxon: Moving this from `docs-app` to `docs-theme` feels a bit strange. I think `docs-theme` is meant to be generalizable outside of the Blueprint repo, and these regex patterns look specific to Blueprint's `docs-app`.

packages/docs-theme/src/components/documentation.tsx
```diff
@@ -156,8 +168,13 @@ export class Documentation extends React.PureComponent<DocumentationProps, Docum
         // build up static map of all references to their page, for navigation / routing
         this.routeToPage = {};
         eachLayoutNode(this.props.docs.nav, (node, parents) => {
-            const { reference } = isPageNode(node) ? node : parents[0]!;
-            this.routeToPage[node.route] = reference;
+            if (isNavSection(node)) {
+                return;
+            } else if (isPageNode(node)) {
```
gluxon: Is the _"components"_ navigation section passing the `isPageNode` check? Could that be a bug in Documentalist or are we setting up the _components_ section in a weird/unexpected way?

packages/docs-theme/src/components/documentation.tsx
evansjohnson: ah I didn't understand what the distinction was for, I suppose some other project could rely on docs-theme to use the blueprint docs styling and provide their own content?

packages/docs-theme/src/components/documentation.tsx
evansjohnson: going on intuition haven't dug into the code much here but I believe "components" is a page node because it has sub-pages - `/docs/#core/components/breadcrumbs`

packages/docs-theme/src/components/documentation.tsx
```diff
@@ -349,7 +357,11 @@ export class Documentation extends React.PureComponent<DocumentationProps, Docum
         const { activeSectionId } = this.state;
         // only scroll nav menu if active item is not visible in viewport.
         // using activeSectionId so you can see the page title in nav (may not be visible in document).
-        const navItemElement = this.navElement.querySelector<HTMLElement>(`a[href="#${activeSectionId}"]`)!;
+        const navItemElement = this.navElement.querySelector<HTMLElement>(`a[href="#${activeSectionId}"]`);
+        if (navItemElement == null) {
+            return;
```
evansjohnson, line 362: this is what was actually crashing, and is all that is needed to avoid crashing. but without the rest of this diff users end up here which isn't really useful/seems like a bug on it's own

Screenshot 2024-06-24 at 2 40 15 PM

packages/docs-theme/src/components/documentation.tsx
gluxon: Yup! I think that's the intention. I'm not sure how many other projects actually do that though.

packages/docs-theme/src/components/documentation.tsx
```diff
@@ -156,8 +156,16 @@ export class Documentation extends React.PureComponent<DocumentationProps, Docum
         // build up static map of all references to their page, for navigation / routing
         this.routeToPage = {};
         eachLayoutNode(this.props.docs.nav, (node, parents) => {
-            const { reference } = isPageNode(node) ? node : parents[0]!;
-            this.routeToPage[node.route] = reference;
+            if (isPageNode(node)) {
+                if (this.props.navigatorExclude?.(node)) {
```
gluxon, line 160: Oh nice! I see we were already setting this prop.

packages/docs-theme/src/components/documentation.tsx
gluxon: Agree!

gluxon (approved):
Thanks! 🥳

PR #6862, opened 2024-06-24
Respect showOutsideDays prop for single month date range

### LINKED ISSUE #4373: DateRangePicker singleMonthOnly does not support showing "outside" days
#### ReactJS

- __Package version(s)__: "@blueprintjs/datetime": "^3.15.1"
- __Browser and OS versions__:   86.0.4240.75 (Official Build) (64-bit), windows 10 2004 

#### Question

 in DateRangePicker, DateRangeInput picker Components, I cannot see "outside days" even I set DayPicker props: showOutsideDays to TRUE,
Could you help me figure this out
many thanks for your support

evansjohnson:
#### Checklist
- [x] Includes tests
- [x] Update documentation

When a DateRangePicker3 is limited to a single month, respect the `showOutsideDays` passed on `dayPickerProps`. Per https://github.com/palantir/blueprint/pull/586/files#r98813760 it was explicitly decided that showing outside days when multiple months are shown is not intended, and I agree that it still leads to some weird questions, but for a single month these issues do not exist.

- Is this the behavior we want?
- Do the updated docs look good?
- Only fixing on `DateRangePicker3` since DateRangePicker is deprecated and frozen.

#### Alternate option
Simply document that `showOutsideDays` is never respected for `DateRangePicker3`

#### Screenshot
Screenshot 2024-06-24 at 4 06 48 PM
Screenshot 2024-06-24 at 4 06 43 PM

### DIFF
--- a/packages/datetime2/src/components/date-range-picker3/_date-range-picker3.scss
+++ b/packages/datetime2/src/components/date-range-picker3/_date-range-picker3.scss
@@ -60,12 +60,11 @@
   }
 
   // showing outside days is allowed for single months
-  &.#{$ns}-datepicker.#{$ns}-daterangepicker-single-month.#{$ns}-daterangepicker-showing-outside-days {
-    .rdp-day_outside {
-      visibility: visible;
-      // weird interactions happen if allowing outside days to be clicked
-      pointer-events: none;
-    }
+  /* stylelint-disable max-line-length */
+  &.#{$ns}-datepicker.#{$ns}-daterangepicker-single-month.#{$ns}-daterangepicker-showing-outside-days .rdp-day_outside {
+    // weird interactions happen if allowing outside days to be clicked
+    pointer-events: none;
+    visibility: visible;
   }
 
   // need some extra specificity to override DatePicker styles
--- a/packages/datetime2/src/components/date-range-picker3/_date-range-picker3.scss
+++ b/packages/datetime2/src/components/date-range-picker3/_date-range-picker3.scss
@@ -60,11 +60,12 @@
   }
 
   // showing outside days is allowed for single months
-  /* stylelint-disable max-line-length */
-  &.#{$ns}-datepicker.#{$ns}-daterangepicker-single-month.#{$ns}-daterangepicker-showing-outside-days .rdp-day_outside {
-    visibility: visible;
-    // weird interactions happen if allowing outside days to be clicked
-    pointer-events: none;
+  &.#{$ns}-datepicker.#{$ns}-daterangepicker-single-month.#{$ns}-daterangepicker-showing-outside-days {
+    .rdp-day_outside {
+      visibility: visible;
+      // weird interactions happen if allowing outside days to be clicked
+      pointer-events: none;
+    }
   }
 
   // need some extra specificity to override DatePicker styles
--- a/packages/datetime2/src/components/date-range-picker3/_date-range-picker3.scss
+++ b/packages/datetime2/src/components/date-range-picker3/_date-range-picker3.scss
@@ -61,10 +61,8 @@
 
   // showing outside days is allowed for single months
   /* stylelint-disable max-line-length */
-  &.#{$ns}-datepicker.#{$ns}-daterangepicker-single-month.#{$ns}-daterangepicker-showing-outside-days .rdp-day {
-    &_outside {
-      visibility: visible;
-    }
+  &.#{$ns}-datepicker.#{$ns}-daterangepicker-single-month.#{$ns}-daterangepicker-showing-outside-days .rdp-day_outside {
+    visibility: visible;
   }
 
   // need some extra specificity to override DatePicker styles
--- a/packages/datetime2/src/components/date-range-picker3/_date-range-picker3.scss
+++ b/packages/datetime2/src/components/date-range-picker3/_date-range-picker3.scss
@@ -65,6 +65,7 @@
     pointer-events: none; // weird interactions happen if allowing outside days to be clicked
     visibility: visible; // showing outside days is allowed for single months
   }
+  /* stylelint-enable max-line-length */
 
   // need some extra specificity to override DatePicker styles
   &.#{$ns}-datepicker .rdp-day {
--- a/packages/datetime2/src/components/date-range-picker3/_date-range-picker3.scss
+++ b/packages/datetime2/src/components/date-range-picker3/_date-range-picker3.scss
@@ -60,6 +60,7 @@
   }
 
   // showing outside days is allowed for single months
+  /* stylelint-disable max-line-length */
   &.#{$ns}-datepicker.#{$ns}-daterangepicker-single-month.#{$ns}-daterangepicker-showing-outside-days .rdp-day {
     &_outside {
       visibility: visible;
--- a/packages/datetime2/src/components/date-range-picker3/_date-range-picker3.scss
+++ b/packages/datetime2/src/components/date-range-picker3/_date-range-picker3.scss
@@ -59,12 +59,11 @@
     }
   }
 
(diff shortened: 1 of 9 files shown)


packages/docs-app/src/examples/datetime2-examples/dateRangePicker3Example.tsx
```diff
@@ -105,6 +105,9 @@ export class DateRangePicker3Example extends React.PureComponent<ExampleProps, D
                     maxDate={maxDate}
                     minDate={minDate}
                     onChange={this.handleDateRangeChange}
+                    dayPickerProps={{
+                        showOutsideDays: true,
+                    }}
```
evansjohnson, line 110: TODO: remove once a PR preview builds

packages/datetime2/test/components/dateRangePicker3Tests.tsx
```diff
@@ -1332,6 +1332,33 @@ describe("<DateRangePicker3>", () => {
...
+            const { left } = render({ initialMonth: new Date(2015, Months.DECEMBER, 2), singleMonthOnly: true, dayPickerProps: { showOutsideDays: true } });
+            assert.equal(left.findOutsideDays().first().getDOMNode().computedStyleMap().get("visibility")?.toString(), "visible");
+        });
+
+        it("does not allow outside days to be selected even when visible", () => {
+            const { left } = render({ initialMonth: new Date(2015, Months.DECEMBER, 2), singleMonthOnly: true, dayPickerProps: { showOutsideDays: true } });
+
+            assert.equal(left.findOutsideDays().first().getDOMNode().computedStyleMap().get("pointer-events")?.toString(), "none");
+
+            // simulated clicks appear to not respect pointerEvents style or else would test with below
```
evansjohnson, line 1373: no activity on it but I found https://github.com/enzymejs/enzyme/issues/2340

open to suggestions for better ways to disable the interaction or better ways to test this

packages/datetime2/test/components/dateRangePicker3Tests.tsx
```diff
@@ -1332,6 +1332,33 @@ describe("<DateRangePicker3>", () => {
...
 
+    describe.only("outside days", () => {
+        it("visually hides outside days when contiguous months are shown, even if showOutsideDays is true", () => {
+            const { left } = render({ initialMonth: new Date(2015, Months.DECEMBER, 2), dayPickerProps: { showOutsideDays: true } });
+            assert.equal(left.findOutsideDays().first().getDOMNode().computedStyleMap().get("visibility")?.toString(), "hidden");
+        });
+
+        it("allows outside days to be shown for single month date range", () => {
+            const { left } = render({ initialMonth: new Date(2015, Months.DECEMBER, 2), singleMonthOnly: true, dayPickerProps: { showOutsideDays: true } });
+            assert.equal(left.findOutsideDays().first().getDOMNode().computedStyleMap().get("visibility")?.toString(), "visible");
```
evansjohnson: open to suggestion for better way to test visibility here

packages/datetime2/test/components/dateRangePicker3Tests.tsx
```diff
@@ -1332,6 +1332,33 @@ describe("<DateRangePicker3>", () => {
         });
     });
 
+    describe("outside days", () => {
```
evansjohnson, line 1335: these tests are very closely tied to the implementation but figured it was good to get something in testing the behavior so this doesn't unintentionally regress

packages/datetime2/src/components/date-range-picker3/_date-range-picker3.scss
```diff
@@ -59,6 +59,13 @@
     }
   }
 
+
+  /* stylelint-disable max-line-length */
+  &.#{$ns}-datepicker.#{$ns}-daterangepicker-single-month.#{$ns}-daterangepicker-showing-outside-days .rdp-day_outside {
+    pointer-events: none; // weird interactions happen if allowing outside days to be clicked
```
evansjohnson, line 65: Note that without this, selecting a date in the earlier "outside days" switches to the previous month. This doesn't seem ideal since there will not be many days ahead of that initial date selected once we switch to the previous month, but the user can reselect the month after making the start date selection.

Also note that we only need to worry about click and not keyboard nav events here because we still want to allow keyboard nav with arrow keys to navigate between months. The difference is that keyboard nav is navigation, then selection, while mouse interaction is selection and navigation which felt a bit jarring.

Open to suggestions for a better solution here.

packages/datetime2/src/components/date-range-picker3/_date-range-picker3.scss
```diff
@@ -59,6 +59,14 @@
     }
   }
 
+
+  /* stylelint-disable max-line-length */
```
gluxon, line 63: nitpick: I usually like to use `stylelint-disable-next-line` whenever possible to avoid the need to re-enable the lint rule further down.

```suggestion
  // stylelint-disable-next-line max-line-length
```

packages/datetime2/src/components/date-range-picker3/_date-range-picker3.scss
```diff
@@ -59,6 +59,14 @@
...
+  &.#{$ns}-datepicker.#{$ns}-daterangepicker-single-month.#{$ns}-daterangepicker-showing-outside-days .rdp-day_outside {
+    pointer-events: none; // weird interactions happen if allowing outside days to be clicked
+    visibility: visible; // showing outside days is allowed for single months
+  }
+  /* stylelint-enable max-line-length */
```
gluxon, line 68: nitpick: It should be possible to remove this after using `// stylelint-disable-next-line` above.

packages/datetime2/test/components/dateRangePicker3Tests.tsx
```diff
@@ -1332,6 +1332,55 @@ describe("<DateRangePicker3>", () => {
...
+        it("visually hides outside days when contiguous months are shown, even if showOutsideDays is true", () => {
+            const warnSpy = stub(console, "warn");
```
gluxon, line 1337: nitpick: Might be useful to clean up this stub at the end of this function with:

```ts
warnSpy.restore();
```

Otherwise this `console.warn` stub stucks around for all tests that run after this one.

packages/datetime2/src/components/date-range-picker3/_date-range-picker3.scss
gluxon: I would be curious for @alex-dales's opinion. Playing around with the UI, I actually found it surprising that I couldn't click on the grayed out dates. It might not be obvious to some users that they have to switch to the previous month intentionally to select these "outside day" dates?

packages/docs-app/src/examples/datetime2-examples/dateRangePicker3Example.tsx
```diff
@@ -131,6 +138,18 @@ export class DateRangePicker3Example extends React.PureComponent<ExampleProps, D
                         label="Single month only"
                         onChange={this.toggleSingleMonth}
                     />
+                    <Tooltip content="Only respected when showing a single month calendar">
+                        <Switch
+                            checked={this.state.showOutsideDays}
+                            className="keep-bottom-margin"
```
gluxon, line 144: super nitpick: It looks like the bottom margin here is getting reset back to `0` by some other CSS selector.

Screenshot 2024-08-06 at 6 08 13 PM

gluxon (approved):
Implementation and tests look good! (Or, the tests look as good as they can be. I don't know of a better way to check for visibility. 😂)

I lean towards allowing the outside days to be click-able, but will let you and @alex-dales make the final call on the right UX for that.

PR #6865, opened 2024-06-25
a11y(Tooltip): "tooltip" role

bvandrc:
#### Checklist

- [X] Includes tests
- [N/A] Update documentation

- apply `role="tooltip"` to the Tooltip's `content` - https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/tooltip_role
- apply an `id` to the tooltip content, and `aria-describedby={id}` to the target
- copy validation tests from Popover to Tooltip to ensure still validates


packages/core/src/components/popover/popover.tsx
```diff
@@ -74,11 +74,6 @@ export interface PopoverProps<TProps extends DefaultPopoverTargetHTMLProps = Def
     /** HTML props for the backdrop element. Can be combined with `backdropClassName`. */
     backdropProps?: React.HTMLProps<HTMLDivElement>;
 
-    /**
```
bvandercar-vt, line 77: moved to `PopoverSharedProps`, since same as `TooltipProps.content`

packages/core/src/components/popover/popover.tsx
```diff
@@ -74,11 +74,6 @@ export interface PopoverProps<TProps extends DefaultPopoverTargetHTMLProps = Def
...
-     * The content displayed inside the popover.
-     */
-    content?: string | React.JSX.Element;
```
bvandrc: moved to `PopoverSharedProps`, since same as `TooltipProps.content`

packages/core/src/components/tooltip/tooltip.tsx
```diff
@@ -42,19 +38,11 @@ export interface TooltipProps<TProps extends DefaultPopoverTargetHTMLProps = Def
     compact?: boolean;
 
     /**
-     * The amount of time in milliseconds the tooltip should remain open after
-     * the user hovers off the trigger. The timer is canceled if the user mouses
-     * over the target before it expires.
-     *
```
bvandrc: If you are just changing the `@default` value, don't need to re-write the whole description when the original is in the `extends`:

image

image

packages/core/src/common/utils/reactUtils.ts
```diff
@@ -26,8 +26,8 @@ import { isEmptyString } from "./jsUtils";
 export function isReactNodeEmpty(node?: React.ReactNode, skipArray = false): boolean {
     return (
         node == null ||
-        node === "" ||
         node === false ||
+        isEmptyString(node) ||
```
evansjohnson: I don't think we should consider " " an empty node - or does that not sound right? This seems like it has a chance to break something existing and I'd rather not try to verify all of the uses.

packages/core/src/components/tooltip/tooltip.tsx
evansjohnson: could we split these docs changes out to their own PR to merge separately?

packages/core/src/components/tooltip/tooltip.tsx
```diff
@@ -114,13 +96,29 @@ export class Tooltip<
...
+            this.props;
+
         const popoverClasses = classNames(Classes.TOOLTIP, Classes.intentClass(intent), popoverClassName, {
             [Classes.COMPACT]: compact,
         });
 
+        const tooltipId = Utils.uniqueId("tooltip");
+
+        const childTarget = Utils.ensureElement(React.Children.toArray(children)[0], undefined, {
+            "aria-describedby": tooltipId,
```
evansjohnson: could we try some version of https://github.com/facebook/react/issues/8205#issuecomment-755866890 and checking [hasOwnProperty](https://github.com/facebook/react/issues/8205#issuecomment-334172163) to make sure we're not overriding an id specified by a consumer of Tooltip?

packages/core/src/components/tooltip/tooltip.tsx
```diff
@@ -114,13 +96,29 @@ export class Tooltip<
         this.popoverRef.current?.reposition();
     }
 
+    protected validateProps(props: TooltipProps<T>) {
+        const childrenCount = React.Children.count(props.children);
+        if (childrenCount > 1) {
+            console.warn(Errors.POPOVER_WARN_TOO_MANY_CHILDREN);
```
evansjohnson, line 133: why repeat this warning here but not the others?

packages/core/test/tooltip/tooltipTests.tsx
```diff
@@ -20,15 +20,65 @@ import * as React from "react";
...
 import { Button, Overlay2 } from "../../src/components";
 import { Popover } from "../../src/components/popover/popover";
 import { Tooltip, type TooltipProps } from "../../src/components/tooltip/tooltip";
 
 const TARGET_SELECTOR = `.${Classes.POPOVER_TARGET}`;
 const TOOLTIP_SELECTOR = `.${Classes.TOOLTIP}`;
-const TEST_TARGET_ID = "test-target";
 
 describe("<Tooltip>", () => {
+    describe("validation", () => {
```
evansjohnson, line 33: Any reason to repeat these tests from popover? It feels like this is just testing the underlying popover component and so is the wrong place for these tests, which already exist on the underlying popover.

evansjohnson (changes requested):
I think we should be able to get something merged here but we should be careful here of any chance of breaking existing behavior and overriding props passed to these components.

packages/core/src/common/utils/reactUtils.ts
bvandrc: fair fair. reverted.

packages/core/src/components/tooltip/tooltip.tsx
bvandrc: Actually `aria-describedby` can have multiple values as a space separated list string (https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-describedby). Updated to concatenate that if passed: https://github.com/palantir/blueprint/pull/6865/commits/730d177bfaa936d6ef973f7fb0e83757e7b889ce

packages/core/src/components/tooltip/tooltip.tsx
bvandrc: We don't want to repeat all of them because A) duplicate code and B) consumer gets duplicate warnings

All the other warnings are caught within `Popover`'s `validateProps` (hence the reason for copying over the tests to `tooltipTests`). This warning needs to be checked here though because now we wrap any passed `children` :

```tsx
const childTarget = Utils.ensureElement(React.Children.toArray(children)[0]);
```

in order to apply some props here in `Tooltip`. So, if we didn't have this `validateProps`, `Popover`'s would never warn for multiple children because it would never receive multiple children.

packages/core/test/tooltip/tooltipTests.tsx
bvandrc: See other comment. Basically we are modifying some of the passed values in `Tooltip`, so wanted to make sure `Popover` is receiving the same values and giving the same warnings.

Adding these tests helped to discover a bug with my original code changes; one was failing. Has since been fixed, but that justifies having them.

packages/core/src/components/tooltip/tooltip.tsx
bvandrc: Sure

bvandrc: https://github.com/palantir/blueprint/pull/6875

packages/core/test/tooltip/tooltipTests.tsx
bvandrc: Just accidentally introduced another failure to these tests with my last change, proved again that we need em!

packages/core/src/components/tooltip/tooltip.tsx
evansjohnson: thanks for the explanation!

packages/core/src/components/tooltip/tooltip.tsx
```diff
@@ -17,22 +17,18 @@
...
 import { TooltipContext, type TooltipContextState, TooltipProvider } from "../popover/tooltipContext";
 
 export interface TooltipProps<TProps extends DefaultPopoverTargetHTMLProps = DefaultPopoverTargetHTMLProps>
-    extends Omit<PopoverSharedProps<TProps>, "shouldReturnFocusOnClose">,
+    extends Omit<PopoverSharedProps<TProps, { tooltipId: string }>, "shouldReturnFocusOnClose">,
         IntentProps {
-    /**
-     * The content that will be displayed inside of the tooltip.
-     */
-    content: React.JSX.Element | string;
```
evansjohnson: same comment as https://github.com/palantir/blueprint/pull/6875/files#r1655441543

packages/core/src/components/tooltip/tooltip.tsx
```diff
@@ -134,6 +149,14 @@ export class Tooltip<
                     },
                 }}
                 {...restProps}
+                // eslint-disable-next-line react/jsx-no-bind
+                renderTarget={renderTarget ? props => renderTarget({ ...props, tooltipId }) : undefined}
+                content={
+                    // want Popover to warn if empty, so don't wrap in element if empty.
+                    Utils.isReactNodeEmpty(content)
+                        ? content
+                        : Utils.ensureElement(content, undefined, { role: "tooltip", id: tooltipId })
```
evansjohnson: I don't think we can do this because a consumer may have already defined an `id` on the `content` which this would override

packages/core/src/components/popover/popover.tsx
```diff
@@ -252,7 +247,8 @@ export class Popover<
         const { disabled, content, placement, position = "auto", positioningStrategy } = this.props;
         const { isOpen } = this.state;
 
-        const isContentEmpty = content == null || (typeof content === "string" && content.trim() === "");
+        const isContentEmpty =
```
evansjohnson: this change means the empty check will be hit if the content is `false` or `[]`, which seems like a good change. previously I imagine we tried to render the popover but it ended up not really displaying anything. I can't imagine anyone would want the current behavior.

evansjohnson (commented):
I think this is getting closer but we need to be very careful here of breaking existing behavior. We'll only be able to merge this if this can be done in a way that is a strict improvement, and doesn't negatively impact users who may have already done the work to work around this.

packages/core/src/components/tooltip/tooltip.tsx
bvandrc: I think you commented this on an old diff. I reverted the change to this type.

packages/core/src/components/tooltip/tooltip.tsx
bvandrc: Highly doubt that would ever be the case, can't imagine what other scenarios they'd be doing this for. There are many cases, such as in `Popover`, where we apply a bunch of props without caring what the consumer may have already applied.

bvandrc: This should account for that case: https://github.com/palantir/blueprint/pull/6865/commits/be12eb309882e8251c8d6e81a170c20f9904cfb0

bvandrc: @evansjohnson  the other alternative solution here would be we add a prop `contentProps` to `Popover` and apply it to the `div` that's already wrapping the `content`. Then we could apply these props via that. Up to you.

packages/core/src/components/popover/popover.tsx
evansjohnson: this should be added back as well

packages/core/src/components/tooltip/tooltip.tsx
evansjohnson: > what other scenarios they'd be doing this for

basically to fix this issue in consumer land is what I'm thinking about, or possibly adding an `id` used to target a tooltip in tests?

packages/core/src/components/popover/popover.tsx
```diff
@@ -301,7 +299,7 @@ export class Popover<
         }
     }
 
-    protected validateProps(props: PopoverProps<T> & { children?: React.ReactNode }) {
```
evansjohnson: any ideas why `{ children?: React.ReactNode }` was here before? preference to remove at least for this PR if we're not sure and if this isn't needed for this PR

packages/core/src/components/tooltip/tooltip.tsx
```diff
@@ -114,13 +115,36 @@ export class Tooltip<
...
-        const { children, compact, disabled, intent, popoverClassName, ...restProps } = this.props;
+        const { children, content, renderTarget, compact, disabled, intent, popoverClassName, ...restProps } =
+            this.props;
+
         const popoverClasses = classNames(Classes.TOOLTIP, Classes.intentClass(intent), popoverClassName, {
             [Classes.COMPACT]: compact,
         });
 
+        const contentElement = Utils.ensureElement(content);
+        const tooltipId = contentElement?.props?.id ?? Utils.uniqueId("tooltip");
```
evansjohnson: could we add a `useMemo` here so this doesn't create a new id every render?

packages/core/src/components/tooltip/tooltip.tsx
evansjohnson: actually - if popover isn't warning it seems like it may not actually be a problem.

if Tooltip "fixes" this warning before it gets to the popover what's the need to warn here?

packages/core/src/components/tooltip/tooltip.tsx
```diff
@@ -134,6 +158,12 @@ export class Tooltip<
                     },
                 }}
                 {...restProps}
+                // eslint-disable-next-line react/jsx-no-bind
```
evansjohnson: nit: preference for `useCallback` over disabling the lint rule

packages/core/src/components/tooltip/tooltip.tsx
```diff
@@ -114,13 +115,36 @@ export class Tooltip<
...
         });
 
+        const contentElement = Utils.ensureElement(content);
+        const tooltipId = contentElement?.props?.id ?? Utils.uniqueId("tooltip");
+        const clonedContent = contentElement
+            ? React.cloneElement(contentElement, { role: "tooltip", id: tooltipId })
+            : contentElement;
+
+        const childTarget = Utils.ensureElement(React.Children.toArray(children)[0]);
+        const clonedTarget = childTarget
```
evansjohnson: if `childTarget` is defined with `ensureElement` then I think this next line will always be truthy which I don't think we want

packages/core/src/components/tooltip/tooltip.tsx
```diff
@@ -114,13 +115,36 @@ export class Tooltip<
...
+            ? React.cloneElement(childTarget, {
+                  "aria-describedby": [tooltipId, childTarget.props["aria-describedby"]].filter(Boolean).join(" "),
```
evansjohnson: `tooltipId` could possibly appear here twice in cases where consumers of `Tooltip` have already added these `id`/`aria-describedby` attributes

I'm not sure of the implications of that but it seems easy enough to guard against

evansjohnson (commented):
Thank you for the iteration so far on this one! In interest of reducing risk of any behavior regressions and to make sure we don't regress in the future on the behavior added here, could we add tests for:
#### protect existing behavior:
- an `id` provided to tooltip `content` should not be overridden
- a provided `aria-describedby` on the tooltip target should not be overridden

#### protect against future regressions:
- for both `children` target and `renderTarget`:
    - an `id` will be added to tooltip `content`, used in `aria-describedby` added to the tooltip target
    - an existing `id` will be used, with relevant `aria-describedby` still added

packages/core/src/components/tooltip/tooltip.tsx
bvandrc: what we should really be doing is using React `useId` hook instead of `uniqueId` and letting that do the work for us.

bvandrc: Done though... we should probably do that in the other cases we use `uniqueId`....

packages/core/src/components/tooltip/tooltip.tsx
bvandrc: Nope, it will be falsy. `Utils.ensureElement`:
image

and its return type:
image


And, as we discussed, having all the tests in `tooltipTests` as `popoverTests`  ensures that the same cases occur.

packages/core/src/components/tooltip/tooltip.tsx
bvandrc: done

packages/core/src/components/tooltip/tooltip.tsx
bvandrc: I don't think that would make a difference in a11y but here: https://github.com/palantir/blueprint/pull/6865/commits/03367967550f47ba9be5d684ca7b4df55c2c30ed

packages/core/src/components/popover/popover.tsx
bvandrc: I don't, `children` is already in `PopoverProps` so it's not needed, and removing it results in no type failures.

packages/core/src/components/tooltip/tooltip.tsx
bvandercar-vt: Actually here: https://github.com/palantir/blueprint/pull/6865/commits/be533a0f056508aa04fbc46bf71dd3f797de080f

evansjohnson (commented):
I'd like to get this merged but there's a lot of separate small stuff going on in this PR right now. Could you split out everything that you think is a simple no-op refactor to set up a smaller actual change to implement this?

bvandrc:
> I'd like to get this merged but there's a lot of separate small stuff going on in this PR right now. Could you split out everything that you think is a simple no-op refactor to set up a smaller actual change to implement this?

https://github.com/palantir/blueprint/pull/6923

packages/core/src/components/tooltip/tooltip.tsx
bvandrc: Refactored, please re review, I think this concern has been solved

bvandrc:
@evansjohnson refactored this one a little bit, which I think fixed some of your concerns. Ready for re-review!

packages/core/src/components/tooltip/tooltip.tsx
```diff
@@ -114,13 +127,43 @@ export class Tooltip<
...
-        const { children, compact, disabled, intent, popoverClassName, ...restProps } = this.props;
+        const { children, content, renderTarget, compact, disabled, intent, popoverClassName, ...restProps } =
+            this.props;
+
         const popoverClasses = classNames(Classes.TOOLTIP, Classes.intentClass(intent), popoverClassName, {
             [Classes.COMPACT]: compact,
         });
 
+        const contentElement = Utils.ensureElement(content);
+        const tooltipId = contentElement?.props?.id ?? React.useMemo(() => Utils.uniqueId("tooltip"), []);
```
ggdouglas, line 148: The use of `React.useMemo` here violates the [rules of hooks](https://react.dev/reference/react/useMemo#caveats) since it is called conditionally within the ternary.

Can we extract this default memoized value out to the top level of this component?

packages/core/src/components/tooltip/tooltip.tsx
```diff
@@ -134,6 +177,14 @@ export class Tooltip<
                     },
                 }}
                 {...restProps}
+                renderTarget={
+                    renderTarget &&
+                    React.useCallback(props => renderTarget({ ...props, tooltipId }), [renderTarget, tooltipId])
```
ggdouglas, line 182: Similar to [above](https://github.com/palantir/blueprint/pull/6865#discussion_r1775732420), can we pull this `React.useCallback` into a variable at the top of the component such that it is not called conditionally?

packages/core/src/components/tooltip/tooltip.tsx
```diff
@@ -114,13 +127,43 @@ export class Tooltip<
...
         });
 
+        const contentElement = Utils.ensureElement(content);
+        const tooltipId = contentElement?.props?.id ?? React.useMemo(() => Utils.uniqueId("tooltip"), []);
+        const tooltipRole = contentElement?.props?.role ?? "tooltip";
+        const clonedContent =
+            contentElement &&
+            React.cloneElement(contentElement, {
+                role: tooltipRole,
+                id: tooltipId,
```
ggdouglas, line 154: **nit:** The linter is complaining that these keys, are not sorted alphabetically. We can fix by sorting them:
```tsx
React.cloneElement(contentElement, {
    id: tooltipId,
    role: tooltipRole,
});
```

ggdouglas (changes requested):
Functionally this looks good to me, but we need to address the lint and hooks errors before merging.

PR #6866, opened 2024-06-25
style: object types

bvandrc:
just a style change, reduce re-declaring constants


evansjohnson (commented):
Are there any benefits here? I default towards less diff churn and commit noise if this is purely a style concern and not something our existing lint/format rules enforce.

bvandrc:
@evansjohnson No benefits, purely style, matches the other 90% of ways this is implemented though.

evansjohnson (approved):
seeing this matches all other places we redeclare a const as a type

PR #6868, opened 2024-06-25
feat(DateTime2): export TimePicker

### LINKED ISSUE #6739: Include TimePicker in Datetime2 package?
`TimePicker` is not currently included the `Datetime2` package, which means that both `Datetime` and `Datetime2` packages must be included as dependencies. Is there a plan to add `TimePicker` to the `Datetime2` package? Would this be a welcome PR?

bvandrc:
#### Checklist

- [N/A] Includes tests
- [N/A] Update documentation

Export `TimePicker` from `datetime2` package, from the original `datetime` package, like is already being done for `TimezoneSelect` as well as other componentes/ types.

Other components/types are exported this way in the same file

### DIFF
--- a/packages/datetime2/src/index.ts
+++ b/packages/datetime2/src/index.ts
@@ -42,6 +42,8 @@ export {
     /** @deprecated import from `@blueprintjs/datetime` instead */
     type DateRangeShortcut,
     /** @deprecated import from `@blueprintjs/datetime` instead */
+    type DatePickerShortcut,
+    /** @deprecated import from `@blueprintjs/datetime` instead */
     getTimezoneMetadata,
     TimePrecision,
     /** @deprecated import from `@blueprintjs/datetime` instead */
--- a/packages/datetime2/src/index.ts
+++ b/packages/datetime2/src/index.ts
@@ -27,28 +27,46 @@ export { Classes as Datetime2Classes, ReactDayPickerClasses } from "./classes";
 /* eslint-disable deprecation/deprecation */
 
 export {
-    /** @deprecated use `Datetime2Classes` instead */
+    /** @deprecated import from `@blueprintjs/datetime` instead, or use `Datetime2Classes` */
     Classes,
+    /** @deprecated import from `@blueprintjs/datetime` instead */
     DateInput as DateInput2,
-    /** @deprecated use `DateInput3Props` instead */
+    /** @deprecated import from `@blueprintjs/datetime` instead */
     type DateInputProps as DateInput2Props,
+    /** @deprecated import from `@blueprintjs/datetime` instead */
     DateRangeInput as DateRangeInput2,
-    /** @deprecated use `DateRangeInput3Props` instead */
+    /** @deprecated import from `@blueprintjs/datetime` instead */
     type DateRangeInputProps as DateRangeInput2Props,
+    /** @deprecated import from `@blueprintjs/datetime` instead */
     TimezoneSelect,
+    /** @deprecated import from `@blueprintjs/datetime` instead */
     type TimezoneSelectProps,
+    /** @deprecated import from `@blueprintjs/datetime` instead */
     TimePicker,
+    /** @deprecated import from `@blueprintjs/datetime` instead */
     type TimePickerProps,
+    /** @deprecated import from `@blueprintjs/datetime` instead */
     type DateRangeShortcut,
+    /** @deprecated import from `@blueprintjs/datetime` instead */
     type DatePickerShortcut,
+    /** @deprecated import from `@blueprintjs/datetime` instead */
     type DateFormatProps,
+    /** @deprecated import from `@blueprintjs/datetime` instead */
     type DateRange,
+    /** @deprecated import from `@blueprintjs/datetime` instead */
     type NonNullDateRange,
+    /** @deprecated import from `@blueprintjs/datetime` instead */
     MonthAndYear,
+    /** @deprecated import from `@blueprintjs/datetime` instead */
     Months,
+    /** @deprecated import from `@blueprintjs/datetime` instead */
     getTimezoneMetadata,
+    /** @deprecated import from `@blueprintjs/datetime` instead */
     TimePrecision,
+    /** @deprecated import from `@blueprintjs/datetime` instead */
     TimeUnit,
+    /** @deprecated import from `@blueprintjs/datetime` instead */
     TimezoneDisplayFormat,
+    /** @deprecated import from `@blueprintjs/datetime` instead */
     type DatePickerLocaleUtils,
 } from "@blueprintjs/datetime";
--- a/packages/datetime2/src/index.ts
+++ b/packages/datetime2/src/index.ts
@@ -48,4 +48,8 @@ export {
     type TimezoneSelectProps,
     /** @deprecated import from `@blueprintjs/datetime` instead */
     TimezoneDisplayFormat,
+    /** @deprecated import from `@blueprintjs/datetime` instead */
+    TimePicker,
+    /** @deprecated import from `@blueprintjs/datetime` instead */
+    type TimePickerProps,
 } from "@blueprintjs/datetime";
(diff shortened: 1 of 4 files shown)


packages/datetime2/src/index.ts
```diff
@@ -48,4 +52,8 @@ export {
     type TimezoneSelectProps,
     /** @deprecated import from `@blueprintjs/datetime` instead */
     TimezoneDisplayFormat,
+    /** @deprecated import from `@blueprintjs/datetime` instead */
+    TimePicker,
+    /** @deprecated import from `@blueprintjs/datetime` instead */
+    type TimePickerProps,
 } from "@blueprintjs/datetime";
```
evansjohnson, line 72: I think the import deprecation messages still make sense because ultimately these components will end up in `datetime` v6.

That said, it does seem unexpected we only offer a subset of the `datetime` exports from `datetime2`, particularly for components that have no `datetime2` equivalent like `TimePicker`.

packages/datetime2/src/index.ts
```diff
@@ -40,6 +40,10 @@ export {
     /** @deprecated import from `@blueprintjs/datetime` instead */
     type DateRangeInputProps as DateRangeInput2Props,
     /** @deprecated import from `@blueprintjs/datetime` instead */
+    type DateRangeShortcut,
```
evansjohnson, line 49: Is there a reason you're exporting these two types specifically? I'm wondering if we should do a more complete version of this PR so we don't need to revisit this.

Full list (a little messy from reformatting but roughly follows order exported) of datetime exports with thoughts (expand below):
<details>
<summary>Excuse the wall of text, possible action items prefixed with *****</summary>
// from ./common

- ***** DatePickerBaseProps, DatePickerModifiers - it feels like these could be re-exported similar to other types
- DateFormatProps, DateRange - already reexported
- ***** NonNullDateRange, Months, TimeUnit - it feels like these could be re-exported similar to other types
- ***** TimePickerProps - if we're exporting `TimePicker` in this PR we may as well export this too, similar to - `TimezoneSelectProps`
- TimePrecision - already reexported
- TimezoneDisplayFormat - already reexported

- getTimezoneMetadata - already reexported

// from index.ts
- ***** DatePickerLocaleUtils, DatePickerDayModifiers - these seem fine/possibly useful to re-export

- DateFormatProps - already reexported
- DateRangeSelectionStrategy, type DateRangeSelectionState - for use in datetime2, no need to re-export
- ***** MonthAndYear - this seems fine/possibly useful to re-export
- TimePickerProps - exported in this PR
- TimePrecision - already reexported

- DatePickerUtils - for use in datetime2, no need to re-export
- DatePickerBaseProps - for use in datetime2, no need to re-export
- DatePickerModifiers - this seems to be specific for the old v7 version of the underlying datepicker
- TimePicker - exported in this PR
- DatePickerShortcutMenu, type DatePickerShortcutMenuProps - only exported for use in datetime2, not intended to be externally consumed
- DatePickerShortcut, DateRangeShortcut - exported in this PR, I assume because these are helpful for typing props passed to the datetime components
- TimezoneSelect, type TimezoneSelectProps - already reexported

---

- DateInput, type DateInputProps - already reexported
- DatePicker, type DatePickerProps - only re-exporting input versions of deprecated datepicker components, let's not change that now
- DateRangeInput, type DateRangeInputProps - already reexported
- DateRangePicker, type DateRangePickerProps - only re-exporting input versions of deprecated datepicker components, let's not change that now
</details>

evansjohnson (commented):
Thanks for looking into this! I've found the lack of consistency here confusing in the past as well.

I imagine the original intent was to push users to import from `datetime`, as that is ultimately the package users will import all of these from, but I think the partial re-export state is more confusing than is worth it to push users to `datetime` imports early.

packages/datetime2/src/index.ts
bvandrc: >I imagine the original intent was to push users to import from datetime, as that is ultimately the package users will import all of these from

The opposite-- most components in the `datetime` package are deprecated and say to import X from `datetime2` (ie `DatePicker` to `DatePicker3`). Which makes it odd that these here all say the opposite (I was just following the convention here). I think it's because these are simply re-exported `datetime` items via `datetime2`. But IMO, it doesn't make sense to recommend doing that, because as [this issue](https://github.com/palantir/blueprint/issues/6739) mentioned, you should be able to import anything you need from `/datetime2` so that you don't have to list both as dependencies in your `package.json`.

In other words, I would be in favor of removing all these ` @deprecated` warnings.

bvandrc: Changed: https://github.com/palantir/blueprint/pull/6868/commits/3669f3e6519cb2184a7867435b72bcf7e2ec2b07

The `as x2` ones still have `@deprecated`:

image

packages/datetime2/src/index.ts
bvandrc: I had just added the types we needed for our project. Thanks for the complete list! Added most of the missing ones, aside from:

- `DatePickerBaseProps`, `DatePickerModifiers` - Many of these are omitted from `DatePicker3Props` so don't really have any relevancy being exported from `datetime2`
- `DatePickerLocaleUtils` - `localeUtils` omitted from `DatePicker3Props`, so has no relevance. Would get the needed type from `DayPickerProps`
- `DatePickerDayModifiers` -`modifiers` omitted from `DatePicker3Props`, so has no relevance. Would get the needed type from `DayPickerProps`

packages/datetime2/src/index.ts
evansjohnson: I think the deprecation warnings in datetime to use the v3 components are because the old components themselves are deprecated. The deprecation warnings here are to push users to import from `datetime` when available, because (I imagine) that is ultimately the package that the v3 components will be promoted to.

Ultimately it doesn't matter too much what we do here since this is just a deprecation warning, but I think leaving the deprecation warning for all of these re-exports is the right thing to do since if followed it will lead to less churn when ultimately updating to the next major blueprint version.

packages/datetime2/src/index.ts
```diff
@@ -27,25 +27,28 @@ export { Classes as Datetime2Classes, ReactDayPickerClasses } from "./classes";
...
 
 export {
-    /** @deprecated import from `@blueprintjs/datetime` or use `Datetime2Classes` instead */
+    /** @deprecated use `Datetime2Classes` instead */
     Classes,
-    type DateFormatProps,
-    /** @deprecated import from `@blueprintjs/datetime` instead */
     DateInput as DateInput2,
-    /** @deprecated import from `@blueprintjs/datetime` instead */
+    /** @deprecated use `DateInput3Props` instead */
```
evansjohnson: I would be in support of improving this deprecation message with an "or preferably use ___3"

evansjohnson (commented):
thanks for taking another pass! happy to approve if we add deprecation messages back 

with that I'm leaning towards respecting existing behavior more so than actually thinking they're helpful, but I do see some argument for them which I left in the comment

packages/datetime2/src/index.ts
bvandrc: I see, didn't realize that these components specific to `datetime2` would be brought into `datetime`

packages/datetime2/src/index.ts
bvandrc: just changed it back to the old one

packages/datetime2/src/index.ts
evansjohnson: That's my understanding, though I am new to working on Blueprint and we may end up doing something else. Thank you for updating to keep the same pattern for now!

packages/datetime2/test/isotest.mjs
```diff
@@ -31,7 +31,13 @@ describe("@blueprintjs/datetime2 isomorphic rendering", () => {
             DateRangePicker3: {},
         },
         {
-            excludedSymbols: ["DateInput2", "DateInput2MigrationUtils", "DateRangeInput2", "TimezoneSelect"],
+            excludedSymbols: [
+                "DateInput2",
+                "DateInput2MigrationUtils",
+                "DateRangeInput2",
+                "MonthAndYear",
```
evansjohnson, line 38: huh, do you know what this was needed for?

bvandrc: The test was failing and `MonthAndYear` is also excluded in the same place in `/datetime/test/isotest.mjs`

evansjohnson (approved):

PR #6875, opened 2024-06-26
style: tooltip props refactor

bvandrc:
If you are just changing the `@default` value, don't need to re-write the whole description when the original is in the `extends`:

image

image

### DIFF
--- a/packages/core/src/components/popover/popover.tsx
+++ b/packages/core/src/components/popover/popover.tsx
@@ -75,11 +75,6 @@ export interface PopoverProps<TProps extends DefaultPopoverTargetHTMLProps = Def
     backdropProps?: React.HTMLProps<HTMLDivElement>;
 
     /**
-     * The content displayed inside the popover.
-     */
-    content?: string | React.JSX.Element;
-
-    /**
      * The kind of interaction that triggers the display of the popover.
      *
      * @default "click"
@@ -301,7 +296,7 @@ export class Popover<
         }
     }
 
-    protected validateProps(props: PopoverProps<T> & { children?: React.ReactNode }) {
+    protected validateProps(props: PopoverProps<T>) {
         if (props.isOpen == null && props.onInteraction != null) {
             console.warn(Errors.POPOVER_WARN_UNCONTROLLED_ONINTERACTION);
         }
@@ -417,7 +412,7 @@ export class Popover<
                 tabIndex: targetTabIndex,
             });
         } else {
-            const childTarget = Utils.ensureElement(React.Children.toArray(children)[0])!;
+            const childTarget = Utils.ensureElement(React.Children.toArray(children)[0]);
 
             if (childTarget === undefined) {
                 return null;
--- a/packages/core/src/components/tooltip/tooltip.tsx
+++ b/packages/core/src/components/tooltip/tooltip.tsx
@@ -29,11 +29,6 @@ export interface TooltipProps<TProps extends DefaultPopoverTargetHTMLProps = Def
     extends Omit<PopoverSharedProps<TProps>, "shouldReturnFocusOnClose">,
         IntentProps {
     /**
-     * The content that will be displayed inside of the tooltip.
-     */
-    content: React.JSX.Element | string;
-
-    /**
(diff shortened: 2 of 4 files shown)


packages/core/src/components/tooltip/tooltip.tsx
```diff
@@ -28,11 +28,6 @@ import { TooltipContext, type TooltipContextState, TooltipProvider } from "../po
 export interface TooltipProps<TProps extends DefaultPopoverTargetHTMLProps = DefaultPopoverTargetHTMLProps>
     extends Omit<PopoverSharedProps<TProps>, "shouldReturnFocusOnClose">,
         IntentProps {
-    /**
-     * The content that will be displayed inside of the tooltip.
-     */
-    content: React.JSX.Element | string;
```
evansjohnson, line 34: Content is currently required on `Tooltip`, which I think is a nice affordance when using on what needs to be provided to get a tooltip showing. There's already easy ways to not render the tooltip (empty content, or disabled) so I don't think we need to loosen the type here.

packages/core/src/components/popover/popover.tsx
```diff
@@ -417,7 +412,7 @@ export class Popover<
                 tabIndex: targetTabIndex,
             });
         } else {
-            const childTarget = Utils.ensureElement(React.Children.toArray(children)[0])!;
```
evansjohnson, line 420: :+1: we check `childTarget === undefined` right after this anyways, no need for the `!`

packages/core/src/components/tooltip/tooltip.tsx
```diff
@@ -42,19 +37,11 @@ export interface TooltipProps<TProps extends DefaultPopoverTargetHTMLProps = Def
     compact?: boolean;
 
     /**
-     * The amount of time in milliseconds the tooltip should remain open after
-     * the user hovers off the trigger. The timer is canceled if the user mouses
-     * over the target before it expires.
-     *
```
evansjohnson: preference to keep this because it impacts the generated docs: it looks like this is done in support of the docs generation: https://output.circle-artifacts.com/output/job/dc1e766c-4230-46d5-84d0-d465ee9e4dd4/artifacts/0/packages/docs-app/dist/index.html#core/components/tooltip

you can click through to the shared type to see the docs but I don't think having it duplicated is that bad if it makes the docs easier to use

packages/core/test/popover/popoverTests.tsx
```diff
@@ -78,7 +78,7 @@ describe("<Popover>", () => {
             assert.isTrue(warnSpy.calledWith(Errors.POPOVER_WARN_TOO_MANY_CHILDREN));
         });
 
-        it("warns if given children and target prop", () => {
+        it("warns if given children and renderTarget prop", () => {
```
evansjohnson, line 81: :+1:

evansjohnson (changes requested):

packages/core/src/components/tooltip/tooltip.tsx
bvandrc: I didn't notice that `Popover` had it required and `Tooltip` didn't, before I consolidated them.

Is there any reason that `Popover` shouldn't have `content` be required? You should always have content, or else it's not a popover...

bvandrc: Changed to be required.

packages/core/src/components/tooltip/tooltip.tsx
bvandrc: Ah good catch. That's annoyng. Alright reverted!

packages/core/src/components/tooltip/tooltip.tsx
evansjohnson: It would be a breaking change (thinking of a build failure from not providing required props) to tighten the typing now, though perhaps this would have been better to do from the start

bvandrc: OK. I'm just going to delete this PR then since none of these changes have gotten approved, so there's nothing really remaining in the PR

PR #6877, opened 2024-06-27
Add affordance to icon selector in docs and fix selection state

evansjohnson:
#### Checklist

- [n/a] Includes tests
- [n/a] Update documentation

- Add better affordance that something needs to be typed for icons to show. I thought the docs were broken and didn't initially try typing anything here.
- Fix selection/active state. https://github.com/palantir/blueprint/commit/f0a9b09f715cfcc36ba237dcbcd4c1c90dfd21f2#diff-8194812817b7f429b5320bb9be65c267dea2308f9a5b914411859fc8d61d18b0R222 seems to be a breaking change (see how some other docs examples were updated). Maybe this was only part of the initial BP 5 release, but didn't seem to be mentioned in the upgrade guide. In any case this has been out long enough that it should be considered expected behavior at this point
- Update docs to mention `selected` only applies with the `"listoption"` role - I had to dig into the code to figure out what was going wrong here and initially thought selection state was entirely broken.

- Anything else related that should be fixed
- Any unintended breaking changes from this PR

#### Screenshot
Current:
I found this to be confusing - [this looks like it came up in the PR that introduced but it](https://github.com/palantir/blueprint/pull/2028#issuecomment-362153856) was a big one and I suppose was never addressed
Screenshot 2024-06-27 at 3 01 48 PM
no indication of selection within menu - note selected icon shown in button:
Screenshot 2024-06-27 at 3 01 55 PM

This PR:
custom placeholder text:
Screenshot 2024-06-27 at 3 02 10 PM
selection state:
Screenshot 2024-06-27 at 3 02 15 PM

### DIFF
--- a/packages/core/src/components/menu/menuItem.tsx
+++ b/packages/core/src/components/menu/menuItem.tsx
@@ -129,7 +129,7 @@ export interface MenuItemProps
     /**
      * Whether this item should appear selected - `roleStructure` must be `"listoption"` for this to be
      * applied. Defining this will set the `aria-selected` attribute and apply a small tick icon if `true`,
-     * and space for a small tick icon if `false`.
+     * and empty space for a small tick icon if `false` or `undefined`.
      *
      * @default undefined
      */
--- a/packages/core/src/components/menu/menuItem.tsx
+++ b/packages/core/src/components/menu/menuItem.tsx
@@ -127,10 +127,9 @@ export interface MenuItemProps
     popoverProps?: Partial<Omit<PopoverProps, "content" | "minimal">>;
 
     /**
-     * Whether this item should appear selected.
-     * Defining this  will set the `aria-selected` attribute and apply a
-     * "check" or "blank" icon on the item (unless the `icon` prop is set,
-     * which always takes precedence).
+     * Whether this item should appear selected - `roleStructure` must be `"listoption"` for this to be
+     * applied. Defining this will set the `aria-selected` attribute and apply a small tick icon if `true`,
+     * and space for a small tick icon if `false`.
      *
      * @default undefined
      */
--- a/packages/docs-app/src/examples/core-examples/common/iconSelect.tsx
+++ b/packages/docs-app/src/examples/core-examples/common/iconSelect.tsx
@@ -43,6 +43,7 @@ export class IconSelect extends React.PureComponent<IconSelectProps> {
                     itemPredicate={this.filterIconName}
                     itemRenderer={this.renderIconItem}
                     noResults={<MenuItem disabled={true} text="No results" />}
+                    placeholder="Start typing to search…"
                     onItemSelect={this.handleIconChange}
                     popoverProps={{ minimal: true }}
                 >
@@ -77,9 +78,6 @@ export class IconSelect extends React.PureComponent<IconSelectProps> {
     };
 
     private filterIconName = (query: string, iconName: IconName | typeof NONE) => {
-        if (iconName === NONE) {
-            return true;
-        }
         if (query === "") {
             return iconName === this.props.iconName;
         }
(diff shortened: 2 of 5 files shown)


evansjohnson:
Noticed https://github.com/palantir/blueprint/issues/6878 while testing but don't want to put more time into this for now

packages/select/src/components/select/select.tsx
```diff
@@ -159,6 +167,7 @@ export class Select<T> extends AbstractPureComponent<SelectProps<T>, SelectState
             filterable = true,
             disabled = false,
             inputProps = {},
+            placeholder = "Filter...",
```
evansjohnson, line 170: I used an ellipsis in `Start typing to search…` but didn't want to convert all ellipsis in BP to the actual character until a major version bump since this could be a breaking change (tests that rely on finding elements by default text)

packages/select/src/components/select/select.tsx
```diff
@@ -81,6 +81,14 @@ export interface SelectProps<T> extends ListItemsProps<T>, SelectPopoverProps {
      */
     menuProps?: React.HTMLAttributes<HTMLUListElement>;
 
+    /**
+     * A placeholder string passed to the filter text input.
+     * Applicable only when `filterable` is `true`.
+     *
+     * @default "Filter..."
+     */
+    placeholder?: string;
```
evansjohnson, line 90: calling out this small new feature. this has some potential to be used inappropriately, but I don't think "Filter..." is always the best indication here particularly since the user is responsible for defining the match predicate and may not strictly implement a filter

packages/docs-app/src/examples/core-examples/common/iconSelect.tsx
```diff
@@ -77,9 +80,6 @@ export class IconSelect extends React.PureComponent<IconSelectProps> {
     };
 
     private filterIconName = (query: string, iconName: IconName | typeof NONE) => {
-        if (iconName === NONE) {
-            return true;
-        }
         if (query === "") {
```
evansjohnson, line 83: this prevents loading the full list on empty query - there's over 600 icons and it's a bit laggy if showing all... and doesn't quite feel worth bringing in a virtualized list for

evansjohnson: I was a bit surprised it was starting to struggle with just 600 items, in case this raises any concerns

packages/core/src/components/menu/menuItem.tsx
```diff
@@ -127,10 +127,9 @@ export interface MenuItemProps
     popoverProps?: Partial<Omit<PopoverProps, "content" | "minimal">>;
 
     /**
-     * Whether this item should appear selected.
-     * Defining this  will set the `aria-selected` attribute and apply a
-     * "check" or "blank" icon on the item (unless the `icon` prop is set,
-     * which always takes precedence).
+     * Whether this item should appear selected - `roleStructure` must be `"listoption"` for this to be
```
gluxon, line 130: Thank you for adding this comment!

> `roleStructure` must be `"listoption"` for this to be applied.

This prop being a no-op depending on `roleStructure` would have surprised me too, especially since `"listoption"` isn't the default.

packages/select/src/components/select/select.tsx
gluxon: I agree just "Filter..." may not be that valuable to a user. I also think we'd want this to be customizable for translations. I'm actually surprised a hard-coded English string has survived this long in Blueprint without someone flagging that a translation is necessary.

Thanks for adding!

gluxon (dismissed):
Thank you!

gluxon (approved):

PR #6881, opened 2024-06-28
Fix DateRangeInput3 focus with timePrecision

### LINKED ISSUE #6879: Focus issues with DateRangeInput3 if timePickerProps.precision is set
#### Steps to reproduce
https://blueprintjs.com/docs/#datetime2/date-range-input3
- enable "Enable time picker"
- set precision to seconds
- open date range input
- click into time input
- use up arrow to increment time
- notice focus is lost

#### Actual behavior
focus is lost on time picker when adjusting time if precision is set through `timePickerProps`

#### Expected behavior
component should work if precision is set through `timePickerProps` as well as `timePrecision`

#### Possible solution
opening PR

evansjohnson:
#### Checklist
- [x] Includes tests
- [n/a] Update documentation

Respect `precision` passed to `timePickerProps` as well as the root `timePrecision` prop when considering for focus.

- Any mistakes
- Anywhere else I may have missed
- Worth sharing the implementation between components?

#### Reproducer
See linked issue

#### Recordings
Develop:
https://github.com/palantir/blueprint/assets/14102129/5e07692f-f433-40aa-af95-3da76cbef0f3

This PR:
https://github.com/palantir/blueprint/assets/14102129/042d91bc-81a2-4198-8066-89484b771387

### DIFF
--- a/packages/datetime2/src/components/date-range-input3/dateRangeInput3.tsx
+++ b/packages/datetime2/src/components/date-range-input3/dateRangeInput3.tsx
@@ -39,6 +39,7 @@ import {
     Errors,
     type NonNullDateRange,
     type TimePrecision,
+    TimePicker,
 } from "@blueprintjs/datetime";
 
 import { Classes } from "../../classes";
@@ -292,7 +293,13 @@ export class DateRangeInput3 extends DateFnsLocalizedComponent<DateRangeInput3Pr
     private getTimePrecision = () => {
         // timePrecision may be set as a root prop or as a property inside timePickerProps, so we need to check both
         const { timePickerProps, timePrecision } = this.props;
-        return timePickerProps?.precision ?? timePrecision;
+        const fromProps = timePickerProps?.precision ?? timePrecision;
+        if (fromProps != null) {
+            return fromProps;
+        }
+
+        // if timePickerProps is defined, the `TimePicker` will render with it's default `precision`
+        return timePickerProps != null ? TimePicker.defaultProps.precision : undefined;
     };
 
     private handleDateRangePickerChange = (selectedRange: DateRange, didSubmitWithEnter = false) => {
--- a/packages/datetime2/src/components/date-range-input3/dateRangeInput3.tsx
+++ b/packages/datetime2/src/components/date-range-input3/dateRangeInput3.tsx
@@ -38,8 +38,8 @@ import {
     DateUtils,
     Errors,
     type NonNullDateRange,
-    type TimePrecision,
     TimePicker,
+    type TimePrecision,
 } from "@blueprintjs/datetime";
 
 import { Classes } from "../../classes";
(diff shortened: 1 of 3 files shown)


packages/datetime2/src/components/date-range-input3/dateRangeInput3.tsx
```diff
@@ -288,6 +290,18 @@ export class DateRangeInput3 extends DateFnsLocalizedComponent<DateRangeInput3Pr
...
+    private getTimePrecision = () => {
+        // timePrecision may be set as a root prop or as a property inside timePickerProps, so we need to check both
+        const { timePickerProps, timePrecision } = this.props;
+        const fromProps = timePickerProps?.precision ?? timePrecision;
+        if (fromProps != null) {
+            return fromProps;
+        }
+
+        // if timePickerProps is defined, the `TimePicker` will render with it's default `precision`
+        return timePickerProps != null ? TimePicker.defaultProps.precision : undefined;
```
gluxon, line 302: A bit confused by the comment above. Will the `TimePicker` still render with its default precision if `timePickerProps` is `undefined`?

I'm curious how important it is to return `TimePicker.defaultProps.precision` vs `undefined` here.

packages/datetime2/test/components/dateRangeInput3Tests.tsx
```diff
@@ -177,45 +177,54 @@ describe("<DateRangeInput3>", () => {
         expectPropValidationError(DateRangeInput3, { ...DATE_FORMAT, value: null! });
     });
 
-    describe("timePrecision prop", () => {
-        it("<TimePicker /> should not lose focus on increment/decrement with up/down arrows", () => {
-            const { root } = wrap(<DateRangeInput3 {...DATE_FORMAT} timePrecision={TimePrecision.MINUTE} />, true);
+    describe("<TimePicker /> focus", () => {
+        // there are multiple ways to render the underlying TimePicker component so run same tests on all
```
gluxon, line 181: Nice coverage of tests! I like that we're checking all these variants.

packages/datetime2/src/components/date-range-input3/dateRangeInput3.tsx
evansjohnson: because of this bit of existing behavior, which I tried to explain more here: https://github.com/palantir/blueprint/pull/6881/files#diff-d2a31417b184106fdf4054399ab49455504ff747c19714aa6df0bb7f8e26bf90R238-R239

let me know any suggestions you can think of to help clarify this

evansjohnson: if `timePickerProps` is not defined, the underlying `DateRangePicker` will have default `timePickerProps`, and therefor not show the time pickers - so in that case we say the precision is `undefined.

if `timePickerProps` is defined, we return the default `precision`, since at this point in this method we already would have returned an explicitly defined precision either directly on this component, or from the `timePickerProps`

Issue #4780, opened 2021-06-24
Table Whites Out With Mouse Scroll Greater Than 3 Rows


S0urcemaster:
#### Environment

Try out on the blueprint table feature docs page..

#### Code Sandbox

https://blueprintjs.com/docs/#table/features

#### Steps to reproduce

Set mouse scroll (windows) to greater than 3 lines.
Scroll the table.

#### Actual behavior

When scrolling up the first two rows sometimes don't get displayed/updated.

![grafik](https://user-images.githubusercontent.com/68502837/123255842-93b0a900-d4f0-11eb-949d-4296f2bdf56b.png)

#### Expected behavior

Every row always gets displayed.

#### Possible solution

Only seen that with blueprint table so I assume the problem is located there.

pulzar:
Any workarounds? For default scroll speed on mac I also get a number of rows that don't render on mouse scroll.

til-schneider:
I have the same problem. My workaround is:

```ts
<Table2
   ...
   onVisibleCellsChange={() => setTimeout(() => this.forceUpdate())}
/>
```

adidahiya:
@til-schneider are you still hitting this issue with the latest versions of `@blueprintjs/table`? There have been some fixes implemented in the past year related to scrolling & rendering.

til-schneider:
Yes, I'm using `@blueprintjs/table` version 4.6.7.

The problem doesn't occur on every scrolling, but maybe 1 out of 10 times, tested in Chrome on a Mac.

jandrew44:
The issue is reproducible with @blueprintjs/table 5.0.22 and Chrome browser on Win 10.
Steps to reproduce:
1) Scroll table down
2) Single click on mouse wheel to enter the scroll mode
3) Make one small and single scroll up move by the mouse wheel

PR #6882, opened 2024-06-28
Fix: Table not keeping focused cell visible

### LINKED ISSUE #1609: Table not keeping focused cell visible
### Bug report

- blueprint.table 1.26.0 & blueprint.core 1.29.0
- Windows / Chrome

#### Steps to reproduce

1. Create table with enableFocus set true.  Include enough rows to enable scroll bar
2. Select first cell and arrow down to observe scroll/visibility behaviour
3. Use up arrow keys to focus on first cell
4. Use scroll bar to scroll up

#### Actual behavior

When using down arrow, focused cell is sometimes not fully visible.

arrow-down-visibility-bug

When using up arrow, top first cell hidden behind heading.

arrow-up-visibility-bug

Top cell becomes visible when scroll bar used to scroll up.

arrow-up-visibility-bug-scroll-fix

#### Expected behavior

Focused cell always fully visible.

filipbalucha:
#### Checklist

- [x] Includes tests
- [ ] Update documentation

- Make sure the focused cell is visible on click or keyboard navigation (including when controlled via props)
- Improve internal docs
- Add pointers for other improvements uncovered as part of the work here (see end of PR)

#### Screenshot

Before:

https://github.com/palantir/blueprint/assets/40673216/fef9d318-f6da-46d2-9032-d0248940d2fe

After:

https://github.com/palantir/blueprint/assets/40673216/322ce3ec-afee-4abe-8839-3f9db9b13318

### Not addressed in this PR

- Polishing interactions with frozen cells. The scrollable part of the table should only scroll if the frozen cell would otherwise not be visible. I file a FR after this merges and maybe FLUP.

  https://github.com/palantir/blueprint/assets/40673216/1a637d89-9d2a-4f01-b509-fae83533a3d4

- `getColumnIndicesInRect` operates on viewport rect but does not ignore headers or take frozen cells into account; this might be feeding into https://github.com/palantir/blueprint/issues/4780 and potentially contaminating the response of `onVisibleCellsChange`. Again, this is a side observation rather than a regression.

### DIFF
--- a/packages/table/src/table2.tsx
+++ b/packages/table/src/table2.tsx
@@ -231,17 +231,15 @@ export class Table2 extends AbstractComponent<Table2Props, TableState, TableSnap
     private refHandlers = {
         cellContainer: (ref: HTMLElement | null) => (this.cellContainerElement = ref),
         columnHeader: (ref: HTMLElement | null) => {
-            this.columnHeaderElement = ref;
             if (ref != null) {
-                this.columnHeaderHeight = Math.max(ref.clientHeight, Grid.MIN_COLUMN_HEADER_HEIGHT);
+                this.setState({ columnHeaderHeight: Math.max(ref.clientHeight, Grid.MIN_COLUMN_HEADER_HEIGHT) });
             }
         },
         quadrantStack: (ref: TableQuadrantStack) => (this.quadrantStackInstance = ref),
         rootTable: (ref: HTMLElement | null) => (this.rootTableElement = ref),
         rowHeader: (ref: HTMLElement | null) => {
-            this.rowHeaderElement = ref;
             if (ref != null) {
-                this.rowHeaderWidth = ref.clientWidth;
+                this.setState({ rowHeaderWidth: ref.clientWidth });
             }
         },
         scrollContainer: (ref: HTMLElement | null) => (this.scrollContainerElement = ref),
@@ -249,17 +247,9 @@ export class Table2 extends AbstractComponent<Table2Props, TableState, TableSnap
 
     private cellContainerElement?: HTMLElement | null;
 
-    private columnHeaderElement?: HTMLElement | null;
-
-    private columnHeaderHeight = Grid.MIN_COLUMN_HEADER_HEIGHT;
-
     private quadrantStackInstance?: TableQuadrantStack;
 
     private rootTableElement?: HTMLElement | null;
-
-    private rowHeaderElement?: HTMLElement | null;
-
-    private rowHeaderWidth = Grid.MIN_ROW_HEADER_WIDTH;
 
     private scrollContainerElement?: HTMLElement | null;
 
@@ -312,6 +302,7 @@ export class Table2 extends AbstractComponent<Table2Props, TableState, TableSnap
 
         this.state = {
             childrenArray,
+            columnHeaderHeight: 0,
             columnIdToIndex,
             columnWidths: newColumnWidths,
             didHeadersMount: false,
@@ -321,6 +312,7 @@ export class Table2 extends AbstractComponent<Table2Props, TableState, TableSnap
             isReordering: false,
             numFrozenColumnsClamped: clampNumFrozenColumns(props),
             numFrozenRowsClamped: clampNumFrozenRows(props),
+            rowHeaderWidth: 0,
             rowHeights: newRowHeights,
             selectedRegions,
             verticalGuides: [],
@@ -734,7 +726,7 @@ export class Table2 extends AbstractComponent<Table2Props, TableState, TableSnap
         }
 
         const rowIndices = this.grid.getRowIndicesInRect({
-            columnHeaderHeight: enableColumnHeader ? this.columnHeaderHeight : 0,
+            columnHeaderHeight: enableColumnHeader ? this.state.columnHeaderHeight : 0,
             includeGhostCells: enableGhostCells!,
             rect: viewportRect,
         });
@@ -893,7 +885,7 @@ export class Table2 extends AbstractComponent<Table2Props, TableState, TableSnap
         // if we have horizontal overflow or exact fit, no need to render ghost columns
         // (this avoids problems like https://github.com/palantir/blueprint/issues/5027)
         const hasHorizontalOverflowOrExactFit = this.locator.hasHorizontalOverflowOrExactFit(
-            enableRowHeader ? this.rowHeaderWidth : 0,
+            enableRowHeader ? this.state.rowHeaderWidth : 0,
             viewportRect,
         );
         const columnIndices = this.grid.getColumnIndicesInRect(
@@ -976,7 +968,7 @@ export class Table2 extends AbstractComponent<Table2Props, TableState, TableSnap
         // if we have vertical overflow or exact fit, no need to render ghost rows
         // (this avoids problems like https://github.com/palantir/blueprint/issues/5027)
         const hasVerticalOverflowOrExactFit = this.locator.hasVerticalOverflowOrExactFit(
-            enableColumnHeader ? this.columnHeaderHeight : 0,
+            enableColumnHeader ? this.state.columnHeaderHeight : 0,
             viewportRect,
         );
         const rowIndices = this.grid.getRowIndicesInRect({
@@ -1073,11 +1065,11 @@ export class Table2 extends AbstractComponent<Table2Props, TableState, TableSnap
         // if we have vertical/horizontal overflow or exact fit, no need to render ghost rows/columns (respectively)
         // (this avoids problems like https://github.com/palantir/blueprint/issues/5027)
         const hasVerticalOverflowOrExactFit = this.locator.hasVerticalOverflowOrExactFit(
-            enableColumnHeader ? this.columnHeaderHeight : 0,
+            enableColumnHeader ? this.state.columnHeaderHeight : 0,
             viewportRect,
         );
         const hasHorizontalOverflowOrExactFit = this.locator.hasHorizontalOverflowOrExactFit(
-            enableRowHeader ? this.rowHeaderWidth : 0,
+            enableRowHeader ? this.state.rowHeaderWidth : 0,
             viewportRect,
         );
         const rowIndices = this.grid.getRowIndicesInRect({
@@ -1478,30 +1470,21 @@ export class Table2 extends AbstractComponent<Table2Props, TableState, TableSnap
     private syncViewportPosition = ({ nextScrollLeft, nextScrollTop }: TableSnapshot) => {
         const { viewportRect } = this.state;
 
-        if (this.scrollContainerElement == null || this.columnHeaderElement == null || viewportRect === undefined) {
+        if (this.scrollContainerElement == null || viewportRect === undefined) {
             return;
         }
 
         if (nextScrollLeft !== undefined || nextScrollTop !== undefined) {
-            // we need to modify the scroll container explicitly for the viewport to shift. in so
-            // doing, we add the size of the header elements, which are not technically part of the
-            // "grid" concept (the grid only consists of body cells at present).
             if (nextScrollTop !== undefined) {
-                const topCorrection = this.shouldDisableVerticalScroll() ? 0 : this.columnHeaderElement.clientHeight;
-                this.scrollContainerElement.scrollTop = nextScrollTop + topCorrection;
+                this.scrollContainerElement.scrollTop = nextScrollTop;
             }
             if (nextScrollLeft !== undefined) {
-                const leftCorrection =
-                    this.shouldDisableHorizontalScroll() || this.rowHeaderElement == null
-                        ? 0
-                        : this.rowHeaderElement.clientWidth;
-
-                this.scrollContainerElement.scrollLeft = nextScrollLeft + leftCorrection;
+                this.scrollContainerElement.scrollLeft = nextScrollLeft;
             }
 
             const nextViewportRect = new Rect(
-                nextScrollLeft ?? 0,
-                nextScrollTop ?? 0,
+                nextScrollLeft ?? viewportRect.left,
+                nextScrollTop ?? viewportRect.top,
                 viewportRect.width,
                 viewportRect.height,
             );
--- a/packages/table/src/table2.tsx
+++ b/packages/table/src/table2.tsx
@@ -37,6 +37,7 @@ import * as ScrollUtils from "./common/internal/scrollUtils";
 import { Rect } from "./common/rect";
 import { RenderMode } from "./common/renderMode";
 import { ScrollDirection } from "./common/scrollDirection";
+import type { TableHeaderDimensions } from "./common/TableHeaderDimensions";
 import { Utils } from "./common/utils";
 import { ColumnHeader } from "./headers/columnHeader";
 import { ColumnHeaderCell, type ColumnHeaderCellProps } from "./headers/columnHeaderCell";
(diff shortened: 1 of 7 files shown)


palantirtech:
Thanks for your interest in [palantir/blueprint](https://github.com/palantir/blueprint), @filipbalucha! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

packages/table/src/table2.tsx
```diff
@@ -231,36 +231,26 @@ export class Table2 extends AbstractComponent<Table2Props, TableState, TableSnap
     private refHandlers = {
         cellContainer: (ref: HTMLElement | null) => (this.cellContainerElement = ref),
         columnHeader: (ref: HTMLElement | null) => {
-            this.columnHeaderElement = ref;
             if (ref != null) {
-                this.columnHeaderHeight = Math.max(ref.clientHeight, Grid.MIN_COLUMN_HEADER_HEIGHT);
+                this.setState({ columnHeaderHeight: Math.max(ref.clientHeight, Grid.MIN_COLUMN_HEADER_HEIGHT) });
```
filipbalucha: Making this part of state so that `TableHotkeys` can access it and factor it into its calculations [here](https://github.com/palantir/blueprint/pull/6882/files#diff-8cd23c90b1959192d311f6112803042130c4c7709017e6cac6fd991ea89221beR316-R317).

packages/table/src/table2.tsx
```diff
@@ -1478,30 +1470,21 @@ export class Table2 extends AbstractComponent<Table2Props, TableState, TableSnap
...
 
-        if (this.scrollContainerElement == null || this.columnHeaderElement == null || viewportRect === undefined) {
+        if (this.scrollContainerElement == null || viewportRect === undefined) {
             return;
         }
 
         if (nextScrollLeft !== undefined || nextScrollTop !== undefined) {
-            // we need to modify the scroll container explicitly for the viewport to shift. in so
-            // doing, we add the size of the header elements, which are not technically part of the
-            // "grid" concept (the grid only consists of body cells at present).
```
filipbalucha, line 1488: This is not true. `scrollTop` should not be dependent on header size here:


https://github.com/palantir/blueprint/assets/40673216/e06954c5-f750-4029-a020-f45745a988d8

packages/table/src/table2.tsx
```diff
@@ -1478,30 +1470,21 @@ export class Table2 extends AbstractComponent<Table2Props, TableState, TableSnap
...
-
-                this.scrollContainerElement.scrollLeft = nextScrollLeft + leftCorrection;
+                this.scrollContainerElement.scrollLeft = nextScrollLeft;
             }
 
             const nextViewportRect = new Rect(
-                nextScrollLeft ?? 0,
-                nextScrollTop ?? 0,
+                nextScrollLeft ?? viewportRect.left,
+                nextScrollTop ?? viewportRect.top,
```
filipbalucha, line 1488: Unless both `nextScrollLeft` and `nextScrollTop` were used, we would set one value to 0 here. However, `viewportRect`'s top-left point is the value by which the scrollable cells have been scrolled. So, we would be scrolling this back to the first row / column (likely causing batcher to schedule the rendering of cells in that region before being corrected by some other prop change).

evansjohnson:
docs-app preview: https://output.circle-artifacts.com/output/job/017ed34c-0b72-43cb-97fe-488fb0795dcf/artifacts/0/packages/docs-app/dist/index.html

table-app preview: https://output.circle-artifacts.com/output/job/017ed34c-0b72-43cb-97fe-488fb0795dcf/artifacts/0/packages/table-dev-app/dist/index.html

packages/table/src/tableState.ts
```diff
@@ -87,8 +91,28 @@ export interface TableState {
     verticalGuides: number[];
 
     /**
-     * The `Rect` bounds of the viewport used to perform virtual viewport
-     * performance enhancements.
+     * The bounds of the table. This includes:
+     * - Column and row headers,
```
evansjohnson: From my understanding, the size of the `viewportRect` includes the headers/frozen cells and we account for this when positioning the scrolling cells, but only the scrolling cells are rendered within the `viewportRect`. I'd expect the headers/frozen cells to be rendered in separate overlayed viewports but haven't confirmed this.

Does that sound correct?

packages/table/src/table2.tsx
evansjohnson: if this is just moved to state to share with `TableHotkeys` I'd prefer to instead have a new way of passing the existing ref storage, ex by passing `new TableHotkeys` a callback that returns `this.columnHeaderHeight`

packages/table/src/table2.tsx
evansjohnson: I think this change makes sense. In `getSnapshotBeforeUpdate` I'm seeing

```
// these will only be defined if they differ from viewportRect
return { nextScrollLeft, nextScrollTop };
```

so using the current viewportRect top/width instead of 0 sounds like a better default

---

I think we could merge this now if you'd like to split this little bit off

packages/table/src/tableHotkeys.ts
```diff
@@ -52,8 +52,16 @@ export class TableHotkeys {
...
 
-    public setState(state: TableState) {
-        this.state = state;
+    public setState(newState: TableState) {
+        if (
+            newState.focusedCell != null &&
+            (this.state.focusedCell == null || this.state.focusedCell !== newState.focusedCell)
+        ) {
+            // Ensure controlled focused cell is scrolled into view
+            this.scrollBodyToFocusedCell(newState.focusedCell);
```
evansjohnson, line 62: it looks like `focusedCell` can be provided in props which may cause this scroll call to be missed

```
// only set focused cell state if not specified in props
        if (this.props.focusedCell == null) {
            this.setState({ focusedCell });
        }
```

evansjohnson: thoughts on adding this `scrollBodyToFocusedCell` here while leaving it in the places it was already? is there no concern with calling this twice?

what are the cases this handled that are currently missed by `handleFocusMove` and `handleFocusMoveInternal` where this is called at the end?

packages/table/src/table2.tsx
evansjohnson: if that's the case then I'd expect the viewportRect to not include space for the headers, re https://github.com/palantir/blueprint/pull/6882/files#diff-c2ca442c1f8367c203de64884a8526876c32365d9dd5f071887fb16183ada632R100

packages/table/src/table2.tsx
```diff
@@ -1478,30 +1470,21 @@ export class Table2 extends AbstractComponent<Table2Props, TableState, TableSnap
...
+                this.scrollContainerElement.scrollTop = nextScrollTop;
             }
             if (nextScrollLeft !== undefined) {
-                const leftCorrection =
-                    this.shouldDisableHorizontalScroll() || this.rowHeaderElement == null
-                        ? 0
-                        : this.rowHeaderElement.clientWidth;
-
-                this.scrollContainerElement.scrollLeft = nextScrollLeft + leftCorrection;
+                this.scrollContainerElement.scrollLeft = nextScrollLeft;
```
evansjohnson, line 1483: the change here removing the "correction" looks like the scrollContainer and viewportRect will have the same scroll top/left now - which seems to be different than what https://github.com/palantir/blueprint/pull/6882/files#diff-8cd23c90b1959192d311f6112803042130c4c7709017e6cac6fd991ea89221beR313-R318 is saying

packages/table/src/tableHotkeys.ts
```diff
@@ -295,19 +303,39 @@ export class TableHotkeys {
...
+
         // sort keys in normal CSS position order (per the trusty TRBL/"trouble" acronym)
         // tslint:disable:object-literal-sort-keys
-        const viewportBounds = {
+        const scrollableContainerBounds = {
             top: viewportRect.top,
-            right: viewportRect.left + viewportRect.width,
-            bottom: viewportRect.top + viewportRect.height,
+            right: viewportRect.left + viewportRect.width - frozenColumnsWidth - this.state.rowHeaderWidth,
+            bottom: viewportRect.top + viewportRect.height - frozenRowsHeight - this.state.columnHeaderHeight,
```
evansjohnson: it seems unexpected these would be subtracted from right and bottom since the frozen cells/headers are rendered in the top and left

packages/table/src/tableHotkeys.ts
```diff
@@ -295,19 +303,39 @@ export class TableHotkeys {
...
+
         const focusedCellBounds = {
-            top: this.grid.getCumulativeHeightBefore(row),
-            right: this.grid.getCumulativeWidthAt(col),
-            bottom: this.grid.getCumulativeHeightAt(row),
-            left: this.grid.getCumulativeWidthBefore(col),
+            top: this.grid.getCumulativeHeightBefore(row) - frozenRowsHeight,
+            right: this.grid.getCumulativeWidthAt(col) - frozenColumnsWidth,
+            bottom: this.grid.getCumulativeHeightAt(row) - frozenRowsHeight,
+            left: this.grid.getCumulativeWidthBefore(col) - frozenColumnsWidth,
```
evansjohnson: this change makes sense as going along with the change of `viewportBounds` -> `scrollableContainerBounds`

evansjohnson (commented):
This is my first time really diving into the table code so doing my best to review here.

Ideally I'd like to see minimal changes, ex if we can do something without introducing a new setState call, that would be preferred.

Adding to my worries is that it looks like some table tests are not running, or may only be running on Table and not Table2: https://github.com/palantir/blueprint/issues?q=is%3Aissue+is%3Aopen+table2 - and that there are no new tests introduced in this PR.

---

As a next step, I think it may help if you can help clarify my comments with your thoughts, and consider if we can avoid refactoring the header size into state.

packages/table/src/tableHotkeys.ts
filipbalucha: > what are the cases this handled that are currently missed by `handleFocusMove` and `handleFocusMoveInternal` where this is called at the end?

BLUF: It helps when the `focusedCell` is controlled and we do not want to rely on native `Table2` navigation logic (`handleFocusMove`, `handleFocusMoveInternal`).

`handleFocusMove` and `handleFocusMoveInternal` correctly call `scrollBodyToFocusedCell`. But they do this even if `focusedCell` is controlled:

https://github.com/palantir/blueprint/blob/f2e9d6e049b3c52eef4ea7d9cb080b4dca23c7a1/packages/table/src/table2Utils.ts#L78-L85

Now, the table should not use its own navigation mechanism when `focusedCell` is controlled.

In our case, navigation logic cannot rely on `handleFocusMove` or `handleFocusMoveInternal`, and so our controlled `focusedCell` would not get scrolled into view.


> thoughts on adding this scrollBodyToFocusedCell here while leaving it in the places it was already?

Done!

> is there no concern with calling this twice?

Assuming ref-consistency of consumers, this should early exit without calling `scrollBodyToFocusedCell`, so I'm not too concerned.

packages/table/src/tableState.ts
filipbalucha: > the size of the `viewportRect` includes the headers/frozen cells

Correct.

> only the scrolling cells are rendered within the `viewportRect`

Not quite, per the above comment. The scrolling cells belong to the main quadrant:

> I'd expect the headers/frozen cells to be rendered in separate overlayed viewports but haven't confirmed this.

The Table is split into 4 quadrants (left, top-left, top, and body. All of them maintain their respective headers; the first 3 allow us to have frozen cells.

packages/table/src/tableHotkeys.ts
filipbalucha: > it seems unexpected these would be subtracted from right and bottom since the frozen cells/headers are rendered in the top and left

I agree. This comes down to the fact that the offset of the scrollable part of `viewportRect` is with respect to the point `O` after headers and frozen cells (see: [ASCII table](https://github.com/palantir/blueprint/pull/6882/files#diff-8cd23c90b1959192d311f6112803042130c4c7709017e6cac6fd991ea89221beR309-R323)). So, `top` and `left` are, in fact, correct; we just need to adjust `bottom` and `right` to get the correct bounds of the scrollable container.

See also:

https://github.com/palantir/blueprint/pull/6882/files#r1659176907

packages/table/src/table2.tsx
filipbalucha: > if that's the case then I'd expect the viewportRect to not include space for the headers

I don't think we can overload the semantics of `viewportRect`. `viewportRect` is interpreted as `Table2`'s rect in other codepaths – for example, here:

https://github.com/palantir/blueprint/blob/f2e9d6e049b3c52eef4ea7d9cb080b4dca23c7a1/packages/table/src/locator.ts#L185-L198


That being said, I do think there is space for refactors. I consider that outside of the scope for this PR though.

filipbalucha:
@evansjohnson This is now ready for another look. I have:
- revamped the logic in `TableHotkeys`,
- removed the state update,
- re-enabled tests
- created a diagram showing how `Table2` (and `Table`) handles the scrollable are and outlining the 4 cases that require scroll ([FigJam here](https://www.figma.com/board/qpJyzGMM2J1Vewjg5Dxlkm/Table2?node-id=0-1&t=zlIVL37PB1W5MnEB-0)):

  image

packages/table/src/table2.tsx
filipbalucha: `viewportRect` does include space for the headers (it contains headers, frozen cells, and the visible "scrollable cells"). The absolute placement of the scrollable cells within the scrollable element `S` is such that `S.scrollTop` need not take frozen cell or header height into account. Hopefully the diagram in https://github.com/palantir/blueprint/pull/6882#issuecomment-2257268989 helps explain this a bit better.

packages/table/src/tableHotkeys.ts
```diff
@@ -52,8 +54,15 @@ export class TableHotkeys {
...
     }
 
-    public setState(state: TableState) {
-        this.state = state;
+    public setState(newState: TableState) {
+        if (
+            newState.focusedCell != null &&
+            (this.state.focusedCell == null || this.state.focusedCell !== newState.focusedCell)
+        ) {
+            this.scrollBodyToFocusedCell(newState.focusedCell);
```
evansjohnson, line 62: is it really true that we _always_ want to scroll the table to the focused cell on _any_ state update? it seems like maybe this should be happening as a more targeted change

ideally we can add this call to the specific internal places needed rather than just assuming this is fine to do on any state update

packages/table/src/table2.tsx
```diff
@@ -1624,4 +1608,19 @@ export class Table2 extends AbstractComponent<Table2Props, TableState, TableSnap
...
     };
+
+    private getHeaderDimensions = (): TableHeaderDimensions => {
+        return {
+            columnHeaderHeight: this.getColumnHeaderHeight(),
+            rowHeaderWidth: this.getRowHeaderWidth(),
+        };
+    };
+
+    private getColumnHeaderHeight = (): number => {
```
evansjohnson, line 1619: my preference would be to merge all no-op refactors like this in a setup PR so we can really be reviewing a minimal diff here

evansjohnson (commented):
going to give final review on this one to @ggdouglas but leaving some last comments for consideration - sorry for the delayed review on this one

packages/table/src/table2.tsx
filipbalucha: I actually tried putting up a PR for noops but they are so minimal they should hardly be a distractor. For instance, `getColumnHeaderHeight` here just encapsulates the implementation from elsewhere in the class.

packages/table/src/tableHotkeys.ts
filipbalucha: > is it really true that we always want to scroll the table to the focused cell on any state update

Totally appreciate the concern. In short, yes.

The focused cell is a UI metaphor for the centre point of user interaction in the table. By that logic, it should always be visible. That's what Excel does, for instance, with the exception of being scrolled away:

https://github.com/user-attachments/assets/b9a2ceec-4990-4c70-90a0-3fb3df2b459b

For full disclosure, not keeping the focused cell in view is _precisely_ what brought my team to this issue. The lack of alignment in the presence of headers and frozen rows and columns was a side-discovery made while working on this.

ggdouglas (dismissed):
Having tested this locally and reviewed the work/conversations here in this PR, I'm satisfied that this meets the bar. Approving so that we can merge this fix in.

filipbalucha:
Thanks @evansjohnson and @ggdouglas for reviewing!

ggdouglas (approved):

PR #6885, opened 2024-07-01
[core] fix(tabs): remove large gap between vertical tabs

gluxon:
## Changes

Fixes an issue reported here https://github.com/palantir/blueprint/pull/6834#issuecomment-2200436245. This was a recent regression in a fix to better support RTL tabs.

### Before

Screenshot 2024-07-01 at 5 08 06 PM

### After

Screenshot 2024-07-01 at 5 08 29 PM


evansjohnson (approved):
LGTM

PR #6886, opened 2024-07-02
a11y(DocsApp): fix version selector a11y

bvandrc:
The version selector in the docs app caused aXe a11y failure. This element is pretty much a `Select` (with a redirect), so just use the `Select` component which has proper a11y already set up within it.


ggdouglas:
@bvandercar-vt can you give specifics on what the "aXe a11y failure" is and what "proper a11y" is in this case? Is the proper method something that can only be achieved with the `<Select>` component, or can we leverage the existing menu here to the same end?

bvandrc:
@ggdouglas We could fix the a11y aXe failure with a much simpler fix to the current implementation, yes. It would be a 1 line fix.

But, I took that change further to not just _fix_  a11y but _improve_ it, because this _is_ more of a `select` than just a popover with a menu, since you select from a list of items. Currently, role-wise, it's just a popover with a menu, but IMO it _is_ a select component. So, making it have `select` related aria roles better reflects what it the component actually is, to screenreaders.

bvandrc:
Replaced by https://github.com/palantir/blueprint/pull/6967 for simpler changes.

PR #6887, opened 2024-07-02
Revert "[core] fix(Tabs): use gap instead of margin to fix RTL spacing (#6834)"

gluxon:
## Changes

Reverts the following PRs

- https://github.com/palantir/blueprint/pull/6834
- https://github.com/palantir/blueprint/pull/6885

We saw a few problems rolling this out:

- A lot of users are overriding `margin` on `.bp5-tab-list` children. The new `gap` value adds additional spacing that's unexpected in this case.
- @evansjohnson caught a `<TabsExpander />` usage that overflowed items with the switch to `gap`.

### DIFF
--- a/packages/core/src/components/tabs/_tabs.scss
+++ b/packages/core/src/components/tabs/_tabs.scss
@@ -88,9 +88,9 @@ $tab-indicator-width: 3px !default;
 .#{$ns}-tab-list {
   align-items: flex-end;
   border: none;
-  column-gap: $pt-grid-size * 2;
   display: flex;
   flex: 0 0 auto;
+  gap: $pt-grid-size * 2;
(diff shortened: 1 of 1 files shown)


evansjohnson:
I opened https://github.com/palantir/blueprint/pull/6894 to better handler the `TabsExpander` case, but I think we should try to push through this change.

If we do a revert we're stuck not supporting RTL gracefully without a direction forward, and the main concerns around breaking changes due to the style change turned out to be non issues (I was previewing in an environment that had both old and new styles applied).

evansjohnson:
closing - planning to release without a full revert of this change and just the fix for the vertical tab spacing since this wasn't as bad as originally feared

PR #6893, opened 2024-07-03
feat(Tabs): pass tabPanelId and tabTitleId to onChange

bvandrc:
Need access to these props for proper a11y when rendering tab panels outside of the `Tabs` component


bvandrc:
Cancelled in liue of https://github.com/palantir/blueprint/pull/6896/files

PR #6894, opened 2024-07-03
Handle flex-expander case in tabs gap

evansjohnson:
See Tabs docs: https://blueprintjs.com/docs/#core/components/tabs - there's currently 40px between the search input and the last tab, because of a double gap on the flex-expander. This was an existing issue before [the change from margin-right -> gap](https://github.com/palantir/blueprint/pull/6834).

This PR adds a style targeting the flex-expander with a negative margin to remove the gap added due to the expander component.

### DIFF
--- a/packages/core/src/components/tabs/_tabs.scss
+++ b/packages/core/src/components/tabs/_tabs.scss
@@ -95,6 +95,13 @@ $tab-indicator-width: 3px !default;
   margin: 0;
   padding: 0;
   position: relative;
+
+  > .#{$ns}-flex-expander {
+    // offset column-gap from parent component above
+    // since flex-expander is taking up _available_ space, we don't want to add a double gap
+    // between the elements before and after the expander
+    margin-right: -$pt-grid-size * 2;
+  }
 }
 
 .#{$ns}-tab {
--- a/packages/core/src/components/tabs/_tabs.scss
+++ b/packages/core/src/components/tabs/_tabs.scss
@@ -96,7 +96,7 @@ $tab-indicator-width: 3px !default;
   padding: 0;
   position: relative;
 
-  .#{$ns}-tabs:not(.#{$ns}-vertical) & > .#{$ns}-flex-expander {
+  .#{$ns}-tabs:not(.#{$ns}-vertical) > .#{$ns}-flex-expander {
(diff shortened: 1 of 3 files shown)


packages/core/src/components/tabs/tabs.tsx
```diff
@@ -353,6 +353,12 @@ export class Tabs extends AbstractPureComponent<TabsProps, TabsState> {
             return (
                 <TabTitle
                     {...child.props}
+                    // add margin-right as inline style to reduce chance of double spacing between tabs
+                    // after switching from margin-right to gap in #6834
+                    // some consumers may have targeted the default margin-right to override it and this
+                    // should wipe out their override and at least reset them to the default gap style
+                    // instead of applying both the gap style and whatever margin-right they have defined
+                    style={{ ...child.props.style, marginRight: 0 }}
```
evansjohnson: tried making this work for the non tab case as well but looks like that will have issues since the `style` prop is not always applied at the top level, going to back this out

Screenshot 2024-07-03 at 3 41 30 PM
Screenshot 2024-07-03 at 3 41 30 PM

ggdouglas (dismissed):
Considering that we're only shrinking the horizontal space here, I feel like this is generally safe.

packages/core/src/components/tabs/_tabs.scss
```diff
@@ -95,6 +95,13 @@ $tab-indicator-width: 3px !default;
   margin: 0;
   padding: 0;
   position: relative;
+
+  .#{$ns}-tabs:not(.#{$ns}-vertical) & > .#{$ns}-flex-expander {
```
ggdouglas: Is the parent selector needed here?
```suggestion
  .#{$ns}-tabs:not(.#{$ns}-vertical) > .#{$ns}-flex-expander {
```

evansjohnson: yea good catch I dunno what I was doing here - maybe originally had it nested with something else and removed the nesting without removing this

these should be equivalent right?

ggdouglas (approved):

evansjohnson:
going to merge this, I agree it seems like it should be safe - we're only possibly removing 20px space, and only when space has already been added by the flex expander component

PR #6896, opened 2024-07-05
feat: export TabPanel

bvandrc:
#### Checklist

- [X] Includes tests
- [X] Update documentation

Extract the tab panel wrapper from `renderTabPanel` to its own component that is exported. This is useful when rendering a tab panel when using `Tabs` in controlled mode.


packages/core/test/tabs/tabsTests.tsx
```diff
@@ -113,7 +113,10 @@ describe("<Tabs>", () => {
                 <Tab id="third" title="Third" className={tabClassName} panel={<Panel title="third" />} />,
             </Tabs>,
         );
-        assert.lengthOf(wrapper.find(`.${tabClassName}`), 9);
```
bvandercar-vt, line 116: Test is not changed to accommodate component changes; result is same as before-- see https://github.com/palantir/blueprint/pull/6898

bvandrc: Tests are not changed to accommodate component changes; result is same as before-- see https://github.com/palantir/blueprint/pull/6898

packages/core/src/components/tabs/tabTitle.tsx
```diff
@@ -91,10 +93,9 @@ export class TabTitle extends AbstractPureComponent<TabTitleProps> {
...
-    return `${Classes.TAB_PANEL}_${parentId}_${tabId}`;
-}
-
-export function generateTabTitleId(parentId: TabId, tabId: TabId) {
-    return `${Classes.TAB}-title_${parentId}_${tabId}`;
+export function generateTabIds(parentId: TabId, tabId: TabId) {
+    return {
+        tabPanelId: `${Classes.TAB_PANEL}_${parentId}_${tabId}`,
+        tabTitleId: `${Classes.TAB}-title_${parentId}_${tabId}`,
+    } satisfies TabIdProps;
```
evansjohnson, line 100: nice!

packages/core/changelog/@unreleased/pr-6865.v2.yml
```diff
@@ -0,0 +1,5 @@
+type: improvement
+fix:
+  description: '[core] a11y(Tooltip): wrap contents in "tooltip" aria role'
```
evansjohnson: ✂️

packages/core/src/components/tabs/tabs.tsx
```diff
@@ -325,25 +326,19 @@ export class Tabs extends AbstractPureComponent<TabsProps, TabsState> {
     }
 
     private renderTabPanel = (tab: TabElement) => {
-        const { className, panel, id, panelClassName } = tab.props;
+        const { className, id, panel, panelClassName } = tab.props;
```
evansjohnson: nit: no diff here please

packages/core/src/components/tabs/tabPanel.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+import { AbstractPureComponent, Classes, Utils } from "../../common";
+
+import { type TabProps } from "./tab";
+import { generateTabIds, type TabTitleProps } from "./tabTitle";
+
+export interface TabPanelProps extends Pick<TabProps, "className" | "id" | "panel">, Pick<TabTitleProps, "parentId"> {
+    /**
+     * Used for setting `aria-hidden` prop.
+     */
```
evansjohnson:
```suggestion
    /**
     * If true, will set required accessibility roles and set `display` to `none`.
     */
```

packages/docs-app/src/examples/core-examples/tabsExample.tsx
```diff
@@ -27,6 +27,7 @@ import {
     Switch,
     Tab,
     type TabId,
+    TabPanel,
```
evansjohnson, line 30: could we have this component show up on the docs page similar to how `Menu item` shows up under `Menu`?

packages/core/test/tabs/tabsTests.tsx
evansjohnson: thank you for splitting out the other tabs test changes

packages/core/src/components/tabs/tabPanel.tsx
evansjohnson: I'm thinking maybe we just make this `required` if the expected usage is that one `TabPanel` is provided for each tab, like is done in the uncontrolled usage of `Tabs`. In your example you essentially share a `TabPanel` but I'm wondering if making this optional encourages that pattern.

packages/core/src/components/tabs/tabPanel.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+import classNames from "classnames";
+import * as React from "react";
+
+import { AbstractPureComponent, Classes, Utils } from "../../common";
+
+import { type TabProps } from "./tab";
+import { generateTabIds, type TabTitleProps } from "./tabTitle";
+
+export interface TabPanelProps extends Pick<TabProps, "className" | "id" | "panel">, Pick<TabTitleProps, "parentId"> {
```
evansjohnson: re `panel` - do we want to inherit the `If omitted, no panel will be rendered for this tab. Can either be an element or a renderer.` behavior?

if so we probably want something similar to https://github.com/palantir/blueprint/pull/6896/files#diff-cf85fdaa862aad83538297ecd6e3a3b6a5a3fee30bd8aaaea4a2294f462c7933R330-R332 where we early return if panel is undefined

evansjohnson (commented):
Some comments here but I'm generally in support of this. There's not too much this component does, but it hooks up some accessibility roles that are otherwise liable to not be set.

packages/core/src/components/tabs/tabPanel.tsx
evansjohnson: I wonder if we should also support the `renderActiveTabPanelOnly` prop here and do an early return when `isHidden` as a convenience without needing to set that up yourself and to keep the same feature set offered by uncontrolled tabs

packages/core/src/components/tabs/tabTitle.tsx
bvandrc: > There's not too much this component does, but it hooks up some accessibility roles that are otherwise liable to not be set.

Bingo, exactly!

packages/core/src/components/tabs/tabPanel.tsx
bvandrc: sure

packages/core/src/components/tabs/tabPanel.tsx
bvandrc: If someone is using `TabPanel` and wants to render active tab panel only, they would be handling the visibility logic themselves outside of the `TabPanel` component.

bvandrc: But it wouldn't hurt anything to add your reccomendation as a bonus 🤷

bvandrc: Yep. I had to remove `TabPanel` from the isomorphic tests then. https://github.com/palantir/blueprint/pull/6896/commits/8f202ae44d84c21a88128ee5781683f51dfab0c1

bvandrc: done https://github.com/palantir/blueprint/pull/6896/commits/fc9d945ef2bf844c84c1fdd73ee4357d8dae76d6

packages/core/src/components/tabs/tabs.tsx
```diff
@@ -330,20 +331,14 @@ export class Tabs extends AbstractPureComponent<TabsProps, TabsState> {
...
-                id={tabPanelId}
+            <TabPanel
+                {...tab.props}
                 key={id}
-                role="tabpanel"
-            >
-                {Utils.isFunction(panel) ? panel({ tabTitleId, tabPanelId }) : panel}
-            </div>
+                className={classNames(className, panelClassName)}
+                isHidden={id !== this.state.selectedTabId}
```
evansjohnson: since any consumer would need to set this same logic up I wonder if we should have the prop be `selectedTabId` instead so that users can't do weird things like have multiple tab panels showing

evansjohnson (commented):
we're you able to figure out getting `TabPanel` to show up in the docs? just one more small suggestion otherwise curious to hear your thoughts on it then this one looks ready to go

packages/docs-app/src/examples/core-examples/tabsExample.tsx
bvandrc: Done

bvandrc:
> we're you able to figure out getting TabPanel to show up in the docs?

Just got that done!

packages/core/src/components/tabs/tabs.tsx
bvandrc: Eh the logic may not always be the same. Take, for example, this case where `isHidden` is always `false` because there's always only 1 `TabPanel` being rendered:

```tsx
import { Tab, Tabs, TabPanel, type TabId } from "@blueprintjs/core";
...

const TABS_PARENT_ID = React.useId();
const [selectedTabId, setSelectedTabId] = React.useState<TabId>("Home");

<Tabs
    id={TABS_PARENT_ID}
    onChange={setSelectedTabId}
    selectedTabId={selectedTabId}
>
    <Tab id="Home" title="Home" />
    <Tab id="Files" title="Files" />
</Tabs>
...
<TabPanel
    id={selectedTabId}
    isHidden={false}
    parentId={TABS_PARENT_ID}
    panel={<p>The current panel id is: "{selectedTabId}"</p>}
/>
```

But then again, if we changed it to `selectedTabId` they could just have it be

```tsx
<TabPanel
    id={selectedTabId}
    selectedTabId={selectedTabId}
    parentId={TABS_PARENT_ID}
    panel={<p>The current panel id is: "{selectedTabId}"</p>}
/>
```

I'm down for either, let me know which you like better.

evansjohnson: yea I think I'm leaning towards `selectedTabId` because then we control more of the implementation - I'm worried about trying to use the same `TabPanel` for multiple tabs causing issues down the line somehow

bvandrc: Done https://github.com/palantir/blueprint/pull/6896/commits/12a665b216bc570ac07824e75a4eb1578490e038

packages/docs-app/src/examples/core-examples/tabsExample.tsx
```diff
@@ -160,6 +133,49 @@ export class TabsExample extends React.PureComponent<ExampleProps, TabsExampleSt
...
+                                    icon={this.state.showIcon ? "build" : undefined}
+                                    tagContent={this.state.showTags ? 4 : undefined}
+                                    tagProps={{ round: this.state.useRoundTags }}
+                                />
+                            </Tabs>
+                        </Navbar.Group>
+                    </Navbar>
+                    <TabPanel
+                        id={this.state.navbarTabId}
+                        selectedTabId={this.state.navbarTabId}
```
evansjohnson, line 169: lol I guess I didn't see this coming where `id` could just also be dynamic 🤷

that's fine I can't think of an easy way to prevent this and maybe it is okay. if we ever actually see an accessibility issue with changing the id dynamically we can update the docs at that time but no need to go searching for if this would be an issue

packages/core/src/components/tabs/tabPanel.tsx
```diff
@@ -0,0 +1,63 @@
...
+import { type TabId, type TabProps } from "./tab";
+import type { TabsProps } from "./tabs";
+import { generateTabIds, type TabTitleProps } from "./tabTitle";
+
+export interface TabPanelProps
+    extends Pick<TabProps, "className" | "id" | "panel">,
+        Pick<TabsProps, "renderActiveTabPanelOnly" | "selectedTabId">,
+        Pick<TabTitleProps, "parentId"> {
+    /**
+     * Used for setting `aria-hidden` prop, if `!== id`
```
evansjohnson:
```suggestion
     * Used for setting visibility. This `TabPanel` will be visibile when `selectedTabId === id`, with proper accessibility attributes set.
```

evansjohnson (approved):
LGTM - thanks for the contribution and getting the docs updated! 🎉

evansjohnson:
overriding policy-bot approval check because it's invalidated just from the one docs update I just pushed

going to merge this now as I'm about to cut a BP release

gluxon (approved):

PR #6898, opened 2024-07-08
test: improve tabs tests

bvandrc:
#### Checklist

- [X] Includes tests
- [N/A] Update documentation

### DIFF
--- a/packages/core/changelog/@unreleased/pr-6865.v2.yml
+++ b/packages/core/changelog/@unreleased/pr-6865.v2.yml
@@ -1,5 +1,0 @@
-type: improvement
-fix:
-  description: '[core] a11y(Tooltip): wrap contents in "tooltip" aria role'
-  links:
-  - https://github.com/palantir/blueprint/pull/6865
--- a/packages/core/changelog/@unreleased/pr-6865.v2.yml
+++ b/packages/core/changelog/@unreleased/pr-6865.v2.yml
@@ -1,0 +1,5 @@
+type: improvement
+fix:
+  description: '[core] a11y(Tooltip): wrap contents in "tooltip" aria role'
+  links:
+  - https://github.com/palantir/blueprint/pull/6865
--- a/packages/core/test/tabs/tabsTests.tsx
+++ b/packages/core/test/tabs/tabsTests.tsx
@@ -76,9 +76,6 @@ describe("<Tabs>", () => {
         assert.lengthOf(wrapper.find(TAB_SELECTOR), 3);
         assert.lengthOf(wrapper.find(TAB_LIST_SELECTOR), 1);
         assert.lengthOf(wrapper.find(TAB_PANEL_SELECTOR), 3);
-        assert.lengthOf(wrapper.find("[role='tab']"), 3);
-        assert.lengthOf(wrapper.find("[role='tablist']"), 1);
-        assert.lengthOf(wrapper.find("[role='tabpanel']"), 3);
     });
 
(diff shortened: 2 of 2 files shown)


packages/core/test/tabs/tabsTests.tsx
```diff
@@ -28,10 +28,9 @@ describe("<Tabs>", () => {
     // default tabs content is generated from these Dsin each test
     const TAB_IDS = ["first", "second", "third"];
 
-    // selectors using ARIA role
-    const TAB = "[role='tab']";
-    const TAB_LIST = "[role='tablist']";
-    const TAB_PANEL = "[role='tabpanel']";
+    const TAB_SELECTOR = `.${Classes.TAB}`;
```
bvandrc: No other test grabs by role-- use `Classes` like the rest.

packages/core/test/tabs/tabsTests.tsx
```diff
@@ -113,7 +115,10 @@ describe("<Tabs>", () => {
                 <Tab id="third" title="Third" className={tabClassName} panel={<Panel title="third" />} />,
             </Tabs>,
         );
-        assert.lengthOf(wrapper.find(`.${tabClassName}`), 9);
+        const NUM_TABS = 3;
+        assert.lengthOf(wrapper.find(TAB_SELECTOR), NUM_TABS);
+        assert.lengthOf(wrapper.find(TAB_PANEL_SELECTOR), NUM_TABS);
+        assert.lengthOf(wrapper.find(`.${tabClassName}`).hostNodes(), NUM_TABS * 2);
```
bvandrc, line 121: Should not have length 9, should have length 6. Need to use `hostNodes` to remove the React nodes.

packages/core/test/tabs/tabsTests.tsx
evansjohnson: I've always heard it's best to select elements in tests as close as possible as to how your users would. Ideally by actual text content, followed by accessibility roles/element types (ex user searches for and clicks a `<button>`), and as a last resort something like targeting by an `id` or special test-id attribute.

So I don't think we should change these tests just because other tests are selecting by class name.

packages/core/changelog/@unreleased/pr-6865.v2.yml
```diff
@@ -0,0 +1,5 @@
+type: improvement
+fix:
+  description: '[core] a11y(Tooltip): wrap contents in "tooltip" aria role'
```
evansjohnson: ✂️

evansjohnson (commented):
lemme know if you think the `hostNodes` change is important, but I'd prefer to not churn how we're targeting elements here

packages/core/test/tabs/tabsTests.tsx
bvandrc: OK! Changed.

packages/core/changelog/@unreleased/pr-6865.v2.yml
bvandrc: > lemme know if you think the `hostNodes` change is important### 

This PR is a precursor to https://github.com/palantir/blueprint/pull/6896 , where this change is more beneficial

evansjohnson (approved):
Thank you for the PR and splitting this out!

I think this is bordering on diff churn without real benefit, but provides a marginally better example for future tests so I'm okay merging. Though I wouldn't want to continue seeing more PRs like this.

PR #6900, opened 2024-07-08
fix: React class component defaultProps types

### LINKED ISSUE #6899: TabsProps has "id" as required, but is not required in component usage
`TabsProps` `id` prop is required, but somewhere this gets changed to optional: 

image

image

I think this has to do with the usage of `React.PureComponent` in `AbstractPureComponent`, it seems this changes _all_ properties to optional, which could lead to major issues.

At somepoint, all component `class`es should probably be migrated to functions, per React's reccomendation: https://react.dev/reference/react/PureComponent#alternatives

bvandrc:
#### Checklist

- [N/A] Includes tests
- [N/A] Update documentation

Setting `defaultProps` to a type of `Partial<Props>` makes **all of that component's props partial**, which could lead to disastrous effects. TypeScript advises against this starting with TS v3.0:

https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-0.html#explicit-types-on-defaultprops

> If an explicit type annotation is added, e.g. `static defaultProps: Partial<Props>;`  the compiler will not be able to identify which properties have defaults

> Use `static defaultProps: Pick<Props, "name">;` as an explicit type annotation instead, or do not add a type annotation as done in the example above.

As an example, see issue https://github.com/palantir/blueprint/issues/6899


bvandrc:
@ggdouglas you may want to look at this one as it does expose and fix some potential issues

PR #6912, opened 2024-07-15
Fix spinner causing scroll bar to appear

shamith09:
#### Checklist

- [ ] Includes tests
- [ ] Update documentation

#### Screenshot

Before: https://github.com/user-attachments/assets/db9dc664-3fb9-4286-ba0a-f4259c580efe

After: https://github.com/user-attachments/assets/d129ec7e-6b89-4e22-9eaf-fe0eed05ed4e

### DIFF
--- a/packages/core/src/components/spinner/_spinner.scss
+++ b/packages/core/src/components/spinner/_spinner.scss
@@ -16,7 +16,7 @@
   justify-content: center;
 
   // allow paths to overflow container -- critical for edges of circles!
-  overflow: visible;
+  overflow: visible hidden;
   vertical-align: middle;
 
   svg {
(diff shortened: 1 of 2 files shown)


packages/core/src/components/spinner/_spinner.scss
```diff
@@ -16,7 +16,7 @@
   justify-content: center;
 
   // allow paths to overflow container -- critical for edges of circles!
-  overflow: visible;
+  overflow: visible hidden;
```
evansjohnson, line 19: any ideas why https://github.com/palantir/blueprint/commit/21db581d9adac5380ed3c77aa9266e639481f330#diff-746703d417092b84b1146be9dc09d0810c0eb37a08d0a867d2f319c197d2e297R32 doesn't apply?

if this is a blind fix because it seems to work I'd prefer if this is fixed where used - ex you may be setting scroll on your parent wrapper for the content that loads after but can hide overflow yourself while loading.

shamith09: got it, closing this

PR #6913, opened 2024-07-16
[Icons] Update CSSStyle and add TableSync

zaini:
#### Checklist

- [ ] Includes tests
- [ ] Update documentation

Update CSSStyle and add TableSync

New icon designs.

#### Screenshot

image

image

Reach out to xinyifu on Slack for design nits.


resources/icons/20px/data-sync.svg
evansjohnson, line 1: I think this should probably be added as a new icon

right now the `data-sync` icon lines up with the `database` and `data-connection` icons and this would break that:
Screenshot 2024-07-16 at 3 45 41 PM
Screenshot 2024-07-16 at 3 46 51 PM

---

perhaps this could be something like `table-sync` or `datasource-sync`

zaini: Added `table-sync` as it's own icon

resources/icons/20px/table-sync.svg
evansjohnson, line 1: a bit of a nit but could we either call this `data-table-sync` or use the same underlying table icon with only one small column and a wider space to the right?

I'm thinking we may end up with a full set of `data-table` icons and it would be nice to either give them a unique prefix, or just stick with the underlying table icon

Screenshot 2024-07-19 at 10 05 23 AM

evansjohnson: decided against this - "Data table" is a bit redundant and there's no need to have a separate set of "data table" icons from "table" just because of the # of columns. # of columns can be dictated by what works best for the particular icon (ex see "Table: disconnect")

evansjohnson (approved):
LGTM - thanks for the update to add table-sync as a new icon!

PR #6914, opened 2024-07-17
Update README.md

evansjohnson:
I hit this during set up and so did another developer who recently set up Blueprint


gluxon (approved):
Thank you!

PR #6915, opened 2024-07-17
Add Support for customTarget in MultiSelect Component

donghee214:
#### Checklist

- [x] Includes tests
- [x] Update documentation

- Support customTarget in MultiSelect, and use that as the target

- MultiSelect functionality is mostly retained even when using `customTarget`. Recognize we will lose the clear all button if customTarget is enabled. Can possibly look into this as a followup, but will need some design brainstorming to come up with a suitable location for this button within the Popover

#### Screenshot
image


packages/select/src/components/multi-select/_multi-select.scss
```diff
@@ -3,12 +3,25 @@
 
 @import "../../common/variables";
 
+$select-padding: $pt-grid-size * 0.5;
```
donghee214, line 6: Copied over from https://github.com/palantir/blueprint/blob/develop/packages/select/src/components/select/_select.scss

This moves the padding from Menu to the Popover-Content component. This is necessary as Menu is no longer the only rendered element in MultiSelect Popover(InputGroup can be present as well).

packages/select/src/components/multi-select/_multi-select.scss
```diff
@@ -3,12 +3,25 @@
...
+  .#{$ns}-popover-content {
+    // use padding on container rather than margin on input group
+    // because top margin leaves some empty space with no background color.
+    padding: $select-padding;
+  }
+
   .#{$ns}-menu {
+      // offset the padding added by the select componenent so menu can remain full width with padding on either side
+    // this is important so menu dividers can be full width, and so that menu items can have padding to the left and right
+    // for menu items
```
donghee214, line 22: Confirmed this doesn't break anything with existing MenuDividers
Current State (without the negative margins and without popover-content padding):
Screenshot 2024-07-17 at 4 57 16 PM

New State (with the negative margins and with popover-content padding):
Screenshot 2024-07-17 at 4 56 38 PM

packages/select/src/components/multi-select/_multi-select.scss
```diff
@@ -3,12 +3,25 @@
...
+    // because top margin leaves some empty space with no background color.
+    padding: $select-padding;
+  }
+
   .#{$ns}-menu {
+    // offset the padding added by the multiselect component so menu can remain full width with padding on either side
+    // this is important so menu dividers can be full width, and so that menu items can have padding to the left and right
+    // for menu items
+    margin-left: -$select-padding;
+    margin-right: -$select-padding;
```
donghee214, line 35: Confirmed this doesn't break anything with existing MenuDividers

#### Current State (without the negative margins and without popover-content padding):
Screenshot 2024-07-17 at 4 57 16 PM

#### New State (with the negative margins and with popover-content padding):
Screenshot 2024-07-17 at 4 56 38 PM

packages/select/src/components/multi-select/multiSelect.tsx
```diff
@@ -352,6 +395,7 @@ export class MultiSelect<T> extends AbstractPureComponent<MultiSelectProps<T>, M
             // scroll active item into view after popover transition completes and all dimensions are stable.
             this.queryList.scrollActiveItemIntoView();
         }
+        this.input?.focus();
```
donghee214: can put this behind a conditional statement to only run if theres a child, but doesn't looks like this does anything to the current implementation

packages/select/src/components/multi-select/multiSelect.tsx
```diff
@@ -221,6 +239,13 @@ export class MultiSelect<T> extends AbstractPureComponent<MultiSelectProps<T>, M
                 className={classNames(listProps.className, popoverProps.className)}
                 content={
                     <div {...popoverContentProps} onKeyDown={handleKeyDown} onKeyUp={handleKeyUp}>
+                        {/* If children is provided, TagInput is not rendered. This loses the search input.
+                        To account for this, render the search input within the popover similar to Select.
+
+                        Clearing all items is still possible since this component is controlled. It just not cannot
+                        be through the default button in this component, rather done through a custom button that can be
+                        rendered from within the popover through the itemListRenderer or from externally. */}
```
donghee214, line 247: For example: 

image

packages/docs-app/src/styles/_examples.scss
```diff
@@ -711,6 +711,18 @@ $docs-hotkey-piano-height: 510px;
   }
 }
 
+#{example("MultiSelect")} .docs-example {
+  .docs-custom-target{
+    display: flex;
+    flex-direction: row;
+    align-items: center;
```
evansjohnson: could we add a min-width here so it doesn't shift around when the tag number changes?

maybe with a `justify-content: space-between` as well so the extra space is in the center

packages/select/src/__examples__/multiSelectCustomTarget.tsx
```diff
@@ -0,0 +1,24 @@
...
+
+import * as React from "react";
+
+import { Tag, Text } from "@blueprintjs/core";
+
+interface IMultiSelectCustomTargetProps {
+    count: number;
+}
+
+export const MultiSelectCustomTarget: React.FC<IMultiSelectCustomTargetProps> = ({ count }) => {
```
evansjohnson: idk why `FilmSelect` was defined in this package but I think this should be defined in the `docs-app` package, maybe just right in the `multiSelectExample.tsx` file

packages/docs-app/src/examples/select-examples/multiSelectExample.tsx
```diff
@@ -51,12 +52,14 @@ export interface MultiSelectExampleState {
     resetOnSelect: boolean;
     showClearButton: boolean;
     tagMinimal: boolean;
+    customTarget: boolean;
 }
 
 export class MultiSelectExample extends React.PureComponent<ExampleProps, MultiSelectExampleState> {
```
evansjohnson, line 59: `Multiselect` docs in general need an update pass, though fine to hold off until review is done

packages/select/src/components/multi-select/_multi-select.scss
evansjohnson: we may need to test with enabling the scrollbar to always be visible in your device settings - we had an issue recently related to this which I fixed here: https://github.com/palantir/blueprint/pull/6804

though it looks like you're doing the same thing as Select so I expect this to be good

packages/select/src/components/multi-select/multiSelect.tsx
```diff
@@ -200,14 +207,25 @@ export class MultiSelect<T> extends AbstractPureComponent<MultiSelectProps<T>, M
...
-        const { disabled, popoverContentProps = {}, popoverProps = {} } = this.props;
+        const { disabled, placeholder, popoverContentProps = {}, popoverProps = {} } = this.props;
         const { handleKeyDown, handleKeyUp } = listProps;
 
         const popoverRef =
             this.props.popoverRef === undefined
                 ? this.refHandlers.popover
                 : mergeRefs(this.refHandlers.popover, this.props.popoverRef);
 
+        const input = (
```
evansjohnson: there's already an `input` ref, could we call this something more specific like `popoverInputForCustomTarget`, and also assign it a unique ref so we consider these cases separately?

packages/select/src/components/multi-select/multiSelect.tsx
evansjohnson: if addressing above comment on `const input = (` this should be handled

packages/select/src/components/multi-select/multiSelect.tsx
```diff
@@ -334,14 +371,20 @@ export class MultiSelect<T> extends AbstractPureComponent<MultiSelectProps<T>, M
...
-            if (this.input != null && !isInputFocused) {
-                // input is no longer focused, we should close the popover
-                this.setState({ isOpen: false });
-            } else if (!this.props.openOnKeyDown) {
-                // we should open immediately on click focus events
-                this.setState({ isOpen: true });
+            // Cannot rely on input to determine popover state as the input will be inside the Popover
+            // if customTarget is provided
+            if (this.props.customTarget != null) {
+                this.setState({ isOpen: nextOpenState });
```
evansjohnson: is this actually needed? it looks like we don't always set the state here currently. maybe we just early return in the `customTarget` case

packages/select/src/components/multi-select/multiSelect.tsx
```diff
@@ -221,6 +239,13 @@ export class MultiSelect<T> extends AbstractPureComponent<MultiSelectProps<T>, M
                 className={classNames(listProps.className, popoverProps.className)}
                 content={
                     <div {...popoverContentProps} onKeyDown={handleKeyDown} onKeyUp={handleKeyUp}>
+                        {/* If customTarget is provided, TagInput is not rendered. This loses the search input.
+                        To account for this, render the search input within the popover similar to Select.
+
+                        Clearing all items is still possible since this component is controlled. It just not cannot
+                        be through the default button in this component, rather done through a custom button that can be
+                        rendered from within the popover through the itemListRenderer or from externally. */}
+                        {this.props.customTarget != null && input}
```
evansjohnson: highlighting this as a TODO - let's catch up on this once everything else is looking good

packages/select/src/components/multi-select/multiSelect.tsx
```diff
@@ -406,6 +450,26 @@ export class MultiSelect<T> extends AbstractPureComponent<MultiSelectProps<T>, M
...
+
+    private getKeyUpHandler = (handleQueryListKeyUp: React.KeyboardEventHandler<HTMLElement>) => {
+        return (e: React.KeyboardEvent<HTMLElement>) => {
+            if (this.state.isOpen) {
+                handleQueryListKeyUp?.(e);
+            }
+
+            this.props.popoverTargetProps?.onKeyDown?.(e);
+        };
+    };
```
evansjohnson: wondering if these should still be wrapping the popover target or the query list? it seems like maybe it should get split up

from the docs app example it seems to be working though so maybe events are bubbling up from the input through the popover? I don't think we'd want to rely on that though

evansjohnson (commented):
looking good! leaving some initial comments for now

packages/select/src/components/multi-select/multiSelect.tsx
donghee214: Yeah this is needed, we control the Popover state via the `isOpen` state.  This is the only place where thats updated

Currently we do update the state from click events here, just not sure why we don't directly use the `nextOpenState` rather than doing these checks for whether whether that TagInput is focused or not

packages/docs-app/src/examples/select-examples/multiSelectExample.tsx
```diff
@@ -183,6 +202,19 @@ export class MultiSelectExample extends React.PureComponent<ExampleProps, MultiS
                         onChange={this.handleShowClearButtonChange}
                     />
                 </PropCodeTooltip>
+                <PropCodeTooltip
+                    content={
+                        <>
+                            <Code>children</Code> is {this.state.customTarget ? "defined" : "undefined"}
```
evansjohnson:
```suggestion
                            <Code>customTarget</Code> is {this.state.customTarget ? "defined" : "undefined"}
```

packages/select/src/components/multi-select/multiSelect.tsx
```diff
@@ -33,12 +34,18 @@ import {
...
 } from "@blueprintjs/core";
-import { Cross } from "@blueprintjs/icons";
+import { Cross, Search } from "@blueprintjs/icons";
 
 import { Classes, type ListItemsProps, type SelectPopoverProps } from "../../common";
 import { QueryList, type QueryListRendererProps } from "../query-list/queryList";
 
 export interface MultiSelectProps<T> extends ListItemsProps<T>, SelectPopoverProps {
+    /**
+     * Element which triggers the multiselect popover. Providing this prop will replace the default TagInput
```
evansjohnson, line 44: could you read through the full docs page? pretty much anything referencing `TagInput` may need to be updated now

packages/docs-app/src/styles/_examples.scss
```diff
@@ -711,6 +711,20 @@ $docs-hotkey-piano-height: 510px;
   }
 }
 
+#{example("MultiSelect")} .docs-example {
+  .docs-custom-target {
+    align-items: center;
+    display: flex;
+    flex-direction: row;
+    justify-content: space-between;
+    min-width: 145px;
```
evansjohnson, line 725: could we override the `rightPadding` here so the radius of the number tag lines up with the edge of the target? (guessing this docs example may end up being copied)

Screenshot 2024-07-18 at 10 09 57 AM

packages/select/src/components/multi-select/multiSelect.tsx
```diff
@@ -402,10 +448,14 @@ export class MultiSelect<T> extends AbstractPureComponent<MultiSelectProps<T>, M
                 handleQueryListKeyUp?.(e);
             }
 
-            this.props.popoverTargetProps?.onKeyDown?.(e);
+            this.props.popoverTargetProps?.onKeyUp?.(e);
```
evansjohnson: yikes - looks like this was always incorrect? let's be very careful with changes to this component because it seems it's current focus interactions may not be extensively tested - could we remove this if not required for this PR and track with an issue?

packages/select/src/components/multi-select/_multi-select.scss
evansjohnson: leaving note for followup, this appears in Select as well - would rather just keep this consistent with Select for now and not try to change both in this PR

---

note 5px above and below scrollbar

this PR
Screenshot 2024-07-18 at 10 15 54 AM

current:
Screenshot 2024-07-18 at 10 15 50 AM

packages/select/src/components/multi-select/multiSelect.tsx
evansjohnson: I see - it seems this `requestAnimationFrame` is for the input focus, which we don't need

maybe in this `requestAnimationFrame` we should be returning early if `customTarget` is defined and instead always directly setting `isOpen` outside of the `requestAnimationFrame` - it looks like that is what the `Select` component does

---

I'm not sure why this is checking input focus and not using the `nextOpenState` but I think it's safest to just let that code be

packages/select/src/components/multi-select/multiSelect.tsx
```diff
@@ -195,19 +206,32 @@ export class MultiSelect<T> extends AbstractPureComponent<MultiSelectProps<T>, M
                 onQueryChange={this.handleQueryChange}
                 ref={this.refHandlers.queryList}
                 renderer={this.renderQueryList}
+                onKeyUp={customTarget != null ? this.keyUpHandler : undefined}
+                onKeyDown={customTarget != null ? this.keyDownHandler : undefined}
```
evansjohnson: I don't think this is the right place for these - currently they're on the element that wraps the target, even with a custom target I think we'd want to keep them there

I'm seeing keyboard nav isn't working on the docs now but I think it was before

packages/select/src/components/multi-select/multiSelect.tsx
```diff
@@ -294,33 +325,41 @@ export class MultiSelect<T> extends AbstractPureComponent<MultiSelectProps<T>, M
...
+                ),
             );
         };
 
     private handleItemSelect = (item: T, evt?: React.SyntheticEvent<HTMLElement>) => {
         if (this.input != null) {
             this.input.focus();
         }
+        if (this.popoverInputForCustomTarget != null) {
+            this.popoverInputForCustomTarget.focus();
```
evansjohnson: is this really needed? I wonder if this similar code above on the `input` is what causes the behavior we often hear complaints about where the list scrolls back to the top after selecting an item

packages/select/test/multiSelectTests.tsx
```diff
@@ -109,6 +109,49 @@ describe("<MultiSelect>", () => {
...
+            popoverProps: { usePortal: false },
+        });
+
+        findTargetButton(wrapper).simulate("click");
+
+        // MultiSelect uses requestAnimationFrame for the PopoverInteraction causing a slight delay
+        await delay(20);
+        wrapper.update();
+
+        let input = wrapper.find("input");
```
evansjohnson, line 148: could we add an assertion that the input is focused here?

evansjohnson (commented):
main focus here on key event handlers/keyboard nav support and updating docs - a few of these comments are just notes not to action now so keep an eye out for that

packages/docs-app/src/examples/select-examples/multiSelectExample.tsx
evansjohnson: +PR title/description

packages/select/src/components/multi-select/multiSelect.tsx
donghee214: I think this is a useful feature fwiw, and keeps the same behavior as TagInput. Feels kind of snappy just being able to search and use arrows + enter to select items, without having to use your mouse to refocus on the input after every selection

the complaint your describing is unrelated to this, that's likely due to after every selection the active item returning to index 0 in the list and being scrolled into view due to `scrollToActiveItem` prop being default true.

packages/select/src/components/multi-select/multiSelect.tsx
donghee214: yeah sounds good- i was concerned this might break existing workflows in the super weird case someone relied on this behavior

donghee214: https://github.com/palantir/blueprint/issues/6917

packages/select/test/multiSelectTests.tsx
```diff
@@ -119,37 +123,44 @@ describe("<MultiSelect>", () => {
...
+
+        const wrapper = mount(<MultiSelect<Film> {...defaultProps} {...handlers} {...props} />, {
+            attachTo: containerElement,
         });
 
         findTargetButton(wrapper).simulate("click");
 
-        // MultiSelect uses requestAnimationFrame for the PopoverInteraction causing a slight delay
-        await delay(20);
+        // There's a slight delay between the Popover rendering and input getting focus
```
donghee214, line 144: This can be visible in actual use, and looks like Select suffers from this delay as well. My guess its due to ref callbacks taking a second after mounting so it can't call focus right away

packages/select/src/components/multi-select/multiSelect.tsx
evansjohnson: yea, I imagine we can change it I just don't want to think through it now alongside this bigger feature change. thank you!

packages/select/src/components/multi-select/multi-select.md
```diff
@@ -1,7 +1,7 @@
 @# MultiSelect
 
 __MultiSelect__ renders a UI to choose multiple items from a list. It renders a
-[__TagInput__](#core/components/tag-input) wrapped in a [__Popover__](#core/components/popover).
+[__TagInput__](#core/components/tag-input) or custom component wrapped in a [__Popover__](#core/components/popover).
```
evansjohnson:
```suggestion
[__TagInput__](#core/components/tag-input), or `customTarget` if defined, wrapped in a [__Popover__](#core/components/popover).
```

packages/select/src/components/multi-select/multiSelect.tsx
```diff
@@ -83,12 +92,14 @@ export interface MultiSelectProps<T> extends ListItemsProps<T>, SelectPopoverPro
...
      *
+     * This prop is ignored if `customTarget` is supplied.
+     *
      * @default false
      */
     openOnKeyDown?: boolean;
 
     /**
-     * Input placeholder text. Shorthand for `tagInputProps.placeholder`.
+     * Placeholder text. Used as placeholder for TagInput or Input depending on whether `customTarget` is supplied.
```
evansjohnson: I don't think we need to commit to this much implementation detail in the docs
```suggestion
     * Placeholder text used for the filter input element.
```

packages/select/src/components/multi-select/multiSelect.tsx
```diff
@@ -332,26 +389,34 @@ export class MultiSelect<T> extends AbstractPureComponent<MultiSelectProps<T>, M
...
-                // we should open immediately on click focus events
-                this.setState({ isOpen: true });
-            }
+    private handlePopoverInteraction = (nextOpenState: boolean, evt?: React.SyntheticEvent<HTMLElement>) => {
+        // Cannot rely on input to determine popover state as the input will be inside the Popover
+        // if customTarget is provided
+        if (this.props.customTarget != null) {
+            this.setState({ isOpen: nextOpenState });
+        } else {
+            this.requestAnimationFrame(() => {
```
evansjohnson: I'd recommend handling the special case early and returning and leaving the existing code as is. right now the `onInteraction` call is moved outside of the `requestAnimationFrame` and I don't want to worry about what changes that could cause

```
if (this.props.customTarget != null) {
    this.setState({ isOpen: nextOpenState });
    this.props.popoverProps?.onInteraction?.(nextOpenState, evt);
    return;
}   
```

packages/select/src/components/multi-select/multiSelect.tsx
evansjohnson: > that's likely due to after every selection the active item returning to index 0 in the list and being scrolled into view due to scrollToActiveItem prop being default true

yea that sounds right. focus state and list scrolling don't need to be tied together because they're not in the same scroll container. we can table this for now, focusing the input does make sense as you describe!

packages/select/test/multiSelectTests.tsx
```diff
@@ -109,6 +109,60 @@ describe("<MultiSelect>", () => {
         assert.isTrue(handleRemove.calledOnceWithExactly(TOP_100_FILMS[3], 1));
     });
 
+    // NOTE: If customTarget is not supplied, MultiSelect will use requestAnimationFrame
+    // for the PopoverInteraction causing a delay that breaks tests. To get around this,
+    // after the click event is simulated you can await for a brief moment then call wrapper.update() to
+    // get latest state, otherwise the wrapper will have stale state and props.
```
evansjohnson: this comment seems out of date now, and I'm not sure we need to comment on the "customTarget is not supplied" case here since we're not testing that

packages/select/test/multiSelectTests.tsx
evansjohnson: it feels immediate to me with `customTarget` since we aren't waiting for the animation frame (I assume to wait for the focus to update?)

is this still needed? if it is a timeout lower than 500 would be nice (like 40/50? allows some buffer over 60fps but still ensures this update is happening quickly which we do expect)

packages/select/src/components/multi-select/multiSelect.tsx
```diff
@@ -200,14 +225,26 @@ export class MultiSelect<T> extends AbstractPureComponent<MultiSelectProps<T>, M
...
+        const { disabled, placeholder, popoverContentProps = {}, popoverProps = {}, inputProps = {} } = this.props;
         const { handleKeyDown, handleKeyUp } = listProps;
 
         const popoverRef =
             this.props.popoverRef === undefined
                 ? this.refHandlers.popover
                 : mergeRefs(this.refHandlers.popover, this.props.popoverRef);
 
+        const popoverInputForCustomTarget = (
+            <InputGroup
```
evansjohnson: should this just be the tag input? otherwise you lose context on what's selected

Screenshot 2024-07-18 at 2 50 37 PM

packages/select/src/components/multi-select/multiSelect.tsx
```diff
@@ -406,6 +471,31 @@ export class MultiSelect<T> extends AbstractPureComponent<MultiSelectProps<T>, M
         };
     };
 
+    private getKeyDownHandler = (handleQueryListKeyDown: React.KeyboardEventHandler<HTMLElement>) => {
+        return (e: React.KeyboardEvent<HTMLElement>) => {
+            if (e.key === "Escape" || e.key === "Tab") {
+                if (this.popoverInputForCustomTarget != null) {
+                    this.popoverInputForCustomTarget.blur();
+                }
+                this.setState({ isOpen: false });
```
evansjohnson: why exclude the bit that opens the popover on other keys? I think that's important for keyboard nav

packages/select/src/components/multi-select/multiSelect.tsx
```diff
@@ -406,6 +471,31 @@ export class MultiSelect<T> extends AbstractPureComponent<MultiSelectProps<T>, M
         };
     };
 
+    private getKeyDownHandler = (handleQueryListKeyDown: React.KeyboardEventHandler<HTMLElement>) => {
+        return (e: React.KeyboardEvent<HTMLElement>) => {
+            if (e.key === "Escape" || e.key === "Tab") {
+                if (this.popoverInputForCustomTarget != null) {
+                    this.popoverInputForCustomTarget.blur();
```
evansjohnson: is this still relevant for the version where the input is inside the popover? asking because it feels like this may just be copying the existing pattern

packages/select/src/components/multi-select/multiSelect.tsx
```diff
@@ -294,33 +338,46 @@ export class MultiSelect<T> extends AbstractPureComponent<MultiSelectProps<T>, M
...
-                    onKeyDown: this.getTagInputKeyDownHandler(handleKeyDown),
-                    onKeyUp: this.getTagInputKeyUpHandler(handleKeyUp),
+                    onKeyDown:
+                        this.props.customTarget != null
+                            ? this.getKeyDownHandler(handleKeyDown)
+                            : this.getTagInputKeyDownHandler(handleKeyDown),
+                    onKeyUp:
+                        this.props.customTarget != null
+                            ? this.getKeyUpHandler(handleKeyUp)
+                            : this.getTagInputKeyUpHandler(handleKeyUp),
```
evansjohnson: seeing the keydown to open the popover seemingly be missed makes me wonder if these should be something like `onKeyDown: this.getTargetKeyDownHandler(handleKeyDown)`

and handle any splits on `customTarget` within

evansjohnson (commented):
thanks for updating docs

biggest things here are keyboard nav seems to still not be working, and something I maybe should have brought up earlier but I'm wondering if it should be a tag input that's rendered at the top of the list - otherwise it's pretty hard to tell what's selected

packages/select/test/multiSelectTests.tsx
donghee214: Slightly worded it better, but going to keep this comment since we're nesting it within requestAnimationFrame

> I'd recommend handling the special case early and returning and leaving the existing code as is. right now the onInteraction call is moved outside of the requestAnimationFrame and I don't want to worry about what changes that could cause

packages/select/src/components/multi-select/multiSelect.tsx
donghee214: hmmm i kinda disagree- if we used a TagInput it feels like we just remade the original MultiSelect component but nested it within a popover instead.

- Think whoever decides to opt in to `customTarget` should be keeping track of the selected items via `handleItemSelect` and updating the `customTarget` component to reflect the changes (and if they want to show the context of what's already selected they can do so)
- Will catch up about other reason through slack

packages/select/src/components/multi-select/multiSelect.tsx
donghee214: hmmm keeping them separate because there is decently separate logic between them, namely how some keystrokes have built in events like space and downarrow (which move the scrollview down).

TagInput is able to automatically ignore it because the target is an input, so space and downarrow will not do it's default behavior

packages/select/test/multiSelectTests.tsx
donghee214: Wasn't able to fully get down to what causes the delay (spent a decent amount digging around) best guess is that ref hypothesis?

Don't think requestAnimationFrame is the culprit though: Playing around with the Select component and the delay on the input focus, it's the same whether the `focus()` is inside or outside of the requestAnimationFrame callback

https://github.com/user-attachments/assets/b6f7b535-a7b6-48e5-970d-8877c5382c2b

packages/select/src/components/multi-select/multiSelect.tsx
donghee214: chatted offline going with TagInput to prioritize functionality over styling (may follow up with something like `tagSummarizerLimit` for TagInput to limit the growth of size with a "+n more" tag)

packages/docs-app/src/examples/select-examples/multiSelectExample.tsx
```diff
@@ -139,6 +139,8 @@ export class MultiSelectExample extends React.PureComponent<ExampleProps, MultiS
                     onItemSelect={this.handleFilmSelect}
                     onItemsPaste={this.handleFilmsPaste}
                     popoverProps={{ matchTargetWidth, minimal: popoverMinimal }}
+                    // Due to usePortal, can't neatly use className from _examples.scss
+                    popoverContentProps={this.state.customTarget ? { style: { maxWidth: 400 } } : undefined}
```
donghee214: Need to explicitly define a maxWidth for the Popover when using `usePortal` + `TagInput` is inside the Popover.

There is no width defined in any of TagInput's grandparents when it's rendered through usePortal so it will never flex-wrap, and infinitely grow horizontally

packages/select/src/components/multi-select/multiSelect.tsx
```diff
@@ -264,7 +239,7 @@ export class MultiSelect<T> extends AbstractPureComponent<MultiSelectProps<T>, M
                         Clearing all items is still possible since this component is controlled. It just not cannot
                         be through the default button in this component, rather done through a custom button that can be
                         rendered from within the popover through the itemListRenderer or from externally. */}
-                        {this.props.customTarget != null && popoverInputForCustomTarget}
+                        {this.props.customTarget != null && this.getTagInput(listProps, CoreClasses.FILL)}
```
donghee214: When the TagInput is rendered inside the Popover, grow the TagInput to be the width of the Popover by default

packages/select/src/components/multi-select/multiSelect.tsx
```diff
@@ -417,7 +378,17 @@ export class MultiSelect<T> extends AbstractPureComponent<MultiSelectProps<T>, M
             // scroll active item into view after popover transition completes and all dimensions are stable.
             this.queryList.scrollActiveItemIntoView();
         }
-        this.popoverInputForCustomTarget?.focus();
+
+        // Unless user explicitly sets autoFocus to false, for customTargets we should by default
+        // focus on the input on open
```
donghee214: select is doing the same https://github.com/palantir/blueprint/blob/develop/packages/select/src/components/select/select.tsx#L336

packages/docs-app/src/examples/select-examples/multiSelectExample.tsx
```diff
@@ -125,13 +139,27 @@ export class MultiSelectExample extends React.PureComponent<ExampleProps, MultiS
                     onItemSelect={this.handleFilmSelect}
                     onItemsPaste={this.handleFilmsPaste}
                     popoverProps={{ matchTargetWidth, minimal: popoverMinimal }}
+                    // Without matchTargetWidth there is no width defined in any of TagInput's
+                    // grandparents when it's rendered through usePortal, so it will never flex-wrap
+                    // and infinitely grow horizontally. To address this, if there is no width guidance
+                    // from matchTargetWidth, explicitly set a width to so Tags will flex-wrap.
```
donghee214: Reason that Default TagInput (not customTarget) doesn't need this is because the TagInput is not rendered within the Popover so it's within the regular DOM hierarchy, meaning it can look to it's ancestors to eventually grab a defined width somewhere to inherit

evansjohnson: Instead of adding this to the docs let's add it into the underlying component. thinking is that _everyone_ will need to do something like this, so we may as well provide a reasonable default.

packages/select/src/components/multi-select/multiSelect.tsx
```diff
@@ -221,6 +231,9 @@ export class MultiSelect<T> extends AbstractPureComponent<MultiSelectProps<T>, M
                 className={classNames(listProps.className, popoverProps.className)}
                 content={
                     <div {...popoverContentProps} onKeyDown={handleKeyDown} onKeyUp={handleKeyUp}>
+                        {/* If customTarget is provided, move the TagInput to within the Popover
+                        to retain core functionalities */}
```
evansjohnson, line 251: nice! it looks like this worked out well, thanks for being down to try it out.

packages/select/src/components/multi-select/multiSelect.tsx
```diff
@@ -346,12 +365,24 @@ export class MultiSelect<T> extends AbstractPureComponent<MultiSelectProps<T>, M
...
+    };
 
     private handlePopoverOpened = (node: HTMLElement) => {
         if (this.queryList != null) {
             // scroll active item into view after popover transition completes and all dimensions are stable.
             this.queryList.scrollActiveItemIntoView();
         }
+
+        // Unless user explicitly sets autoFocus to false, for customTargets we should by default
+        // focus on the input when the Popover opens.
```
evansjohnson: nit: I find it better to pull out to better names so it reads like what you would comment, ex
```
const hasCustomTarget = this.props.customTarget != null;
if (hasCustomTarget && this.input != null) {
  const shouldAutofocus = this.props.tagInputProps?.inputProps?.autoFocus !== false;
  if (shouldAutofocus) {
    this.input.focus();
  }
}
```
personally I don't think `this.input != null` needs anything since that's just a ref guard that we always expect to be set by the time this is called

code comments risk getting out of sync with the code, while pulling out smaller variables allows each step to be grokked as you go

packages/select/src/components/multi-select/multiSelect.tsx
```diff
@@ -379,6 +410,11 @@ export class MultiSelect<T> extends AbstractPureComponent<MultiSelectProps<T>, M
                 }
                 this.setState({ isOpen: false });
             } else if (!(e.key === "Backspace" || e.key === "ArrowLeft" || e.key === "ArrowRight")) {
+                // Custom target might not be an input, so certain keystrokes might have other effects
+                // (space pushing the scrollview down), prevent default events before opening the popover
+                if (this.props.customTarget != null && (e.key === " " || e.key === "ArrowDown")) {
+                    e.preventDefault();
+                }
```
evansjohnson: could we instead make this:
- only open the popover on " " or "Enter"
    - thinking this is different for custom target because it's an element that opens the popover (on keyboard action), not something a user should be typing in (could re-think if custom targets need to actually be inputs but I wouldn't really expect this)
- not prevent default on `ArrowDown`
    - when focusing on a native button and hitting down the page scrolls, so I think that's okay/expected

packages/select/src/components/multi-select/multiSelect.tsx
evansjohnson: sorry I'm realizing this may not have been clear but I was expecting to see this early return block outside of the `requestAnimationFrame` - have you tried that by any chance? wondering if that's why the tests need the `await`

evansjohnson (commented):
nice this is looking much cleaner with using the tag input in the popover! just a few cleanup comments here but otherwise think we're very close. my only real open question is around why we're setting the state inside the animation frame still if we're not checking the `activeElement`

packages/select/src/components/multi-select/multiSelect.tsx
```diff
@@ -220,7 +230,22 @@ export class MultiSelect<T> extends AbstractPureComponent<MultiSelectProps<T>, M
                 {...popoverProps}
                 className={classNames(listProps.className, popoverProps.className)}
                 content={
-                    <div {...popoverContentProps} onKeyDown={handleKeyDown} onKeyUp={handleKeyUp}>
+                    <div
+                        // In the case where customTarget is supplied and the TagInput is rendered within the Popover,
+                        // without matchTargetWidth there is no width defined in any of TagInput's
+                        // grandparents when it's rendered through usePortal, so it will never flex-wrap
+                        // and infinitely grow horizontally. To address this, if there is no width guidance
+                        // from matchTargetWidth, explicitly set a default width to so Tags will flex-wrap.
```
evansjohnson, line 240: nice - I don't think we want to add this class when `matchTargetWidth` prop is set though

Screenshot 2024-07-19 at 4 27 22 PM

packages/select/src/components/multi-select/multiSelect.tsx
```diff
@@ -332,7 +356,15 @@ export class MultiSelect<T> extends AbstractPureComponent<MultiSelectProps<T>, M
 
     // Popover interaction kind is CLICK, so this only handles click events.
     // Note that we defer to the next animation frame in order to get the latest activeElement
-    private handlePopoverInteraction = (nextOpenState: boolean, evt?: React.SyntheticEvent<HTMLElement>) =>
+    private handlePopoverInteraction = (nextOpenState: boolean, evt?: React.SyntheticEvent<HTMLElement>) => {
+        if (this.props.customTarget != null) {
+            // Cannot rely on input to determine popover state as the input will be inside the Popover
+            // if customTarget is provided
```
evansjohnson:
```suggestion
```
I think this comment only makes sense right now if you assume relying on the input is the "correct" way - I don't think we need to explain handling this separately. If anything the input focus handling could be explained, but I'd rather not just slap a comment on there now since we may miss some nuance about why it was done that way.

packages/select/test/multiSelectTests.tsx
```diff
@@ -109,6 +109,64 @@ describe("<MultiSelect>", () => {
...
+        };
+
+        const wrapper = mount(<MultiSelect<Film> {...defaultProps} {...handlers} {...props} />, {
+            attachTo: containerElement,
+        });
+
+        findTargetButton(wrapper).simulate("click");
+
+        // There's a slight delay between the Popover rendering and input getting focus
+        await delay(500);
```
evansjohnson, line 145: talked offline about this - delay is to wait for the opening animation

evansjohnson (commented):
almost there 😄

packages/select/src/components/multi-select/multiSelect.tsx
donghee214: sounds good!

gluxon:
Small visual nitpick: The box shadow on the `<TagInput>` is cut off by the solid white background of the list below. One way to fix this would be to add some spacing.


https://github.com/user-attachments/assets/13783081-f38d-4d38-9a06-cf61057eb507

packages/docs-app/src/examples/select-examples/multiSelectExample.tsx
```diff
@@ -132,6 +146,11 @@ export class MultiSelectExample extends React.PureComponent<ExampleProps, MultiS
                         tagProps: getTagProps,
                     }}
                     selectedItems={this.state.films}
+                    customTarget={
+                        this.state.customTarget ? (
+                            <MultiSelectCustomTarget count={this.state.films.length} />
+                        ) : undefined
+                    }
```
gluxon: Apologies if this was already discussed, but was there reasons for or against a `renderTarget`-like API that passes in the current list selection to a render function?

The `renderTarget` API prevents mistakes where the state of the list and the data the `customTarget` is reading from can differ.

evansjohnson: ah good catch, this was not discussed

packages/select/src/components/multi-select/multiSelect.tsx
```diff
@@ -303,7 +304,7 @@ export class MultiSelect<T> extends AbstractPureComponent<MultiSelectProps<T>, M
                     ref,
                     role: "combobox",
                 },
-                this.props.customTarget != null ? this.props.customTarget : this.getTagInput(listProps),
+                this.props.renderTarget != null ? this.props.renderTarget(targetProps) : this.getTagInput(listProps),
```
evansjohnson: I don't think we want to pass `targetProps` since we're already wrapping the custom target in an element that is handling spreading those.

I think the idea is to have a `renderTarget` callback that is passed `selectedItems` to ensure we always are rendering an up to date target. We could also use that callback to pass anything else that could be helpful, such as passing `isOpen` if the user would want to style differently based on that without separately keeping track of the loading state.

packages/docs-app/src/examples/select-examples/multiSelectCustomTarget.tsx
evansjohnson, line 1: rename file if keeping `renderTarget`

packages/docs-app/src/examples/select-examples/multiSelectExample.tsx
```diff
@@ -51,12 +63,14 @@ export interface MultiSelectExampleState {
     resetOnSelect: boolean;
     showClearButton: boolean;
     tagMinimal: boolean;
+    customTarget: boolean;
```
evansjohnson, line 56: rename in docs to avoid confusion if keeping `renderTarget`

evansjohnson (commented):
I think it may be better to revert before the last change and keep the `customTarget` naming, while still updating the API style per last comment. The comment was to create a `renderTarget` style API, but I'm not sure we actually want to reuse that name since we won't be threading through the props that are on `Popover`s `renderTarget` which may be unexpected given the name.

evansjohnson (approved):
thanks for the great contribution and working through a few changes in direction and getting some tests added!

PR #6923, opened 2024-07-23
style(popover): clean up code

bvandrc:


bvandrc:
@evansjohnson this is a prelude to https://github.com/palantir/blueprint/pull/6865 that you asked me to make. Can you take a look at this?

packages/core/src/components/popover/popover.tsx
```diff
@@ -300,7 +295,7 @@ export class Popover<
         }
     }
 
-    protected validateProps(props: PopoverProps<T> & { children?: React.ReactNode }) {
+    protected validateProps(props: PopoverProps<T>) {
```
evansjohnson, line 298: this seems fine - `{ children?: React.ReactNode }` is the same as what's on `PopoverSharedProps`

packages/core/test/popover/popoverTests.tsx
```diff
@@ -81,7 +81,7 @@ describe("<Popover>", () => {
             assert.isTrue(warnSpy.calledWith(Errors.POPOVER_WARN_TOO_MANY_CHILDREN));
         });
 
-        it("warns if given children and target prop", () => {
+        it("warns if given children and renderTarget prop", () => {
```
evansjohnson, line 84: :+1:

packages/core/test/tooltip/tooltipTests.tsx
```diff
@@ -159,7 +158,7 @@ describe("<Tooltip>", () => {
     function renderTooltip(props?: Partial<TooltipProps>) {
         return mount<TooltipProps>(
             <Tooltip content={<p>Text</p>} hoverOpenDelay={0} {...props} usePortal={false}>
-                <Button id={TEST_TARGET_ID} text="target" />
```
evansjohnson: this isn't needed - just leave it

packages/core/src/components/popover/popoverSharedProps.ts
```diff
@@ -90,6 +90,11 @@ export interface PopoverSharedProps<TProps extends DefaultPopoverTargetHTMLProps
     /** Interactive element which will trigger the popover. */
     children?: React.ReactNode;
 
+    /**
+     * The content displayed inside the popover.
+     */
+    content?: string | React.JSX.Element;
```
evansjohnson, line 96: this shouldn't happen yet, tooltip does nothing with this prop but this would document it

packages/core/src/components/popover/popover.tsx
```diff
@@ -417,7 +412,7 @@ export class Popover<
                 tabIndex: targetTabIndex,
             });
         } else {
-            const childTarget = Utils.ensureElement(React.Children.toArray(children)[0])!;
```
evansjohnson, line 420: this seems fine, since we handle the undefined case just below

```
if (childTarget === undefined) {
    return null;
}
```

packages/core/src/components/popover/popover.tsx
```diff
@@ -791,7 +786,7 @@ export class Popover<
 
     private getIsContentEmpty() {
         const { content } = this.props;
-        return content == null || (typeof content === "string" && content.trim() === "");
+        return Utils.isReactNodeEmpty(content) || Utils.isEmptyString(content);
```
evansjohnson: I don't think we should change this. If consumers want this behavior it's trivial for them to call `.trim`.

We could have done this from the beginning but on the off chance something relies on displaying an empty popover for empty space this doesn't feel worth even a 0.1% chance of a fire drill to address this.

evansjohnson (commented):
Sorry for the delay, I haven't been as active with Blueprint stuff in recent weeks. Left comments on individual diffs - if you could address and rebase the main PR onto this I can give the main one a full review as well!

packages/core/src/components/popover/popoverSharedProps.ts
bvandrc: tooltip absolutely uses the `content` prop

packages/core/test/tooltip/tooltipTests.tsx
bvandrc: it doesn't do anything, is never used

packages/core/src/components/popover/popover.tsx
bvandrc: >  If consumers want this behavior it's trivial for them to call `.trim`

Existing code is already calling `trim`. This refactor is really just using existing functions to replace the same code that does the same thing.

 _Utils.isEmptyString:_
```tsx
 return typeof val === "string" && val.trim().length === 0;
```

It's literally the same code as previous. And then `isReactNodeEmpty` accounts for the left side of previous.

bvandrc: But if you _are_ interested in changing the functionality, I do agree with your statement that it makes sense for it to be on the consumer to not pass empty content, not us to determine that.

packages/core/src/components/popover/popoverSharedProps.ts
evansjohnson: yes seeing that now - didn't see anything removed from tooltip in this diff which threw me off

why are we moving this to shared props?

packages/core/test/tooltip/tooltipTests.tsx
evansjohnson: it's used here - we have no lint rule against defining constants separately and using once so there's no need to make this change

bvandrc: I'm saying it's never used by any test

packages/core/src/components/popover/popover.tsx
evansjohnson: okay right, seeing

`typeof content === "string" && content.trim() === ""`

is functionally equal to

```
export function isEmptyString(val: any) {
    return typeof val === "string" && val.trim().length === 0;
}
```

packages/core/test/tooltip/tooltipTests.tsx
bvandrc: But sure, reverted

packages/core/src/components/popover/popoverSharedProps.ts
bvandrc: Because it _is_ shared.

bvandrc: `TooltipProps` defines its own `content` prop to make it required, but having it in `PopoverSharedProps` ensures that it's the same type. Which makes sense, because `Tooltip` is really just a fancy wrapper for a `Popover`.

 `content` truly is a shared prop between the two, so it makes sense to have it in `PopoverSharedProps`.

packages/core/src/components/popover/popover.tsx
evansjohnson: I think I was looking into both sides of this and got mixed up when writing a response

previously `[]`, or something like `["", "", undefined, null, []]` would render an empty content popover, now we'd consider this empty

evansjohnson: this seems less likely to be bad but this still feels like something we don't need to be messing with outside of a major version

bvandrc: > previously `[]`, or something like `["", "", undefined, null, []]` would render an empty content popover, now we'd consider this empty

That's probably a good thing, because that is truly empty. Think of this as a bug fix

evansjohnson:
Yes but it’s also fixable in consumer land by changing what’s passed to
content. We shouldn’t change behavior in non major versions just because we
think the new behavior is better, even if we expect it would be an
appreciated change for 99%+ of uses.

packages/core/src/components/popover/popover.tsx
bvandrc: @evansjohnson  reverted

packages/core/src/components/popover/popoverSharedProps.ts
evansjohnson: makes sense to me - thanks for explaining!

evansjohnson (approved):

PR #6925, opened 2024-07-25
Update README.md

evansjohnson:
Updating with yarn setup steps that helped me while setting up Blueprint for a new laptop

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,7 +77,7 @@ Builds are orchestrated via [Nx's task runner](https://nx.dev/getting-started/in
 
 **Prerequisites**: Node.js v20.11+ (see version specified in `.nvmrc`), Yarn v4.x (see version specified in `package.json`)
 - note `yarn` may need to be updated from outside of the Blueprint repo or you may see an error like `This project's package.json defines "packageManager": "yarn@4.0.2". However the current global version of Yarn is 1.22.22`
-- update `yarn` with `yarn set version stable`
+- update `yarn` with `yarn set version <version from package.json>`
 
 ### One-time setup
 
--- a/README.md
+++ b/README.md
@@ -76,8 +76,6 @@ Builds are orchestrated via [Nx's task runner](https://nx.dev/getting-started/in
 [Lerna-Lite](https://github.com/lerna-lite/lerna-lite) is used to prepare releases.
 
 **Prerequisites**: Node.js v20.11+ (see version specified in `.nvmrc`), Yarn v4.x (see version specified in `package.json`)
-- note `yarn` may need to be updated from outside of the Blueprint repo or you may see an error like `This project's package.json defines "packageManager": "yarn@4.0.2". However the current global version of Yarn is 1.22.22`
-- update `yarn` with `yarn set version <version from package.json>`
 
 ### One-time setup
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -75,7 +75,9 @@ then [check out the "help wanted" label](https://github.com/palantir/blueprint/l
 Builds are orchestrated via [Nx's task runner](https://nx.dev/getting-started/intro) and NPM scripts.
 [Lerna-Lite](https://github.com/lerna-lite/lerna-lite) is used to prepare releases.
 
-**Prerequisites**: Node.js v18+ (see version specified in `.nvmrc`), Yarn v4.x (see version specified in `package.json`)
+**Prerequisites**: Node.js v20.11+ (see version specified in `.nvmrc`), Yarn v4.x (see version specified in `package.json`)
+- note `yarn` may need to be updated from outside of the Blueprint repo or you may see an error like `This project's package.json defines "packageManager": "yarn@4.0.2". However the current global version of Yarn is 1.22.22`
+- update `yarn` with `yarn set version stable`
```
evansjohnson:
```suggestion
- update `yarn` with `yarn set version <version from package.json>`
```

evansjohnson: this does need to be the exact same version I think. originally did stable and the yarn lock file was changing slightly

README.md
```diff
@@ -75,7 +75,9 @@ then [check out the "help wanted" label](https://github.com/palantir/blueprint/l
 Builds are orchestrated via [Nx's task runner](https://nx.dev/getting-started/intro) and NPM scripts.
 [Lerna-Lite](https://github.com/lerna-lite/lerna-lite) is used to prepare releases.
 
-**Prerequisites**: Node.js v18+ (see version specified in `.nvmrc`), Yarn v4.x (see version specified in `package.json`)
+**Prerequisites**: Node.js v20.11+ (see version specified in `.nvmrc`), Yarn v4.x (see version specified in `package.json`)
+- note `yarn` may need to be updated from outside of the Blueprint repo or you may see an error like `This project's package.json defines "packageManager": "yarn@4.0.2". However the current global version of Yarn is 1.22.22`
+- update `yarn` with `yarn set version <version from package.json>`
```
gluxon: The `packageManager` field is for `corepack`. Since it's set up in the Blueprint repo, should we encourage developers to run `corepack enable` instead of `yarn set version`?

evansjohnson: hmm we do have that below but I don't think that worked for me. I needed to back out of BP and run this before I was able to go back into BP and run `yarn` to install deps

no urgency to merge this and would err towards not doing anything, I may reclone and see what happens but wonder if now that I have yarn 4 downloaded it would work

gluxon: Thanks for answering — I think for now I lean towards recommending developers have a working `corepack` setup. Corepack is nice because it'll automatically download and use newer version of yarn as the `packageManager` field changes.

If someone does `yarn set version <version>` they might get out of sync as someone else changes the `packageManager` field in this repo.

evansjohnson: sounds good, going to back this part out for now

```suggestion
**Prerequisites**: Node.js v20.11+ (see version specified in `.nvmrc`), Yarn v4.x (see version specified in `package.json`)
```

gluxon (approved):
Thanks!

PR #6934, opened 2024-08-05
Fix node-build-scripts-test failing in validate

### LINKED ISSUE #6926: Some tests failing when running `yarn verify`
#### Environment
- __Operating System__: macOS Sonoma 14.5
- __Browser name and version__: n/a

#### Steps to reproduce
Follow README.md setup steps up until `yarn verify`

#### Actual behavior
The following errors:
Screenshot 2024-07-25 at 1 41 40 PM
Screenshot 2024-07-25 at 1 41 24 PM

#### Expected behavior
the tests should not error locally but seemingly pass running in CI (TODO: confirm these tests actually are running in CI)

#### Possible solution
Wondering if all scroll related tests are related to running this on a macbook where the default is to hide scrollbars unless scrolling. I did try changing this operating system setting but saw the same failures (though perhaps a restart would be needed for the changes to take effect here, or I hit some test cache)

ggdouglas:
#### Checklist

- [x] Includes tests
- [ ] Update documentation

This PR resolves failing tests encountered while running the introductory `yarn validate` command (from the [README](https://github.com/palantir/blueprint/blob/1a5c82bcdfec804b9b65d44e4ac75b3af9aafe1e/README.md?plain=1#L94)).

> **tl;dr** The test uses a variable that should exist in the version of Node we're running but doesn't exist during the build.

| Before | After    |
| ------  | ------  |
| before | after |

The "fix" involved here replaces the use of `import.meta.dirname` in `cssVariables.test.ts` with an alternative method of accessing the current folder path. This method was considered one of the ways to expose `__dirname` in an ES module prior to it being added in [v20.11.0](https://nodejs.org/en/blog/release/v20.11.0) (see [this article](https://www.sonarsource.com/blog/dirname-node-js-es-modules/) for more context). 

For some reason, `import.meta.dirname` is `undefined` when running the test scripts. According to the [docs](https://nodejs.org/docs/latest-v20.x/api/esm.html#importmetadirname) it _should_ be there. The typings assume that it is, but in reality only `import.meta.url` is defined.

It might be worth investigating why the build system doesn't expose these variables at runtime when they're supposed to exist. Perhaps there is something deeper in webpack config that needs addressing.

I did not encounter the scrollbar test flake described in #6926, so those tests are not addressed here.

### DIFF
--- a/packages/node-build-scripts/src/__tests__/cssVariables.test.ts
+++ b/packages/node-build-scripts/src/__tests__/cssVariables.test.ts
@@ -4,11 +4,15 @@
 
 import { describe, expect, test } from "@jest/globals";
 import { readFileSync } from "node:fs";
-import { join, resolve } from "node:path";
+import { join, resolve, dirname } from "node:path";
+import { fileURLToPath } from "node:url";
 
 import { generateLessVariables, generateScssVariables, getParsedVars } from "../cssVariables.mjs";
 
-const FIXTURES_DIR = join(import.meta.dirname, "__fixtures__");
+const __filename = fileURLToPath(import.meta.url);
+const __dirname = dirname(__filename);
+
+const FIXTURES_DIR = join(__dirname, "__fixtures__");
 const INPUT_DIR = resolve(FIXTURES_DIR, "input");
 const EXPECTED_DIR = resolve(FIXTURES_DIR, "expected");
 


packages/node-build-scripts/src/__tests__/cssVariables.test.ts
```diff
@@ -4,11 +4,15 @@
...
 import { readFileSync } from "node:fs";
-import { join, resolve } from "node:path";
+import { join, resolve, dirname } from "node:path";
+import { fileURLToPath } from "node:url";
 
 import { generateLessVariables, generateScssVariables, getParsedVars } from "../cssVariables.mjs";
 
-const FIXTURES_DIR = join(import.meta.dirname, "__fixtures__");
+const __filename = fileURLToPath(import.meta.url);
+const __dirname = dirname(__filename);
```
gluxon, line 13: Does `__dirname` need to be defined? Curious if this works without it?

https://nodejs.org/docs/latest/api/modules.html#__dirname

ggdouglas: Yeah, this was the first thing I tried but it sadly does not seem to work:

Screenshot 2024-08-05 at 17 15 05@2x

gluxon: Oh I see. It looks like that's because we're already opting into Jest's ESM support based on the `node:80071` warning. It makes sense that `__dirname` (a CommonJS feature) isn't available given that. Thanks for trying.

gluxon (approved):
Thank you @ggdouglas!

PR #6936, opened 2024-08-07
fix(Popover a11y): target only gets aria-expanded if is button type (not hover)

bvandrc:
#### Checklist

- [N/A] Includes tests
- [N/A] Update documentation

Never apply `aria-expanded` prop if is hover type. This prop being included causes aXe a11y failure when the item target isn't an interactive type (ie `button`, but if it's only hoverable it definitely shouldn't be `button` type)

Thread: https://stackoverflow.com/questions/59207174/is-aria-expanded-only-for-click-interaction-or-can-it-be-used-for-focus-as

Fixes a11y aXe failure


ggdouglas (dismissed):

bvandrc:
@ggdouglas this issue is creating a bunch of aXe a11y failures in our application, can this PR please be looked at?

ggdouglas (approved):

PR #6940, opened 2024-08-12
feat(a11y): give RadioGroup aria "radiogroup" role

bvandrc:
#### Checklist

- [N/A] Includes tests
- [N/A] Update documentation

`RadioGroup` component should be getting `role="radiogroup"`. Currently, it's just a `div`. Everything else about it works in terms of a radiogroup, ie keypresses, attributes, etc.


packages/core/src/components/forms/radioGroup.tsx
```diff
@@ -17,15 +17,22 @@
...
 import * as Errors from "../../common/errors";
-import { isElementOfType } from "../../common/utils";
+import { isElementOfType, uniqueId } from "../../common/utils";
 import { RadioCard } from "../control-card/radioCard";
 
 import type { ControlProps } from "./controlProps";
 import { Radio, type RadioProps } from "./controls";
 
-export interface RadioGroupProps extends Props {
+export interface RadioGroupProps extends Props, Pick<React.HTMLAttributes<Element>, "role">, React.AriaAttributes {
```
ggdouglas: Could you explain the reasoning around using `Pick<React.HTMLAttributes<Element>` and `React.AriaAttributes` for this prop interface?

Could we alternately achieve this using something like `HTMLDivProps`? https://github.com/palantir/blueprint/blob/4111fc89158a72c614c52a35f5d901015f2dddd8/packages/core/src/common/props.ts#L29

packages/core/src/components/forms/radioGroup.tsx
```diff
@@ -86,11 +93,21 @@ export class RadioGroup extends AbstractPureComponent<RadioGroupProps> {
...
+        const labelId = uniqueId("label");
         return (
-            <div className={classNames(Classes.RADIO_GROUP, this.props.className)}>
-                {label == null ? null : <label className={Classes.LABEL}>{label}</label>}
-                {Array.isArray(this.props.options) ? this.renderOptions() : this.renderChildren()}
+            <div
+                className={classNames(Classes.RADIO_GROUP, className)}
+                role="radiogroup"
+                aria-labelledby={label ? labelId : undefined}
+                {...removeNonHTMLProps(htmlProps)}
```
ggdouglas, line 105: I think we're missing a few other props here not included in `removeNonHTMLProps`. Namely, `selectedValue` and `disabled` are included on this div where they shouldn't.

Can you include `"selectedValue"` in the `INVALID_PROPS` array of `removeNonHTMLProps` and also add `disabled` to the destructuring statement above so that it is omitted from the rendered div?

https://github.com/palantir/blueprint/blob/4111fc89158a72c614c52a35f5d901015f2dddd8/packages/core/src/common/props.ts#L130-L163

packages/core/src/components/forms/radioGroup.tsx
```diff
@@ -86,11 +93,21 @@ export class RadioGroup extends AbstractPureComponent<RadioGroupProps> {
...
+        const { label, options, className, children, name, onChange, ...htmlProps } = this.props;
+        const labelId = uniqueId("label");
         return (
-            <div className={classNames(Classes.RADIO_GROUP, this.props.className)}>
-                {label == null ? null : <label className={Classes.LABEL}>{label}</label>}
-                {Array.isArray(this.props.options) ? this.renderOptions() : this.renderChildren()}
+            <div
+                className={classNames(Classes.RADIO_GROUP, className)}
+                role="radiogroup"
+                aria-labelledby={label ? labelId : undefined}
```
ggdouglas, line 104: Is it problematic that we now can pass both `aria-labelledby` and `label` to this component? Take the following example:
```tsx
<label id="external-label">My Label</label>
<RadioGroup
    label={<label id="prop-node-label">Determine lunch</label>}
    aria-labelledby="external-label"
    ...
>
    <Radio label="Soup" value="one" />
    <Radio label="Salad" value="two" />
    <Radio label="Sandwich" value="three" />
</RadioGroup>
```

Which "label" should take precedence?

Furthermore, we can theoretically achieve the proper aria labeling by passing in a label manually.
```tsx
<RadioGroup
    aria-labelledby="my-label"
    ...
>
    <Label id="my-label">My Label</Label>
    <Radio label="Soup" value="one" />
    <Radio label="Salad" value="two" />
    <Radio label="Sandwich" value="three" />
</RadioGroup>
```

IMO, the functionality of the `label` prop itself seems a bit ambiguous for this case. Perhaps it would ultimately be better if `<RadioGroup>` did away with this prop and instead associated any `<Label>` components passed as children with the radio group.

packages/core/src/components/forms/radioGroup.tsx
```diff
@@ -124,15 +141,15 @@ export class RadioGroup extends AbstractPureComponent<RadioGroupProps> {
     }
 
     private getRadioProps(optionProps: OptionProps): Omit<RadioProps, "ref"> {
-        const { name } = this.props;
+        const { name, inline, disabled: groupDisabled, selectedValue, onChange } = this.props;
```
ggdouglas: Are the changes here functional or cleanup? If it's the latter, can we scope these changes into another PR?

ggdouglas (changes requested):
I'm generally supportive of the direction of this PR since it improves accessibility, but I also think we should more carefully consider how we approach the composition of the label component in the radio group.

I also support adding `role="radiogroup"` to the component since it should generally be there. Could you split that part of this PR out so that we can expedite getting that fixed?

packages/core/src/components/forms/radioGroup.tsx
bvandrc: It's just a style change to remove the mix of destructuring and not destructuring `this.props` in this function.

Not worth the change, reverted.

packages/core/src/components/forms/radioGroup.tsx
bvandrc: > Could we alternately achieve this using something like HTMLDivProps?

Absolutely! Changed.

packages/core/src/components/forms/radioGroup.tsx
bvandrc: Done

packages/core/src/components/forms/radioGroup.tsx
bvandrc: > Is it problematic that we now can pass both aria-labelledby and label to this component? 

No, this allows a consumer to pass a custom `aria-labelledby` if they wish to use an external label.

> Which "label" should take precedence?

In your example, `external-label` takes precedence because the consumer intentionally passed the `aria-labelledby` prop. which is exactly the way they should do it if they want an external label. The internal label no longer labels it to a screenreader, but that's exactly what they wanted when they chose to pass `aria-labelledby`.

> Furthermore, we can theoretically achieve the proper aria labeling by passing in a label manually.

Yes your second example is exactly what is happening. Via `uniqueId`. The code you wrote here is the code that it is.

> IMO, the functionality of the label prop itself seems a bit ambiguous for this case.

I disagree, it's pretty clear as to what it does. It creates a label for the radiogroup.

>  and instead associated any <Label> components passed as children with the radio group.

You would really just want 1 label.

To summarize, this change keeps all the existing implementation.

bvandrc: @ggdouglas looks like this PR has been forgotten about, can you take a look?

ggdouglas (approved):

PR #6947, opened 2024-08-23
fix: Respect icon intent color passed to Dialog

### LINKED ISSUE #6824: [BUG] Icon intent ignored in <Dialog/> title
#### Environment

- __Package version(s)__: @blueprintjs/core 5.10.3
- __Operating System__: MacOS
- __Browser name and version__: Chrome 125.0.6422.76

#### Steps to reproduce

- Make a dialog
- Put an icon in the header with an intent
- Expect icon to render with intent

#### Actual behavior

- Icon renders as grey regardless of intent

image

#### Expected behavior

- Icon renders with intent if set

image

#### Possible solution

Currently manually setting `color` prop to intent colour

ggdouglas:
#### Checklist

- [N/A] Includes tests
- [X] Update documentation

Fixes an issue where dialog header would always overwrite the intent of an icon passed into the dialog, either though the `icon` prop:
```tsx
<Dialog icon={<Icon icon={IconNames.INFO_SIGN} intent={Intent.PRIMARY} ... />}
```
or embedded in the title:
```tsx
<Dialog title={<>Info <Icon icon={IconNames.INFO_SIGN} intent={Intent.PRIMARY} /></>} ... />}
```

> **Note:** Icons with the `color` property set (e.g. `<Icon icon={IconNames.INFO_SIGN} color={Colors.BLUE3} />`) are not affected by this bug since they set color on the fill of the icon's SVG directly.

##### Light Mode
| Before | After    |
| ------  | ------  |
| before-light | after-light |

##### Dark Mode
| Before | After    |
| ------  | ------  |
| before-dark | after-dark |

### DIFF
--- a/packages/core/src/components/dialog/_dialog.scss
+++ b/packages/core/src/components/dialog/_dialog.scss
@@ -111,11 +111,8 @@ $dialog-header-padding: math.div($pt-grid-size, 2);
     margin-right: $dialog-padding * 0.5;
 
     // only apply light color to icons that are not intent-specific
-    :not([class*="#{$ns}-intent"]) {
-      &.#{$ns}-icon-large,
-      &.#{$ns}-icon {
-        color: $pt-icon-color;
-      }
+    &:not([class*="#{$ns}-intent"]) {
+      color: $pt-icon-color;
     }
   }
 
--- a/packages/core/src/components/dialog/_dialog.scss
+++ b/packages/core/src/components/dialog/_dialog.scss
@@ -110,6 +110,12 @@ $dialog-header-padding: math.div($pt-grid-size, 2);
     flex: 0 0 auto;
     margin-left: -3px;
     margin-right: $dialog-padding * 0.5;
+
+    @each $intent, $color in $pt-intent-colors {
+      &.#{$ns}-intent-#{$intent} {
+        color: $color;
+      }
+    }
   }
 
   .#{$ns}-heading {
@@ -130,6 +136,12 @@ $dialog-header-padding: math.div($pt-grid-size, 2);
     .#{$ns}-icon-large,
     .#{$ns}-icon {
       color: $pt-dark-icon-color;
+
+      @each $intent, $color in $pt-intent-colors {
+        &.#{$ns}-intent-#{$intent} {
+          color: $color;
+        }
+      }
     }
   }
 
--- a/packages/core/src/components/dialog/_dialog.scss
+++ b/packages/core/src/components/dialog/_dialog.scss
@@ -106,14 +106,15 @@ $dialog-header-padding: math.div($pt-grid-size, 2);
 
   .#{$ns}-icon-large,
   .#{$ns}-icon {
-    color: $pt-icon-color;
     flex: 0 0 auto;
     margin-left: -3px;
     margin-right: $dialog-padding * 0.5;
 
-    @each $intent, $color in $pt-intent-colors {
-      &.#{$ns}-intent-#{$intent} {
-        color: $color;
+    // only apply light color to icons that are not intent-specific
+    .#{$ns}-icon:not([class*="#{$ns}-intent"]) {
+      &.#{$ns}-icon-large,
+      &.#{$ns}-icon {
+        color: $pt-icon-color;
       }
     }
   }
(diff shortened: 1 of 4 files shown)


packages/core/src/components/dialog/_dialog.scss
```diff
@@ -130,6 +136,12 @@ $dialog-header-padding: math.div($pt-grid-size, 2);
     .#{$ns}-icon-large,
     .#{$ns}-icon {
       color: $pt-dark-icon-color;
+
+      @each $intent, $color in $pt-dark-intent-text-colors {
```
ggdouglas: Tried out both `$pt-intent-colors` and `$pt-dark-intent-text-colors` here for dark mode. The `$pt-intent-colors` that we use for light mode icon intents looks a bit too low contrast, so opted for the dark "text" intent colors instead. The text colors are lighter, and thus they stand out from the dark background a bit more.

pt-intent-colors

pt-dark-intent-text-colors

---

The weird thing here is using something named "intent-text-colors" for icon intent. The colors do align with the original intended dark icon intent colors, but there are multiple levels of selector overriding:

Screenshot 2024-08-23 at 11 11 09@2x

packages/core/src/components/dialog/_dialog.scss
```diff
@@ -110,6 +110,12 @@ $dialog-header-padding: math.div($pt-grid-size, 2);
     flex: 0 0 auto;
     margin-left: -3px;
     margin-right: $dialog-padding * 0.5;
+
+    @each $intent, $color in $pt-intent-colors {
+      &.#{$ns}-intent-#{$intent} {
+        color: $color;
+      }
+    }
```
invliD: Could we avoid having to redefine all intent colors if we qualified the override in L109 to only apply to intent-less icons?

ggdouglas: Yes! I originally went with the following to do just that:
```scss
.#{$ns}-icon:not([class*="#{$ns}-intent"]) {
  &.#{$ns}-icon-large,
  &.#{$ns}-icon {
    color: $pt-icon-color;
  }
}
```
This works, and it has the advantage of preventing the styles from being overridden in the first place. However, the selector in the example above feels a bit brittle to me. This is why I opted change it and explicitly loop over the intent colors.

Totally happy to change it back to be qualified by the above, just wanted to make sure the selector was optimized first.

ggdouglas: Added here: 48ecd122255085690d8e36e999334a4eb7c25c97

packages/core/src/components/dialog/_dialog.scss
```diff
@@ -106,10 +106,17 @@ $dialog-header-padding: math.div($pt-grid-size, 2);
...
   .#{$ns}-icon {
-    color: $pt-icon-color;
     flex: 0 0 auto;
     margin-left: -3px;
     margin-right: $dialog-padding * 0.5;
+
+    // only apply light color to icons that are not intent-specific
+    .#{$ns}-icon:not([class*="#{$ns}-intent"]) {
+      &.#{$ns}-icon-large,
+      &.#{$ns}-icon {
```
invliD: Are you sure this is correct? I read this (simplified) as `.#{$ns}-icon .#{$ns}-icon.#{$ns}-icon:not([class*="#{$ns}-intent"])` (for the standard size case), which (1) nests the icon class in itself, and (2) duplicates the selector on the inner icon element. Unless there are specificity issues, I feel like `:not([class*="#{$ns}-intent"])` should be enough here (i.e. drop the part before the colon in R114, and drop R115-R116.

ggdouglas: Ah, yep, you're absolutely right! The additional icon class is duplicative. Removed here 3f710af985ab56268b3e61c6d3a0bb9e0baad5a3

packages/core/src/components/dialog/_dialog.scss
```diff
@@ -106,10 +106,17 @@ $dialog-header-padding: math.div($pt-grid-size, 2);
...
   .#{$ns}-icon {
-    color: $pt-icon-color;
     flex: 0 0 auto;
     margin-left: -3px;
     margin-right: $dialog-padding * 0.5;
+
+    // only apply light color to icons that are not intent-specific
+    :not([class*="#{$ns}-intent"]) {
+      &.#{$ns}-icon-large,
+      &.#{$ns}-icon {
```
invliD: These shouldn't be necessary either (except maybe for specificity); you're assembling a selector like `.#{$ns}-icon:not([class*="#{$ns}-intent"]).#{$ns}-icon`, so you have the icon class in there twice.

ggdouglas: 🤦‍♂️ Not quite sure how I missed that one.. Thanks for the catch! 2d4ae487a3c6b54ca4894576e91b34a0218c39b0

invliD (approved):

PR #6950, opened 2024-08-26
Add focus styles to FileInput

### LINKED ISSUE #6859: FileInput has no visual focus state
#### Steps to reproduce
See https://blueprintjs.com/docs/#core/components/file-input

#### Actual behavior
`FileInput` component can be focused, and the file selector opened with an enter keypress, but there is no visual focus state

#### Expected behavior
`FileInput` should support a visual focus indication similar to other input components

#### Possible solution
Add same visual focus indication as `InputGroup`/`NumericInput`/`TextArea` and others

ggdouglas:
#### Checklist

- [N/A] Includes tests
- [N/A] Update documentation

Add visual focus styles to [FileInput](https://blueprintjs.com/docs/#core/components/file-input) to match other input components.

##### Focus (light mode)
| Before | After    |
| ------  | ------  |
| before | after |

##### Focus (dark mode)
| Before | After |
| ------  | ------  |
| before | after |

Screenshot 2024-08-26 at 10 11 32

### DIFF
--- a/packages/core/src/components/forms/_file-input.scss
+++ b/packages/core/src/components/forms/_file-input.scss
@@ -65,6 +65,7 @@ $file-input-button-padding-small: ($pt-input-height-small - $pt-button-height-sm
       }
     }
 
+    // borrows focus styles from pt-input and pt-dark-input mixins (see _common.scss)
     &:focus + .#{$ns}-file-upload-input {
       box-shadow: input-transition-shadow($input-shadow-color-focus, true),
         $input-box-shadow-focus;
--- a/packages/core/src/components/forms/_file-input.scss
+++ b/packages/core/src/components/forms/_file-input.scss
@@ -64,6 +64,16 @@ $file-input-button-padding-small: ($pt-input-height-small - $pt-button-height-sm
         }
       }
     }
+
+    &:focus + .#{$ns}-file-upload-input {
+      box-shadow: input-transition-shadow($input-shadow-color-focus, true),
+        $input-box-shadow-focus;
+
+      .#{$ns}-dark & {
+        box-shadow: dark-input-transition-shadow($dark-input-shadow-color-focus, true),
+          $pt-dark-input-box-shadow;
+      }
+    }
   }
 
   &.#{$ns}-file-input-has-selection {


packages/core/src/components/forms/_file-input.scss
```diff
@@ -64,6 +64,16 @@ $file-input-button-padding-small: ($pt-input-height-small - $pt-button-height-sm
...
+
+    &:focus + .#{$ns}-file-upload-input {
+      box-shadow: input-transition-shadow($input-shadow-color-focus, true),
+        $input-box-shadow-focus;
+
+      .#{$ns}-dark & {
+        box-shadow: dark-input-transition-shadow($dark-input-shadow-color-focus, true),
+          $pt-dark-input-box-shadow;
+      }
+    }
```
evansjohnson, line 77: is this a paired down version of the `pt-input-intent` mixin? maybe we could comment that if so to at least loosely link these

also wondering if we should be using the same `2px` found in that mixin for consistency? is there a reason these should be different?

ggdouglas: The box shadows here were pulled from the base `pt-input` mixin:
https://github.com/palantir/blueprint/blob/dc1b16237900c9b49e37f31b7a95523c16ae07e0/packages/core/src/components/forms/_common.scss#L109-L112 https://github.com/palantir/blueprint/blob/dc1b16237900c9b49e37f31b7a95523c16ae07e0/packages/core/src/components/forms/_common.scss#L59-L73 https://github.com/palantir/blueprint/blob/c50a4a49b736b87dfa36284565b6975ef0b00dbb/packages/core/src/common/_mixins.scss#L100-L102 https://github.com/palantir/blueprint/blob/dc1b16237900c9b49e37f31b7a95523c16ae07e0/packages/core/src/components/forms/_common.scss#L30-L32
and the base `pt-dark-input` mixin:
https://github.com/palantir/blueprint/blob/dc1b16237900c9b49e37f31b7a95523c16ae07e0/packages/core/src/components/forms/_common.scss#L200-L202 https://github.com/palantir/blueprint/blob/dc1b16237900c9b49e37f31b7a95523c16ae07e0/packages/core/src/components/forms/_common.scss#L75-L89

The default focus box shadow on input groups without an intent appears to be 2px total (1px inset + 1px).
Screenshot 2024-08-28 at 11 10 20@2x

For input groups with an intent, it appears to be 3px (1px inset + 2px)
Screenshot 2024-08-28 at 11 10 11@2x

I chose to go with the non-intent style box shadow for file input since it currently does not have the ability to have our intent styles applied. I do think it's a bit odd that the default, non-intent focus outline doesn't match the width of the focused outlines. Perhaps that's something to address in a FLUP? For now, I can comment to reference where these are coming from.

evansjohnson: > I do think it's a bit odd that the default, non-intent focus outline doesn't match the width of the focused outlines.

ahh nice catch - I just recently updated these focus styles. looking back at this comment: https://github.com/palantir/blueprint/pull/6854#issuecomment-2200929002 I'm seeing originally I used a lighter color for the 2px which then became the same color (and a that point I probably just missed keeping it consistent at 2px)

evansjohnson: oh - maybe I did that because it already starts out with the 1px colored border when not focused

agreed without intent it makes sense for this to just be the 2px focus state!

evansjohnson (approved):

PR #6952, opened 2024-08-27
Add minimal prop to Callout which hides the background fill color

ayxliu19:
#### Checklist

- [x] Includes tests
- [x] Update documentation

New minimal prop on the Callout component that hides the background fill color

Screenshot 2024-08-27 at 2 34 50 PM

1. Go to http://localhost:9001/#core/components/callout
2. Toggle the minimal option and observe that the background color turns on and off.

### DIFF
--- a/packages/core/src/components/callout/_callout.scss
+++ b/packages/core/src/components/callout/_callout.scss
@@ -103,11 +103,11 @@ $callout-padding-compact: $pt-grid-size;
 
   @each $intent, $color in $pt-intent-colors {
     &.#{$ns}-intent-#{$intent} {
+      color: map-get($pt-intent-text-colors, $intent);
+
       &:not(.#{$ns}-minimal) {
         background-color: rgba($color, 0.1);
       }
-
-      color: map-get($pt-intent-text-colors, $intent);
 
       @media (forced-colors: active) and (prefers-color-scheme: dark) {
         border: 1px solid $pt-high-contrast-mode-border-color;
@@ -120,11 +120,11 @@ $callout-padding-compact: $pt-grid-size;
       }
 
       .#{$ns}-dark & {
+        color: map-get($pt-dark-intent-text-colors, $intent);
+
         &:not(.#{$ns}-minimal) {
           background-color: rgba($color, 0.2);
         }
-
-        color: map-get($pt-dark-intent-text-colors, $intent);
 
         &[class*="#{$ns}-icon-"]::before,
         > .#{$ns}-icon:first-child,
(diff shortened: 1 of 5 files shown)


packages/core/src/components/callout/_callout.scss
```diff
@@ -87,6 +87,14 @@ $callout-padding-compact: $pt-grid-size;
...
+  &.#{$ns}-minimal,
+  &.#{$ns}-minimal.#{$ns}-intent-primary,
+  &.#{$ns}-minimal.#{$ns}-intent-success,
+  &.#{$ns}-minimal.#{$ns}-intent-warning,
+  &.#{$ns}-minimal.#{$ns}-intent-danger {
+    background-color: transparent;
+  }
+
   .#{$ns}-dark & {
     background-color: rgba($gray3, 0.2);
```
ggdouglas: It looks like we're missing overrides for the dark mode variants of Callout.

Screenshot 2024-08-27 at 15 03 26

packages/core/src/components/callout/_callout.scss
```diff
@@ -87,6 +87,14 @@ $callout-padding-compact: $pt-grid-size;
     }
   }
 
+  &.#{$ns}-minimal,
+  &.#{$ns}-minimal.#{$ns}-intent-primary,
+  &.#{$ns}-minimal.#{$ns}-intent-success,
+  &.#{$ns}-minimal.#{$ns}-intent-warning,
+  &.#{$ns}-minimal.#{$ns}-intent-danger {
+    background-color: transparent;
+  }
```
ggdouglas: Instead of explicitly overwriting styles, could we instead use a negation pseudo-class to prevent the background color from being applied in the first case if the callout is minimal? e.g.

```scss
.#{$ns}-callout {
  &:not(.#{$ns}-minimal) {
    background-color: rgba($gray3, 0.15);
  }

  .#{$ns}-dark & {
    &:not(.#{$ns}-minimal) {
      background-color: rgba($gray3, 0.2);
    }
  }

  @each $intent, $color in $pt-intent-colors {
    &.#{$ns}-intent-#{$intent} {
      &:not(.#{$ns}-minimal) {
        background-color: rgba($color, 0.1);
      }

      .#{$ns}-dark & {
        &:not(.#{$ns}-minimal) {
          background-color: rgba($color, 0.2);
        }
      }
    }
  }
}
```

ggdouglas (changes requested):

ggdouglas (dismissed):
LGTM. Thanks!

ggdouglas (approved):

PR #6959, opened 2024-09-03
Make sure toast max width doesn't exceed screen size on mobile devices

evansjohnson:
#### Checklist

- [-] Includes tests
- [-] Update documentation

I could try to update the tests to avoid regressing but think this may be pretty hard to test for, particularly given how long this went without being a big issue

Modify toast "max-width" to be the minimum of page width - 20px and 500px, rather than the max width always being 500px which may exceed the screen size, particularly on mobile devices.

any concerns/ possible negative impact from this change

## this PR
Screenshot 2024-09-13 at 10 12 49 AM

layout gets messed up but I think this is an improvement over current behavior, particularly for toasts that don't have an action button

## current
Screenshot 2024-09-13 at 10 13 56 AM

### DIFF
--- a/packages/core/src/components/toast/_toast.scss
+++ b/packages/core/src/components/toast/_toast.scss
@@ -104,7 +104,7 @@ $toast-intent-colors: (
   display: flex;
   margin: $toast-margin 0 0;
   max-width: min($toast-max-width, 100%);
-  min-width: min($toast-min-width, 100%));
+  min-width: min($toast-min-width, 100%);
 
   // toast is interactive even though container isn't
   pointer-events: all;
--- a/packages/core/src/components/toast/_toast.scss
+++ b/packages/core/src/components/toast/_toast.scss
@@ -103,7 +103,7 @@ $toast-intent-colors: (
   box-shadow: $pt-toast-box-shadow;
   display: flex;
   margin: $toast-margin 0 0;
-  max-width: $toast-max-width;
+  max-width: min($toast-max-width, calc(100vw - $toast-margin));
   min-width: $toast-min-width;
 
   // toast is interactive even though container isn't
--- a/packages/core/src/components/toast/_toast.scss
+++ b/packages/core/src/components/toast/_toast.scss
@@ -104,7 +104,7 @@ $toast-intent-colors: (
   display: flex;
   margin: $toast-margin 0 0;
   max-width: min($toast-max-width, calc(100vw - $toast-margin));
-  min-width: $toast-min-width;
+  min-width: max($toast-min-width, calc(100vw - $toast-margin));
 
   // toast is interactive even though container isn't
   pointer-events: all;
--- a/packages/core/src/components/toast/_toast.scss
+++ b/packages/core/src/components/toast/_toast.scss
@@ -103,8 +103,8 @@ $toast-intent-colors: (
   box-shadow: $pt-toast-box-shadow;
   display: flex;
   margin: $toast-margin 0 0;
-  max-width: min($toast-max-width, calc(100vw - $toast-margin));
-  min-width: min($toast-min-width, calc(100vw - $toast-margin));
+  max-width: min($toast-max-width, 100%);
+  min-width: min($toast-min-width, 100%));
 
   // toast is interactive even though container isn't
   pointer-events: all;
--- a/packages/core/src/components/toast/_toast.scss
+++ b/packages/core/src/components/toast/_toast.scss
@@ -104,7 +104,7 @@ $toast-intent-colors: (
   display: flex;
   margin: $toast-margin 0 0;
   max-width: min($toast-max-width, calc(100vw - $toast-margin));
-  min-width: max($toast-min-width, calc(100vw - $toast-margin));
+  min-width: min($toast-min-width, calc(100vw - $toast-margin));
 
   // toast is interactive even though container isn't
   pointer-events: all;


packages/core/src/components/toast/_toast.scss
```diff
@@ -103,7 +103,7 @@ $toast-intent-colors: (
   box-shadow: $pt-toast-box-shadow;
   display: flex;
   margin: $toast-margin 0 0;
-  max-width: $toast-max-width;
+  max-width: min($toast-max-width, calc(100vw - $toast-margin));
```
evansjohnson: probably should handle considering `$toast-min-width` here as well 🤔

evansjohnson: from https://www.w3.org/TR/CSS21/visudet.html#min-max-widths

> The following algorithm describes how the two properties influence the [used value](https://www.w3.org/TR/CSS21/cascade.html#computed-value) of the ['width'](https://www.w3.org/TR/CSS21/visudet.html#propdef-width) property:
> The tentative used width is calculated (without ['min-width'](https://www.w3.org/TR/CSS21/visudet.html#propdef-min-width) and ['max-width'](https://www.w3.org/TR/CSS21/visudet.html#propdef-max-width)) following the rules under ["Calculating widths and margins"](https://www.w3.org/TR/CSS21/visudet.html#Computing_widths_and_margins) above.
If the tentative used width is greater than ['max-width'](https://www.w3.org/TR/CSS21/visudet.html#propdef-max-width), the rules [above](https://www.w3.org/TR/CSS21/visudet.html#Computing_widths_and_margins) are applied again, but this time using the computed value of ['max-width'](https://www.w3.org/TR/CSS21/visudet.html#propdef-max-width) as the computed value for ['width'](https://www.w3.org/TR/CSS21/visudet.html#propdef-width).
If the resulting width is smaller than ['min-width'](https://www.w3.org/TR/CSS21/visudet.html#propdef-min-width), the rules [above](https://www.w3.org/TR/CSS21/visudet.html#Computing_widths_and_margins) are applied again, but this time using the value of ['min-width'](https://www.w3.org/TR/CSS21/visudet.html#propdef-min-width) as the computed value for ['width'](https://www.w3.org/TR/CSS21/visudet.html#propdef-width).

so I guess I should do something similar to this for the `min-width` field as well (though we're _propobably_ safe with 300px...)

packages/core/src/components/toast/_toast.scss
```diff
@@ -103,7 +103,7 @@ $toast-intent-colors: (
...
   min-width: $toast-min-width;
```
evansjohnson:
```suggestion
  min-width: max($toast-min-width, calc(100vw - $toast-margin));
```

packages/core/src/components/toast/_toast.scss
```diff
@@ -103,8 +103,8 @@ $toast-intent-colors: (
   box-shadow: $pt-toast-box-shadow;
   display: flex;
   margin: $toast-margin 0 0;
-  max-width: $toast-max-width;
-  min-width: $toast-min-width;
+  max-width: min($toast-max-width, calc(100vw - $toast-margin));
+  min-width: max($toast-min-width, calc(100vw - $toast-margin));
```
ggdouglas: This sets `min-width` to be `100vw - $toast-margin` by default, which will end up stretching toasts across the screen.

Screenshot 2024-09-11 at 10 26 56@2x


I don't think this was the intended effect for the default experience. Would a `min()` or `clamp()` function work here instead?
```suggestion
  min-width: min($toast-min-width, calc(100vw - $toast-margin));
```

ggdouglas:
Is there a way we can we leverage the existing padding on the Overlay surrounding Toasts vs. hardcoding a `calc(100vw - $toast-margin)` in like this? The `20px` side padding that the overlay container provides seems like the appropriate amount to account for a toast on a mobile device.

Screenshot 2024-09-11 at 10 49 26@2x

Screenshot 2024-09-11 at 10 51 14@2x

packages/core/src/components/toast/_toast.scss
evansjohnson: gah yea I blindly made this change and hadn't tested it yet... thanks for actually testing (I'd like to think I would have before merging but 😅)

but yes this makes sense, we still want `min-width: $toast-min-width;`, unless the screen is smaller than that in which case we want `calc(100vw - $toast-margin)`

ggdouglas (dismissed):
LGTM!

evansjohnson:
> leverage the existing padding on the Overlay surrounding Toasts

good call! this toast margin has nothing to do with space outside the toast right now so no good reason I'm using that. it looks like just setting it to 100% does the trick.

Screenshot 2024-09-12 at 5 06 22 PM

of course if you add a button and have a _really_ small screen size things start to look a bit silly but I think this is still a directionally a good change. to take this further we could wrap the text and action button in it's own flex wrapper and set `flex-direction: column`

packages/core/src/components/toast/_toast.scss
```diff
@@ -103,8 +103,8 @@ $toast-intent-colors: (
   box-shadow: $pt-toast-box-shadow;
   display: flex;
   margin: $toast-margin 0 0;
-  max-width: $toast-max-width;
-  min-width: $toast-min-width;
+  max-width: min($toast-max-width, calc(100vw - $toast-margin));
+  min-width: min($toast-min-width, calc(100vw - $toast-margin));
```
evansjohnson:
```suggestion
  max-width: min($toast-max-width, 100%);
  min-width: min($toast-min-width, 100%));
```

packages/core/src/components/toast/_toast.scss
```diff
@@ -103,8 +103,8 @@ $toast-intent-colors: (
   box-shadow: $pt-toast-box-shadow;
   display: flex;
   margin: $toast-margin 0 0;
-  max-width: $toast-max-width;
-  min-width: $toast-min-width;
+  max-width: min($toast-max-width, 100%);
+  min-width: min($toast-min-width, 100%));
```
evansjohnson:
```suggestion
  min-width: min($toast-min-width, 100%);
```

evansjohnson:
## this PR
Screenshot 2024-09-13 at 10 12 49 AM

layout gets messed up but I think this is an improvement over current behavior, particularly for toasts that don't have an action button

## current
Screenshot 2024-09-13 at 10 13 56 AM

ggdouglas (approved):

PR #6960, opened 2024-09-04
fix(timezoneSelect): a11y aria-label

bvandrc:

Add aria-label to timezone select to fix a11y aXe failure.


packages/datetime/src/components/timezone-select/timezoneSelect.tsx
```diff
@@ -162,6 +162,7 @@ export class TimezoneSelect extends AbstractPureComponent<TimezoneSelectProps, T
 
         return (
             <Select<TimezoneWithNames>
+                aria-label="timezone"
```
ggdouglas: On what element in Select does this `aria-label` get applied?

ggdouglas:
@bvandercar-vt could you provide a bit more context as to the accessibility failure you're experiencing and how adding this label addresses it?

packages/datetime/src/components/timezone-select/timezoneSelect.tsx
bvandrc: > On what element in Select does this aria-label get applied?

> could you provide a bit more context as to the accessibility failure you're experiencing and how adding this label addresses it?

Here is the failure:

image

Every type of input needs an `aria-label` or `aria-labelledby`, including a `select` (which in this case is the `combobox`).

ggdouglas: It doesn't look like providing the `aria-label` prop to Select does anything. Inside of `<Select>`, it will be passed via `restProps` to `<QueryList>`, where it is then discarded:
https://github.com/palantir/blueprint/blob/3f4a89445aaffd2c3f3d863eea56b5c11716f404/packages/select/src/components/select/select.tsx#L137-L142 https://github.com/palantir/blueprint/blob/3f4a89445aaffd2c3f3d863eea56b5c11716f404/packages/select/src/components/query-list/queryList.tsx#L220

Did you mean to pass this `aria-label` via `popoverTargetProps`?
```tsx
<Select<TimezoneWithNames>
    popoverTargetProps={{
        "aria-label": "timezone",
    }}
    ...
/>
```

---

Regardless, it feels a bit redundant to have to add an additional label when the Select's input already has a placeholder value of "Select timezone...". Is there any way we can leverage the existing placeholder to improve accessibility?

bvandrc: > Did you mean to pass this aria-label via popoverTargetProps?

Good catch, that worked perfectly!

> it feels a bit redundant to have to add an additional label when the Select's input already has a placeholder value of "Select timezone...". Is there any way we can leverage the existing placeholder to improve accessibility?

If the placeholder was enough, we wouldn't be getting the aXe failure. A `select`/`combobox` component needs dedicated label for _what it is_, it can't just use its text content because that is just its current value, which doesn't always tell a screenreader what the component is actually _for_.

bvandrc:
@jscheiny @ggdouglas I rebased this PR, it's ready.

ggdouglas (approved):

ggdouglas:
Thanks for improving the accessibility here, @bvandercar-vt. Much appreciated!

PR #6966, opened 2024-09-10
Fix border overlap between Buttons in outlined ButtonGroup

ggdouglas:
#### Checklist

- [ ] Includes tests
- [x] Update documentation

Fixes an issue where outlined button styles were not rendering properly when wrapped in a Popover inside of ButtonGroup. Also adds an `outlined` variant prop example to the "Usage with popovers" ButtonGroup example to allow for visual testing of this issue in the future.

#### New `outlined` prop toggle on ButtonGroup's ["Usage with popovers" ](https://blueprintjs.com/docs/#core/components/button-group.usage-with-popovers) example.

Screenshot 2024-09-09 at 21 30 56

This change fixes borders on outlined buttons with popovers in two instances:

1. The first is where the `outlined` prop is applied to the `ButtonGroup` parent container:
```tsx
<ButtonGroup outlined>
  <Popover content={...}>
    <Button icon="git-branch" rightIcon="caret-down" text="Main" />
  </Popover>
  <Button icon="more" />
</ButtonGroup>
```
In this case, buttons surrounded by popovers weren't being styled as outlined at all. The `.bp5-popover-target` class breaks the selector applying the outlined styling.
| Before | After    |
| ------  | ------  |
| before | after |

2. The other is where the `outlined` prop is applied to the individual buttons within the group
```tsx
<ButtonGroup>
  <Popover content={...}>
    <Button icon="git-branch" rightIcon="caret-down" text="Main" outlined />
  </Popover>
  <Button icon="more" outlined />
</ButtonGroup>
```
In this case, the outline styles render, but there is a slight overlap between the intermediate borders of the button group. This overlap is noticeable due to the semi-transparent styling of the borders on outlined buttons.

Screenshot 2024-09-09 at 21 29 04@2x

- Various possible Button variants/permutations that can be contained within a ButtonGroup wrapper
- Buttons surrounded by Popovers
- Buttons surrounded by Tooltips

### DIFF
--- a/packages/core/src/components/button/_button-group.scss
+++ b/packages/core/src/components/button/_button-group.scss
@@ -142,12 +142,13 @@ Styleguide button-group
   }
 
   &.#{$ns}-outlined {
-    .#{$ns}-button {
+    > .#{$ns}-popover-target > .#{$ns}-button,
+    > .#{$ns}-button {
       @include pt-button-outlined();
     }
 
     &:not(.#{$ns}-vertical) {
-      > .#{$ns}-popover-target:not(:last-child) .#{$ns}-button,
+      > .#{$ns}-popover-target:not(:last-child) > .#{$ns}-button,
       > .#{$ns}-button:not(:last-child) {
         border-right: none;
       }
--- a/packages/core/src/components/button/_button-group.scss
+++ b/packages/core/src/components/button/_button-group.scss
@@ -265,6 +265,11 @@ Styleguide button-group
         border-bottom: none;
       }
     }
+
+    > .#{$ns}-popover-target:not(:last-child) .#{$ns}-button.#{$ns}-outlined,
+    > .#{$ns}-button.#{$ns}-outlined:not(:last-child) {
+      border-bottom: none;
+    }
   }
 
   &.#{$ns}-align-left .#{$ns}-button {
--- a/packages/core/src/components/button/_button-group.scss
+++ b/packages/core/src/components/button/_button-group.scss
@@ -142,14 +142,22 @@ Styleguide button-group
   }
 
   &.#{$ns}-outlined {
-    > .#{$ns}-button {
+    .#{$ns}-button {
       @include pt-button-outlined();
     }
 
     &:not(.#{$ns}-vertical) {
+      > .#{$ns}-popover-target:not(:last-child) .#{$ns}-button,
       > .#{$ns}-button:not(:last-child) {
         border-right: none;
       }
+    }
+  }
+
+  &:not(.#{$ns}-vertical) {
+    > .#{$ns}-popover-target:not(:last-child) .#{$ns}-button.#{$ns}-outlined,
+    > .#{$ns}-button.#{$ns}-outlined:not(:last-child) {
+      border-right: none;
     }
   }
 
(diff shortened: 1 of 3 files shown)


packages/core/src/components/button/_button-group.scss
```diff
@@ -142,17 +142,25 @@ Styleguide button-group
   }
 
   &.#{$ns}-outlined {
-    > .#{$ns}-button {
+    .#{$ns}-button {
```
invliD: We likely need to keep the child combinator here, otherwise a button inside an inline overlay within the button group would be affected by these styles. We'll need to be specific for popovers, similar to L100-L112 above.

packages/core/src/components/button/_button-group.scss
```diff
@@ -142,17 +142,25 @@ Styleguide button-group
...
       @include pt-button-outlined();
     }
 
     &:not(.#{$ns}-vertical) {
+      > .#{$ns}-popover-target:not(:last-child) .#{$ns}-button,
```
invliD: same

packages/core/src/components/button/_button-group.scss
```diff
@@ -142,17 +142,25 @@ Styleguide button-group
...
       }
     }
   }
 
+  &:not(.#{$ns}-vertical) {
+    > .#{$ns}-popover-target:not(:last-child) .#{$ns}-button.#{$ns}-outlined,
+    > .#{$ns}-button.#{$ns}-outlined:not(:last-child) {
+      border-right: none;
+    }
+  }
```
invliD, line 163: Do we need this for vertical button groups too (with border-bottom)?

ggdouglas: Ah, good catch! Added in 79e3441e9138d848d591f34b974e630dfba5ecce

packages/core/src/components/button/_button-group.scss
ggdouglas: Hm, good point. I'd neglected to test for inline overlays since it's a bit more involved to set up an example for those. Does this same issue not apply to the `&.#{$ns}-minimal, &.#{$ns}-outlined` selector above also? https://github.com/palantir/blueprint/blob/1038d7f8604428b56c4204b40d6cc04a84e39371/packages/core/src/components/button/_button-group.scss#L121-L122

Added back child combinator and selected for popovers here: 84f92db44a95ab6cc4739fa4d507711c74e93427

packages/core/src/components/popover/_popover-in-button-group.scss
```diff
@@ -15,7 +15,12 @@
     > .#{$ns}-popover-target:not(:last-child) .#{$ns}-button {
       border-bottom-right-radius: 0;
       border-top-right-radius: 0;
-      margin-right: -$button-border-width;
+    }
+
+    &:not(.#{$ns}-outlined) {
+      > .#{$ns}-popover-target:not(:last-child) .#{$ns}-button:not(.#{$ns}-outlined) {
```
invliD, line 21: There seems to be a lot of duplicate (or at least similar) styles for popovers here and in the other file. We should probably consolidate them and also merge the selectors for popover-target and popover-wrapper, since they should likely apply the exact same styles (the latter only existing for back-compat anyway).

ggdouglas: Yeah, I noticed this as well, but was hesitant to tackle reorganizing those styles in this PR since it's a more involved change. I'm also not super familiar with the historical use of the popover-wrapper class, so that also made the prospect of that change seem more brittle to me.

Happy to address cleaning these up though. Would you recommend this PR or a FLUP?

invliD (approved):

PR #6967, opened 2024-09-10
a11y(DocsApp): fix version selector a11y

bvandrc:
The version selector in the docs app causes aXe a11y failure. Fix with this PR.

Replaces [this PR](https://github.com/palantir/blueprint/pull/6886) with simpler changes to accomplish the a11y failure fix.

### DIFF
--- a/packages/docs-app/src/components/navHeader.tsx
+++ b/packages/docs-app/src/components/navHeader.tsx
@@ -17,7 +17,7 @@
 import type { NpmPackageInfo } from "@documentalist/client";
 import * as React from "react";
 
-import { Classes, HotkeysTarget2, type Intent, Menu, MenuItem, NavbarHeading, Popover, PopupKind, Tag } from "@blueprintjs/core";
+import { Classes, HotkeysTarget2, type Intent, Menu, MenuItem, NavbarHeading, Popover, Tag } from "@blueprintjs/core";
 import { NavButton } from "@blueprintjs/docs-theme";
 
 import { Logo } from "./logo";
@@ -109,7 +109,6 @@ export class NavHeader extends React.PureComponent<NavHeaderProps> {
                     </Menu>
                 }
                 placement="bottom"
-                popupKind={PopupKind.MENU}
             >
                 <Tag
                     interactive={true}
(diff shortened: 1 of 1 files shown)


packages/docs-app/src/components/navHeader.tsx
```diff
@@ -101,8 +102,24 @@ export class NavHeader extends React.PureComponent<NavHeaderProps> {
...
-            <Popover content={<Menu className="docs-version-list">{releaseItems}</Menu>} placement="bottom">
-                <Tag interactive={true} minimal={true} round={true} rightIcon="caret-down">
+            <Popover
+                content={
+                    <Menu aria-label="docs version" className="docs-version-list" id={VERSION_MENU_ID}>
+                        {releaseItems}
+                    </Menu>
+                }
+                placement="bottom"
+                popupKind={PopupKind.MENU}
```
ggdouglas: Do we need to specify `popupKind` here since `"menu"` is the default value?

packages/docs-app/src/components/navHeader.tsx
```diff
@@ -17,7 +17,7 @@
 import type { NpmPackageInfo } from "@documentalist/client";
 import * as React from "react";
 
-import { Classes, HotkeysTarget2, type Intent, Menu, MenuItem, NavbarHeading, Popover, Tag } from "@blueprintjs/core";
+import { Classes, HotkeysTarget2, type Intent, Menu, MenuItem, NavbarHeading, Popover, PopupKind, Tag } from "@blueprintjs/core";
```
ggdouglas: Breaking this import out into multiple lines should satisfy prettier.

ggdouglas (commented):
Thanks for your changes here! Two minor comments, but I'm supportive of getting this change in to the docs.

packages/docs-app/src/components/navHeader.tsx
bvandrc: I didn't realize the default was `menu`, can remove.

packages/docs-app/src/components/navHeader.tsx
bvandrc: Not needed after removing popupkind

ggdouglas (approved):

PR #6972, opened 2024-09-12
fix eslint config

bvandrc:
Getting lint error in CircleCI:

image

The fix: https://stackoverflow.com/questions/72780296/warning-react-version-not-specified-in-eslint-plugin-react-settings-while-run


ggdouglas (approved):
Thanks for adding this

PR #6980, opened 2024-09-16
[Draft] Box Component

ggdouglas:

This PR introduces a new `<Box>` component, providing flexible layout utilities for margin, padding, alignment, and positioning. The component is highly customizable and supports dynamic generation of class names based on a set of configurable props.

CleanShot 2024-09-17 at 13 59 33@2x

This component provides flexible and reusable spacing utilities that integrate seamlessly with Blueprint components, allowing for consistent alignment of smaller components within a broader layout structure.

For instance, multiple Buttons can be arranged horizontally:
```tsx
<Box display="flex" gap={2}>
  <Button intent="primary">Button</Button>
  <Button intent="success">Button</Button>
  <Button intent="warning">Button</Button>
  <Button intent="danger">Button</Button>
</Box>
```
CleanShot 2024-09-17 at 14 42 02@2x

Or stacked vertically:
```tsx
<Box display="flex" flexDirection="column" alignItems="center" gap={2}>
  <Button intent="primary">Button</Button>
  <Button intent="success">Button</Button>
  <Button intent="warning">Button</Button>
  <Button intent="danger">Button</Button>
</Box>
```
CleanShot 2024-09-17 at 14 44 20@2x

It can also be used to create more molecular components (such as [EntityTitle](https://blueprintjs.com/docs/#core/components/entity-title)):
```tsx
<Box display="flex" gap={2}>
    <Icon className={Classes.TEXT_MUTED} icon="shop" />
    Buy groceries on my way home
    <Box asChild={true} marginXStart="auto">
        <Tag intent="danger" minimal={true}>
            Due today
        </Tag>
    </Box>
</Box>
```
CleanShot 2024-09-17 at 14 56 02@2x


packages/core/src/components/box/boxProps.ts
```diff
@@ -0,0 +1,330 @@
...
+     *
+     * Defaults to "div" when omitted.
+     */
+    as?: As;
+
+    /**
+     * When true, `<Box>` will **not** render an extra wrapper. Instead, it will
+     * clone & enhance its single child element, merging class names and props.
+     */
+    asChild?: boolean;
```
jscheiny: Wondering if we can combine these by having `"child"` be an option for `As` so that we don't have a weird situation where a user has both `asChild` and `as` and we need to determine what that means.

packages/core/src/components/box/boxProps.ts
```diff
@@ -0,0 +1,330 @@
...
+
+/**
+ * The range of values for spacing tokens.
+ */
+type SpacingRange = 0 | 0.5 | 1 | 1.5 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9 | 10;
+
+/**
+ * The range of values for size as a percentage.
+ */
+type SizeRange = 25 | 50 | 75 | 100;
```
mm-wang, line 32: **question**: Why limited to quarters? Could you do 33/67? Or eighths?

packages/core/src/components/box/boxProps.ts
```diff
@@ -0,0 +1,330 @@
...
+export type Position = "static" | "absolute" | "relative" | "fixed" | "sticky";
+
+type BoxOwnProps = Readonly<{
+    /**
+     * Underlying element or component to render. Accepts any intrinsic
+     * element name (e.g. "div", "span") or a React component.
+     *
+     * Defaults to "div" when omitted.
+     */
+    as?: As;
```
mm-wang: **note**: I find it a little odd to think of this as a prop; I think it would have made sense to me to call this `wrapperElement` and `asChild` to be `includeWrapper` but with opposite logic. Also combining them makes sense to me.

packages/core/src/components/box/boxProps.ts
```diff
@@ -0,0 +1,330 @@
...
+
+    /**
+     * When true, `<Box>` will **not** render an extra wrapper. Instead, it will
+     * clone & enhance its single child element, merging class names and props.
+     */
+    asChild?: boolean;
+
+    /** Shorthand for the CSS `gap` property (applies to flex and grid layouts). */
+    gap?: Gap;
```
mm-wang, line 194: **question**: Do we want to limit this value? I think not, but curious

packages/core/src/components/box/boxProps.ts
```diff
@@ -0,0 +1,330 @@
...
+    /** CSS `position` */
+    position?: Position;
+
+    /** CSS `overflow` */
+    overflow?: Overflow;
+    /** CSS `overflow-x` */
+    overflowX?: Overflow;
+    /** CSS `overflow-y` */
+    overflowY?: Overflow;
+}>;
```
mm-wang: **suggestion**: For sake of organization, would you want to separate these props out into something like `OverflowProps`, `PositionProps`, etc. and use an intersection type?

packages/core/src/components/box/box.md
```diff
@@ -0,0 +1,46 @@
+@# Box
+
+**Box** is a primative, low-level layout component that exposes a subset of the CSS box-model and flexbox APIs as props. It enables layout composition using Blueprint's spacing and sizing tokens.
```
mm-wang:
```suggestion
**Box** is a primitive, low-level layout component that exposes a subset of the CSS box-model and flexbox APIs as props. It enables layout composition using Blueprint's spacing and sizing tokens.
```

packages/docs-app/src/examples/core-examples/boxExamples.tsx
```diff
@@ -0,0 +1,56 @@
...
+        <Box display="flex" flexDirection="column" alignItems="center" gap={2}>
+            <Box as="span">Span</Box>
+            <Box as={Button} intent="primary">
+                Button
+            </Box>
+        </Box>`;
+    return (
+        <CodeExample previewCode={code} sourceCode={code} {...props}>
+            <Box display="flex" flexDirection="column" alignItems="center" gap={2}>
+                <Box as="span">Span</Box>
```
mm-wang: **note**: It feels a bit odd to me to use a string to denote an html element, if you can also accept a custom component like `Button`. I'd expect to be able to define it as `<span>`.

packages/core/src/components/box/_box.scss
```diff
@@ -0,0 +1,480 @@
+/* !
+ * (c) Copyright 2025 Palantir Technologies Inc. All rights reserved.
+ */
+
+@import "../../common/variables";
+
+$sizes: (
```
mm-wang, line 7: **suggestion**: `multiples`? It's not quite the same semantically as how relative-sizes is being used, I feel. Could also change the other.

mm-wang (commented):
This was interesting to read through! I'm interested in the progress. I found some of the interface a little confusing when using similar terms, but otherwise love that we will have this component!

packages/core/src/components/box/box.md
ggdouglas: f7ebb59e471e4760275898c3be1488c81f8c615a

packages/core/src/components/box/boxProps.ts
ggdouglas: It makes sense to me to limit this value to be the same as the other spacing values (margin/padding/etc).

ggdouglas:
Replaced by #7595

PR #6981, opened 2024-09-17
fix(Icon): Support null tagName in Icon component

### LINKED ISSUE #6974: Icon component does not support null value in tagName prop despite documentation and types say it does
#### Environment

- __Package version(s)__: "@blueprintjs/core": "^5.12.0"
- __Operating System__: Arch Linux
- __Browser name and version__: Firefox 129.0.2

#### Code Sandbox

Link to a minimal repro (fork [this code sandbox](https://codesandbox.io/p/sandbox/blueprint-v5-x-sandbox-react-16-wy0ojy)): https://codesandbox.io/p/devbox/magical-dream-5fjckr

#### Actual behavior

Error in console `Uncaught Error: Element type is invalid: expected a string (for built-in components) or a class/function (for composite components) but got: null.`

#### Expected behavior

Icon rendered without `span` wrapper

ggdouglas:
#### Checklist

- [x] Includes tests
- [ ] Update documentation

Fixes a condition where a null element gets passed into `React.createElement` while the icon is loading paths asynchronously. This prevents an error in the Icon component when `null` is passed in as the `tagName`.

### DIFF
--- a/packages/core/src/components/icon/icon.tsx
+++ b/packages/core/src/components/icon/icon.tsx
@@ -171,7 +171,7 @@ export const Icon: IconComponent = React.forwardRef(function <T extends Element>
                 : size === IconSize.LARGE
                   ? Classes.ICON_LARGE
                   : undefined;
-        return React.createElement(tagName!, {
+        return React.createElement(tagName || "span", {
             "aria-hidden": title ? undefined : true,
             ...removeNonHTMLProps(htmlProps),
             className: classNames(
--- a/packages/core/test/icon/iconTests.tsx
+++ b/packages/core/test/icon/iconTests.tsx
@@ -132,7 +132,8 @@ describe("<Icon>", () => {
 
     it("allows specifying the root element as <svg> when tagName={null}", () => {
         const handleClick: React.MouseEventHandler<SVGSVGElement> = () => undefined;
-        mount(<Icon<SVGSVGElement> icon="add" onClick={handleClick} tagName={null} />);
+        const wrapper = mount(<Icon<SVGSVGElement> icon="add" onClick={handleClick} tagName={null} />);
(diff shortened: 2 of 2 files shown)


packages/core/src/components/icon/icon.tsx
```diff
@@ -171,7 +171,7 @@ export const Icon: IconComponent = React.forwardRef(function <T extends Element>
                 : size === IconSize.LARGE
                   ? Classes.ICON_LARGE
                   : undefined;
-        return React.createElement(tagName!, {
```
ggdouglas, line 174: I think it makes sense to ensure this value is always a valid element. The alternative would be to do something like a null check and return early, e.g.
```tsx
if (tagName == null) {
    return null;
}
return React.createElement(tagName, { ... });
```

The former will show a blip of the fallback while the icon paths load:

Screenshot 2024-09-17 at 16 21 42

whereas the later will appear and shift in without any placeholder at all:

Screenshot 2024-09-17 at 16 22 34

jscheiny: Any particular reason to use `||` here over `??`?

ggdouglas: Hm, I suppose the only main difference is that `||` will use `"span"` as the default for _any_ falsy value, including `""`, `0`, and `false`. If the user is consuming this component in TypeScript, there is functionally no difference since `tagName` is either type of`React.JSX.IntrinsicElements` or `null`.

Do we care about this edge case for non-TypeScript consumers?

jscheiny: Yeah okay, that makes sense. I think its fair to keep this for non-TS consumers.

jscheiny (approved):

PR #6982, opened 2024-09-17
style: HTMLElement casting style

bvandrc:
Better to use say, `elem.closest<HTMLInputElement>("input")` than `elem.closest("input") as HTMLInputElement`

### DIFF
--- a/packages/core/src/common/utils/domUtils.ts
+++ b/packages/core/src/common/utils/domUtils.ts
@@ -37,7 +37,7 @@ export function elementIsTextInput(elem: HTMLElement) {
         return false;
     }
 
-    const editable = elem.closest("input, textarea, [contenteditable=true]");
+    const editable = elem.closest<HTMLInputElement>("input, textarea, [contenteditable=true]");
 
     if (editable == null) {
         return false;
@@ -45,14 +45,14 @@ export function elementIsTextInput(elem: HTMLElement) {
 
     // don't let checkboxes, switches, and radio buttons prevent hotkey behavior
     if (editable.tagName.toLowerCase() === "input") {
-        const inputType = (editable as HTMLInputElement).type;
+        const inputType = editable.type;
         if (inputType === "checkbox" || inputType === "radio") {
             return false;
         }
     }
 
     // don't let read-only fields prevent hotkey behavior
-    if ((editable as HTMLInputElement).readOnly) {
+    if (editable.readOnly) {
         return false;
     }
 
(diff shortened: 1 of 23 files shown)


packages/core/src/common/utils/domUtils.ts
```diff
@@ -63,12 +63,8 @@ export function elementIsTextInput(elem: HTMLElement) {
...
  */
 export function getActiveElement(element?: HTMLElement | null, options?: GetRootNodeOptions) {
-    if (element == null) {
-        return document.activeElement;
-    }
-
-    const rootNode = (element.getRootNode(options) ?? document) as DocumentOrShadowRoot & Node;
-    return rootNode.activeElement;
+    const rootNode = (element?.getRootNode(options) ?? document) as DocumentOrShadowRoot & Node;
+    return rootNode.activeElement as HTMLElement | null;
```
ggdouglas: Can we use a double assertion here to specify the type of `rootNode`?
```ts
export function getActiveElement(element?: HTMLElement | null, options?: GetRootNodeOptions): HTMLElement | null {
    const rootNode = (element?.getRootNode(options) ?? document) as unknown as DocumentOrShadowRoot;
    const activeElement = rootNode.activeElement;
    return activeElement instanceof HTMLElement ? activeElement : null;
}
```
I think this makes the uncertainty of the type conversion a bit more explicit/clear.

packages/docs-theme/src/components/documentation.tsx
```diff
@@ -395,18 +395,18 @@ export class Documentation extends React.PureComponent<DocumentationProps, Docum
...
  */
 function getScrolledReference(offset: number, scrollContainer: HTMLElement = document.documentElement) {
-    const headings = Array.from(scrollContainer.querySelectorAll(".docs-title"));
+    const headings = Array.from(scrollContainer.querySelectorAll<HTMLElement>(".docs-title"));
     while (headings.length > 0) {
         // iterating in reverse order (popping from end / bottom of page)
         // so the first element below the threshold is the one we want.
-        const element = headings.pop() as HTMLElement;
+        const element = headings.pop()!;
         if (element.offsetTop < scrollContainer.scrollTop + offset) {
```
ggdouglas: Can we use a null check vs. a non-null assertion here?
```suggestion
        if (element != null && element.offsetTop < scrollContainer.scrollTop + offset) {
```

packages/core/src/common/utils/domUtils.ts
bvandrc: I like that!

packages/docs-theme/src/components/documentation.tsx
bvandrc: Yeah that's much better.

ggdouglas (approved):
Thanks for cleaning these types up! 🙂

PR #6984, opened 2024-09-17
Select: use shouldReturnFocusOnClose

bvandrc:
In `Select`, all of the `previousFocusedElement` code does what `Popover`'s `shouldReturnFocusOnClose` already does. Just use that.


packages/select/src/components/select/select.tsx
```diff
@@ -122,8 +122,6 @@ export class Select<T> extends AbstractPureComponent<SelectProps<T>, SelectState
 
     private queryList: QueryList<T> | null = null;
 
-    private previousFocusedElement: HTMLElement | undefined;
```
bvandrc, line 125: All of this `previousFocusedElement` code does what `Popover`'s `shouldReturnFocusOnClose` already does. Just use that.

PR #6985, opened 2024-09-18
fix breadcrumbs collapsed a11y failure

bvandrc:
collapsed button should be tabbable and have `role="button"`


packages/core/src/components/breadcrumbs/breadcrumbs.tsx
```diff
@@ -97,7 +97,7 @@ export class Breadcrumbs extends AbstractPureComponent<BreadcrumbsProps> {
             <OverflowList
                 collapseFrom={collapseFrom}
                 minVisibleItems={minVisibleItems}
-                tagName="ul"
+                tagName="ol"
```
ggdouglas: Is there a reason we are changing the default element type here from an unordered to an ordered list element? Is it possible to separate this change from the other a11y failure?

packages/core/src/components/breadcrumbs/breadcrumbs.tsx
```diff
@@ -127,7 +127,12 @@ export class Breadcrumbs extends AbstractPureComponent<BreadcrumbsProps> {
                     content={<Menu>{orderedItems.map(this.renderOverflowBreadcrumb)}</Menu>}
                     {...popoverProps}
                 >
-                    <span className={Classes.BREADCRUMBS_COLLAPSED} />
+                    <span
+                        aria-label="collapsed breadcrumbs"
```
ggdouglas, line 136: While adding an `aria-label` is a great accessibility improvement, we should avoid hardcoding the string "collapsed breadcrumbs" directly in the component. This approach doesn't support localization for international users.

Instead, can we introduce a new prop, something like `expandText`, to allow consumers of this component to provide a localized or customized string? We can set the default value of that prop to a string that best describes the action of this button.

```suggestion
                        aria-label={this.props.expandText}
```

packages/core/src/components/breadcrumbs/breadcrumbs.tsx
bvandrc: Breadcrumbs _are_ ordered so they should be tagged as such.

Here is the aria example: https://www.w3.org/WAI/ARIA/apg/patterns/breadcrumb/examples/breadcrumb/

bvandrc: > Is it possible to separate this change

https://github.com/palantir/blueprint/pull/7007

packages/core/src/components/breadcrumbs/breadcrumbs.tsx
bvandrc: Are you down for a _default_ of `"collapsed breadcrumbs"`? Or do you want no `aria-label` at all?

bvandrc: Can set via `overflowButtonProps`: https://github.com/palantir/blueprint/pull/6985/commits/85519823178f115b902d4572a1cbbfbd43658ac8

ggdouglas: > Are you down for a _default_ of `"collapsed breadcrumbs"`? Or do you want no `aria-label` at all?

Having `"collapsed breadcrumbs"` as the default seems fine to me.

ggdouglas (approved):

PR #6986, opened 2024-09-18
refactor: getArrowKeyDirection

bvandrc:
Small refactor that allows usage of `getArrowKeyDirection` in more places.

### DIFF
--- a/packages/core/src/common/utils/keyboardUtils.ts
+++ b/packages/core/src/common/utils/keyboardUtils.ts
@@ -22,24 +22,25 @@ export function isKeyboardClick(event: React.KeyboardEvent<HTMLElement>) {
     return event.key === "Enter" || event.key === " ";
 }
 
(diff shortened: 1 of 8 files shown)


packages/core/src/common/utils/keyboardUtils.ts
```diff
@@ -22,24 +22,25 @@ export function isKeyboardClick(event: React.KeyboardEvent<HTMLElement>) {
     return event.key === "Enter" || event.key === " ";
 }
 
+const ARROW_KEYS = ["ArrowUp" as const, "ArrowDown" as const, "ArrowLeft" as const, "ArrowRight" as const];
```
ggdouglas: Could we write this as:
```suggestion
const ARROW_KEYS = ["ArrowUp", "ArrowDown", "ArrowLeft", "ArrowRight"] as const;
```

bvandrc: Yes my b, good catch

ggdouglas (approved):
Thank you for the change!

PR #6987, opened 2024-09-18
refactor(mergeRefs): allow undefined

bvandrc:
Allow undefined arg so don't have to do ternaries when used.


packages/core/src/common/refs.ts
```diff
@@ -40,7 +40,7 @@ export function setRef<T>(refTarget: React.Ref<T> | undefined, ref: T | null): v
  * Utility for merging refs into one singular callback ref.
  * If using in a functional component, would recomend using `useMemo` to preserve function identity.
  */
-export function mergeRefs<T>(...refs: Array<React.Ref<T>>): React.RefCallback<T> {
+export function mergeRefs<T>(...refs: Array<React.Ref<T> | undefined>): React.RefCallback<T> {
```
ggdouglas, line 43: Can we filter out any falsy refs passed into this function? Also, it might be worth handling the case where only a single ref is passed in. Something like:
```tsx
export function mergeRefs<T>(...refs: Array<React.Ref<T> | undefined>): React.Ref<T> {
    const filteredRefs = refs.filter(Boolean);

    if (filteredRefs.length <= 1) {
        return filteredRefs[0] || null;
    }

    return value => {
        refs.forEach(ref => {
            setRef(ref, value);
        });
    };
}
```

bvandrc: That seems like it makes the current implementation more complicated, and also making it less performant since the type is already checked/filtered inside of `setRef`. `setRef` already allows passing `undefined` so this filtering is not needed.

I'm just not sure what you're hoping to gain via this change?

ggdouglas: In my opinion, if this function receives only falsy values, it should also return `null`, e.g.
```ts
mergeRefs()
mergeRefs(null, null);
// => both should return null
```

That said, if you feel that the current implementation is more straightforward and sufficient, I'm okay with keeping it as is.

ggdouglas (approved):
Thanks for the cleanup!

PR #6991, opened 2024-09-19
fix(Popover): fix autoFocus on target focus

bvandrc:
Fixes autofocus to the `Overlay`.


packages/core/src/components/overlay/overlay.tsx
```diff
@@ -215,18 +215,11 @@ export class Overlay extends AbstractPureComponent<OverlayProps, OverlayState> {
         }
 
         // decorate the child with a few injected props
-        const tabIndex = this.props.enforceFocus || this.props.autoFocus ? 0 : undefined;
```
bvandrc, line 217: This `tabIndex={0}` doesn't do anything because it's not on an interactive element. And `autoFocus` brings focus to `startFocusTrapElement`, not this element.

packages/core/src/components/overlay/overlay.tsx
```diff
@@ -290,7 +283,7 @@ export class Overlay extends AbstractPureComponent<OverlayProps, OverlayState> {
      * the `startFocusTrapElement`), depending on whether the element losing focus is inside the
      * Overlay.
      */
-    private handleStartFocusTrapElementFocus = (e: React.FocusEvent<HTMLDivElement>) => {
```
bvandrc, line 292: can cast `relatedTarget` type using the 2nd param

packages/core/src/components/overlay2/overlay2.tsx
```diff
@@ -408,7 +408,6 @@ export const Overlay2 = React.forwardRef<OverlayInstance, Overlay2Props>((props,
                 // IMPORTANT: only inject our ref if the user didn't specify childRef or childRefs already. Otherwise,
                 // we risk clobbering the user's ref (which we cannot inspect here while cloning/decorating the child).
                 ref: userChildRef === undefined ? localChildRef : undefined,
-                tabIndex: enforceFocus || autoFocus ? 0 : undefined,
```
bvandrc, line 413: This `tabIndex={0}` doesn't do anything because it's not on an interactive element. And `autoFocus` brings focus to `startFocusTrapElement`, not this element.

packages/core/src/components/popover/popover.tsx
```diff
@@ -612,30 +616,27 @@ export class Popover<
...
 
-    private handleTargetBlur = (e: React.FocusEvent<HTMLElement>) => {
+    private handleTargetBlur = (e: React.FocusEvent<HTMLElement, HTMLElement>) => {
         if (this.props.openOnTargetFocus && this.isHoverInteractionKind()) {
             if (e.relatedTarget != null) {
-                // if the next element to receive focus is within the popover, we'll want to leave the
-                // popover open.
-                if (
-                    e.relatedTarget !== this.popoverElement &&
-                    !this.isElementInPopover(e.relatedTarget as HTMLElement)
```
bvandrc, line 557: This is the main fix in this PR. `Overlay2` brings focus to its `startFocusTrapElement` upon `autoFocus`, and this element is **not** _within_ the popover element. It is just _outside_ the popover element, within the `overlay` element. So, use `isElementInOverlay` instead of `isElementInPopover`

bvandrc:
@ggdouglas Made some updates to this PR after finding another bug with this issue.

PR #6999, opened 2024-09-24
Marks Select arrays as readonly

### LINKED ISSUE #4976: Select's types don't accept readonly item arrays

cybae0804:
#### Checklist

- [x] Includes tests
- [x] Update documentation

Previous attempt + reversion https://github.com/palantir/blueprint/pull/5171

Select components don't actually mutate the arrays provided, but because it is not marked as readonly, the consumers sometimes have to cast the prop to not be readonly. This fixes the issue by marking the props correctly as readonly.

### DIFF
--- a/packages/docs-app/src/examples/select-examples/multiSelectExample.tsx
+++ b/packages/docs-app/src/examples/select-examples/multiSelectExample.tsx
@@ -251,7 +251,9 @@ export class MultiSelectExample extends React.PureComponent<ExampleProps, MultiS
         );
     }
 
-    private renderCustomTarget = (selectedItems: Film[]) => <MultiSelectCustomTarget count={selectedItems.length} />;
+    private renderCustomTarget = (selectedItems: readonly Film[]) => (
+        <MultiSelectCustomTarget count={selectedItems.length} />
+    );
 
     private renderTag = (film: Film) => film.title;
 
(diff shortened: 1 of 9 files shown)


bvandrc (dismissed):

invliD (commented):
I don't like that this change forces consumers to deal with readonly arrays. Most code I have seen does not bother declaring arrays readonly, so returning readonly arrays in callbacks would break many such usages.

We're experimenting with a version that uses generics to return the same kind of array you pass in (readonly or not), and if that works out we will go with that approach instead. Otherwise, we'll come back to this.

ggdouglas:
@cybae0804 @invliD Here's an implementation that leverages generics as described: #6999

This accomplishes the same goal, while also having the advantage of being backwards compatible. This will hopefully make it much easier to transition existing code to use readonly arrays.

PR #7002, opened 2024-09-27
Revert: Fix border overlap between Buttons in outlined ButtonGroup

ggdouglas:
This reverts the changes added in #6966 (while retaining additions to the docs). The removal of borders between Buttons in ButtonGroup had the unintentional consequence of regressing styles for ButtonGroups that contain mixed intents or mixed types of buttons.


evansjohnson (approved):
Screenshot 2024-09-27 at 4 08 31 PM

leaving example of observed undesired behavior

thanks for the fix!

PR #7005, opened 2024-10-01
Collapse borders between buttons in outlined ButtonGroup

ggdouglas:
#### Checklist

- [ ] Includes tests
- [x] Update documentation

This change applies a subset of functionality from #6966 (reverted in #7002) in order to collapse borders between Buttons in a ButtonGroup iff the `outlined` prop is applied on ButtonGroup. Conversely, the collapsed border behavior is not present if `outlined` is applied to individual buttons within the ButtonGroup. This allows consumers to either opt-in or opt-out of the collapsed border behavior. It also prevents a regression in styles to the default ButtonGroup where consumers are already using mixed variant/intent buttons within a ButtonGroup.

Example of where the collapsing border behavior **does** apply:
```tsx
<ButtonGroup outlined={true}>
    <Button icon="floppy-disk" intent="success" text="Save" />
    <Button icon="undo" intent="primary" text="Undo" disabled={true} />
</ButtonGroup>
```
collapsed

Examples of where the collapsing border behavior **does not** apply:
```tsx
<ButtonGroup>
    <Button icon="floppy-disk" intent="success" text="Save" outlined={true} />
    <Button icon="undo" intent="primary" text="Undo" disabled={true} outlined={true} />
</ButtonGroup>
```
not-collapsed

ButtonGroups with mixed use buttons (e.g. `minimal` and `outlined`) can also still co-exist if the variants are applied to individual Buttons:
```tsx
<ButtonGroup>
    <Button icon="chevron-left" minimal={true} />
    <Button rightIcon="caret-down" outlined={true}>
        1
    </Button>
    <Button icon="chevron-right" minimal={true} />
</ButtonGroup>
```

Screenshot 2024-10-01 at 15 11 44@2x

#### Split Buttons
Additionally, this allows consumers of Blueprint to specify a split button without the double border between buttons:
```tsx
<ButtonGroup outlined={true}>
    <Button text="Squash and merge" />
    <Popover
        minimal={true}
        position="bottom-left"
        content={...}
        fill={true}
    >
        <Button icon="caret-down" />
    </Popover>
</ButtonGroup>
```

A "split button" example has been added to the ["Usage with popovers"](https://blueprintjs.com/docs/#core/components/button-group.usage-with-popovers) section to demonstrate this:

Screenshot 2024-10-01 at 14 26 31

All possible combinations of `<ButtonGroup>` variants and `<Button>` children variants to avoid regressions in styles.

### DIFF
--- a/packages/core/src/components/button/_button-group.scss
+++ b/packages/core/src/components/button/_button-group.scss
@@ -142,11 +142,20 @@ Styleguide button-group
   }
 
   &.#{$ns}-outlined {
+    > .#{$ns}-popover-target > .#{$ns}-button,
     > .#{$ns}-button {
       @include pt-button-outlined();
     }
 
+    &.#{$ns}-vertical {
+      > .#{$ns}-popover-target:not(:last-child) .#{$ns}-button,
+      > .#{$ns}-button:not(:last-child) {
+        border-bottom: none;
+      }
+    }
+
     &:not(.#{$ns}-vertical) {
+      > .#{$ns}-popover-target:not(:last-child) > .#{$ns}-button,
       > .#{$ns}-button:not(:last-child) {
         border-right: none;
       }
--- a/packages/core/src/components/button/button-group.md
+++ b/packages/core/src/components/button/button-group.md
@@ -1,12 +1,12 @@
 @# Button group
 
-**ButtonGroup** arranges multiple buttons in a horizontal or vertical group.
+__ButtonGroup__ arranges multiple buttons in a horizontal or vertical group.
 
 @reactExample ButtonGroupExample
 
 @## Usage
 
-Most of **ButtonGroup**'s props are also supported by **Button** directly; setting these props on **ButtonGroup** will
+Most of __ButtonGroup__'s props are also supported by __Button__ directly; setting these props on __ButtonGroup__ will
 apply the same value to all buttons in the group. Note that most modifiers, once enabled on the group, cannot be
 overridden on child buttons (due to the cascading nature of CSS).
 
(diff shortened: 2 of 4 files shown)


packages/core/src/components/button/_button-group.scss
```diff
@@ -142,11 +142,20 @@ Styleguide button-group
...
+      > .#{$ns}-button:not(:last-child) {
+        border-bottom: none;
+      }
+    }
+
     &:not(.#{$ns}-vertical) {
+      > .#{$ns}-popover-target:not(:last-child) > .#{$ns}-button,
       > .#{$ns}-button:not(:last-child) {
         border-right: none;
       }
```
ggdouglas, line 161: Changes to this file include a subset of the same changes originally applied in #6966, but scoped only to the `outlined` ButtonGroup.

packages/core/src/components/button/button-group.md
```diff
@@ -1,12 +1,12 @@
...
 
-__ButtonGroup__ arranges multiple buttons in a horizontal or vertical group.
+**ButtonGroup** arranges multiple buttons in a horizontal or vertical group.
 
 @reactExample ButtonGroupExample
 
 @## Usage
 
-Most of __ButtonGroup__'s props are also supported by __Button__ directly; setting these props on __ButtonGroup__ will
+Most of **ButtonGroup**'s props are also supported by **Button** directly; setting these props on **ButtonGroup** will
```
ggdouglas: Formatting changes were added here by prettier on editor save.

packages/core/src/components/button/button-group.md
```diff
@@ -26,32 +26,33 @@ The component also supports all HTML `<div>` props.
 
 @## Usage with popovers
 
-__Button__ elements inside a __ButtonGroup__ can trivially be wrapped with a [__Popover__](#core/components/popover) to
-create complex toolbars.
+**Button** elements inside a **ButtonGroup** can be wrapped with a [**Popover**](#core/components/popover) to create
+complex toolbars or to provide split button functionality, allowing the action of a **Button** to be changed.
```
evansjohnson: I don't think "split button" is the unique thing to call out here in the use with popovers section, maybe "menu button"?

Also not sure about `allowing the action of a **Button** to be changed` - pretty much anything could be done inside the popover so it feels weird to call out a specific case here. I think it would be weird if the popover just contained options that would change the wrapped button's behavior (implying it needs to be clicked again?).

evansjohnson (commented):
LGTM - thanks for the docs updates to go along with this. Will approve when you're ready with the docs changes just had one small comment to consider.

packages/core/src/components/button/button-group.md
ggdouglas: Yeah, fair! I originally modeled this after GitHub's own merge options button, which I've seen as a pretty common dropdown use pattern in other places. Totally fine with omitting this and leaving the docs as is if the example isn't adding value.

ggdouglas: Reverted 77763571d4711d507d4456c23cdcf12c06107296

PR #7007, opened 2024-10-03
feat(Breadcrumbs): use ol instead of ul

bvandrc:
Breadcrumbs _are_ ordered so they should be tagged as such.

Here is the aria example: https://www.w3.org/WAI/ARIA/apg/patterns/breadcrumb/examples/breadcrumb/

It uses `ol`.

### DIFF
--- a/packages/core/src/components/breadcrumbs/_breadcrumbs.scss
+++ b/packages/core/src/components/breadcrumbs/_breadcrumbs.scss
@@ -8,13 +8,13 @@
 Breadcrumbs
 
 Markup:
-<ul class="#{$ns}-breadcrumbs">
+<ol class="#{$ns}-breadcrumbs">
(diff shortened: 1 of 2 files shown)


ggdouglas (approved):
I fundamentally agree with this change. However, I want to explicitly highlight that it could potentially break consumers of this component who style directly using the `ul` element tag. While this scenario should be uncommon, I believe internal elements should not be considered a breaking API change. Nonetheless, we will make an explicit note of the change in the changelog upon release.

bvandrc:
> I fundamentally agree with this change. However, I want to explicitly highlight that it could potentially break consumers of this component who style directly using the `ul` element tag. While this scenario should be uncommon, I believe internal elements should not be considered a breaking API change. Nonetheless, we will make an explicit note of the change in the changelog upon release.

Well said, I did consider that with this change!

PR #7021, opened 2024-10-21
Add isComposing check to EditableText toggleEdit

natsuozawa:
#### Checklist

- [x] Includes tests (no need for update)
- [x] Update documentation (no need for update)

EditableText will check that the user is not composing before sending out onConfirm event.
This is necessary because in some IMEs, such as Japanese, users will press Enter to convert text (ie from hiragana to kanji). In these cases, we don't want EditableText to exit from editing mode.  

https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/isComposing

### DIFF
--- a/packages/core/src/components/editable-text/editableText.tsx
+++ b/packages/core/src/components/editable-text/editableText.tsx
@@ -357,7 +357,7 @@ export class EditableText extends AbstractPureComponent<EditableTextProps, Edita
         }
 
         const hasModifierKey = altKey || ctrlKey || metaKey || shiftKey;
-        if (event.key === "Enter" && !event.isComposing) {
+        if (event.key === "Enter" && !event.nativeEvent.isComposing) {
             // prevent browsers (Edge?) from full screening with alt + enter
             // shift + enter adds a newline by default
             if (altKey || shiftKey) {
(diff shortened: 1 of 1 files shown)


ggdouglas (dismissed):

ggdouglas:
Hey @natsuozawa, thanks for the PR. This looks good to me as far as preserving existing functionality, but I don't have ready access to a Japanese keyboard to test the added functionality here. I'll assume that you have verified these changes locally. Is there any way we can add a test in code as well?

packages/core/src/components/editable-text/editableText.tsx
```diff
@@ -357,7 +357,7 @@ export class EditableText extends AbstractPureComponent<EditableTextProps, Edita
         }
 
         const hasModifierKey = altKey || ctrlKey || metaKey || shiftKey;
-        if (event.key === "Enter") {
+        if (event.key === "Enter" && !event.nativeEvent.isComposing) {
```
invliD, line 360: I think we actually need to do this earlier, before we handle escape. Normally, at least on macOS, pressing escape while composing cancels the _current_ sequence. Both with and without this change, it cancels the entire edit, i.e. everything that was changed will be abandoned, instead of just the current sequence discarded and other changes and editing mode preserved.

natsuozawa: Hey you're right that *maybe* we should not escape when the IME is composing, but in my understanding people don't usually press escape during typing. Personally, I would expect to escape even when composing

PR #7024, opened 2024-10-24
Do not toggle edit mode with escape key during IME composition

natsuozawa:
Following up on #7021 

#### Checklist

- [x] Includes tests
- [x] Update documentation

Do not toggle edit mode with escape key during IME composition


packages/core/src/components/editable-text/editableText.tsx
```diff
@@ -351,7 +351,7 @@ export class EditableText extends AbstractPureComponent<EditableTextProps, Edita
 
     private handleKeyEvent = (event: React.KeyboardEvent<HTMLElement>) => {
         const { altKey, ctrlKey, metaKey, shiftKey } = event;
-        if (event.key === "Escape") {
+        if (event.key === "Escape" && !event.nativeEvent.isComposing) {
```
invliD: This entire handler now does nothing when composing. Instead of having this here and below, add a `if (event.nativeEvent.isComposing) { return; }` at the very top, ideally with a comment explaining that enter and escape have special meanings during composing that we should not override.

invliD (approved):

PR #7025, opened 2024-10-24
fix(KeyComboTag): Better rendering on non-Mac operating systems

### LINKED ISSUE #6211: KeyCombo renders mod as ^ instead of Ctrl on Windows
#### Environment

- __Package version(s)__: 4.20.1
- __Operating System__: macOS Ventura 13.4 (22F66)
- __Browser name and version__: Version 114.0.5735.106 (Official Build) (arm64)

#### Code Sandbox

Link to a minimal repro (fork [this code sandbox](https://codesandbox.io/p/sandbox/blueprint-sandbox-2023-fjo3z4)): 

#### Steps to reproduce

N/A

#### Actual behavior

`^` is rendered instead of `Ctrl` on Windows.

#### Expected behavior

I expected `Ctrl` to be rendered instead of `^`.

#### Possible solution

Check for `mod` instead of `meta` [here](https://github.com/palantir/blueprint/blob/develop/packages/core/src/components/hotkeys/hotkeyParser.ts#L292-L298)?

jscheiny:
#### Checklist

- [x] Includes tests
- [x] Update documentation

In this PR we update the `KeyComboTag` component to render properly in non-Mac environments. Namely, all modifier keys no longer show their icons (which are mac specific). In addition the minimal state has been updated to show a more windows friendly UI. Finally, the arrow keys were broken and not showing their icons, they do now.

I implemented a `KeyComboTagInternal` component for testing purposes. What do we think of this pattern?

#### Screenshot

Previously, this is how the `mod+shift+x` key combo looked like on all platforms. This is still how it looks on Mac. (Minimal state on bottom)
Screenshot 2024-10-24 at 12 26 24 PM

This is how it looks non-Mac platforms now:
Screenshot 2024-10-24 at 12 26 15 PM

### DIFF
--- a/packages/core/src/components/hotkeys/hotkeyParser.ts
+++ b/packages/core/src/components/hotkeys/hotkeyParser.ts
@@ -240,7 +240,7 @@ export const normalizeKeyCombo = (combo: string, platformOverride?: string): str
     });
 };
 
-function isMac(platformOverride?: string) {
+export function isMac(platformOverride?: string) {
     // HACKHACK: see https://github.com/palantir/blueprint/issues/5174
     // eslint-disable-next-line deprecation/deprecation
     const platform = platformOverride ?? (typeof navigator !== "undefined" ? navigator.platform : undefined);
--- a/packages/core/src/components/hotkeys/hotkeyParser.ts
+++ b/packages/core/src/components/hotkeys/hotkeyParser.ts
@@ -50,7 +50,7 @@ export const CONFIG_ALIASES: KeyMap = {
     esc: "escape",
     escape: "escape",
     minus: "-",
-    mod: isMac() ? "meta" : "ctrl",
+    mod: isMac(undefined) ? "meta" : "ctrl",
     option: "alt",
     plus: "+",
     return: "enter",
@@ -232,7 +232,7 @@ export const getKeyCombo = (e: KeyboardEvent): KeyCombo => {
  * Unlike the parseKeyCombo method, this method does NOT convert shifted
  * action keys. So `"@"` will NOT be converted to `["shift", "2"]`).
  */
-export const normalizeKeyCombo = (combo: string, platformOverride?: string): string[] => {
+export const normalizeKeyCombo = (combo: string, platformOverride: string | undefined): string[] => {
     const keys = combo.replace(/\s/g, "").split("+");
     return keys.map(key => {
         const keyName = CONFIG_ALIASES[key] != null ? CONFIG_ALIASES[key] : key;
@@ -240,7 +240,7 @@ export const normalizeKeyCombo = (combo: string, platformOverride?: string): str
     });
 };
 
-export function isMac(platformOverride?: string) {
+export function isMac(platformOverride: string | undefined) {
     // HACKHACK: see https://github.com/palantir/blueprint/issues/5174
     // eslint-disable-next-line deprecation/deprecation
     const platform = platformOverride ?? (typeof navigator !== "undefined" ? navigator.platform : undefined);
--- a/packages/core/src/components/hotkeys/keyComboTag.tsx
+++ b/packages/core/src/components/hotkeys/keyComboTag.tsx
@@ -33,20 +33,20 @@ import {
 import { AbstractPureComponent, Classes, DISPLAYNAME_PREFIX, type Props } from "../../common";
 import { Icon } from "../icon/icon";
 
-import { normalizeKeyCombo } from "./hotkeyParser";
+import { isMac, normalizeKeyCombo } from "./hotkeyParser";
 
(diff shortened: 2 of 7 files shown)


packages/core/src/components/hotkeys/hotkeyParser.ts
```diff
@@ -232,15 +232,15 @@ export const getKeyCombo = (e: KeyboardEvent): KeyCombo => {
...
+export const normalizeKeyCombo = (combo: string, platformOverride: string | undefined): string[] => {
     const keys = combo.replace(/\s/g, "").split("+");
     return keys.map(key => {
         const keyName = CONFIG_ALIASES[key] != null ? CONFIG_ALIASES[key] : key;
         return keyName === "meta" ? (isMac(platformOverride) ? "cmd" : "ctrl") : keyName;
     });
 };
 
-function isMac(platformOverride?: string) {
+export function isMac(platformOverride: string | undefined) {
```
jscheiny, line 243: This function is duplicative with another in the codebase, will be looking into refactoring that in a future change.

packages/core/test/hotkeys/keyComboTagTests.tsx
```diff
@@ -18,11 +18,24 @@ import { render, screen } from "@testing-library/react";
 import { expect } from "chai";
 import * as React from "react";
 
-import { KeyComboTag } from "../../src/components/hotkeys";
+import { KeyComboTagInternal } from "../../src/components/hotkeys/keyComboTag";
 
 describe("KeyCombo", () => {
     it("renders key combo", () => {
-        render(<KeyComboTag combo="cmd+C" />);
+        render(<KeyComboTagInternal combo="cmd+C" platformOverride="Mac" />);
```
ggdouglas, line 25: Are we using this `platformOverride` flag solely for testing purposes? If so, that feels a slight bit icky to me. Is there a way we could potentially mock the return of the `isMac` function instead? This would potentially save us from having to wire this override flag in multiple places.

packages/core/src/components/hotkeys/keyComboTag.tsx
```diff
@@ -94,8 +104,22 @@ export class KeyComboTag extends AbstractPureComponent<KeyComboTagProps> {
         );
     };
 
-    private renderMinimalKey = (key: string, index: number) => {
-        const icon = KEY_ICONS[key];
-        return icon == null ? key : <Icon icon={icon.icon} title={icon.iconTitle} key={`key-${index}`} />;
+    private renderMinimalKey = (key: string, index: number, isLastKey: boolean) => {
+        const icon = this.getKeyIcon(key);
+        if (icon == null) {
+            return isLastKey ? key : <React.Fragment key={`key-${index}`}>{key}&nbsp;+&nbsp;</React.Fragment>;
```
ggdouglas, line 110: Should we add the "+" between tokens regardless of icon/platform? IMO, the minimal version looks a bit better with it on mac. I also notice there's a bit of extra space after the first token if the next token happens to be another icon.

Screenshot 2024-10-25 at 16 43 28@2x

Screenshot 2024-10-25 at 16 50 48@2x

ggdouglas: Following that, I wonder if we could insert the plus signs an alternative way, perhaps either via a CSS `::before` pseudo-element or by inserting them after construction of the array, e.g.
```tsx
const normalizedKeys = normalizeKeyCombo(combo, platformOverride);
const upperKeys = normalizedKeys.map(key => (key.length === 1 ? key.toUpperCase() : key));
const keys = upperKeys.map(this.renderKey);
const minimalKeys = upperKeys
    .map(this.renderMinimalKey)
    .flatMap((value, index, array) =>
        array.length - 1 !== index ? [value, <span key={index}>+</span>] : value,
    );
return <span className={classNames(Classes.KEY_COMBO, className)}>{minimal ? minimalKeys : keys}</span>;
```

packages/core/test/hotkeys/keyComboTagTests.tsx
jscheiny: Agreed its ugly. However, I took a look around and wasn't able to find a good way to mock `isMac`, but let me know if I missed something here. I did make the argument required on these functions so that we wouldn't forget to pass it and miss a spot.

packages/core/src/components/hotkeys/keyComboTag.tsx
jscheiny: I wanted to mirror the ways these different platforms display key combos. Namely on Windows/Linux, will show the plus signs and on Macs they will not. Agreed on the extra space, will fix that.

In terms of putting in the plus signs, I attempted to use the pseudo elements earlier but getting the styles to work on those turned out to be unwieldy. Its not clear to me that the code above is particularly better than what we have currently.

ggdouglas (approved):

PR #7027, opened 2024-10-25
Rewrite Button docs with more discrete examples

ggdouglas:
#### Checklist

- [ ] Includes tests
- [x] Update documentation

**Preview:** Here is the Button documentation page [before](https://blueprintjs.com/docs/#core/components/buttons) and [after](https://output.circle-artifacts.com/output/job/a70533f7-5415-4948-9dcd-0443b3fa5bf5/artifacts/0/packages/docs-app/dist/index.html#core/components/buttons)

#### Proposed Changes

The intent of this example is to serve as a template for rewriting more of our documentation to enhance comprehension and utility towards developers/consumers of Blueprint. Here are some basic principles:

- **Modular Examples:** Instead of centralizing all properties in a single ["kitchen sink"](https://idioms.thefreedictionary.com/everything+but+the+kitchen+sink) example, this proposal aims to break down these monolithic examples into more individual and discrete examples.
- **Concise Code Examples:** Code examples should ideally be concise and illustrate just one or two central concepts per example. This makes it easier for developers to use these examples as direct references or to copy them directly into their code.
- **Recipe Documentation:** The individual examples provide an opportunity to offer ["recipe"](https://diataxis.fr/how-to-guides) documentation that explains how to compose components together to solve more abstract problems. For instance, how to use `AnchorButton` and `Tooltip` together to address opening tooltips around a disabled button.
- **Expositional Writing:** More discrete examples allow us to include more exposition, offering suggestions on how certain props should be used by developers. This also provides room to explain the design system itself, discussing the "why" behind the design of some components.
- **Direct Linking:** Splitting examples allows us to link developers directly to an example of a certain prop or concept. This is currently impossible with the multifaceted/multi-state examples.

#### Screenshot

Frame 1


packages/core/src/components/button/buttons.md
```diff
@@ -1,55 +1,110 @@
...
+
+@## Ellipsized text
+
+The `ellipsizeText` prop allows text within a button to be truncated with an ellipsis if it exceeds the available space. This is useful for cases where the button needs to remain compact without overflowing, especially when the text content is dynamic or potentially lengthy.
+
+@reactCodeExample ButtonEllipsizeTextExample
+
+@## Icons with text
+
+Buttons can include icons alongside text for extra context or visual cues. Icons can be added to either the left or right side of text/children with the `icon` and `rightIcon` props respectively. These icons can either be specified as string identifiers (e.g. `"arrow-right"`), dynamically-loaded [`<Icon>` components](https://blueprintjs.com/docs/#core/components/icon), [static icon components](#core/components/icon.static-components) (e.g. `<ArrowRight />`), or any custom JSX element.
```
jscheiny:
```suggestion
Buttons can include icons alongside text for extra context or visual cues. Icons can be added to either the left or right side of text/children with the `icon` and `rightIcon` props respectively. These icons can either be specified as string identifiers (e.g. `"arrow-right"`), dynamically-loaded [`<Icon>` components](#core/components/icon), [static icon components](#core/components/icon.static-components) (e.g. `<ArrowRight />`), or any custom JSX element.
```

packages/core/src/components/button/buttons.md
```diff
@@ -1,55 +1,110 @@
...
+
+Icon buttons display only an icon without any accompanying text. Icon buttons are used when an action can be clearly conveyed through a visual symbol, making the interface more compact and visually appealing. They are ideal for toolbars or areas with limited space.
+
+@reactCodeExample ButtonIconExample
+
+@## Button states
+
+Buttons have different states to show their interaction status. The `active`, `disabled`, and `loading` props provide visual feedback to help users understand available actions and when to wait.
+
+-   **Active**: Indicates the button is currently active.
```
jscheiny: This kind of tautological. Instead perhaps something like:
```suggestion
-   **Active**: Indicates the button is currently being pressed or interacted with.
```

packages/docs-app/src/examples/core-examples/buttonExamples.tsx
```diff
@@ -0,0 +1,175 @@
...
+            <Button text="Loading..." loading={true} />
+        </CodeExample>
+    );
+};
+
+export const ButtonAnchorButtonExample: React.FC<ExampleProps> = props => {
+    const code = `<AnchorButton href="#core/components/buttons.anchorbutton" rightIcon="share" text="Link" />`;
+    return (
+        <CodeExample code={code} {...props}>
+            <AnchorButton href="#core/components/buttons.anchorbutton" rightIcon="share" text="Link" />
```
jscheiny: I think we should include a full url here, something like 
```suggestion
            <AnchorButton href="https://blueprintjs.com/" rightIcon="share" text="Link" />
```

As this will help make it clear for new users what this means.

packages/docs-app/src/examples/core-examples/buttonExamples.tsx
```diff
@@ -0,0 +1,175 @@
...
+ * (c) Copyright 2024 Palantir Technologies Inc. All rights reserved.
+ */
+
+import * as React from "react";
+
+import { AnchorButton, Button, Icon, Tooltip } from "@blueprintjs/core";
+import { CodeExample, type ExampleProps } from "@blueprintjs/docs-theme";
+
+export const ButtonBasicExample: React.FC<ExampleProps> = props => {
+    const code = `<Button text="Click Me" />`;
```
jscheiny, line 11: This seems okay for now, but would love to see if we could find a way to keep these in sync.

ggdouglas: Yep! Will continue to iterate here on an optimal way to define these

jscheiny (approved):

PR #7028, opened 2024-10-26
Fix IME bug in QueryList

AaronJRubin:
#### Checklist

- [ ] Includes tests
- [ ] Update documentation

This PR fixes a bug in the query list component (as used in Select) where IME enter and arrow key keyboard events are not ignored, as they should be, causing IME confirmation to result in selecting a value and other sadness.

It's a little weird that we already had the `isEnterKeyPressed` logic mostly in place to catch IME events, but didn't have the actual "is composing" check needed to make it actually do what it was supposed to do. Maybe just an oversight?

#### Before

https://github.com/user-attachments/assets/301433eb-af58-4739-bd6d-99fffeb9c7f0

#### After

https://github.com/user-attachments/assets/f340609b-2fee-47a2-a15b-3a958028c99c

### DIFF
--- a/packages/select/changelog/@unreleased/pr-7028.v2.yml
+++ b/packages/select/changelog/@unreleased/pr-7028.v2.yml
@@ -1,0 +1,5 @@
+type: fix
+fix:
+  description: Fix buggy behavior with IME input in QueryList component
+  links:
+  - https://github.com/palantir/blueprint/pull/7028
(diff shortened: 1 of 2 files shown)


invliD (approved):
I think there's still a separate bug here where pressing Escape while composing closes the Select dropdown, though I think that needs to be fixed in the Overlay component.

ggdouglas (approved):

PR #7033, opened 2024-10-30
[DateRangePicker] Allow date picker popover to be closed by esc key press

ggdouglas:
#### Checklist

- [x] Includes tests
- [ ] Update documentation

- Closes the `<DatePicker>` popover on `<DateRangeInput>` when the `"Escape"` key is pressed.
- Aligns `<DateRangeInput>` popover behavior with the existing Escape-to-close functionality already implemented in [DateInput](https://blueprintjs.com/docs/#datetime2/date-input3)
- Ensures users can close the DatePicker without requiring a mouse click, enhancing accessibility and user experience.

#### Screenshot

| Before | After    |
| ------  | ------  |
| before | after |

### DIFF
--- a/packages/datetime/src/components/date-range-input/dateRangeInput.tsx
+++ b/packages/datetime/src/components/date-range-input/dateRangeInput.tsx
@@ -636,13 +636,8 @@ export class DateRangeInput extends AbstractPureComponent<DateRangeInputProps, D
         const { selectedStart, selectedEnd } = this.state;
 
         if (isEscapeKeyPressed) {
-            const { isStartInputFocused, isEndInputFocused } = this.state;
-            if (isStartInputFocused) {
-                this.startInputElement?.blur();
-            }
-            if (isEndInputFocused) {
-                this.endInputElement?.blur();
-            }
+            this.startInputElement?.blur();
+            this.endInputElement?.blur();
             this.setState({ isOpen: false, isStartInputFocused: false, isEndInputFocused: false });
             return;
         }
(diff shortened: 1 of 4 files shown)


packages/datetime/src/components/date-range-input/dateRangeInput.tsx
```diff
@@ -630,10 +630,23 @@ export class DateRangeInput extends AbstractPureComponent<DateRangeInputProps, D
...
         const { selectedStart, selectedEnd } = this.state;
 
+        if (isEscapeKeyPressed) {
+            const { isStartInputFocused, isEndInputFocused } = this.state;
+            if (isStartInputFocused) {
+                this.startInputElement?.blur();
+            }
+            if (isEndInputFocused) {
+                this.endInputElement?.blur();
+            }
```
jscheiny: Any reason we can't just optimistically blur both components without doing the state check? I think this would simplify things and also make sure that in case of weird bugs where the state gets out of sync we still handle this case properly.

packages/datetime2/src/components/date-range-input3/dateRangeInput3.tsx
```diff
@@ -484,10 +484,23 @@ export class DateRangeInput3 extends DateFnsLocalizedComponent<DateRangeInput3Pr
...
         const { selectedStart, selectedEnd } = this.state;
 
+        if (isEscapeKeyPressed) {
+            const { isStartInputFocused, isEndInputFocused } = this.state;
+            if (isStartInputFocused) {
+                this.startInputElement?.blur();
+            }
+            if (isEndInputFocused) {
+                this.endInputElement?.blur();
+            }
```
jscheiny: Same question here

packages/datetime/src/components/date-range-input/dateRangeInput.tsx
ggdouglas: Fair point! Added in b4490dcaf83017a49f2e4b23dcb386e0856a4c7f

jscheiny (approved):

PR #7036, opened 2024-10-30
Use generics for Select arrays

ggdouglas:
#### Checklist

- [x] Includes tests
- [ ] Update documentation

This change is an alternative approach to #6999, allowing  arrays passed to Select (as well as other components that take arrays) to be marked as readonly. It also preserves backwards compatibility with previous types such that the new prop types will not break existing consumers of these components where they are already passing mutable arrays. This will make it easier to transition existing implementations to use readonly arrays.

### DIFF
--- a/packages/docs-app/src/examples/select-examples/multiSelectExample.tsx
+++ b/packages/docs-app/src/examples/select-examples/multiSelectExample.tsx
@@ -291,14 +291,14 @@ export class MultiSelectExample extends React.PureComponent<ExampleProps, MultiS
 
     private selectFilms(filmsToSelect: readonly Film[]) {
         this.setState(({ createdItems, films, items }) => {
-            let nextCreatedItems = createdItems.slice();
-            let nextFilms = films.slice();
-            let nextItems = items.slice();
+            let nextCreatedItems = createdItems;
+            let nextFilms = films;
+            let nextItems = items;
 
             filmsToSelect.forEach(film => {
                 const results = maybeAddCreatedFilmToArrays(nextItems, nextCreatedItems, film);
-                nextItems = results.items.slice();
-                nextCreatedItems = results.createdItems.slice();
+                nextItems = results.items;
+                nextCreatedItems = results.createdItems;
                 // Avoid re-creating an item that is already selected (the "Create
                 // Item" option will be shown even if it matches an already selected
                 // item).
--- a/packages/docs-app/src/examples/select-examples/multiSelectExample.tsx
+++ b/packages/docs-app/src/examples/select-examples/multiSelectExample.tsx
@@ -40,13 +40,13 @@ const INTENTS = [Intent.NONE, Intent.PRIMARY, Intent.SUCCESS, Intent.DANGER, Int
 
 export interface MultiSelectExampleState {
     allowCreate: boolean;
-    createdItems: Film[];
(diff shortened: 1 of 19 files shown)


packages/docs-app/src/examples/select-examples/multiSelectExample.tsx
```diff
@@ -40,13 +40,13 @@ const INTENTS = [Intent.NONE, Intent.PRIMARY, Intent.SUCCESS, Intent.DANGER, Int
 
 export interface MultiSelectExampleState {
     allowCreate: boolean;
-    createdItems: Film[];
+    createdItems: readonly Film[];
```
ggdouglas, line 43: The modification of the examples here in this PR isn't strictly necessary, though it helps to serve as a practical example of how to migrate existing Select/MultiSelect components to use readonly arrays. It's also worth noting that the examples here still compile fine if they are not switched to use readonly arrays, demonstrating the backwards compatibility of these type changes.

All the docs/test changes are contained within: 19cdbeec25bdc51d8382f611920f4f6660bd5347

packages/docs-app/src/examples/select-examples/multiSelectExample.tsx
```diff
@@ -124,7 +124,7 @@ export class MultiSelectExample extends React.PureComponent<ExampleProps, MultiS
 
         return (
             <Example options={this.renderOptions()} {...this.props}>
-                <MultiSelect<Film>
```
ggdouglas, line 127: The omission of the specified `Film` generic parameter to the component is necessary here, otherwise we'll get a type error:
```
'readonly Film[]' is 'readonly' and cannot be assigned to the mutable type 'Film[]'
```
Another way to solve this is to explicitly define the second type parameter, e.g.
```ts
<MultiSelect<Film, readonly Film[]>
```

packages/docs-app/src/examples/select-examples/multiSelectExample.tsx
```diff
@@ -287,16 +289,16 @@ export class MultiSelectExample extends React.PureComponent<ExampleProps, MultiS
...
             let nextCreatedItems = createdItems.slice();
             let nextFilms = films.slice();
             let nextItems = items.slice();
 
             filmsToSelect.forEach(film => {
                 const results = maybeAddCreatedFilmToArrays(nextItems, nextCreatedItems, film);
-                nextItems = results.items;
-                nextCreatedItems = results.createdItems;
+                nextItems = results.items.slice();
+                nextCreatedItems = results.createdItems.slice();
```
invliD: Do we have to slice these? We're not actually mutating either of them below.

invliD (dismissed):
Nice!

jscheiny (dismissed):

packages/docs-app/src/examples/select-examples/multiSelectExample.tsx
ggdouglas: Ah, yeah, good call. I originally added the slice here since, otherwise, we'd be assigning the readonly arrays from results (`results.items` and `results.createdItems`) to mutable variables, which results in a type error. We can remove the slice altogether and make everything consistently readonly. 5b8c385361d358ba90e18980189eb5ddc683f6b1

invliD (approved):

PR #7037, opened 2024-10-31
feat(Table2) Row focus mode

jscheiny:
#### Checklist

- [x] Includes tests
- [x] Update documentation

This PR introduces a new focus API for the table2 component (the existing table component does not take advantage of this API).

Previously, focus was controlled via `enableFocusedCell`, `focusedCell`, and `onFocusedCell` props. These have now been updated to `focusMode`, `focusedRegion`, and `onFocusedRegion` which supersede the previous props. If values are passed to these props they take precedence of the values passed to the old ones. `FocusMode` is a enum of row and column. Passing undefined to the `focusMode` prop disables focus like `enableFocusedCell={false}` would previously.


packages/table/src/common/internal/focusedCellUtils.ts
```diff
@@ -15,116 +15,228 @@
...
+ * Returns the inferred focus mode from the table props. This prefers the new focus mode API, falling back to the
+ * deprecated enableFocusedCell API if that is not provided.
+ */
+export function getFocusModeFromProps(props: TableProps): FocusMode | undefined {
+    // eslint-disable-next-line deprecation/deprecation
+    const { enableFocusedCell, focusMode } = props;
+    return focusMode ?? getFocusModeFromEnabled(enableFocusedCell);
+}
+
+function getFocusModeFromEnabled(enableFocusedCell: boolean | undefined): FocusMode | undefined {
```
ggdouglas: **nit:** Can we use a default function parameter to infer the boolean type here?
 ```suggestion
function getFocusModeFromEnabled(enableFocusedCell = false): FocusMode | undefined {
```

File: packages/table/src/table.tsx [hunk 2]
```diff
@@ -1498,13 +1506,13 @@ export class Table extends AbstractComponent<TableProps, TableState, TableSnapsh
...
 
         const { viewportRect } = this.state;
-        this.setState({ viewportRect: nextViewportRect });
 
         const didViewportChange =
             (viewportRect != null && !viewportRect.equals(nextViewportRect)) ||
             (viewportRect == null && nextViewportRect != null);
 
         if (didViewportChange) {
+            this.setState({ viewportRect: nextViewportRect });
```
ggdouglas: Is this change related to the other focus mode changes?

packages/table/src/table.tsx
```diff
@@ -1565,4 +1573,39 @@ export class Table extends AbstractComponent<TableProps, TableState, TableSnapsh
     private getRowHeaderWidth = (): number => {
         return this.props.enableRowHeader ? this.rowHeaderElement?.clientWidth ?? Grid.MIN_ROW_HEADER_WIDTH : 0;
     };
+
+    private getFocusMode(): FocusMode | undefined {
+        return this.props.enableFocusedCell ? FocusMode.CELL : undefined;
+    }
+}
+
+function getFocusedRegionFromCell(cell: FocusedCellCoordinates | undefined): FocusedCell | undefined {
```
ggdouglas: Should this be returning a `FocusedRegion`?

packages/table/src/common/cellTypes.ts
```diff
@@ -22,3 +22,20 @@ export interface CellCoordinates {
...
+}
+
+export interface FocusedCell extends FocusedCellCoordinates {
+    type: FocusMode.CELL;
+}
+
+export interface FocusedRow {
+    type: FocusMode.ROW;
+    row: number;
+    focusSelectionIndex: number;
```
ggdouglas, line 38: I'm a bit confused as to what the purpose of `focusSelectionIndex` is. Can you explain? Under which conditions is it non-zero?

packages/table/src/common/internal/focusedCellUtils.ts
```diff
@@ -151,3 +263,16 @@ function getExpandedRegionIndices(
...
 }
+
+export function areFocusedRegionsEqual(left: FocusedRegion, right: FocusedRegion) {
+    if (left.type !== right.type) {
+        return false;
+    }
+
+    switch (left.type) {
+        case FocusMode.CELL:
+            return left.row === right.row && left.col === getFocusedColumn(right);
```
ggdouglas: **nit:** Can we use some type guards here to avoid the call to `getFocusedColumn`? e.g.

```ts
function isFocusedCell(region: FocusedRegion): region is FocusedCell {
    return region.type === FocusMode.CELL;
}

function isFocusedRow(region: FocusedRegion): region is FocusedRow {
    return region.type === FocusMode.ROW;
}

export function areFocusedRegionsEqual(left: FocusedRegion, right: FocusedRegion) {    
    if (isFocusedCell(left) && isFocusedCell(right)) {
        return left.row === right.row && left.col === right.col;
    } else if (isFocusedRow(left) && isFocusedRow(right)) {
        return left.row === right.row;
    } else {
        return false;
    }
}
```

packages/table/src/table.tsx
jscheiny: `FocusedCell` is a subtype of `FocusedRegion`. Will change to `getFocusedCellFromCoordinates` to be more clear.

packages/table/src/common/cellTypes.ts
jscheiny: I had a lot of trouble understanding it too. Best I can tell it has to do with maintaining order when tabbing the focused cell through multiple selections. Given that we don't support that behavior for rows, it should be irrelevant to row focus mode. 🤞

packages/table/src/table.tsx
jscheiny: Oops forgot to revert this. This was from when I was trying to fix tests.

ggdouglas (approved):
Nice work!

PR #7038, opened 2024-10-31
Add arrows-arc

ryannono:
#### Changes proposed in this pull request:

Add `arrows-arc` icon

#### Screenshot

image

PR #7039, opened 2024-10-31
Add rotation icons

ryannono:

Add `rotate-clockwise`, and `rotate-counter-clockwise` icons

#### Screenshot

rotate


packages/icons/icons.json
```diff
@@ -4520,5 +4520,26 @@
         "tags": "currency, money",
         "group": "miscellaneous",
         "codepoint": 62342
+    },
+    {
+        "displayName": "Arrows: arc",
+        "iconName": "arrows-arc",
```
ggdouglas: Is the `"arrows-arc"` icon from #7038 added in this PR intentionally?

packages/icons/icons.json
```diff
@@ -4527,5 +4527,19 @@
         "tags": "rotate cursor, curved, arrows",
         "group": "interface",
         "codepoint": 62343
+    },
+    {
+        "displayName": "Square: rotate clockwise",
```
jscheiny: Why the "square:" before these? I feel like just "Rotate {counter}clockwise" would be sufficient.

jscheiny (dismissed):
Approving to unblock, but consider changing display names

packages/icons/icons.json
ggdouglas: I feel like there is a reasonable argument for future-proofing the naming here in case we want to support a more generic "rotate" icon in the future, something like this:

Screenshot 2024-11-04 at 15 22 36@2x

The more descriptive (and less subjective) the naming conventions that we apply, the less likely that we'll run into naming collisions down the line as we continue to add more icons over time. To illustrate this, here's what a search for "rotate" turns up in other popular icon libraries:
- https://fontawesome.com/search?q=rotate&o=r
- https://mui.com/material-ui/material-icons/?query=rotate
- https://feathericons.com/?query=rotate
- https://icons.getbootstrap.com/?q=rotate
- https://www.streamlinehq.com/?search=rotate
- https://lucide.dev/icons/?focus=&search=rotate
- https://react-icons.github.io/react-icons/search/#q=rotate
- https://icon-sets.iconify.design/?query=rotate

ggdouglas: Reverted naming changes to ""Rotate {counter}clockwise" in 79484d984a52ea6df157523dec55868a6c87c559

PR #7048, opened 2024-11-07
fix(OverlayToaster): Don't dismiss excess toasts when updating a toast while maxToasts is set

jscheiny:
#### Checklist

- [x] Includes tests
- [ ] Update documentation (N/A)

Something of a niche case, but noticed it while iterating on a different toaster issue and wanted to quickly fix it. This issue occurs if there are a number of toasts on the screen equal to the `maxToasts` prop. Previously, if you updated one of those toasts, a toast would be dismissed unnecessarily (which could be bad if its meant to be an indefinitely present toast). Now, if a toast is updated, we don't do the max toasts check since the number of toasts hasn't changed.

This PR also contains some changes that will help with further refactors to the overlay toaster component.

### DIFF
--- a/packages/core/src/components/toast/overlayToaster.tsx
+++ b/packages/core/src/components/toast/overlayToaster.tsx
@@ -146,21 +146,35 @@ export class OverlayToaster extends AbstractPureComponent<OverlayToasterProps, O
     };
 
     public show(props: ToastProps, key?: string) {
+        const options = this.createToastOptions(props, key);
+        const updatedExistingToast = this.maybeUpdateExistingToast(options, key);
+        if (updatedExistingToast) {
+            return options.key;
+        }
+
         if (this.props.maxToasts) {
             // check if active number of toasts are at the maxToasts limit
             this.dismissIfAtLimit();
         }
-        const options = this.createToastOptions(props, key);
+
+        this.updateToastsInState(toasts => [options, ...toasts]);
+        return options.key;
+    }
+
+    private maybeUpdateExistingToast(options: ToastOptions, key: string | undefined) {
+        if (key == null || this.isNewToastKey(key)) {
+            return false;
+        }
+
+        this.updateToastsInState(toasts => toasts.map(t => (t.key === key ? options : t)));
+        return true;
+    }
+
+    private updateToastsInState(getNewToasts: (toasts: ToastOptions[]) => ToastOptions[]) {
         this.setState(prevState => {
-            const toasts =
-                key === undefined || this.isNewToastKey(key)
-                    ? // prepend a new toast
-                      [options, ...prevState.toasts]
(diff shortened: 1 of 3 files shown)


packages/core/src/components/toast/overlayToaster.tsx
```diff
@@ -146,21 +146,35 @@ export class OverlayToaster extends AbstractPureComponent<OverlayToasterProps, O
     };
 
     public show(props: ToastProps, key?: string) {
+        const options = this.createToastOptions(props, key);
+        const updatedExistingToast = this.maybeUpdateExistingToast(options, key);
```
ggdouglas: **nit:** can we name this variable to something like `isUpdatingExistingToast` to help clarify that it's a boolean?

packages/core/src/components/toast/overlayToaster.tsx
```diff
@@ -146,21 +146,35 @@ export class OverlayToaster extends AbstractPureComponent<OverlayToasterProps, O
...
+    private maybeUpdateExistingToast(options: ToastOptions, key: string | undefined) {
+        if (key == null || this.isNewToastKey(key)) {
+            return false;
+        }
+
+        this.updateToastsInState(toasts => toasts.map(t => (t.key === key ? options : t)));
+        return true;
+    }
+
+    private updateToastsInState(getNewToasts: (toasts: ToastOptions[]) => ToastOptions[]) {
```
ggdouglas, line 173: I've noticed that the current abstraction with this function taking another function as an argument is making the code a bit harder for me to follow. It seems a bit premature to abstract it in this way.

Would it be possible to simplify by pulling the check for updating an existing toast to the top and using it directly in the conditional for dismissing? e.g.
```ts
public show(props: ToastProps, key?: string) {
    const options = this.createToastOptions(props, key);
    const isUpdatingExistingToast = key != null && !this.isNewToastKey(key);

    if (this.props.maxToasts && !isUpdatingExistingToast) {
        // check if active number of toasts are at the maxToasts limit
        this.dismissIfAtLimit();
    }

    this.setState(prevState => {
        const toasts = isUpdatingExistingToast
            ? // update a specific toast
              prevState.toasts.map(t => (t.key === key ? options : t))
            : // prepend a new toast
              [options, ...prevState.toasts];

        return { toasts, toastRefs: this.getToastRefs(toasts) };
    });
    return options.key;
}
```

jscheiny: I agree that its a bit premature in the context of this PR, but with my FLUP change for queuing, it becomes more needed as we're doing these state updates in a couple of places where its not so easy to refactor. If you're down, lets leave this as is and maybe in the context of the other PR we can reevaluate?

ggdouglas (approved):

PR #7049, opened 2024-11-07
fix(OverlayToaster): Fix toasts being cut off if show() called too quickly

### LINKED ISSUE #7034: Showing multiple toasts in quick succession flows over top of screen
#### Environment

- __Package version(s)__: @blueprintjs/core 5.13.1
- __Operating System__: Windows Server 2016
- __Browser name and version__: Chrome 130.0.6723.70

#### Code Sandbox

https://codesandbox.io/p/sandbox/jovial-galois-x8727d

Relevant code is in `src/App.tsx` - if you increase the `waitInterval` to one second then the toasts animate smoothly, but if we don't wait (`waitInterval = 0`) then they get stuck.

#### Steps to reproduce

1. create an `<OverlayToaster>`
2. send multiple toast messages to it in quick succession

#### Actual behavior

Animation seems to stop early and some toast messages flow over the top of the screen.

#### Expected behavior

All toast should finish animating correctly into the correct position.

#### Possible solution

This seems to be some timing issue: if we insert an arbitrary wait between sending toasts then they animate correctly.

jscheiny:
#### Checklist

- [x] Includes tests
- [ ] Update documentation (N/A)

This issue appears to be a react css transition group issue, but we can mitigate it on our side by effectively debouncing the `show()` call. Now when a user calls `show()` if there's been a toasted shown in the last 50ms, the new toast will be added to a queue and only shown after that time has elapsed. Any other `show()` calls that happen within that window will also be added to the queue and only one new toast will be shown after each 50ms increment.
#### Screenshots

Without this change we get this ugly behavior:
Screenshot 2024-11-07 at 1 29 05 PM

With the change it animates in properly:
Screenshot 2024-11-07 at 1 29 16 PM

### DIFF
--- a/packages/core/src/components/toast/overlayToaster.tsx
+++ b/packages/core/src/components/toast/overlayToaster.tsx
@@ -49,7 +49,8 @@ interface OverlayToasterQueueState {
     toasts: ToastOptions[];
 }
 
-const QUEUE_TIMEOUT_MS = 50;
+export const OVERLAY_TOASTER_DELAY_MS = 50;
+
 /**
  * OverlayToaster component.
  *
@@ -174,7 +175,7 @@ export class OverlayToaster extends AbstractPureComponent<OverlayToasterProps, O
         } else {
             // If we have not recently shown a toast, we can immediately show the given toast
             this.immediatelyShowToast(options);
-            this.maybeStartQueueTimeout();
+            this.startQueueTimeout();
         }
 
         return options.key;
@@ -209,21 +210,18 @@ export class OverlayToaster extends AbstractPureComponent<OverlayToasterProps, O
         this.updateToastsInState(toasts => [options, ...toasts]);
     }
 
-    private maybeStartQueueTimeout() {
-        if (this.queue.isRunning) {
-            return;
-        }
-
+    private startQueueTimeout() {
         this.queue.isRunning = true;
-        this.queue.cancel = this.setTimeout(this.handleQueueTimeout, QUEUE_TIMEOUT_MS);
+        this.queue.cancel = this.setTimeout(this.handleQueueTimeout, OVERLAY_TOASTER_DELAY_MS);
     }
 
     private handleQueueTimeout = () => {
         const nextToast = this.queue.toasts.shift();
-        this.queue.isRunning = false;
         if (nextToast != null) {
             this.immediatelyShowToast(nextToast);
-            this.maybeStartQueueTimeout();
+            this.startQueueTimeout();
+        } else {
+            this.queue.isRunning = false;
         }
     };
 
--- a/packages/core/src/components/toast/overlayToaster.tsx
+++ b/packages/core/src/components/toast/overlayToaster.tsx
@@ -43,6 +43,13 @@ export interface OverlayToasterState {
 
 export type OverlayToasterCreateOptions = DOMMountOptions<OverlayToasterProps>;
 
+interface OverlayToasterQueueState {
+    cancel: (() => void) | undefined;
+    isRunning: boolean;
+    toasts: ToastOptions[];
+}
+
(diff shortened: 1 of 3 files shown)


packages/core/src/components/toast/overlayToaster.tsx
```diff
@@ -43,6 +43,13 @@ export interface OverlayToasterState {
 
 export type OverlayToasterCreateOptions = DOMMountOptions<OverlayToasterProps>;
 
+interface OverlayToasterQueueState {
+    cancel: (() => void) | undefined;
+    isRunning: boolean;
+    toasts: ToastOptions[];
+}
+
+const QUEUE_TIMEOUT_MS = 50;
```
ggdouglas: What determines that the timeout should be 50ms? Is this something that we should potentially make configurable?

packages/core/test/toast/overlayToasterTests.tsx
```diff
@@ -108,13 +108,31 @@ describe("OverlayToaster", () => {
                 });
             });
 
-            it("multiple show()s renders them all", () => {
+            it("multiple show()s renders them all", async () => {
                 toaster.show({ message: "one" });
                 toaster.show({ message: "two" });
                 toaster.show({ message: "six" });
+                await delay(150);
```
ggdouglas: Is there any way we can tie these hardcoded values to the value of `QUEUE_TIMEOUT_MS`? That might make them a bit less ambiguous in the context of these tests. Perhaps we could leave a comment or even export reference the constant itself in code:
```ts
await delay(3 * QUEUE_TIMEOUT_MS);
```

Also, should we be using [timer mocks](https://sinonjs.org/releases/latest/fake-timers/) in our tests?

packages/core/src/components/toast/overlayToaster.tsx
```diff
@@ -152,24 +167,66 @@ export class OverlayToaster extends AbstractPureComponent<OverlayToasterProps, O
...
 
-        this.updateToastsInState(toasts => toasts.map(t => (t.key === key ? options : t)));
-        return true;
+        this.queue.isRunning = true;
+        this.queue.cancel = this.setTimeout(this.handleQueueTimeout, QUEUE_TIMEOUT_MS);
     }
 
+    private handleQueueTimeout = () => {
+        const nextToast = this.queue.toasts.shift();
+        this.queue.isRunning = false;
```
ggdouglas: Do we have to set `isRunning` to `false` during each execution of the timeout? Would it make sense to check that the queue is empty first? Something like:
```ts
if (this.queue.toasts.length === 0) {
    this.queue.isRunning = false;
}
```

packages/core/src/components/toast/overlayToaster.tsx
jscheiny: This was based on experimentation on my part, it was a value that seemed to never exhibit issues with this problem (+ some buffer to be safe) and not have too much apparent effect on the UX of the component. I don't see a particular reason right now to make this configurable.

packages/core/test/toast/overlayToasterTests.tsx
jscheiny: Yeah good call on the delay, will export that. I attempted to user the timer mocks and I wasn't able to get them to work, will give them another shot.

jscheiny: Figured out what I messed up and switched to fake timers 👍

packages/core/src/components/toast/overlayToaster.tsx
jscheiny: Yeah good call. I ended up implementing this slightly differently but with the same intent.

ggdouglas (approved):

PR #7050, opened 2024-11-07
feat(SegmentedControl): allow for passing stricter type for options

bvandrc:
#### Checklist

- [N/A] Includes tests
- [X] Update documentation

Allow for stricter type on `SegmentedControl` than just `string`, which reduces the need for type casting in many cases (such as `onChange` call).

For example, you can still have just
```tsx
<SegmentedControl
```

as before, for general string Options, or you can have 
```tsx
<SegmentedControl<OptionsEnum>
```
to make sure that all passed option values, and the `onChange`, have the `OptionsEnum` type.

Improved type casting abilities, and mainly, the reduction of having to type cast.

#### Screenshot

No visual changes.


bvandrc:
commit list was funky, replacement: https://github.com/palantir/blueprint/pull/7051

PR #7051, opened 2024-11-07
feat(SegmentedControl): allow for passing stricter type for options

bvandrc:
#### Checklist

- [N/A] Includes tests
- [X] Update documentation

Allow for stricter type on `SegmentedControl` than just `string`, which reduces the need for type casting in many cases (such as `onChange` call).

For example, you can still have just
```tsx
<SegmentedControl
```

as before, for general string Options, or you can have 
```tsx
<SegmentedControl<OptionsEnum>
```
to make sure that all passed option values, and the `onChange`, have the `OptionsEnum` type.

Improved type casting abilities, and mainly, the reduction of having to type cast.

#### Screenshot

No visual changes.


packages/core/src/components/segmented-control/segmentedControl.tsx
```diff
@@ -83,133 +83,151 @@ export interface SegmentedControlProps
     small?: boolean;
 }
 
+// This allows the ability to pass a more strict type for `options`/`onValueChange`
+// i.e. <SegmentedControl<Intent> />
+interface ReactFCWithGeneric extends React.FC<SegmentedControlProps> {
```
jscheiny: Nit this is a specific interface for `SegmentedControl` so lets name it something in accordance with that.
```suggestion
interface GenericSegmentedControl extends React.FC<SegmentedControlProps> {
```

packages/docs-app/src/examples/core-examples/common/sizeSelect.tsx
```diff
@@ -27,25 +27,21 @@ export interface SizeSelectProps {
...
-                ]}
-                onValueChange={handleChange}
-                value={size}
-            />
-        </FormGroup>
-    );
-};
+export const SizeSelect: React.FC<SizeSelectProps> = ({ label, size, optionLabels, onChange }) => (
+    <FormGroup label={label}>
+        <SegmentedControl<Size>
```
jscheiny: Out of curiosity, in these types of cases if you omit `<Size>` is the generic type parameter properly inferred from the `options` array? I'm guessing not because there's no `as const`.

packages/docs-app/src/examples/core-examples/common/sizeSelect.tsx
```diff
@@ -27,25 +27,21 @@ export interface SizeSelectProps {
...
+            fill={true}
+            small={true}
+            options={[
+                { label: optionLabels[0], value: "small" },
+                { label: optionLabels[1], value: "regular" },
+                { label: optionLabels[2], value: "large" },
+            ]}
+            onValueChange={onChange}
```
jscheiny: Another point of curiosity (since we ran into this issue with another generic component recently), if you provide an arrow function here, does the type of the argument get properly inferred?

packages/core/src/components/segmented-control/segmentedControl.tsx
bvandrc: ✔️

packages/docs-app/src/examples/core-examples/common/sizeSelect.tsx
bvandrc: >  I'm guessing not because there's no as const.

Correct, it can't be inferred from the `options` in that case. But it _is_ inferred from the type of the `value` prop.

packages/docs-app/src/examples/core-examples/common/sizeSelect.tsx
bvandrc: Yes.

If you specify a type in the generic, it will make sure the `onChange` arg has the proper type`.

If you don't specify a generic, the arrow function argument is just `string`

packages/core/src/components/segmented-control/segmentedControl.tsx
bvandercar-vt: Also just as an FYI, I created this type to match the fact that all other components in this repo specify the type `React.FC`. But really you don't need to manually add a type at all, you could just have 

```tsx
export const SegmentedControl = React.forwardRef(
    <T extends string>(props: SegmentedControlProps<T>, ref: React.ForwardedRef<HTMLDivElement>) => {
```

bvandercar-vt:
@ggdouglas @jscheiny Can you take a look at this one?

packages/core/src/components/segmented-control/segmentedControl.tsx
bvandrc: Renamed and commented to match the other cases.

bvandrc:
@ggdouglas @jscheiny -- rebased this one.

bvandrc:
Going to replace with a newer PR.

PR #7052, opened 2024-11-07
fix(SegmentedControl): option label defaults to value

bvandrc:
#### Checklist

- [X] Includes tests
- [X] Update documentation

Per `OptionProps` JSDocs:

```tsx
/** Label text for this option. If omitted, `value` is used as the label. */
    label?: string;
```

However, this "`value` as default" was not occurring for SegmentedControl option props. This fixes that.

Fixes the bug so that implementation matches the doc

#### Screenshot

No visible change


bvandrc:
@jscheiny @ggdouglas I rebased this bug fix PR, it's ready.

ggdouglas (approved):
This change seems reasonable to me. Thanks for the catch!

PR #7060, opened 2024-11-12
fix(Tag): Add keyboard accessibility

### LINKED ISSUE #4782: Tag onClick does not fire with spacebar or enter press when focused

jscheiny:

In this PR we pull out the common accessibility functionality from the button components to be shared with the tag component. This enables the tag component to be keyboard accessible when `interactive={true}`

### DIFF
--- a/packages/core/src/accessibility/useInteractiveAttributes.ts
+++ b/packages/core/src/accessibility/useInteractiveAttributes.ts
@@ -23,8 +23,9 @@ export function useInteractiveAttributes<E extends HTMLElement>(
     interactive: boolean,
     props: InteractiveComponentProps,
     ref: React.Ref<E>,
+    defaultTabIndex?: number,
 ): [active: boolean, interactiveProps: InteractiveAttributes<E>] {
-    const { active, onClick, onFocus, onKeyDown, onKeyUp, onBlur, tabIndex } = props;
+    const { active, onClick, onFocus, onKeyDown, onKeyUp, onBlur, tabIndex = defaultTabIndex } = props;
     // the current key being pressed
     const [currentKeyPressed, setCurrentKeyPressed] = React.useState<string | undefined>();
     // whether the button is in "active" state
--- a/packages/core/src/accessibility/useInteractiveAttributes.ts
+++ b/packages/core/src/accessibility/useInteractiveAttributes.ts
@@ -1,0 +1,87 @@
+/* !
+ * (c) Copyright 2024 Palantir Technologies Inc. All rights reserved.
+ */
+
+import * as React from "react";
+
+import { mergeRefs, Utils } from "../common";
+
+type InteractiveHTMLAttributes<E extends HTMLElement> = Pick<
+    React.HTMLAttributes<E>,
+    "onBlur" | "onClick" | "onFocus" | "onKeyDown" | "onKeyUp" | "tabIndex"
+>;
+
+interface InteractiveComponentProps extends InteractiveHTMLAttributes<HTMLElement> {
+    active?: boolean | undefined;
+}
+
+interface InteractiveAttributes<E extends HTMLElement> extends InteractiveHTMLAttributes<E> {
+    ref: React.Ref<E>;
+}
+
+export function useInteractiveAttributes<E extends HTMLElement>(
+    interactive: boolean,
+    props: InteractiveComponentProps,
+    ref: React.Ref<E>,
+): [active: boolean, interactiveProps: InteractiveAttributes<E>] {
+    const { active, onClick, onFocus, onKeyDown, onKeyUp, onBlur, tabIndex = 0 } = props;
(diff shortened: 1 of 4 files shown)


packages/core/src/accessibility/useInteractiveAttributes.ts
```diff
@@ -0,0 +1,88 @@
...
+
+interface InteractiveComponentProps extends InteractiveHTMLAttributes<HTMLElement> {
+    active?: boolean | undefined;
+}
+
+interface InteractiveAttributes<E extends HTMLElement> extends InteractiveHTMLAttributes<E> {
+    ref: React.Ref<E>;
+}
+
+export function useInteractiveAttributes<E extends HTMLElement>(
```
ggdouglas, line 22: **nit:** can we add tests for this hook with `@testing-library/react-hooks` in a follow-up PR?

packages/core/src/components/button/buttons.tsx
```diff
@@ -73,69 +72,17 @@ AnchorButton.displayName = `${DISPLAYNAME_PREFIX}.AnchorButton`;
  */
 function useSharedButtonAttributes<E extends HTMLAnchorElement | HTMLButtonElement>(
     props: E extends HTMLAnchorElement ? AnchorButtonProps : ButtonProps,
-    ref: React.Ref<E>,
+    passedRef: React.Ref<E>,
```
ggdouglas: Why are we renaming this?

packages/core/src/accessibility/useInteractiveAttributes.ts
```diff
@@ -0,0 +1,88 @@
...
+    const resolvedActive = interactive && (active || isActive);
+
+    return [
+        resolvedActive,
+        {
+            onBlur: handleBlur,
+            onClick: interactive ? onClick : undefined,
+            onFocus: interactive ? onFocus : undefined,
+            onKeyDown: handleKeyDown,
+            onKeyUp: handleKeyUp,
```
ggdouglas, line 83: Would this function the same if we were to conditionally render the inclusion of all the handlers based on the `interactive` flag? e.g.

```ts
const handlers = {
    onBlur: handleBlur,
    onClick,
    onFocus,
    onKeyDown: handleKeyDown,
    onKeyUp: handleKeyUp,
};

return [
    resolvedActive,
    {
        ...(interactive ? handlers : {}),
        ref: mergeRefs(elementRef, ref),
        tabIndex: interactive ? tabIndex : -1,
    },
];
```

packages/core/src/accessibility/useInteractiveAttributes.ts
```diff
@@ -0,0 +1,88 @@
...
+    props: InteractiveComponentProps,
+    ref: React.Ref<E>,
+    defaultTabIndex?: number,
+): [active: boolean, interactiveProps: InteractiveAttributes<E>] {
+    const { active, onClick, onFocus, onKeyDown, onKeyUp, onBlur, tabIndex = defaultTabIndex } = props;
+    // the current key being pressed
+    const [currentKeyPressed, setCurrentKeyPressed] = React.useState<string | undefined>();
+    // whether the button is in "active" state
+    const [isActive, setIsActive] = React.useState(false);
+    // our local ref for the interactive element, merged with the consumer's own ref (if supplied) in this hook's return value
```
ggdouglas: **nit:** since `ref` is not an optional argument within this hook, can we change the wording of this comment?
```suggestion
    // our local ref for the interactive element, merged with the consumer's own ref in this hook's return value
```

packages/core/src/accessibility/useInteractiveAttributes.ts
jscheiny: Yeah, sgtm

packages/core/src/components/button/buttons.tsx
jscheiny: Oops, left over from a previous iteration, no longer necessary.

packages/core/src/accessibility/useInteractiveAttributes.ts
jscheiny: Very likely I think, but I wanted to quite safe to avoid any unintentional side effects so I wanted to match the button behavior as close as possible.

ggdouglas (approved):

PR #7069, opened 2024-11-14
fix(Tag) No more accidentally passing active as html prop

jscheiny:
The tag saga continues, this PR prevents us from accidentally passing `active` as a prop to the tag `span`

### DIFF
--- a/packages/core/src/components/tag/tag.tsx
+++ b/packages/core/src/components/tag/tag.tsx
@@ -73,6 +73,7 @@ export interface TagProps
  */
 export const Tag: React.FC<TagProps> = React.forwardRef((props, ref) => {
     const {
+        active: _active,
         children,
         className,
         fill,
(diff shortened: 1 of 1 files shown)


packages/core/src/components/tag/tag.tsx
```diff
@@ -73,6 +73,7 @@ export interface TagProps
  */
 export const Tag: React.FC<TagProps> = React.forwardRef((props, ref) => {
     const {
+        active: _active,
```
ggdouglas: Would wrapping `htmlProps` in `removeNonHTMLProps` achieve the same result?
```ts
{...removeNonHTMLProps(htmlProps)}
```

https://github.com/palantir/blueprint/blob/18e55c67918f2f16a4d806c9965d00d1a150b2b0/packages/core/src/common/props.ts#L130-L131

jscheiny: I think so, will do that too to be safe

ggdouglas (approved):

PR #7070, opened 2024-11-15
Add RTL support for checkbox/radio controls

### LINKED ISSUE #7053: Checkboxes and radio buttons work porly with RTL
#### Environment

- __Package version(s)__:  v5
- __Operating System__:  Mac and Windows
- __Browser name and version__:  Island, chrome, all tested

#### Steps to reproduce

1. Create a list using checkboxes or radio buttons
2. Set dir="rtl"

#### Actual behavior

image

The labels get written over the checkboxes

If AlignIndicator is set to right it looks normal, but this is not ideal for apps with multiple locales

#### Expected behavior

I expect the checkboxes to either stay in place like for ltr, or for alignment to automatically change from left to right.

#### Possible solution

Either automatically toggle alignment as well based on dir or change the overlap.

ggdouglas:
#### Checklist

- [ ] Includes tests
- [ ] Update documentation

Checkbox/Radio (and all other controls) as well as components that use them, like ControlCard.

#### Screenshot

**With `dir="rtl"`**

| Before | After    |
| ------  | ------  |
| before | after |

**With `dir="rtl"` and `alignIndicator="right"`**

| Before | After    |
| ------  | ------  |
| before-2 | after-2 |

### DIFF
--- a/packages/core/src/components/forms/_controls.scss
+++ b/packages/core/src/components/forms/_controls.scss
@@ -112,6 +112,16 @@ $control-indicator-spacing: $pt-grid-size !default;
       margin-right: -$padding;
     }
   }
+
+  // special handling for left-aligned indicators in RTL
+  [dir="rtl"] &.#{$ns}-align-left {
+    padding-inline-start: 0;
+    padding-left: $padding;
+
+    .#{$ns}-control-indicator {
+      margin-left: -$padding;
+    }
+  }
 }
 
 .#{$ns}-control {
@@ -206,6 +216,11 @@ $control-indicator-spacing: $pt-grid-size !default;
     margin-top: 1px;
   }
 
+  [dir="rtl"] &.#{$ns}-align-left .#{$ns}-control-indicator {
+    float: left;
+    margin-top: 1px;
+  }
+
   &.#{$ns}-large {
     @include indicator-position($control-indicator-size-large);
     // larger text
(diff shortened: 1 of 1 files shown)


jscheiny:
What happens if you manually provide `alignIndicator="left"` while in RTL mode?

ggdouglas:
> What happens if you manually provide `alignIndicator="left"` while in RTL mode?

Here's a comparison of a checkbox with all the different `alignIndicator` values before and after this change:

| Before | After |
| ------ | ------ |
| before | after |

As you can see, setting `alignIndicator="left"` in RTL mode is essentially a no-op. I think this is acceptable behavior for the following reasons:

1. It keeps this change small and targeted at fixing the crux of the [original issue](https://github.com/palantir/blueprint/issues/7053).
2. It doesn't worsen the existing experience (it arguably functions better).
3. Adding logic to ensure the indicator is always on the left, regardless of text direction, would likely overcomplicate the already complex control styling logic.

There's more we could do to clean up these control styles and make them more RTL compatible. However, I’d argue that should be part of a breaking change to `alignIndicator`. I suggest we rename the prop language to `"start/end"` instead of `"left/right"`. This terminology is more consistent with logical properties and better handles bidirectional text.

ggdouglas:
@jscheiny added special handling for left aligned controls in RTL pages 88ded4939eb1d36a63096d0aac51755e553eba13

Here's an overview of the behavior between LTR/RTL

| LTR | RTL    |
| ------  | ------  |
| ltr | rtl |

In this instance, the indicator will always align properly "left" or "right" if it is explicitly set by the consumer (regardless of text directionality). However, if alignment is not specified (`undefined`, likely the majority of cases) the indicator will align at the start based on text direction.

jscheiny (approved):

PR #7073, opened 2024-11-15
fix(Tag): Mark tag as interactive when onClick is set

jscheiny:
Ensures that when onClick is passed, the interactive prop is set so we get the best possible accessibility story.

### DIFF
--- a/packages/core/src/components/tag/tag.tsx
+++ b/packages/core/src/components/tag/tag.tsx
@@ -53,6 +53,16 @@ export interface TagProps
     htmlTitle?: string;
 
     /**
+     * Whether the tag should visually respond to user interactions. If set to `true`, hovering over the
+     * tag will change its color and mouse cursor.
+     *
+     * Tags will be marked as interactive automatically if an onClick handler is provided and this prop is not.
+     *
+     * @default false
+     */
+    interactive?: boolean;
+
+    /**
      * Name of a Blueprint UI icon (or an icon element) to render on the left side of the tag,
      * before the child nodes.
      */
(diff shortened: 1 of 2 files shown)


packages/core/src/components/tag/tag.tsx
```diff
@@ -99,8 +99,9 @@ export const Tag: React.FC<TagProps> = React.forwardRef((props, ref) => {
     } = props;
 
     const isRemovable = Utils.isFunction(onRemove);
+    const isInterative = interactive || htmlProps.onClick != null;
```
ericjeney: It might be useful to allow someone to still explicitly set `interactive: false` even when they provide an `onClick` handler, in case that's behavior they're already relying on for some reason.

To do that, should we change this to `interactive ?? htmlProps.onClick != null` and remove the default `= false` on line 89 above?

jscheiny: Yeah good call

packages/core/test/tag/tagTests.tsx
```diff
@@ -57,6 +57,20 @@ describe("<Tag>", () => {
         assert.isTrue(handleRemove.calledOnce);
     });
 
+    it("should be interactive when onClick is provided", () => {
```
ericjeney, line 60: Thanks for adding tests!

ericjeney (dismissed):
Can you update the comment about the default value in the docs as well? Otherwise LGTM.

packages/core/src/components/tag/tag.tsx
evansjohnson: should this `interactive ?? htmlProps.onClick != null` defaulting behavior happen in `useInteractiveAttributes`? right now I think every consumer of this hook would need to do something similar

and maybe it should be a `disabled` option instead? if a component is using this hook I think we should infer it's interactive based on being provided an `onClick`, then `disabled` can be offered as a feature so a consumer doesn't have to do something like conditionally provide `onClick`

it seems like we ended up with `interactive` as an option to this hook since the `Tag` originally had it, but now we're walking back on `interactive` needing to be set to get the interactive attributes

PR #7074, opened 2024-11-17
Switch from @import to @use

### LINKED ISSUE #7031: @import rule is now deprecated.  Migrate to @use and @forward
Sass 1.80.0 has recently [deprecated](https://sass-lang.com/documentation/breaking-changes/import/) the `@import` rule, causing warnings to appear during compile.  Now might be a good time to migrate blueprint's scss to `@use` and `@forward` rules, which have been supported since v1.23.0 (2019).  Support for `@import` will be removed entirely in 2 years.  

No doubt migration is non-trivial, but I believe the changes necessary are:

1. For any `@import` occurring for the purpose of importing variables, mixins, or functions, use `@use` instead.  Multiple files may need to be imported using `@use`, since Sass only includes variables that were defined in the file directly imported.  The variables/functions will be available via a namespace, although it is possible to import them as global for an even easier migration.
2. For any `@import` occurring for the purpose of including the contents of another file (e.g. as seen in `core/src/blueprint.scss`), use `@forward` instead.

They provide [a migration tool](https://github.com/sass/migrator), although I have not tried it so I cannot speak on it.

justinbhopper:
#### Fixes #7031

This PR allows consumers to import Blueprint SCSS using `@use` and `@forward`.  Without this change, consumers must continue to use the old `@import` rule, which has recently been [deprecated](https://sass-lang.com/documentation/breaking-changes/import/).

#### Checklist

- [ ] Includes tests
- [ ] Update documentation

Converts all scss files from the now [deprecated](https://sass-lang.com/documentation/breaking-changes/import/) usage of `@import` to `@use` and `@forward`.  

The uses of `@import` still remaining are only those importing plain css, which SASS [interprets as the native CSS import rule](https://sass-lang.com/documentation/at-rules/import/#plain-css-imports).  These types of imports are only for the Blueprint apps (docs-app, landing-app, and table-dev-app) and follow the recommended guideline on how to import Blueprint's exported css.  

Documentation still recommends using the legacy `@import` rule, but I'll be happy to update it to recommend `@use` or `@forward` for any snippets illustrating scss importing.


justinbhopper:
@ggdouglas Just checking if there's anything further I need to do before this can be reviewed?

ggdouglas:
Hey @justinbhopper, these changes are looking great—thanks so much for taking this on! Before we move forward with merging/releasing, I’d like to [clarify](https://github.com/palantir/blueprint/issues/7031#issuecomment-2540016977) where the warnings we’re addressing are originating. This will help ensure we’re covering everything appropriately. Appreciate your efforts on this!

justinbhopper:
@ggdouglas Following up on this -- please see my response in #7031.  Thank you!

ggdouglas (approved):
Looks great—thanks for taking the initiative on this! 🙂

PR #7078, opened 2024-11-18
Update Stylelint to 16.10.0

MariaSolOs:

In an effort of keeping dev dependencies up to date, fixing the setup here to use the latest Stylelint version.

N/A? I guess that if CI is happy we should be okay 😸

### DIFF
--- a/packages/core/src/common/_mixins.scss
+++ b/packages/core/src/common/_mixins.scss
@@ -38,11 +38,8 @@ $pt-dark-intent-text-colors: (
 }
 
 @mixin position-all($position, $value) {
-  bottom: $value;
-  left: $value;
+  inset: $value $value $value $value;
   position: $position;
-  right: $value;
-  top: $value;
 }
 
 @mixin base-typography() {
--- a/packages/core/src/common/_mixins.scss
+++ b/packages/core/src/common/_mixins.scss
@@ -38,7 +38,7 @@ $pt-dark-intent-text-colors: (
 }
 
 @mixin position-all($position, $value) {
-  inset: $value $value $value $value;
+  inset: $value;
   position: $position;
 }
 
--- a/packages/table-dev-app/src/index.scss
+++ b/packages/table-dev-app/src/index.scss
@@ -21,7 +21,7 @@ body {
 
 /* stylelint-disable-next-line selector-max-id */
 #page-content {
-  inset: $pt-navbar-height 0 0 0;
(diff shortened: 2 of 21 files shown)


packages/core/src/common/_mixins.scss
```diff
@@ -38,11 +38,8 @@ $pt-dark-intent-text-colors: (
 }
 
 @mixin position-all($position, $value) {
-  bottom: $value;
-  left: $value;
+  inset: $value $value $value $value;
```
ggdouglas: **nit:** can we write this as:
```suggestion
  inset: $value;
```

packages/table-dev-app/src/index.scss
```diff
@@ -24,11 +21,8 @@ body {
 
 /* stylelint-disable-next-line selector-max-id */
 #page-content {
-  bottom: 0;
-  left: 0;
+  inset: $pt-navbar-height 0 0 0;
```
ggdouglas: **nit:** can we write this as:
```suggestion
  inset: $pt-navbar-height 0 0;
```

packages/table/src/headers/_headers.scss
```diff
@@ -294,10 +294,8 @@ $dark-selectable-header-menu-selected-hover-background: menu-background(
   &.#{$ns}-editable-text {
     &::before {
       border-radius: 0;
-      bottom: -$cell-padding-vertical;
-      left: -$cell-padding-horizontal + $cell-border-width - 2px;
-      right: -$cell-padding-horizontal + $cell-border-width - 1px;
-      top: -$cell-padding-vertical + $cell-border-width - 2px;
+      inset: -$cell-padding-vertical + $cell-border-width - 2px -$cell-padding-horizontal + $cell-border-width -
+        1px -$cell-padding-vertical -$cell-padding-horizontal + $cell-border-width - 2px;
```
ggdouglas: Is the value of this `inset` property equivalent to the above?

Also, the formatted result here is a bit harder to read vs. the original. Is there any way we can break this back out into multiple lines?

MariaSolOs: We can keep the original and just disable the rule that prefers a single `inset` property.

ggdouglas (approved):
Thanks for tackling this! 🙂

MariaSolOs:
> Thanks for tackling this! 🙂

Thank you for the review and welcoming the change!

PR #7080, opened 2024-11-19
feat(DateRangeInput3) Add keyboard accessibility

jscheiny:

This PR adds improved keyboard accessibility to the `DateRangeInput3` component. In particular we align with other date range picker so that when one of the input fields is focused, the user can then use their arrow keys to navigate the calendar and make selections. This properly respects min and max dates and allow single selection only and properly navigates the calendar view based on user selections.

#### Screenshot

https://github.com/user-attachments/assets/6f09a85f-25d3-403f-9d07-16a46750c7ee

### DIFF
--- a/packages/datetime2/src/components/date-range-input3/dateRangeInput3.tsx
+++ b/packages/datetime2/src/components/date-range-input3/dateRangeInput3.tsx
@@ -554,6 +554,7 @@ export class DateRangeInput3 extends DateFnsLocalizedComponent<DateRangeInput3Pr
     };
 
     private handleInputArrowKeyDown = (e: React.KeyboardEvent<HTMLInputElement>, boundary: Boundary) => {
+        const { minDate, maxDate } = this.props;
         const { isOpen } = this.state;
         const inputElement = boundary === Boundary.START ? this.startInputElement : this.endInputElement;
 
@@ -561,29 +562,34 @@ export class DateRangeInput3 extends DateFnsLocalizedComponent<DateRangeInput3Pr
             return;
         }
 
-        // We've commited to moving the selection, prevent the default arrow key interactions
-        e.preventDefault();
-
-        const newDate =
+        const shiftedDate =
             this.getNextDateForArrowKeyNavigation(e.key, boundary) ??
             this.getDefaultDateForArrowKeyNavigation(e.key, boundary);
 
+        if (shiftedDate == null) {
+            return;
+        }
+
+        // We've commited to moving the selection, prevent the default arrow key interactions
+        e.preventDefault();
+
+        const clampedDate = clampDate(shiftedDate, minDate, maxDate);
         const { keys } = this.getStateKeysAndValuesForBoundary(boundary);
         const nextState: Partial<DateRangeInput3State> = {
-            [keys.inputString]: this.formatDate(newDate),
+            [keys.inputString]: this.formatDate(clampedDate),
             shouldSelectAfterUpdate: true,
         };
 
         if (!this.isControlled()) {
-            nextState[keys.selectedValue] = newDate;
+            nextState[keys.selectedValue] = clampedDate;
         }
 
-        this.props.onChange?.(this.getDateRangeForCallback(newDate, boundary));
+        this.props.onChange?.(this.getDateRangeForCallback(clampedDate, boundary));
         this.setState(nextState);
     };
 
     private getNextDateForArrowKeyNavigation(arrowKey: string, boundary: Boundary) {
-        const { allowSingleDayRange, maxDate, minDate } = this.props;
+        const { allowSingleDayRange } = this.props;
         const [selectedStart, selectedEnd] = this.getSelectedRange();
         const initialDate = boundary === Boundary.START ? selectedStart : selectedEnd;
         if (initialDate == null) {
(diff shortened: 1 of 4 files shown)


packages/datetime2/src/components/date-range-input3/dateRangeInput3.tsx
```diff
@@ -545,6 +553,64 @@ export class DateRangeInput3 extends DateFnsLocalizedComponent<DateRangeInput3Pr
...
+        }
+
+        const relativeDate = shiftDateByArrowKey(initialDate, arrowKey);
+
+        // Ensure that we don't move onto a single day range selection if that is disallowed
+        const adjustedStart =
+            selectedStart == null || allowSingleDayRange ? selectedStart : shiftDateByDays(selectedStart, 1);
+        const adjustedEnd = selectedEnd == null || allowSingleDayRange ? selectedEnd : shiftDateByDays(selectedEnd, -1);
+
+        return boundary === Boundary.START
```
ggdouglas, line 612: There seems to be some unexpected behavior here with `allowSingleDayRange={true}` set.
```tsx
  <DateRangeInput3 allowSingleDayRange={true} />
``` 

When I select the dates by clicking on on the picker, the inputs allow overlap on the same day:
Screenshot 2024-11-19 at 19 24 24

However, if I try to do the same with the arrow key selection, I get an "Overlapping dates" error.
Screenshot 2024-11-19 at 19 25 07

jscheiny: This should be fixed now

packages/datetime2/src/components/date-range-input3/dateRangeInput3.tsx
```diff
@@ -545,6 +559,79 @@ export class DateRangeInput3 extends DateFnsLocalizedComponent<DateRangeInput3Pr
...
+        const otherBoundary = boundary === Boundary.START ? selectedEnd : selectedStart;
+
+        if (otherBoundary == null) {
+            return getTodayAtMidnight();
+        }
+
+        const isForwardArrowKey = arrowKey === "ArrowRight" || arrowKey === "ArrowDown";
+        // If the arrow key direction is in the same direction as the boundary, then moving that way will not create an
+        // overlapping date range
+        if (isForwardArrowKey === (boundary === Boundary.END)) {
```
ggdouglas, line 628: The equivalence of a boolean to a boolean here reads a bit weird. Can we change this to something like:
```ts
const isForwardArrowKey = arrowKey === "ArrowRight" || arrowKey === "ArrowDown";
const isEndBoundary = boundary === Boundary.END;
// If the arrow key direction is in the same direction as the boundary, then moving that way will not create an
// overlapping date range
if (isForwardArrowKey && isEndBoundary) {
    return shiftDateByArrowKey(otherBoundary, arrowKey);
}
```

packages/datetime2/src/components/date-range-input3/dateRangeInput3.tsx
```diff
@@ -545,6 +559,79 @@ export class DateRangeInput3 extends DateFnsLocalizedComponent<DateRangeInput3Pr
...
+        if (initialDate == null) {
+            return undefined;
+        }
+
+        const relativeDate = shiftDateByArrowKey(initialDate, arrowKey);
+
+        // Ensure that we don't move onto a single day range selection if that is disallowed
+        const adjustedStart =
+            selectedStart == null || allowSingleDayRange ? selectedStart : shiftDateByDays(selectedStart, 1);
+        const adjustedEnd = selectedEnd == null || allowSingleDayRange ? selectedEnd : shiftDateByDays(selectedEnd, -1);
```
ggdouglas, line 610: Looks like the behavior now prevents arrow-key selection on the start input when the end input value is already set. Here's the steps I took:

1. Given a range input with a value set for the end date
2. Move focus to the start input
3. Try to arrow key select a date
4. Expected a new date to be selected based on arrow key direction, but no behavior change

https://github.com/user-attachments/assets/3469bb64-1078-458f-8c71-9639fd952208

packages/datetime2/src/components/date-range-input3/dateRangeInput3.tsx
jscheiny: We still need equality here for this to work properly, `isForwardArrowKey && isEndBoundary` is not equivalent to `isForwardArrowKey === isEndBoundary`. I thought about doing something like:

```ts
const arrowKeyDirection = arrowKey === "ArrowRight" || arrowKey === "ArrowDown" ? "forward" : "backward";
const boundaryDirection = boundary === Boundary.END ? "forward" : "backward";
if (arrowKeyDirection === boundaryDirection) {
```

But this is just a lot of extra stuff for the same thing and minimal readability improvement. I'll pull `isEndBoundary` out I like that, but I'm not sure how to make it much more readable beyond that.

packages/datetime2/src/components/date-range-input3/dateRangeInput3.tsx
jscheiny: Hmmm, I'm not able to repro this, seems to work fine for me:

https://github.com/user-attachments/assets/f8b176d7-ce30-4219-bd2a-0f1d38ab6431

ggdouglas (approved):

PR #7081, opened 2024-11-19
Split Prettier from ESLint configuration

MariaSolOs:
#### Checklist

- [x] Update documentation

This PR removes the Prettier plugins from ESLint and instead configures the editor settings and other CI configuration to run Prettier directly.

#### Motivation
- ESLint has deprecated its formatting rules: https://eslint.org/blog/2023/10/deprecating-formatting-rules/
- Performance: Prettier recommends not using the linting integrations and instead running the formatting CLI directly (https://prettier.io/docs/en/integrating-with-linters#notes).
- [Stylelint 15 has also deprecated its formatting rules](https://stylelint.io/migration-guide/to-15/#deprecated-stylistic-rules) (and v16 removed them), and so this change also allows formatting SCSS with Prettier.

The editor experience. How does the experience of format-on-save feel when using [the Prettier VS Code extension](https://marketplace.visualstudio.com/items?itemName=esbenp.prettier-vscode)?

I suggest reviewing each commit separately (the last one is the chunky one that includes the formatting changes of running Prettier in SCSS files).


packages/eslint-config/eslint-builtin-rules.json
```diff
@@ -1,50 +1,19 @@
 {
-    "array-bracket-newline": "off",
```
MariaSolOs, line 2: Note that the rules removed here were [already being disabled](https://github.com/prettier/eslint-config-prettier/blob/5687e7de69ac132f053b4bc5474d5963aa3bf6d6/index.js) by `eslint-config-prettier`, and will now be enforced by Prettier.

.stylelintrc
```diff
@@ -11,17 +11,12 @@
...
-    "max-line-length": [
-      100,
-      { "ignore": ["comments"] }
-    ],
     "no-invalid-position-at-import-rule": [true, {
         "ignoreAtRules": ["use"]
     }],
     "scss/media-feature-value-dollar-variable": null,
-    "selector-max-universal": 1
+    "selector-max-universal": 1,
```
MariaSolOs, line 18: Note that we should be able to remove these rules in https://github.com/palantir/blueprint/pull/7078 when we update `stylelint-config-palantir`.

packages/datetime2/src/components/date-range-picker3/_date-range-picker3.scss
```diff
@@ -114,7 +114,8 @@
   }
 
   /* stylelint-disable max-line-length */
```
ggdouglas: Should we replace this `stylelint-disable` with `prettier-ignore` as well?

packages/table/src/interactions/_interactions.scss
```diff
@@ -53,7 +53,9 @@ $reorder-handle-width: 22px !default;
 }
 
 /* stylelint-disable-next-line max-line-length */
```
ggdouglas: Should this be replaced with `prettier-ignore` to preserve the line length here?

packages/core/src/components/card/_card-variables.scss
```diff
@@ -16,9 +16,11 @@ $card-list-border-width: 1px !default;
 // CardList Card item min-height is calculated as height of a button + vertical padding.
 // We need to add an extra pixel to account for the bottom border.
 $card-list-item-padding-vertical: $pt-grid-size !default;
-$card-list-item-min-height: $pt-button-height + ($card-list-item-padding-vertical * 2) + $card-list-border-width !default;
+$card-list-item-min-height: $pt-button-height + ($card-list-item-padding-vertical * 2) +
```
ggdouglas: Ditto here regarding line-length and the `stylelint-disable max-line-length` rule above.

packages/datetime2/src/components/date-range-picker3/_date-range-picker3.scss
MariaSolOs: Good catch, I had only replaced the ones that resulted in invalid selectors.

MariaSolOs: I've replaced all the `stylelint-disable max-line-length` comments by `prettier-ignore`s :)

ggdouglas (approved):

PR #6724, opened 2024-02-23
[core] feat: more dark theme color aliases

adidahiya:
Ported relevant preflight changes from #6711 to help with a smoother migration to the newer dark theme colors

#### Checklist

- [ ] Includes tests
- [ ] Update documentation

#### Changes proposed in this pull request:

- Add new dark theme color aliases
- Use some of these aliases across core components and docs-theme package

#### Reviewers should focus on:

No regressions / visual changes in docs styling

PR #6711, opened 2024-02-14
[core] feat: darken dark theme colors

adidahiya:
Supersedes https://github.com/palantir/blueprint/pull/6708 and https://github.com/palantir/blueprint/pull/6707

#### Checklist

- [ ] Includes tests
- [x] Update documentation

#### Changes proposed in this pull request:

Update dark theme background colors to be one shade darker for many components based on the following proposal:

image

PR #7085, opened 2024-11-20
Fix new color alias

mud:
#### Fixes #0000
Just bumped Blueprint core to 5.15.0 and noticed there was some duplicate to the color alias where the new values were being overridden by the old duplicated variables.

#### Checklist

- [ ] Includes tests
- [ ] Update documentation

Remove the duplicate variables that use the old dark colors

### DIFF
--- a/packages/core/changelog/@unreleased/pr-7085.v2.yml
+++ b/packages/core/changelog/@unreleased/pr-7085.v2.yml
@@ -1,0 +1,6 @@
+type: fix
+fix:
+  description: Remove duplicate variables which overrides new dark theme secondary
+    and elevated colors with old values.
+  links:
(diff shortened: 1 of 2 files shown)


ggdouglas:
Tracking that this links back to [this comment](https://github.com/palantir/blueprint/pull/6711#issuecomment-1960155281) and originally added in https://github.com/palantir/blueprint/pull/6724

mud:
> Tracking that this links back to [this comment](https://github.com/palantir/blueprint/pull/6711#issuecomment-1960155281) and originally added in #6724

Gotcha, maybe having an inline comment suggesting the phased rollout would be useful since it currently looks like a bug.

PR #7105, opened 2024-12-02
Fix invalid code snippet example in Select documentation

Deuteu:

Prop `placeholder` does not exist on `Button`

#### Screenshot

Error when copy-pasting the example:
image


ggdouglas (approved):
Thanks for the fix!

PR #7106, opened 2024-12-02
Update button tests to use RTL

ggdouglas:
### Proposed changes

This PR serves as a proof of concept for how we can begin to migrate some of our existing [Enzyme](https://enzymejs.github.io/enzyme/) tests to [React Testing Library](https://testing-library.com/docs/react-testing-library). It covers some of the base testing cases such as: testing for rendering, testing for the presence of classes, and simulating mouse and keyboard events on components.


ggdouglas:
Getting seemingly unreleated failures from tests in Overlay/Overlay2 as a result of these changes to Button tests:

```
 <Overlay>
    Focus management
       does not crash while trying to return focus to overlay if user clicks outside the document
          1) Error: Uncaught TypeError: Cannot read properties of undefined (reading 'toLowerCase') (/var/folders/n1/whj_5mt92t10sltc2xkhhhwx08g4by/T/_karma_webpack_456877/commons.js:229994)
             at Context.<anonymous> (/var/folders/n1/whj_5mt92t10sltc2xkhhhwx08g4by/T/_karma_webpack_456877/commons.js:211319:26)
```
https://github.com/palantir/blueprint/blob/ffb7284b20bef8cc7e588c0fd69ef827fd8dbbb6/packages/core/test/overlay/overlayTests.tsx#L422-L446

```
 <Overlay2>
    Focus management
       does not crash while trying to return focus to overlay if user clicks outside the document
          2) Error: Uncaught TypeError: Cannot read properties of undefined (reading 'toLowerCase') (/var/folders/n1/whj_5mt92t10sltc2xkhhhwx08g4by/T/_karma_webpack_456877/commons.js:229994)
             at Context.<anonymous> (/var/folders/n1/whj_5mt92t10sltc2xkhhhwx08g4by/T/_karma_webpack_456877/commons.js:212391:26)
```
https://github.com/palantir/blueprint/blob/ffb7284b20bef8cc7e588c0fd69ef827fd8dbbb6/packages/core/test/overlay2/overlay2Tests.tsx#L476-L500

This is a bit concerning since the tests (in theory) have no obvious relation to each other. It might be worth noting that there is a callout that these tests have been flaky in the past due to their polluting the DOM with leftover elements:
https://github.com/palantir/blueprint/blob/ffb7284b20bef8cc7e588c0fd69ef827fd8dbbb6/packages/core/test/overlay/overlayTests.tsx#L36-L43

#### There are a few possible solutions:
1. Trace down the root change/cause of the new test failure/flake (currently attempting / in-progress)
2. Re-write the Overlay tests to also use RTL (desirable to have these updated, but will be very time-consuming)
3. Skip these old failing tests for now
4. Abandon the RTL test refactoring process for now

jscheiny (approved):

PR #7108, opened 2024-12-05
[docs] Update CardExample

ggdouglas:
#### Checklist

- [ ] Includes tests
- [x] Update documentation

Updates CardExample: [Before](https://blueprintjs.com/docs/#core/components/card) / [After](https://output.circle-artifacts.com/output/job/0e516b3b-88cf-4325-aed7-95b9658d5a64/artifacts/0/packages/docs-app/dist/index.html#core/components/card)

### DIFF
--- a/packages/docs-app/src/examples/core-examples/cardExample.tsx
+++ b/packages/docs-app/src/examples/core-examples/cardExample.tsx
@@ -1,5 +1,5 @@
 /*
- * Copyright 2017 Palantir Technologies, Inc. All rights reserved.
+ * Copyright 2024 Palantir Technologies, Inc. All rights reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


packages/docs-app/src/examples/core-examples/cardExample.tsx
```diff
@@ -16,64 +16,47 @@
...
+    const options = (
         <>
             <H5>Props</H5>
-            <Switch checked={this.state.interactive} label="Interactive" onChange={this.handleInteractiveChange} />
-            {this.state.interactive && (
-                <Switch checked={this.state.selected} label="Selected" onChange={this.handleSelectedChange} />
-            )}
-            <Switch checked={this.state.compact} label="Compact" onChange={this.handleCompactChange} />
+            <Switch checked={interactive} label="Interactive" onChange={handleBooleanChange(setInteractive)} />
+            {interactive && <Switch checked={selected} label="Selected" onChange={handleBooleanChange(setSelected)} />}
```
jscheiny: Nit: For more transparency and I think consistency with other examples, I think we should make this disabled instead of hidden when interactive is false.

ggdouglas: Good idea! Added in 097990c4944edb2dd3c2fe9ea4a4fd2bc4ca7c1d

jscheiny (approved):

PR #7117, opened 2024-12-09
export focusable element selector

bvandrc:
#### Checklist

- [N/A] Includes tests
- [X] Update documentation

Export the selector for all focusable elements


packages/core/src/common/utils/domUtils.ts
```diff
@@ -158,3 +158,20 @@ export function clickElementOnKeyPress(keys: string[]) {
...
+ */
+export const SELECTOR_FOCUSABLE = [
+    'a[href]:not([tabindex="-1"])',
+    'button:not([disabled]):not([tabindex="-1"])',
+    'details:not([tabindex="-1"])',
+    'input:not([disabled]):not([tabindex="-1"])',
+    'select:not([disabled]):not([tabindex="-1"])',
+    'textarea:not([disabled]):not([tabindex="-1"])',
+    '[tabindex]:not([tabindex="-1"])',
+].join(",");
```
ggdouglas, line 177: Instead of exporting this array, could we instead export a utility function that returns the focusable elements? e.g.

```ts
/**
 * Gets all focusable elements within the given element.
 *
 * @param {HTMLElement} element - The element to search within.
 * @returns {HTMLElement[]} An array of focusable elements.
 */
export function getFocusableElements(element: HTMLElement): HTMLElement[] {
    return Array.from(element.querySelectorAll(SELECTOR_FOCUSABLE));
}
```

packages/datetime/src/components/date-input/dateInput.tsx
```diff
@@ -732,11 +732,7 @@ function getKeyboardFocusableElements(popoverContentRef: React.MutableRefObject<
         return [];
     }
 
-    const elements = Array.from(
-        popoverContentRef.current.querySelectorAll<HTMLElement>(
-            "button:not([disabled]),input,[tabindex]:not([tabindex='-1'])",
-        ),
-    );
+    const elements = Array.from(popoverContentRef.current.querySelectorAll<HTMLElement>(Utils.SELECTOR_FOCUSABLE));
```
ggdouglas: Why are these additional selectors necessary for the date input components? Can you provide a brief summary of why we want to make this change?

packages/core/src/common/utils/domUtils.ts
bvandrc: No problem! https://github.com/palantir/blueprint/pull/7117/commits/7c0402592614df7cb2070a1c4641b816b4f8f40e

packages/datetime/src/components/date-input/dateInput.tsx
bvandrc: They are definitely not necessary, there is no functional change either. I had just found this and thought we could re-use the "get focusable elements" code.

ggdouglas: Looking into it a bit, it looks like there is a slight functional change to the default tab focusing behavior in DateInput3, but the change appears to be beneficial. We can see the before/after here where the current tab behavior skips over the month and year in focusing the elements within the picker:

https://github.com/user-attachments/assets/38f95a1a-e9f3-4a93-98bc-454ecf51a2c6

https://github.com/user-attachments/assets/8262b589-1ee3-48bf-930e-fc0a12a5ffa6

I'm in favor of us merging this since it also seems to fix this subtle bug.

ggdouglas (approved):

packages/datetime/src/components/date-input/dateInput.tsx
bvandrc: Awesome, didn't know we were fixing a bug with that change but yep that change is definitely what we want!

PR #7121, opened 2024-12-11
fix bug related to ticket #6703 placeholder not updating due to wrong…

### LINKED ISSUE #6703: Controlled DateInput3 does not display new values if popover is closed without selecting a date
#### Environment

- __Package version(s)__:
  "@blueprintjs/core": "^5.9.1",
  "@blueprintjs/datetime": "^5.3.1",
  "@blueprintjs/datetime2": "^2.3.1",
- __Operating System__: Windows 11
- __Browser name and version__: Firefox 122.0

#### Code Sandbox

Link to a minimal repro: https://codesandbox.io/p/devbox/eloquent-swanson-h79ky6?file=%2Fsrc%2FDatetimeExample.tsx

#### Steps to reproduce

1. Click within the input to open the popover
1. Click outside the popover to close it, without selecting a date
1. Click `Now` to set the date

#### Actual behavior

The displayed date in the DateInput still shows the placeholder text, and does not change to the new value.

This only occurs if the popover was opened without a date already set, and closed without selecting a date within the popover.

#### Expected behavior

The displayed text in the DateInput should change to 'now'.

sc-mitton:
… focus state

#### Fixes #6703

Fix a bug where the value isn't updating from controlled input.

Screenshot 2024-12-10 at 6 22 24 PM

### DIFF
--- a/packages/datetime2/src/components/date-input3/dateInput3.tsx
+++ b/packages/datetime2/src/components/date-input3/dateInput3.tsx
@@ -374,6 +374,7 @@ export const DateInput3: React.FC<DateInput3Props> = React.memo(function _DateIn
     const handleInputBlur = React.useCallback(
         (e: React.FocusEvent<HTMLInputElement>) => {
             if (inputValue == null || valueAsDate == null) {
+                setIsInputFocused(false);
                 return;
             }
 


palantirtech:
Thanks for your interest in [palantir/blueprint](https://github.com/palantir/blueprint), @sc-mitton! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

packages/datetime2/src/components/date-input3/dateInput3.tsx
```diff
@@ -373,6 +373,7 @@ export const DateInput3: React.FC<DateInput3Props> = React.memo(function _DateIn
     const handleInputBlur = React.useCallback(
         (e: React.FocusEvent<HTMLInputElement>) => {
             if (inputValue == null || valueAsDate == null) {
+                setIsInputFocused(false);
```
ggdouglas, line 377: This seems to fix the bug where a controlled prop passed to DateInput3 is set to a particular value, but it still seems to miss when the value is cleared, e.g.
Screenshot 2025-01-07 at 12 14 49
Can we also handle that related case here?

I think it should be a matter of affecting one of the useEffects above to handle the case where `valueFromProps` gets set to to a nullish value. Something like:
```ts
React.useEffect(() => {
    if (isControlled || valueFromProps == null) {
        setValue(valueFromProps);
    }
}, [isControlled, valueFromProps]);
```

sc-mitton: `const isControlled = valueFromProps !== undefined;`

The isControlled variable is already defined as valueFromProps !== undefined. So isControlled || valueFromProps == null would be redundant no? If valueFromProps == null is true, then isControlled is already true.

sc-mitton: I'm also not having this issue you mentioned in my local dev. In the hook you mentioned the value should be cleared.

ggdouglas: Apologies! I didn't have access to the original repro from the issue. It's possible that I have some other confounding variable in my local example that is causing this.

ggdouglas (approved):
Thanks for the fix 🙂

PR #7123, opened 2024-12-11
[docs] Rewrite Collapse docs

ggdouglas:
#### Checklist

- [ ] Includes tests
- [x] Update documentation

Rewrites Collapse docs with additional examples: [Before](https://blueprintjs.com/docs/#core/components/collapse) / [After](https://output.circle-artifacts.com/output/job/6437c2e8-baa3-4e29-9262-b9550d22d95d/artifacts/0/packages/docs-app/dist/index.html#core/components/collapse)


packages/docs-app/src/examples/core-examples/collapseExamples.tsx
```diff
@@ -0,0 +1,109 @@
...
+                        <p>
+                            This is an example of collapsible content. It could include detailed explanations, extra
+                            information, or supporting text for the primary content on the page. When expanded, the content is
+                            fully visible and takes up space in the document layout. When collapsed, it smoothly slides out of
+                            view.
+                        </p>
+                        <p>
+                            You can include multiple paragraphs, images, or any other elements inside a collapse. This example
+                            demonstrates how a longer content block behaves when toggled.
+                        </p>
```
jscheiny: This second paragraph is not in the actual code below

packages/docs-app/src/examples/core-examples/collapsePlaygroundExample.tsx
```diff
@@ -0,0 +1,59 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import * as React from "react";
+
+import { Button, Collapse, H5, Pre, Switch } from "@blueprintjs/core";
+import { Example, type ExampleProps, handleBooleanChange } from "@blueprintjs/docs-theme";
+
+export const CollapsePlaygroundExample: React.FC<ExampleProps> = props => {
```
jscheiny, line 22: Would be nice to update this example to make the keep children mounted toggle actually make a difference. As it stands currently, this toggle appears to have no effect on the example.

packages/docs-app/src/examples/core-examples/collapseExamples.tsx
ggdouglas: Ah, good catch! This was leftover from a previous iteration. Fixed in cf59c0b9737189948044edd96a9442059ec45b48

packages/docs-app/src/examples/core-examples/collapsePlaygroundExample.tsx
ggdouglas: Hm, I'm open to ideas on how best to demo it. IMO, the new "Keeping children mounted" section demonstrates this pretty well with the form input state either being mounted/unmounted.

I've added an input field to the playground example to the same effect: f625df3062f0c8cc477f4ff2e73fb0118e20086f

Let me know if that seems adequate!

jscheiny: Yeah, that should do the trick just fine.

jscheiny (approved):

PR #7127, opened 2024-12-12
[docs] Rewrite EditableText docs

ggdouglas:
#### Checklist

- [ ] Includes tests
- [x] Update documentation

Rewrites EditableText docs with additional examples: [Before](https://blueprintjs.com/docs/#core/components/editable-text) / [After](https://output.circle-artifacts.com/output/job/ae6261e8-86b1-427c-9288-805e50749372/artifacts/0/packages/docs-app/dist/index.html#core/components/editable-text)


packages/core/src/components/editable-text/editable-text.md
```diff
@@ -1,18 +1,34 @@
...
+
+**EditableText** can be used in both controlled and uncontrolled modes, similar
+to a standard React [`<input>` element](https://react.dev/reference/react-dom/components/input).
+Use the `value` prop for controlled usage, and `defaultValue` for uncontrolled usage. Use `onChange` to listen to
+ongoing updates and use `onConfirm` and `onCancel` to listen only to completed or canceled edits.
+
+The `onConfirm` and `onCancel` callbacks are invoked based on user interaction. The user presses <kbd>Enter</kbd>
+(or <kbd>Mod + Enter</kbd> when multiline) or blurs the input to confirm the current value, or presses
+<kbd>Escape</kbd> to cancel. Canceling resets the field to the last confirmed value. Neither callback is invoked if the
+value is unchanged.
```
jscheiny: The phrasing here is a bit difficult to parse. I think I'd prefer something like:

>
>```
>The `onConfirm` callback is invoked when a user presses <kbd>Enter</kbd>
>(or <kbd>Mod + Enter</kbd> when multiline) or when the user blurs the input.
>The `onCancel` callback is invoked when user presses <kbd>Escape</kbd>.
>Canceling resets the field to the last confirmed value. Neither callback is
>invoked if the value is unchanged.
>```

packages/core/src/components/editable-text/editable-text.md
```diff
@@ -23,34 +39,34 @@ you should center the component via flexbox or with `position` and `transform: t
...
-
-```tsx
-<EditableText multiline={true} minLines={3} maxLines={12} {...props} />
-```
+By default, **EditableText** supports a single line of text and resizes horizontally as needed.
+Enabling the `multiline` prop transforms it into a `<textarea>`, which grows and shrinks vertically
+as content changes. Use the `minLines` and `maxLines` props to constrain the height of the component.
 
 Users may confirm text in multiline mode by pressing <kbd>Ctrl + Enter</kbd> or <kbd>Command + Enter</kbd> rather than
 <kbd>Enter</kbd>. (Pressing the <kbd>Enter</kbd> key by itself moves the cursor to the next line.) This behavior
```
jscheiny:
```suggestion
<kbd>Enter</kbd>. Pressing the <kbd>Enter</kbd> key by itself moves the cursor to the next line. This behavior
```

packages/core/src/components/editable-text/editable-text.md
```diff
@@ -23,34 +39,34 @@ you should center the component via flexbox or with `position` and `transform: t
...
-__EditableText__ is used like an [`<input>` element](https://facebook.github.io/react/docs/forms.html) and supports
-controlled or uncontrolled usage through the `value` or `defaultValue` props, respectively. Use `onChange` to listen to
-ongoing updates and use `onConfirm` and `onCancel` to listen only to completed or canceled edits.
+@## Intent
 
-The `onConfirm` and `onCancel` callbacks are invoked based on user interaction. The user presses <kbd>Enter</kbd>
-(or <kbd>Command + Enter</kbd> when multiline) or blurs the input to confirm the current value, or presses
-<kbd>Escape</kbd> to cancel. Canceling resets the field to the last confirmed value. Neither callback is invoked if the
-value is unchanged.
+The `intent` prop controls the visual appearance of **EditableText**, similar to
```
jscheiny, line 57: We have notes on what intents mean in the button docs but not here. I don't know that we should necessarily add them here, but I think the lack of consistency is confusing. Curious if you think its worth repeating the meanings of intents in each place that we reference them. My intuition is that its worthwhile since we don't know where folks will look first.

packages/docs-app/src/examples/core-examples/editableTextExamples.tsx
```diff
@@ -0,0 +1,58 @@
...
+import { CodeExample, type ExampleProps } from "@blueprintjs/docs-theme";
+
+export const EditableTextBasicExample: React.FC<ExampleProps> = props => {
+    const code = `<EditableText placeholder="Click to edit..." onConfirm={...} onCancel={...} />`;
+    return (
+        <CodeExample code={code} {...props}>
+            <EditableText
+                placeholder="Click to edit..."
+                // eslint-disable-next-line no-console
+                onConfirm={value => console.log("Confirmed:", value)}
```
jscheiny: It would be nice to have these do something more visible. I played around with this example and didn't realize that it was logging to the console.

packages/core/src/components/editable-text/editable-text.md
ggdouglas: b7fbdefe8cf35e0204b322060f0bac5a753f2780

packages/core/src/components/editable-text/editable-text.md
ggdouglas: b7fbdefe8cf35e0204b322060f0bac5a753f2780

packages/core/src/components/editable-text/editable-text.md
ggdouglas: Yeah, that's a fair point. I'd included them at one point, but removed them while editing. I'm fine with including something for consistency though. Added in b7fbdefe8cf35e0204b322060f0bac5a753f2780

packages/docs-app/src/examples/core-examples/editableTextExamples.tsx
ggdouglas: Hm, we could use something like an alert, but I fear that might be a bit too disruptive. Do you have a suggestion on something more visible?

jscheiny: Maybe a toast?

ggdouglas: Added in b12eedf5e42525cb6f4f5e812368c10f0845c00c / 2f97df674ae0db4680948912732ef9257ae8a9bc

jscheiny (approved):

PR #7129, opened 2024-12-13
Apply outline styles to outlined ButtonGroups with Popover children

ggdouglas:
#### Fixes #0000

Related to #6966

#### Checklist

- [ ] Includes tests
- [ ] Update documentation

Fixes a case where ButtonGroups that have Buttons nested in Popovers would not apply outlined styles when `outlined={true}` was applied to the ButtonGroup parent. For example:

```tsx
<ButtonGroup outlined={true}>
    <Popover content={<FileMenu />}>
        <Button text="Click Me" />
    </Popover>
</ButtonGroup>
```

This behavior can be seen when toggling the `outlined` prop on the the [Usage with popovers ](http://localhost:9001/#core/components/button-group.usage-with-popovers)section of the Button docs:

| Before | After    |
| ------  | ------  |
| before | after |

### DIFF
--- a/packages/core/src/components/button/_button-group.scss
+++ b/packages/core/src/components/button/_button-group.scss
@@ -144,7 +144,8 @@ Styleguide button-group
   }
 
   &.#{$ns}-outlined {
-    > .#{$ns}-button {
+    > .#{$ns}-button,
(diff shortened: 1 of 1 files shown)


ggdouglas:
Replaced by https://github.com/palantir/blueprint/pull/8058

PR #7131, opened 2024-12-17
Remove reference to old Cobalt colours in docs

Athenodoros:
I haven't filed an issue, but I think this is a pretty simple change.

#### Checklist

- [ ] Includes tests
- [ ] Update documentation

The qualitative colour array used to reference the Cobalt3, but is [rendering empty](https://blueprintjs.com/docs/#core/colors.qualitative-color-schemes) now that it's been removed. This replaces it with cerulean, which seems like the closest option now.


ggdouglas (approved):
Thanks for the fix!

PR #7134, opened 2024-12-20
Fix code snippet in RadioCard documentation

Deuteu:
To be coherent with [actual code](https://github.com/palantir/blueprint/blob/fd422a9ab1cd5191483abeb84a43d257242ea97b/packages/docs-theme/src/common/eventHandlerUtils.ts#L50)

And also in my experience `currentValue` was `null`

### DIFF
--- a/packages/core/changelog/@unreleased/pr-7134.v2.yml
+++ b/packages/core/changelog/@unreleased/pr-7134.v2.yml
@@ -1,0 +1,5 @@
+type: improvement
+improvement:
+  description: Fix code snippet in RadioCard documentation
+  links:
+  - https://github.com/palantir/blueprint/pull/7134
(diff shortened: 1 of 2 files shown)


ggdouglas:
> And also in my experience `currentValue` was `null`

Can you elaborate on where you are seeing `currentValue` as `null`? From my understanding, `event.target` and `currentTarget` should be equivalent in this particular case (example [here](https://codesandbox.io/p/sandbox/radiogroup-event-target-vs-currenttarget-dskmvg)).

This is because `RadioGroup` internally applies the onChange event handler directly to each of its children: https://github.com/palantir/blueprint/blob/894a7d89810025296f8efaf270ad98f728d3a7b9/packages/core/src/components/forms/radioGroup.tsx#L155

Therefore, `event.currentTarget` seems relevant since it references the element to which the event handler has been attached.

Is there an advantage towards suggesting this alternative approach of casting as `HTMLInputElement`?

PR #7141, opened 2025-01-09
feat(SegmentedControl): menu role option

bvandrc:
Add `menu` role option, as well as associated children roles.


bvandrc:
@ggdouglas @jscheiny @ryannono Can one of you please take a look at this, we need this change for some correct a11y purposes.

ggdouglas (approved):
The addition of menu seems reasonable here. Thanks for adding the child roles as well!

PR #7142, opened 2025-01-09
React 18 Upgrade

ggdouglas:
This feature branch PR serves as the central hub for upgrading Blueprint's documentation and internal dependencies to **React 18**. The initial changes in this PR involve updating dependencies relevant to the React 18 upgrade. Subsequent changes will be made in separate PRs that will be be merged into `develop` and rebased into this feature branch.

**Tracking progress in the list below:**

##### Fixes type, lint, and compile breaks:
- [x] #7147
- [x] #7150
- [x] #7145
- [x] ~#7146~
- [x] #7173
- [x] #7174
- [x] #7161
- [x] #7159
- [x] #7160
- [x] #7162
- [x] #7177

##### Fixes/skips test breaks
- [x] #7148
- [x] #7169
- [x] #7170
- [x] #7171

##### Related issues:
- #7165
- #7166
- #7167
- #7168


packages/core/test/controls/numericInputTests.tsx
```diff
@@ -1098,7 +1097,7 @@ describe("<NumericInput>", () => {
             incrementButton.simulate("mousedown", { shiftKey: true });
             expect(component.find("input").prop("value")).to.equal("1.101");
 
-            TestUtils.act(() => {
```
ggdouglas, line 1101: Fixes a deprecation in React 18. Act is now exported from React as of `v18.3.1`.

Screenshot 2025-01-14 at 17 12 11@2x

https://react.dev/warnings/react-dom-test-utils#reactdomtestutilsact-warning

> act from react-dom/test-utils has been deprecated in favor of act from react.

packages/core/src/components/icon/icon.tsx
```diff
@@ -101,7 +101,7 @@ export interface DefaultIconProps extends IntentProps, Props, DefaultSVGIconProp
  * @see https://stackoverflow.com/a/73795494/7406866
  */
 export interface IconComponent extends React.FC<IconProps<Element>> {
-    <T extends Element = Element>(props: IconProps<T>): React.ReactElement | null;
```
ggdouglas: Attempted to cherry-pick this type change into develop, but unfortunately encountered errors. This type fix is valid in React 18, but fails with React 16 types.

See type failures on branch based off of develop: https://github.com/palantir/blueprint/compare/gdouglas/test-Icon-compile-break

packages/icons/src/svgIconContainer.tsx
```diff
@@ -42,7 +42,7 @@ export type SVGIconContainerProps<T extends Element> = Omit<SVGIconProps<T>, "ch
  * @see https://stackoverflow.com/a/73795494/7406866
  */
 export interface SVGIconContainerComponent extends React.FC<SVGIconContainerProps<Element>> {
-    <T extends Element = Element>(props: SVGIconContainerProps<T>): React.ReactElement | null;
```
ggdouglas: Similar to [type compile issue with Icon](https://github.com/palantir/blueprint/pull/7142#discussion_r1915691062), this fix compiles in React 18, but not 16.

See: https://github.com/palantir/blueprint/compare/gdouglas/test-SVGIconContainer-compile-break

packages/datetime/package.json
```diff
@@ -68,10 +68,10 @@
         "karma": "^6.4.2",
         "mocha": "^10.2.0",
         "npm-run-all": "^4.1.5",
-        "react": "^16.14.0",
```
ggdouglas, line 71: Should we be updating the react `devDependencies` in the datetime package to 18 if it has `peerDependencies` of 16/17?

Wondering if I should leave these be instead of changing them.

packages/test-commons/bootstrap.mjs
```diff
@@ -2,8 +2,10 @@
  * Copyright 2017 Palantir Technologies, Inc. All rights reserved.
  */
 
+import ReactEighteenAdapter from "@cfaester/enzyme-adapter-react-18";
```
ggdouglas, line 5: https://github.com/cfaester/enzyme-adapter-react-18
> "A **very** unofficial adapter for React 18 for [Enzyme](https://enzymejs.github.io/enzyme/)."

Very unofficial, but it at least gets us moving forward until we can replace Enzyme.

packages/docs-app/src/examples/core-examples/toastCreateAsyncExample.tsx
```diff
@@ -88,12 +88,13 @@ async function showMessageFromNewToaster() {
...
  * @param containerElement The container argument passed to OverlayToaster.create/OverlayToaster.createAsync
  */
-function unmountReact16Toaster(containerElement: HTMLElement) {
+function unmountToaster(containerElement: HTMLElement) {
     const toasterRenderRoot = containerElement.firstElementChild;
     if (toasterRenderRoot == null) {
         throw new Error("No elements were found under Toaster container.");
     }
-    ReactDOM.unmountComponentAtNode(toasterRenderRoot);
+    const root = ReactDOM.createRoot(toasterRenderRoot);
```
gluxon: I believe this calls `ReactDOM.createRoot` on a DOM node that previously had a react root created on it, which leaks memory. The old root never gets unmounted.

The `OverlayToaster.createAsync` call above doesn't seem to override the `domRenderer` field. I think the React root that was previously created on `toasterRenderRoot` is actually a React 17 legacy root. Without changes to how `createAsync` is called above, I think `ReactDOM.unmountComponentAtNode` is most correct here?

To fully remove `unmountComponentAtNode`, there would need to be a change to the `createAsync` call to create a React root, store a reference to it in the example here, and then call unmount in the cleanup effect.

packages/core/src/components/icon/icon.tsx
gluxon: I know I suggested the `React.ReactElement` to `React.ReactNode` change, but I'm realizing we might need something more dynamic to maintain both React 16 and 18 compatibility.

We could try this:

```ts
<T extends Element = Element>(props: IconProps<T>): ReturnType<typeof React.FC< IconProps<Element>>> | null;
```

I think that should return `React.ReactNode` on `@types/react@18` and `React.ReactElement` on `@types/react@16`?

When Blueprint drops React 16/17 compatibility, the `ReturnType` hack above could be replaced with `React.ReactNode` again.

packages/docs-app/src/examples/core-examples/toastCreateAsyncExample.tsx
ggdouglas: Ah, yup, good catch! I'll split this out, ignore the deprecated ReactDOM lint errors, and track this in #7166

ggdouglas: 595f50be28e84d6845240a90416855faef095c65 / 89ce662ec2694accab04244454a9225bc0b527e5

packages/core/src/components/icon/icon.tsx
ggdouglas: I think this will work! Thanks for the suggestion. #7177

jscheiny (approved):

PR #7143, opened 2025-01-09
Remove tab hotkey when in ROW FocusMode

evansjohnson:
#### Fixes no tracked issue

Quick follow on to the recent PR at https://github.com/palantir/blueprint/pull/7037 after testing an initial implementation.

#### Checklist
- [-] Includes tests
- [-] Update documentation

Remove the tab and shift+tab hotkeys from Table2 ROW FocusMode. The reasoning is that with a table with a large number of rows, focus effectively becomes trapped in the table if navigating by tab key, since a user would need to tab through every row to proceed to later elements. Without these hotkeys, users will still see the focus indication, and can use the arrow keys to adjust the focused row.

Do we need to expose a boolean toggle for this? I suppose for tables with small #s of rows this isn't really a problem and maybe someone would want tab to navigate through the rows

#### Screenshot
n/a

### DIFF
--- a/packages/table/src/index.ts
+++ b/packages/table/src/index.ts
@@ -30,6 +30,7 @@ export {
     type AnyRect,
     type CellCoordinates,
     Clipboard,
+    type FocusMode,
     type FocusedCellCoordinates,
     Grid,
     Rect,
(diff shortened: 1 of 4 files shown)


packages/table/src/index.ts
```diff
@@ -30,6 +30,7 @@ export {
     type AnyRect,
     type CellCoordinates,
     Clipboard,
+    type FocusMode,
```
evansjohnson, line 33: also sneaking this export of `FocusMode` into this PR

ggdouglas (approved):

PR #7145, opened 2025-01-09
Replace `defaultProps` with destructuring in function components

ggdouglas:

#### Proposed changes:

Replaces `defaultProps` with destructuring in anticipation of lint errors that will be encountered in React 18.3.0.

Fixes lint errors such as the following:
```
error: 'defaultProps' is deprecated. Use {@link https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#default_valuedefault values for destructuring assignments instead} (deprecation/deprecation) at src/components/card-list/cardList.tsx:61:10:
         59 |     );
         60 | });
       > 61 | CardList.defaultProps = {
            |          ^
         62 |     bordered: true,
         63 |     compact: false,
         64 | };
```

#### See also:
- https://github.com/reactjs/rfcs/blob/createlement-rfc/text/0000-create-element-changes.md#deprecate-defaultprops-on-function-components
- https://sophiabits.com/blog/stop-using-defaultprops


packages/core/src/components/dialog/dialog.tsx
```diff
@@ -146,7 +146,7 @@ export class Dialog extends AbstractPureComponent<DialogProps> {
                     <div
                         className={classNames(Classes.DIALOG, className)}
                         role={role}
-                        aria-modal={overlayProps.enforceFocus ?? Overlay2.defaultProps?.enforceFocus}
```
gluxon, line 149: Would it be useful to have `Overlay2` still export its default props if we need to reference them internally?

```tsx
// Overlay2.tsx

export const OVERLAY2_DEFAULT_PROPS = {
  enforceFocus: true
};
```

I think that makes it easier to see the intent of the code here. Otherwise it's more difficult to know why someone chose to default `aria-modal` to `true`.

packages/datetime/test/components/dateInputTests.tsx
```diff
@@ -660,7 +660,7 @@ describe("<DateInput>", () => {
             focusInput(wrapper);
             changeInput(wrapper, "4/77/2016");
             blurInput(wrapper);
-            assert.strictEqual(wrapper.find(InputGroup).prop("value"), DateInput.defaultProps?.invalidDateMessage);
+            assert.strictEqual(wrapper.find(InputGroup).prop("value"), "Invalid date");
```
gluxon: Similar comment around potentially exporting default props from `DateInput`.

packages/datetime2/src/components/date-input3/dateInput3.tsx
```diff
@@ -568,6 +568,9 @@ export const DateInput3: React.FC<DateInput3Props> = React.memo(function _DateIn
     );
 });
 DateInput3.displayName = `${DISPLAYNAME_PREFIX}.DateInput3`;
+
+// TODO: Removing `defaultProps` here breaks tests. Investigate why.
+// eslint-disable-next-line deprecation/deprecation
```
gluxon, line 573: That's fun. 😂

Thanks for leaving a comment so others have context on this in the future.

gluxon (commented):
Thanks! Scrolled through and I believe we've properly copied the defaults over. Only question is around whether we should share default values through an export somehow.

packages/core/src/components/dialog/dialog.tsx
ggdouglas: Yep, can do! I 👍 had considered something like this originally, but wavered at the time of implementing. I think this is a good idea for posterity since these can technically be consumed externally. Having the export set of values to point to will give us a migration path in case there are any other consumers we are unaware of.

packages/datetime2/src/components/date-input3/dateInput3.tsx
ggdouglas: Yeah.. 🥲

I am working on updating/refactoring a lot of the datetime2 tests currently, so will hopefully uncover the cause of this during the process.

packages/core/src/components/dialog/dialog.tsx
ggdouglas: 5084a87cee82f326fafdff3f6e0f7f414b221364

packages/datetime/test/components/dateInputTests.tsx
ggdouglas: d491cf1bd57faeb751a1124cf33c95935f7e0904

packages/select/src/components/suggest/suggest.tsx
```diff
@@ -182,7 +182,7 @@ export class Suggest<T> extends AbstractPureComponent<SuggestProps<T>, SuggestSt
         if (this.state.isOpen === false && prevState.isOpen === true) {
             // just closed, likely by keyboard interaction
             // wait until the transition ends so there isn't a flash of content in the popover
-            const timeout = this.props.popoverProps?.transitionDuration ?? Popover.defaultProps.transitionDuration;
+            const timeout = this.props.popoverProps?.transitionDuration ?? 300;
```
gluxon: nitpick: Do we want to use a `POPOVER_DEFAULT_PROPS` export here as well?

gluxon (dismissed):

packages/select/src/components/suggest/suggest.tsx
ggdouglas: Hm, I'm more inclined to revert my original change here since Popover is a class component (and thus should be exempt from the deprecation). aefacb1fa9115485bebaf01e5cb0b2e0726223a7

gluxon: That works too!

gluxon (approved):

PR #7148, opened 2025-01-10
Wrap tests in TestUtils.act(...)

ggdouglas:

Updates instances where we call `setState` in tests that lead to warnings such as this:
```
When testing, code that causes React state updates should be wrapped into act(...):

act(() => {
  /* fire events that update state */
});
/* assert on the output */

This ensures that you're testing the behavior the user would see in the browser. Learn more at https://reactjs.org/link/wrap-tests-with-act
```
as well as failures on the main feature branch when running in React 18.


gluxon (approved):
Looks good!

PR #7150, opened 2025-01-10
Update types and props for compatibility with React 18

ggdouglas:

#### Proposed changes:

Updates types for forward compatibility with React 18.

### DIFF
--- a/packages/core/src/components/entity-title/entityTitle.tsx
+++ b/packages/core/src/components/entity-title/entityTitle.tsx
@@ -177,7 +177,7 @@ EntityTitle.displayName = `${DISPLAYNAME_PREFIX}.EntityTitle`;
 /**
  * Construct header class name from H{*}. Returns `undefined` if `heading` is not a Blueprint heading.
  */
-function getClassNameFromHeading(heading: React.FC<unknown>) {
+function getClassNameFromHeading(heading: React.FC<any>) {
     const headerIndex = [H1, H2, H3, H4, H5, H6].findIndex(header => header === heading);
     if (headerIndex < 0) {
         return undefined;
--- a/packages/core/src/components/entity-title/entityTitle.tsx
+++ b/packages/core/src/components/entity-title/entityTitle.tsx
@@ -22,7 +22,7 @@ import { type IconName, IconNames } from "@blueprintjs/icons";
 import { Classes, DISPLAYNAME_PREFIX, type MaybeElement, type Props } from "../../common";
 import { H1, H2, H3, H4, H5, H6 } from "../html/html";
 import { Icon } from "../icon/icon";
-import { Text } from "../text/text";
+import { Text, type TextProps } from "../text/text";
 
 export interface EntityTitleProps extends Props {
     /**
@@ -177,7 +177,7 @@ EntityTitle.displayName = `${DISPLAYNAME_PREFIX}.EntityTitle`;
 /**
  * Construct header class name from H{*}. Returns `undefined` if `heading` is not a Blueprint heading.
  */
-function getClassNameFromHeading(heading: React.FC<any>) {
+function getClassNameFromHeading(heading: React.FC<TextProps>) {
     const headerIndex = [H1, H2, H3, H4, H5, H6].findIndex(header => header === heading);
     if (headerIndex < 0) {
         return undefined;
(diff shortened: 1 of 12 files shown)


packages/core/src/components/entity-title/entityTitle.tsx
```diff
@@ -177,7 +177,7 @@ EntityTitle.displayName = `${DISPLAYNAME_PREFIX}.EntityTitle`;
 /**
  * Construct header class name from H{*}. Returns `undefined` if `heading` is not a Blueprint heading.
  */
-function getClassNameFromHeading(heading: React.FC<unknown>) {
+function getClassNameFromHeading(heading: React.FC<any>) {
```
gluxon: It can be valuable to use `unknown` rather than `any` since the former is more future-proof. If someone refactors this function, they might not realize there's some type safety loss with the argument.

Does `() => React.ReactNode` work as an alternative to `React.FC<unknown>`?

packages/core/src/components/hotkeys/hotkeys.tsx
```diff
@@ -24,7 +24,7 @@ import { H4 } from "../html/html";
 
 import { Hotkey, type HotkeyProps } from "./hotkey";
 
-export interface HotkeysProps extends Props {
+export interface HotkeysProps extends Props, React.PropsWithChildren<object> {
```
gluxon: It looks like we're expecting children to be `React.ReactElement<HotkeyProps>`. Should we use that type over `object`?

https://github.com/palantir/blueprint/blob/0c815e909e86f511d9b057e7fa4f2dd80c91a3c8/packages/core/src/components/hotkeys/hotkeys.tsx#L57

Or was the goal to preserve the same type definition of `HotkeyProps` and avoid compile breaks on consumers?

gluxon: This one is a nitpick, but I've personally been avoiding `React.PropsWithChildren<...>` since I see the React docs usually write the `children` prop directly.

https://react.dev/learn/typescript#typing-children

Having the `children` prop directly in the interface also makes it easier to add a JSDoc on the field.

packages/table/src/headers/columnHeader.tsx
```diff
@@ -36,7 +36,7 @@ export interface ColumnWidths {
     defaultColumnWidth: number;
 }
 
-export interface ColumnHeaderProps extends HeaderProps, ColumnWidths, ColumnIndices {
+export interface ColumnHeaderProps extends React.PropsWithChildren<HeaderProps>, ColumnWidths, ColumnIndices {
```
gluxon, line 39: nitpick: Similar comment around potentially writing the `children` field directly.

gluxon (dismissed):

packages/core/src/components/entity-title/entityTitle.tsx
ggdouglas: Ack, will try this out!

ggdouglas: Changing to `() => React.ReactNode` does not seem to work:
```
Argument of type 'FC<any> | FC<TextProps>' is not assignable to parameter of type '() => ReactNode'.
  Type 'FC<any>' is not assignable to type '() => ReactNode'.
    Target signature provides too few arguments. Expected 1 or more, but got 0.ts(2345)
```
but I do think this leads us to a better solution to use `FC<TextProps>` instead. This function is only used internally in EntityTitle.
```suggestion
function getClassNameFromHeading(heading: React.FC<TextProps>) {
```

ggdouglas: f50354994ce29779ce51af89b0d30f87d4b527de

packages/core/src/components/hotkeys/hotkeys.tsx
ggdouglas: fd385b9a1e5e30308df5d7637f2f0aa279ae493b

packages/table/src/headers/columnHeader.tsx
ggdouglas: 21a62fcc8a12121739b04430f25dc05d1e14da19

packages/core/src/components/entity-title/entityTitle.tsx
gluxon: Nice!

gluxon (approved):

PR #7152, opened 2025-01-13
chore: Migrate tslint sort keys rule to eslint

jscheiny:
This PR moves us from the tslint `object-literal-sort-keys` rule to the eslint `sort-keys` rule.


packages/core/src/components/popover/customModifiers.ts
```diff
@@ -21,7 +21,7 @@
 
 import type { Modifier } from "@popperjs/core";
 
-// tslint:disable object-literal-sort-keys
```
ggdouglas, line 24: Any clue why we're disabling sorting in this file?

packages/core/test/forms/fileInputTests.tsx
```diff
@@ -79,7 +79,7 @@ describe("<FileInput>", () => {
         const onChange = sinon.spy();
         const onInputChange = sinon.spy();
 
-        const wrapper = shallow(<FileInput {...{ onChange, onInputChange, inputProps }} />);
+        const wrapper = shallow(<FileInput {...{ inputProps, onChange, onInputChange }} />);
```
ggdouglas: **nit:** Spreading these three props seems a bit odd. Could we inline them instead?

packages/datetime2/test/components/dateRangePicker3Tests.tsx
```diff
@@ -237,13 +237,13 @@ describe("<DateRangePicker3>", () => {
 
             it("calls onDayMouseEnter", () => {
                 const onDayMouseEnter = sinon.spy();
-                render({ defaultValue, dayPickerProps: { onDayMouseEnter } }).left.mouseEnterDay(14);
```
ggdouglas, line 240: Why weren't these caught before? I assume that we weren't running the tslint checks on these tests?

packages/core/src/components/popover/customModifiers.ts
jscheiny: I'm guessing that was so that `matchReferenceWidthModifier` matched its definition with the reference material.

packages/datetime2/test/components/dateRangePicker3Tests.tsx
jscheiny: The previous lint rule did not check objects that were contained to a single line, it only ran when the object spanned multiple lines. Most of the changes in this PR result from that difference in rule behavior.

packages/core/test/forms/fileInputTests.tsx
jscheiny: Unclear to me why this was like this, will try inlining them since its a test

ggdouglas (approved):

PR #7154, opened 2025-01-13
Revert "Switch from @import to @use"

ggdouglas:
### Reverts palantir/blueprint#7074

We are reverting this PR due to style regressions discovered while testing `develop` before release. Notably, the borders and shadows around text inputs and popovers are missing. Additionally, there are instances in the documentation where margins and padding differ. Given the wide scope of these changes, we have decided to revert the original change for now and revisit it in a follow-up PR with more thorough manual regression testing of Blueprint's styles.

### Examples:
| Expected | Actual    |
| ------  | ------  |
| expected | actual |

| Expected | Actual    |
| ------  | ------  |
| expected | actual |

diff

---

@justinbhopper - Apologies for not catching this at the time of review. I'd be happy to revisit and re-review your changes. We will push this forward again while being more mindful of style regressions..


gluxon (approved):
Agree with reverting for now and spending more time (unrushed) to debug later. Thanks @ggdouglas.

jscheiny (approved):

justinbhopper:
@ggdouglas Sounds good!

I did a quick comparison and see that some imports were missed for function calls in a few places. SASS compile did not complain because I guess it assumes they might be native CSS functions.

image
image

I can put up a new PR with corrections.

PR #7157, opened 2025-01-13
Switch from @import to @use

### LINKED ISSUE #7031: @import rule is now deprecated.  Migrate to @use and @forward

justinbhopper:
#### Fixes #7031

This PR re-introduces the changes from #7074 and includes fixes to issues mentioned in #7154.

In the original PR, some imports were missed for function calls in a few places. SASS compile did not complain because I guess it assumes they might be native CSS functions.

I have double-checked all function usage to make sure no other imports were missing, and I've compared the results of the docs-app locally to make sure the issues shown in #7154 are resolved.

cc: @ggdouglas 

#### Checklist

- [ ] Includes tests
- [ ] Update documentation


packages/core/src/common/_variables.scss
```diff
@@ -1,8 +1,8 @@
 // Copyright 2015 Palantir Technologies, Inc. All rights reserved.
 // Licensed under the Apache License, Version 2.0.
 
-@import "color-aliases";
-@import "mixins";
```
ggdouglas, line 5: It looks like updating `@import "mixins"` here causes the `border-shadow` mixin to not be available in certain component style files. Particularly, it causes the elevation shadows on [Cards](https://output.circle-artifacts.com/output/job/69b89d04-aab0-4e11-8db8-07fab4374a7d/artifacts/0/packages/docs-app/dist/index.html#core/components/card.elevation) and [wherever else](https://output.circle-artifacts.com/output/job/69b89d04-aab0-4e11-8db8-07fab4374a7d/artifacts/0/packages/docs-app/dist/index.html#datetime2/date-picker3) we use the elevation classes specifically to become malformed.

| Before | After    |
| ------  | ------  |
| before | after |

| Before | After    |
| ------  | ------  |
| before | after |

Screenshot 2025-01-27 at 11 42 33
```css
/* `border-shadow` mixin appears in compiled CSS */
.bp5-elevation-4 {
    box-shadow: border-shadow(.1), 0 4px 8px rgba(17, 20, 24, .2), 0 18px 46px 6px rgba(17, 20, 24, .2);
}
```

ggdouglas (changes requested):
Hey @justinbhopper, thanks for re-introducing this! After a more thorough review, it looks like we still need to address some stylistic changes to elevation styles. Otherwise, the remainder here looks good, and I can approve once we address fixing the elevation classes.

packages/core/src/common/_variables.scss
justinbhopper: @ggdouglas Great catch!  Somehow, I missed this function usage.  

In order to solve this, I had to remove the usage of `border-shadow` in `_variables.css`.  This is because we can't import a reference to `_mixins.scss` in this file without creating a circular dependency loop.  Since `border-shadow` is pretty straight forward, I figured it would be fine.

I _triple-checked_ all the other functions and I feel pretty confident this was the only thing I missed.  Famous last words? 🤣

ggdouglas: Yeah, I had also noticed the circular dependency between `_variables.css` and `_mixins.scss`. Substituting the mixin seems like a reasonable change to me. 👍

ggdouglas (approved):
Going to do one last visual regression pass here before merging, but generally this LGTM. Thanks for tackling this! 🙂

PR #7158, opened 2025-01-14
Docs: Update examples that reference ReactDOM.render

ggdouglas:
#### Proposed Changes

Updates stale documentation that references old/deprecated ReactDOM API methods to have better documented support for use with React 18.

There are no actual changes to working code in this PR, only examples in docs.

#### Checklist

- [ ] Includes tests
- [x] Update documentation

### DIFF
--- a/packages/core/src/components/portal/portal.md
+++ b/packages/core/src/components/portal/portal.md
@@ -40,15 +40,16 @@ components which use portals (popovers, tooltips, dialogs, etc.). You can do so 
 ```tsx
 import { Button, Popover, PortalProvider } from "@blueprintjs/core";
 import * as React from "react";
-import * as ReactDOM from "react-dom";
+import * as ReactDOM from "react-dom/client";
 
-ReactDOM.render(
+const root = ReactDOM.createRoot(document.getElementById("root"));
+
+root.render(
     <PortalProvider portalClassName="my-custom-class">
         <Popover content="My portal has a custom class">
             <Button text="Example" />
         </Popover>
-    </PortalProvider>
-    document.querySelector("#app"),
+    </PortalProvider>,
 );
 ```
 
--- a/packages/core/src/context/hotkeys/hotkeys-provider.md
+++ b/packages/core/src/context/hotkeys/hotkeys-provider.md
@@ -39,28 +39,29 @@ To use **HotkeysProvider**, wrap your application with it at the root level:
 ```tsx
 import { HotkeysProvider } from "@blueprintjs/core";
 import * as React from "react";
-import * as ReactDOM from "react-dom";
+import * as ReactDOM from "react-dom/client";
 
-ReactDOM.render(
+const root = ReactDOM.createRoot(document.getElementById("root"));
+
+root.render(
     <HotkeysProvider>
         <div>My app has hotkeys 😎</div>
     </HotkeysProvider>,
-    document.querySelector("#app"),
 );
 ```
 
 @## Advanced usage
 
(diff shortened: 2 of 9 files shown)


packages/core/src/components/portal/portal.md
```diff
@@ -40,15 +40,16 @@ components which use portals (popovers, tooltips, dialogs, etc.). You can do so
 ```tsx
 import { Button, Popover, PortalProvider } from "@blueprintjs/core";
-import * as ReactDOM from "react-dom";
+import * as ReactDOM from "react-dom/client";
 
-ReactDOM.render(
+const root = ReactDOM.createRoot(document.getElementById("root"));
```
ggdouglas, line 45: Opting to use language most similar to what is used in the core [React docs](https://react.dev/reference/react-dom/client/createRoot#reference).

packages/docs-app/src/getting-started.md
```diff
@@ -110,31 +110,27 @@ For more information, see [Understanding TypeScript](#blueprint/reading-the-docs
 @## Vanilla JS APIs
 
 JS components are built using React, but that does not limit their usage to only React applications.
-You can render any component in any JavaScript application with `ReactDOM.render`. Think of it like
+You can render any component in any JavaScript application with `render`. Think of it like
```
ggdouglas, line 113: These docs generally seem a bit unnecessary to have included for Blueprint. Perhaps we should replace this entire section with links to the core React docs?

packages/core/src/context/hotkeys/hotkeys-provider.md
```diff
@@ -103,15 +104,20 @@ function Plugin() {
     );
 }
 
-function PluginSlot(props) {
+function PluginSlot({ children }: { children: React.ReactNode }) {
```
ggdouglas, line 107: N.B. I haven't actually tested/executed this code, so I'm not 100% sure if it is right for this given example for plugins. At the very least, the code does compile TS.

packages/docs-app/src/getting-started.md
jscheiny: Yeah agreed with that

packages/icons/src/loading-icons.md
```diff
@@ -119,24 +119,24 @@ importing icon modules for you**.  Let's take a look at this required configurat
...
+                      `@blueprintjs/icons/lib/esm/generated/20px/paths/${name}`
+                  )
         ).default;
     };
 
     Icons.setLoaderOptions({ loader: lazyLoader });
     ```
 
-4. Load some icon paths up front (dynamically) with network requests, and the rest lazily/on-demand.
+5. Load some icon paths up front (dynamically) with network requests, and the rest lazily/on-demand.
```
jscheiny, line 139: Why the number change here?

jscheiny (approved):

packages/icons/src/loading-icons.md
ggdouglas: Hrm, I think Prettier did this. Will look into it

ggdouglas: Oh, it's because we have two "4."s in the markdown (but the docs correct for this when published)

https://github.com/palantir/blueprint/blob/c8570835d81382313c2bcf3a09aafe614f3cd854/packages/icons/src/loading-icons.md?plain=1#L111

https://github.com/palantir/blueprint/blob/c8570835d81382313c2bcf3a09aafe614f3cd854/packages/icons/src/loading-icons.md?plain=1#L139

https://blueprintjs.com/docs/#icons/loading-icons.using-dynamic-imports

So the new order is more correct.

PR #7159, opened 2025-01-14
Ignore deprecated ReactDOM methods in Context Menu singleton

ggdouglas:
Cherry-picked from **React 18 Upgrade feature branch** https://github.com/palantir/blueprint/pull/7142

Part of #7165

Preemptively ignores linter errors in anticipation of internal React 18 upgrade. See above issue for more detail.

### DIFF
--- a/packages/core/src/components/context-menu/contextMenuSingleton.tsx
+++ b/packages/core/src/components/context-menu/contextMenuSingleton.tsx
@@ -50,7 +50,9 @@ export function showContextMenu(
 ) {
     const {
         container = document.body,
+        // eslint-disable-next-line deprecation/deprecation
         domRenderer = ReactDOM.render,
+        // eslint-disable-next-line deprecation/deprecation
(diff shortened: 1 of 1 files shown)


jscheiny (dismissed):

packages/core/src/components/context-menu/contextMenuSingleton.tsx
```diff
@@ -50,7 +50,9 @@ export function showContextMenu(
 ) {
     const {
         container = document.body,
+        // eslint-disable-next-line deprecation/deprecation
```
jscheiny, line 54: Can we add some sort of react 18 note to these comments so that we can easily track these down later?

ggdouglas: Yep! Added here eb3f5e367b647d93dedf4ae448c11d88bf6b1167

jscheiny (approved):

PR #7161, opened 2025-01-14
Fix additional types for React 18

ggdouglas:
Cherry-picked from **React 18 Upgrade feature branch** https://github.com/palantir/blueprint/pull/7142

Fixes some additional remaining type compile breaks in anticipation of internal React 18 upgrade.

### DIFF
--- a/packages/core/test/overflow-list/overflowListTests.tsx
+++ b/packages/core/test/overflow-list/overflowListTests.tsx
@@ -238,6 +238,8 @@ describe("<OverflowList>", function (this) {
 
         /** Assert ordered IDs of overflow items. */
         wrapper.assertOverflowItems = (...ids: number[]) => {
+            // enzyme's wrapper.find returns `any` type here, so we need to cast to the correct type
+            // see: https://github.com/palantir/blueprint/pull/7161/files#r1915372750
             const overflowItems: TestItemProps[] = wrapper.find(TestOverflow).prop("items");
             assert.sameMembers(
                 overflowItems.map(it => it.id),
--- a/packages/core/test/overflow-list/overflowListTests.tsx
+++ b/packages/core/test/overflow-list/overflowListTests.tsx
@@ -238,7 +238,7 @@ describe("<OverflowList>", function (this) {
 
         /** Assert ordered IDs of overflow items. */
         wrapper.assertOverflowItems = (...ids: number[]) => {
-            const overflowItems = wrapper.find(TestOverflow).prop("items");
+            const overflowItems: TestItemProps[] = wrapper.find(TestOverflow).prop("items");
             assert.sameMembers(
                 overflowItems.map(it => it.id),
                 ids,
--- a/packages/core/test/utils.ts
+++ b/packages/core/test/utils.ts
@@ -16,7 +16,7 @@
 
 import { ReactWrapper } from "enzyme";
 
-import { Portal } from "../src";
+import { Portal, type PortalProps } from "../src";
 
 export function findInPortal<P>(overlay: ReactWrapper<P>, selector: string) {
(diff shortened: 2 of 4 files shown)


packages/core/test/overflow-list/overflowListTests.tsx
```diff
@@ -238,7 +238,7 @@ describe("<OverflowList>", function (this) {
 
         /** Assert ordered IDs of overflow items. */
         wrapper.assertOverflowItems = (...ids: number[]) => {
-            const overflowItems = wrapper.find(TestOverflow).prop("items");
+            const overflowItems: TestItemProps[] = wrapper.find(TestOverflow).prop("items");
```
jscheiny, line 243: Out of curiosity, what is the default type here?

packages/core/test/utils.ts
```diff
@@ -27,7 +27,9 @@ export function findInPortal<P>(overlay: ReactWrapper<P>, selector: string) {
 
     // React 15: unstable_renderSubtree does not preserve tree so we must create new wrapper.
     const portal = overlay.find(Portal).instance();
-    const portalChildren = new ReactWrapper(portal.props.children as React.JSX.Element[]);
+    const portalChildren = new ReactWrapper(
+        (portal as React.Component<PortalProps>).props.children as React.JSX.Element[],
```
jscheiny: If we're typing `portal` do we still need the cast for `children`?

packages/core/test/overflow-list/overflowListTests.tsx
ggdouglas: On `develop`, the default type is inferred as `TestItemProps[]` (as the code would lead you to expect).

However, in the React 18 upgrade branch, the type for `overflowItems` evaluates to `any`. I'm assuming this is an issue with the React 18 enzyme wrapper that we're using to support React 18 in these tests.

Screenshot 2025-01-14 at 14 39 49@2x

Screenshot 2025-01-14 at 14 40 10@2x

I'll add an inline code comment to reference this one.

packages/core/test/utils.ts
ggdouglas: In this specific case, yes. Again, I think there is something odd going on with the types from the enzyme wrapper (or perhaps I'm limited in my own understanding)

Breaking this down:

1.  `overlay.find(Portal).instance()` returns `React. Component<{}, {}, any>`

Screenshot 2025-01-14 at 14 59 32@2x

2. Since `portal.props` is of type `Readonly<{}>`, we get this error when trying to access `children`:

Screenshot 2025-01-14 at 15 03 30@2x

3. Even when we cast portal to `React.Component<PortalProps>`, `ReactWrapper` still expects to be passed a `nodes` argument of type `JSX.Element | JSX.Element[]`, which doesn't appear to be sympatico with the `ReactNode` type of `children`.

Screenshot 2025-01-14 at 15 06 00@2x

---

Perhaps we could keep the `React.Component<PortalProps>` cast and wrap children in a fragment to satisfy both the linter and enzyme?
```tsx
const portal = overlay.find(Portal).instance() as React.Component<PortalProps>;
const portalChildren = new ReactWrapper(<>{portal.props.children}</>);
```

packages/core/test/overflow-list/overflowListTests.tsx
ggdouglas: 46d1e0d69a7e11094514b19a3753f55ddbec3178

packages/core/test/utils.ts
ggdouglas: 76161cab24d8d4841b095728610d3e8b0d9ccbd3

jscheiny (approved):

PR #7164, opened 2025-01-14
chore: Remove remaining tslint references

jscheiny:
This removes remaining tslint references with the exception of not removing the `tslint-config` package. We will do that when we major version bump


ggdouglas (approved):
Nice! 👍

PR #7177, opened 2025-01-16
Update return types for Icon and SVGIconContainer

ggdouglas:

#### Proposed changes:

Modifies the return type of Icon and SVGIconContainer for compatibility with React 16/17/18.

### DIFF
--- a/packages/core/src/components/icon/icon.tsx
+++ b/packages/core/src/components/icon/icon.tsx
@@ -101,6 +101,11 @@ export interface DefaultIconProps extends IntentProps, Props, DefaultSVGIconProp
  * @see https://stackoverflow.com/a/73795494/7406866
  */
 export interface IconComponent extends React.FC<IconProps<Element>> {
+    /**
+     * ReturnType here preserves type compatability with React 16 while we migrate to React 18.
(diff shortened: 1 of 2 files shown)


packages/core/src/components/icon/icon.tsx
```diff
@@ -101,7 +101,7 @@ export interface DefaultIconProps extends IntentProps, Props, DefaultSVGIconProp
  * @see https://stackoverflow.com/a/73795494/7406866
  */
 export interface IconComponent extends React.FC<IconProps<Element>> {
-    <T extends Element = Element>(props: IconProps<T>): React.ReactElement | null;
+    <T extends Element = Element>(props: IconProps<T>): ReturnType<React.FC<IconProps<Element>>> | null;
```
gluxon, line 109: Might be worth a comment that this is written in a certain way to guarantee React 16 + 18 compatability!

gluxon (dismissed):

packages/core/src/components/icon/icon.tsx
ggdouglas: Done! bf41f431b07d31f469b88d0a2933097ff124dabd

Issue #6289, opened 2023-07-19
Refactor simple class components into functions


adidahiya:
#### Environment

- __Package version(s)__: @blueprintjs/core v5.x

#### Feature request

Request to refactor simple class components into functions, which would allow:

- Forwarding refs via the `ref` prop
- Spreading DOM attributes like `data-*` - https://github.com/palantir/blueprint/issues/3763
- Rendering as server components - https://github.com/palantir/blueprint/issues/6263#issuecomment-1746745749

As of @blueprintjs/core v5.1.x, these include:

- [x] Alert | https://github.com/palantir/blueprint/pull/7184
- [x] Breadcrumbs | https://github.com/palantir/blueprint/pull/7236
- [x] Callout | https://github.com/palantir/blueprint/pull/6800
- [x] Collapse | https://github.com/palantir/blueprint/pull/7437
- [x] Dialog | https://github.com/palantir/blueprint/pull/7438
- [x] DialogBody | https://github.com/palantir/blueprint/pull/7180
- [x] DialogFooter | https://github.com/palantir/blueprint/pull/7180
- [x] Divider | https://github.com/palantir/blueprint/pull/6814
- [ ] Drawer | https://github.com/palantir/blueprint/pull/6815
- [x] FileInput | https://github.com/palantir/blueprint/pull/6818
- [x] FormGroup | https://github.com/palantir/blueprint/pull/7185
- [ ] InputGroup | https://github.com/palantir/blueprint/pull/6819
- [x] KeyComboTag | https://github.com/palantir/blueprint/pull/7237
- [x] Menu | https://github.com/palantir/blueprint/pull/6827
- [x] MenuDivider | https://github.com/palantir/blueprint/pull/6826
- [x] Navbar | https://github.com/palantir/blueprint/pull/7183
- [x] NavbarDivider | https://github.com/palantir/blueprint/pull/7183
- [x] NavbarGroup | https://github.com/palantir/blueprint/pull/6830
- [x] NonIdealState | https://github.com/palantir/blueprint/pull/7235
- [x] ProgressBar | https://github.com/palantir/blueprint/pull/7182
- [x] RadioGroup | https://github.com/palantir/blueprint/pull/7186
- [x] Spinner | https://github.com/palantir/blueprint/pull/7181
- [ ] TextArea | https://github.com/palantir/blueprint/pull/6838

This would be a ⚠️ breaking change ⚠️

mm-wang:
- [ ] datetime package

Issue #7179, opened 2025-01-17
Infinite re-rendering of popover


jandrew44:
#### Environment

- __Package version(s)__:
@blueprintjs/core: 5.16.4
- __Operating System__: 
Windows 10 Home Edition
- __Browser name and version__:
Google Chrome 131.0.6778.266

#### Code Sandbox

Can be reproduced on the Blueprint docs page:
https://blueprintjs.com/docs/#core/components/popover

#### Steps to reproduce

1. Open the link mentioned above
2. Select "Hover" interaction kind
3. Reduce browser window width so that "Po" is the only visible part of the button "Popover target"
4. Move mouse cursor to the visible part of the button

https://github.com/user-attachments/assets/83f639bf-3551-4e79-a34a-8dd7293969bb

#### Actual behavior

Popover rendering loop.
Also in development environment I am getting a couple of runtime errors:

![Image](https://github.com/user-attachments/assets/776238b2-8bab-42d8-a9ff-1bc450134dbd)

#### Expected behavior

Single render of popover.

#### Possible solution

Set position: "left" for the popover.

rafaellichen:
this isn't reproducible in v6

PR #7181, opened 2025-01-17
chore: Refactor Spinner from a class-based to functional component

ggdouglas:
Part of #6289

#### Changes:

Refactors `Spinner` from a class-based to functional component. All functionality should remain preserved.

**Example:** [Before](https://blueprintjs.com/docs/#core/components/spinner) / [After](https://output.circle-artifacts.com/output/job/3b3263af-6c0c-45b5-bdd0-77eb0b146340/artifacts/0/packages/docs-app/dist/index.html#core/components/spinner)

### DIFF
--- a/packages/core/src/components/spinner/spinner.tsx
+++ b/packages/core/src/components/spinner/spinner.tsx
@@ -85,21 +85,7 @@ export const Spinner: React.FC<SpinnerProps> = props => {
         }
     }, [className, size]);
 
-    /**
-     * Resolve size to a pixel value.
-     * Size can be set by className, props, default, or minimum constant.
-     */
-    const sizePx = React.useMemo(() => {
-        if (size == null) {
-            if (className.indexOf(Classes.SMALL) >= 0) {
-                return SpinnerSize.SMALL;
-            } else if (className.indexOf(Classes.LARGE) >= 0) {
-                return SpinnerSize.LARGE;
-            }
-            return SpinnerSize.STANDARD;
-        }
-        return Math.max(MIN_SIZE, size);
-    }, [className, size]);
+    const sizePx = getSize(size, className);
 
     // keep spinner track width consistent at all sizes (down to about 10px).
     const strokeWidth = Math.min(MIN_STROKE_WIDTH, (STROKE_WIDTH * SpinnerSize.LARGE) / sizePx);
@@ -152,3 +138,19 @@ export const Spinner: React.FC<SpinnerProps> = props => {
 };
 
 Spinner.displayName = `${DISPLAYNAME_PREFIX}.Spinner`;
+
+/**
+ * Resolve size to a pixel value.
+ * Size can be set by className, props, default, or minimum constant.
+ */
+const getSize = (size: number | undefined, className: string): number => {
+    if (size == null) {
+        if (className.indexOf(Classes.SMALL) >= 0) {
+            return SpinnerSize.SMALL;
+        } else if (className.indexOf(Classes.LARGE) >= 0) {
+            return SpinnerSize.LARGE;
+        }
+        return SpinnerSize.STANDARD;
+    }
+    return Math.max(MIN_SIZE, size);
+};
--- a/packages/core/src/components/spinner/spinner.tsx
+++ b/packages/core/src/components/spinner/spinner.tsx
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015 Palantir Technologies, Inc. All rights reserved.
+ * Copyright 2025 Palantir Technologies, Inc. All rights reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -17,7 +17,7 @@
 import classNames from "classnames";
 import * as React from "react";
 
-import { AbstractPureComponent, Classes } from "../../common";
+import { Classes } from "../../common";
 import { SPINNER_WARN_CLASSES_SIZE } from "../../common/errors";
 import { DISPLAYNAME_PREFIX, type IntentProps, type Props } from "../../common/props";
 import { clamp } from "../../common/utils";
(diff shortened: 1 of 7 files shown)


packages/core/src/components/spinner/spinner.tsx
```diff
@@ -159,13 +99,56 @@ export class Spinner extends AbstractPureComponent<SpinnerProps> {
...
+
+    const classes = classNames(
+        Classes.SPINNER,
+        Classes.intentClass(intent),
+        { [Classes.SPINNER_NO_SPIN]: value != null },
+        className,
+    );
+
+    // multiple DOM elements around SVG are necessary to properly isolate animation:
+    // - SVG elements in IE do not support anim/trans so they must be set on a parent HTML element.
```
ggdouglas, line 105: > // - SVG elements in IE do not support anim/trans so they must be set on a parent HTML element.

Not sure if this comment is still relevant since we haven't supported IE11 for a while now. We could potentially remove this nesting complexity in a separate body of work. Regardless, I have tested these changes to Spinner in Edge.

packages/core/src/components/spinner/spinner.tsx
```diff
@@ -76,81 +76,21 @@ export interface SpinnerProps<T extends HTMLElement = HTMLElement> extends Props
...
-    }
+    }, [className, size]);
 
     /**
      * Resolve size to a pixel value.
      * Size can be set by className, props, default, or minimum constant.
      */
-    private getSize() {
-        const { className = "", size } = this.props;
+    const sizePx = React.useMemo(() => {
```
jscheiny: This component isn't memoized and this doesn't look expensive, I think we should remove the `useMemo` here and either extract this as a function below the component or make this an IIFE

packages/core/src/components/spinner/spinner.tsx
```diff
@@ -159,13 +99,56 @@ export class Spinner extends AbstractPureComponent<SpinnerProps> {
...
-    }
+    }, [className, size]);
+
+    // keep spinner track width consistent at all sizes (down to about 10px).
+    const strokeWidth = Math.min(MIN_STROKE_WIDTH, (STROKE_WIDTH * SpinnerSize.LARGE) / sizePx);
+    const strokeOffset = PATH_LENGTH - PATH_LENGTH * (value == null ? 0.25 : clamp(value, 0, 1));
 
     /** Compute viewbox such that stroked track sits exactly at edge of image frame. */
-    private getViewBox(strokeWidth: number) {
+    const viewBox = React.useMemo(() => {
```
jscheiny: Same note about `useMemo` here

packages/core/src/components/spinner/spinner.tsx
```diff
@@ -76,81 +76,21 @@ export interface SpinnerProps<T extends HTMLElement = HTMLElement> extends Props
...
-                        strokeDasharray={`${PATH_LENGTH} ${PATH_LENGTH}`}
-                        strokeDashoffset={strokeOffset}
-                    />
-                </svg>,
-            ),
-        );
-    }
-
-    protected validateProps({ className = "", size }: SpinnerProps) {
+    React.useEffect(() => {
```
jscheiny: This was happening in `validateProps` before which is only called if the node env is not production: https://github.com/palantir/blueprint/blob/f374289ad883ff00f71b84cc77180183f0166a5e/packages/core/src/common/abstractComponent.ts#L53-L55

We should make sure we're not doing this check in production. Maybe we can add a `useValidateProps` hooks that only calls the callback if the env is not production?

ggdouglas: Good catch! Will add this 👍

packages/core/src/components/spinner/spinner.tsx
ggdouglas: d04cc1da26e00cd49f4b23e72df64dadc94eb7df

packages/core/src/components/spinner/spinner.tsx
ggdouglas: ef212d93b9f492ac2c0d48ffb7217c801eaf145a

packages/core/src/components/spinner/spinner.tsx
ggdouglas: Extracted to a hook here: b36f43a819e965447d596173bbbeb0cca2686f6b

packages/demo-app/src/examples/Examples.tsx
```diff
@@ -80,6 +81,7 @@ const ExamplesContainer: React.FC<{ isDark?: boolean }> = ({ isDark = false }) =
             <NonIdealStateExample />
             <PopoverExample />
             <SliderExample />
+            <SpinnerExample />
```
ggdouglas, line 84: I noticed that we didn't have demo examples for Spinner, and I wanted to test the validation warnings locally, so I added some examples.

packages/demo-app/src/examples/TooltipExample.tsx
```diff
@@ -31,7 +31,6 @@ export const TooltipExample = React.memo(() => {
                         Always open tooltip
                     </span>
                 }
-                isOpen={true}
```
ggdouglas, line 34: Discovered this while running demo-app. This appears to be crashing things. possibly related to #7179?

CleanShot 2025-01-17 at 16 23 16@2x

packages/core/test/hooks/useValidatePropsTests.tsx
```diff
@@ -0,0 +1,47 @@
...
+        return null;
+    };
+
+    it("calls validator in development environment", () => {
+        render(<TestComponent />);
+        expect(validatorSpy.called).to.be.true;
+    });
+
+    it.skip("does not call validator in production environment", () => {
+        // TODO: figure out how to test this
```
ggdouglas, line 30: FWIW, I tried mocking out `process.env.NODE_ENV` to `"production"` in these tests, but ran into trouble with karma/webpack. I'm open to writing this test if there is an easy/standard way of doing this, but I'm not going to spend too much time worrying about it at the moment.

Setting `process.env.NODE_ENV = "production";` (or to any value) causes tests to emit the following error:
```
"SyntaxError: Invalid left-hand side in assignment"
```

I tried wrapping with a test helper to no avail:
```ts
export function withProductionEnvironment(testFn: () => void) {
    const originalNodeEnv = process.env.NODE_ENV;
    Object.defineProperty(process, 'env', {
        value: { ...process.env, NODE_ENV: 'production' }
    });
    testFn();
    Object.defineProperty(process, 'env', {
        value: { ...process.env, NODE_ENV: originalNodeEnv }
    });
}
```

jscheiny (approved):

Issue #3763, opened 2019-10-01
Spread data-* attributes onto elements


roryg11:
#### Feature request

Currently not all blueprint components propagate their data attributes to the rendered element and it would be great if they would, but for now we'd like to request data-* attributes on the Tab component.

We will can make more feature requests on other elements as we discover which ones are important. 😄

rootulp:
+1 in order to query Blueprint elements via [`getByTestId`](https://testing-library.com/docs/dom-testing-library/api-queries#bytestid) which uses the attribute `data-testid`.

eastside:
+1 for this

It would be really nice to get this out. It's extremely helpful for integration tests to be able to store/access data-* attributes. Without it, I'm stuck with rather ugly hacks, like folding data into classes.

adidahiya:
This was fixed for the Tab component in https://github.com/palantir/blueprint/pull/3766.

If there are other components where you need this feature, please list those out here.

eastside:
I believe the Select<> component needs it as well.

adidahiya:
I'd be open to a PR for this feature in `<Select>`, but it's not as simple as `<Tab>` because it's a more complicated component.

Can you show me an example of what you're doing in your test code? I'm potentially concerned that simply creating an easy selector for the wrapper element of a Select will not be sufficient for tests, and that you may end up reaching into the internals of the component, at which point you might as well use the exported class names:

```ts
import { Classes } from "@blueprintjs/select";
```

petejkim:
@adidahiya Please make it possible for `<Alert>` to have a `data-testid` as well, and it would be even better if we can assign `data-testid` for the cancel and confirm button in the `<Alert>` dialog.

luigilake:
@adidahiya I agree with @petejkim . Could it possible to send in a `data-h` (`dataH`) into `<Alert>` specifically for the `Cancel` and `Confirm/[CUSTOM TEXT]` buttons? Thank you!

adidahiya:
Btw we fixed this for Spinner in https://github.com/palantir/blueprint/pull/5321

bossa-jpg:
Please make it possible for `<TimePicker>` and `<NonIdealState>` components as well.

ghost:
It would be nice to include it for `<MultiSelect2>` aswell.

adidahiya:
> Please make it possible for `<TimePicker>` and `<NonIdealState>` components as well.

[...]

PR #7182, opened 2025-01-17
chore: Refactor ProgressBar from a class to functional component

ggdouglas:
Part of #6289

#### Changes:

Refactors `ProgressBar` from a class-based to functional component. All functionality should remain preserved.

Also allows `htmlProps` to be spread to component to allow passing of `data-*` props. (see #3763)

**Example:** [Before](https://blueprintjs.com/docs/#core/components/progress-bar) / [After](https://output.circle-artifacts.com/output/job/565af5a7-8830-4196-b9db-1b43dc45c20b/artifacts/0/packages/docs-app/dist/index.html#core/components/progress-bar)

### DIFF
--- a/packages/core/src/components/progress-bar/progressBar.tsx
+++ b/packages/core/src/components/progress-bar/progressBar.tsx
@@ -17,11 +17,11 @@
 import classNames from "classnames";
 import * as React from "react";
 
-import { AbstractPureComponent, Classes } from "../../common";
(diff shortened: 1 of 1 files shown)


jscheiny (approved):

p--b:
FYI as an added bonus, this also makes it possible for the progressbar to pass the `aria-progressbar-name` rule, which afaict was not easily possible before.

PR #7197, opened 2025-01-24
Add `variant` prop to Button and ButtonGroup

ggdouglas:
#### Proposed changes:

Adds a new `variant` prop to replace the (now deprecated) `minimal` and `outlined` props on `Button`/`AnchorButton` and `ButtonGroup`.


packages/docs-app/src/examples/core-examples/buttonExamples.tsx
```diff
@@ -51,12 +65,12 @@ export const ButtonOutlinedExample: React.FC<ExampleProps> = props => {
     const code = dedent`
         <Button text="Outlined" outlined={true} />
         <Button text="Primary" outlined={true} intent="primary" />
-        <Button text="Disabled" minimal={true} disabled={true} />`;
```
ggdouglas, line 54: This was a typo in the current Button outlined example (unrelated to the rest of these changes)

packages/core/src/common/errors.ts
```diff
@@ -120,3 +120,11 @@ export const OVERLAY_WITH_MULTIPLE_CHILDREN_REQUIRES_CHILD_REFS =
     ns + ` <Overlay2> requires childRefs prop when rendering multiple child elements`;
 export const OVERLAY_CHILD_REQUIRES_KEY =
     ns + ` <Overlay2> requires each child element to have a unique key prop when childRefs is used`;
+
+// prop deprecation warnings
```
ggdouglas: There will be quite a few other warnings related to prop deprecations in upcoming PRs, so creating a new section to house them here. https://github.com/palantir/blueprint/blob/61b79df070ba3e432bce2ec8a3d42ccca4f37aed/packages/core/src/common/errors.ts#L130-L151

packages/core/src/components/button/buttons.tsx
```diff
@@ -94,8 +106,8 @@ function useSharedButtonAttributes<E extends HTMLAnchorElement | HTMLButtonEleme
             [Classes.FILL]: fill,
             [Classes.LARGE]: large,
             [Classes.LOADING]: loading,
-            [Classes.MINIMAL]: minimal,
-            [Classes.OUTLINED]: outlined,
+            [Classes.MINIMAL]: minimal || variant === "minimal",
+            [Classes.OUTLINED]: outlined || variant === "outlined",
```
jscheiny: This bit of code has some unintended side effects. I think that we should make the behavior be that `variant` supersedes both `outlined` and `minimal`.  However, if we do the following:

```tsx
<Button text="Button" variant="minimal" outlined={true} />
```

We end up with an outlined button instead of a minimal button as would be hoped for. The same issue occurs with `ButtonGroup`s.

ggdouglas: Technically, this behavior [already exists](https://codesandbox.io/p/sandbox/button-minimal-outlined-d99k9v) between the `minimal` and `outlined` props where `outlined` overwrites the styles of `minimal` if both are applied simultaneously.
```tsx
<Button text="Button" minimal={true} outlined={true} />
```

I think it is fair, however, that we wouldn't want to continue this behavior!

a2889bd88b73f8372e994f20c562763d47f1a330 wraps the logic into a utility function, providing the proper application of classes based on precedence. ~~This technically changes the behavior of the presence of classes (for the case described above), but in practice it doesn't matter much since, as we've established, the `outlined` classes overwrite `minimal` anyways.~~

**Edit:** fd62b36890579b440eb28e1b74562e821e067868 preserves the existing behavior

jscheiny (dismissed):
LGTM! Excited about this change

packages/core/src/components/button/button-group.md
```diff
@@ -22,14 +22,29 @@ visual style, it’s recommended to apply the same `intent` to all buttons withi
 
 @reactCodeExample ButtonGroupIntentExample
 
+@## Variant
+
+Most of **ButtonGroup**'s props are also supported by [**Button**](#core/components/buttons) directly. Setting these props on **ButtonGroup** will apply the same value to all buttons in the group. Note that most modifiers, once enabled on the group, cannot be overridden on child buttons (due to the cascading nature of CSS).
```
jscheiny: This bit feels a bit incongruous in this position, which is refering to a specific prop. Its also duplicated in the outlined and minimal section below. I think we should take this section and move it to the top of the page and remove it from both of these sections.

ggdouglas: d1befe6fc3d2477170d238b832cc4ac184290dcd

jscheiny (dismissed):

bvandrc:
`Tag` component has deprecated `size` but not deprecated variants.

PR #7226, opened 2025-02-04
Add `size` prop to Button and ButtonGroup

ggdouglas:
#### Proposed changes:

Adds a new `size` prop to replace the (now deprecated) `small` and `large` props on `Button`/`AnchorButton` and `ButtonGroup`.


packages/core/src/common/classes.ts
```diff
@@ -439,3 +440,20 @@ export function positionClass(position: Position | undefined) {
     }
     return `${NS}-position-${position}`;
 }
+
+export function sizeClass(
+    size: Size | Omit<Size, "small">,
```
jscheiny:
```suggestion
    size: Size,
```

This type is equivalent, a type of `Omit<Size, "small">` should be assignable to `Size`.

packages/core/src/common/errors.ts
```diff
@@ -120,3 +120,16 @@ export const OVERLAY_WITH_MULTIPLE_CHILDREN_REQUIRES_CHILD_REFS =
     ns + ` <Overlay2> requires childRefs prop when rendering multiple child elements`;
 export const OVERLAY_CHILD_REQUIRES_KEY =
     ns + ` <Overlay2> requires each child element to have a unique key prop when childRefs is used`;
+
+function deprecatedSizeWarning(component: string, size: "large" | "small") {
```
jscheiny: Thoughts on just using this function directly from components? I feel like we're going to have a lot of constants resulting from this that will all just be used in one place, might be nicer to just use this function in `useValidateProps`.

jscheiny: If we wanted to get extra fancy we could do something like:


```ts
function logDeprecatedSizeWarning(
    component: string,
    props: Partial<Record<"large" | "small", boolean>>,
) { ... }
```

which would automatically log the right messages based on the passed props which are defined. So in components we could do something like:

```ts
    useValidateProps(() => {
        logDeprecatedSizeWarning("Button", { large, small });
    }, [large, small]);
```

Just a thought, might be too much for a simple case like this.

packages/core/src/common/classes.ts
```diff
@@ -439,3 +440,20 @@ export function positionClass(position: Position | undefined) {
...
+    large: boolean = false,
+    small: boolean = false,
```
jscheiny: Having these two boolean optional props next to each other feels ripe for being messed up. I would suggest using an object parameter here:
```suggestion
    legacyProps: Partial<Record<"large" | "small", boolean>>,
```

packages/core/src/common/errors.ts
ggdouglas: Fair point, I like it! Added in 098390192fc65445791daed58bc49d862142a4aa

packages/core/src/common/classes.ts
ggdouglas: bba8d5aec50ccfb1791117edf6bf63dee97a570c

packages/core/src/common/classes.ts
ggdouglas: bba8d5aec50ccfb1791117edf6bf63dee97a570c

packages/core/src/common/errors.ts
```diff
@@ -138,3 +138,17 @@ export const OVERLAY_WITH_MULTIPLE_CHILDREN_REQUIRES_CHILD_REFS =
...
+    if (large && small) {
+        console.warn(
+            ns +
+                ` <${component}> large and small props are mutually exclusive. Please use size="large" or size="small" instead.`,
+        );
+    } else if (large) {
+        console.warn(ns + ` <${component}> large is deprecated. Please use size="large" instead.`);
+    } else if (small) {
+        console.warn(ns + ` <${component}> small is deprecated. Please use size="small" instead.`);
+    }
```
jscheiny: I think we want to log if these props are non null rather than just true. That way if someone is doing something like `<Button small={someBooleanVariable} />` we properly log even when passing `false`.

ggdouglas: Ah, yeah, good call. Added in 5337c009ee15aa86138578302443ac0d2eb29de6

jscheiny (approved):

PR #7228, opened 2025-02-04
Add `size` prop to more components

ggdouglas:
#### Proposed changes:

FLUP to #7226 that adds the `size` prop to the following components:
- `<FileInput>`
- `<InputGroup>`
- `<NumericInput>`
- `<TextArea>`
- `<Menu>`
- `<SegmentedControl>`

### DIFF
--- a/packages/core/src/components/forms/inputGroup.tsx
+++ b/packages/core/src/components/forms/inputGroup.tsx
@@ -18,13 +18,14 @@ import classNames from "classnames";
 import * as React from "react";
 
 import { AbstractPureComponent, Classes } from "../../common";
-import * as Errors from "../../common/errors";
+import { INPUT_WARN_LEFT_ELEMENT_LEFT_ICON_MUTEX, logDeprecatedSizeWarning } from "../../common/errors";
 import {
     type ControlledValueProps,
     DISPLAYNAME_PREFIX,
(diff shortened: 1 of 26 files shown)


packages/core/src/components/forms/inputGroup.tsx
```diff
@@ -117,8 +145,8 @@ export class InputGroup extends AbstractPureComponent<InputGroupProps, InputGrou
                 [Classes.DISABLED]: disabled,
                 [Classes.READ_ONLY]: readOnly,
                 [Classes.FILL]: fill,
-                [Classes.LARGE]: large,
-                [Classes.SMALL]: small,
+                [Classes.LARGE]: large || size === "large",
+                [Classes.SMALL]: small || size === "small",
```
jscheiny: Should we be using `sizeClass()` here?

jscheiny (commented):
This PR in general seems to not be using the `sizeClass` function from its parent PR.

packages/core/src/components/forms/inputGroup.tsx
ggdouglas: Rebased to include changes and utilities from previous base branches

jscheiny (dismissed):

PR #7230, opened 2025-02-05
fix(CompoundTag): Right icon alignment

### LINKED ISSUE #7203: CompoundTag rightElement & remove icon not aligned to right when fill is enabled
#### Environment
BlueprintJS docs:
- https://blueprintjs.com/docs/#core/components/tag
- https://blueprintjs.com/docs/#core/components/compound-tag

#### Steps to reproduce

Enable the following options for the Tag and CompoundTag components on the BlueprintJS docs site:
- Fill
- Removable
- Right icon

#### Actual behavior

Right icon and the remove icon is aligned to the right in the Tag component.
It is not in CompoundTag component.

#### Expected behavior

These elements should be aligned to the right in the CompoundTag component.

#### Possible solution

CompoundTag should have the same rules present for Tag in `.bp5-tag > .bp5-fill`.

jscheiny:

Looks like at some point the `$ns-compound-tag-right-text` class name was renamed to `$ns-compound-tag-right-content` but the CSS was never properly updated.

#### Screenshot

Screenshot 2025-02-05 at 10 14 36 AM

### DIFF
--- a/packages/core/src/components/tag/_compound-tag.scss
+++ b/packages/core/src/components/tag/_compound-tag.scss
@@ -75,7 +75,7 @@ $compound-tag-left-intent-colors: (
       margin-left: 4px;
     }
 
-    .#{$ns}-compound-tag-right-text {
+    .#{$ns}-compound-tag-right-content {
       flex-grow: 1;
     }
 


ggdouglas (approved):
LGTM 👍

PR #7234, opened 2025-02-05
Add `size` prop to `large` variant components

ggdouglas:
#### Proposed changes:

FLUP to #7226 that adds the `size` prop to the following components:
- `<Checkbox>`
- `<CompoundTag>`
- `<Radio>`
- `<Switch>`
- `<Tabs>`
- `<Tag>`
- `<TagInput>`

### DIFF
--- a/packages/core/src/common/classes.ts
+++ b/packages/core/src/common/classes.ts
@@ -447,7 +447,7 @@ export function positionClass(position: Position | undefined) {
 }
 
 export function sizeClass(
-    size: Size,
+    size: Size | Omit<Size, "small">,
     legacyProps: Partial<Record<"large" | "small", boolean>>,
 ): string | Record<string, boolean> {
     if (size === "small") {
--- a/packages/core/src/common/classes.ts
+++ b/packages/core/src/common/classes.ts
@@ -19,7 +19,7 @@ import type { ButtonVariant } from "./buttonVariant";
 import { Elevation } from "./elevation";
 import { Intent } from "./intent";
 import { Position } from "./position";
-import type { Size } from "./size";
+import type { NonSmallSize, Size } from "./size";
 
 // injected by webpack.DefinePlugin
 declare let BLUEPRINT_NAMESPACE: string | undefined;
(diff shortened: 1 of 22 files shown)


packages/core/src/components/control-card/controlCard.tsx
```diff
@@ -91,7 +91,7 @@ export const ControlCard: React.FC<ControlCardProps> = React.forwardRef((props,
 
     // use a container element to achieve a good flex layout
     const labelElement = <div className={Classes.CONTROL_CARD_LABEL}>{children ?? label}</div>;
-    const controlProps: ControlProps = {
+    const controlProps: Omit<ControlProps, "size"> = {
```
jscheiny: This type feels used often enough that I think it would be worth naming it so that consumers can interact with it directly, rather than having to type out the clunk `Omit` type or the props index type (e.g. `ControlProps["size"]`). The only issue is what to call it? I think something simple like `NonSmallSize` is probably sufficient, but definitely open to bikeshedding on this.

jscheiny: @ggdouglas Any further thoughts on this?

ggdouglas: Not sure I quite understand what you mean. This particular `Omit` is needed due to [`size`](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/size) being a valid attribute on HTML input, which comes in conflict with the `"size"` attribute that we are adding to controls. Without it, we get the following compilation error:
```
Types of property 'size' are incompatible.
    Type 'number | undefined' is not assignable to type 'Omit<Size, "small"> | undefined'.
      Type 'number' is not assignable to type 'Omit<Size, "small">'.
```

ggdouglas: 0fc046db9dbbc3afeaa799fb4e59f12c32241cf3

packages/core/src/common/classes.ts
```diff
@@ -447,7 +447,7 @@ export function positionClass(position: Position | undefined) {
 }
 
 export function sizeClass(
-    size: Size | Omit<Size, "small">,
+    size: Size | NonSmallSize,
```
jscheiny: This is equivalent to just `Size`, we can simplify this.

jscheiny (dismissed):
One minor nit, approving

packages/core/src/common/classes.ts
ggdouglas: f6716efb581069d04a00e9567818e82516bba303

jscheiny (approved):

PR #7237, opened 2025-02-05
[chore] Refactor KeyComboTag to a functional component

ggdouglas:
Refactors `<KeyComboTag>` from a class-based to a functional component.

[Before](https://blueprintjs.com/docs/#core/hooks/use-hotkeys.key-combo-tester) | [After](https://output.circle-artifacts.com/output/job/9c912c99-b919-4b28-83e3-5de7e938355a/artifacts/0/packages/docs-app/dist/index.html#core/hooks/use-hotkeys.key-combo-tester)

**Part of:**
- #6289

**Supersedes:**
- https://github.com/palantir/blueprint/pull/6822


packages/core/src/components/hotkeys/keyComboTag.tsx
```diff
@@ -76,50 +76,55 @@ interface KeyComboTagInternalProps extends KeyComboTagProps {
...
+    const getKeyIcon = useCallback(
+        (key: string) => {
+            const icon = KEY_ICONS[key.toLowerCase()];
+            if (icon?.isMacOnly && !isMac(platformOverride)) {
+                return undefined;
+            }
+            return icon;
+        },
+        [platformOverride],
+    );
```
cameronjoyner, line 91: I could be mistaken, but isn't this a good opportunity for `useMemo`?

```suggestion
    const getKeyIcon = useMemo(
        (key: string) => {
            const icon = KEY_ICONS[key.toLowerCase()];
            if (icon?.isMacOnly && !isMac(platformOverride)) {
                return undefined;
            }
            return icon;
        },
        [platformOverride],
    );
```

packages/core/src/components/hotkeys/keyComboTag.tsx
```diff
@@ -76,50 +76,55 @@ interface KeyComboTagInternalProps extends KeyComboTagProps {
...
+
+    const renderKey = useCallback(
```
cameronjoyner, line 93: I would also argue for a `useMemo` here, but thats based solely on my own experiences and preferences.

packages/core/src/components/hotkeys/keyComboTag.tsx
```diff
@@ -76,50 +76,55 @@ interface KeyComboTagInternalProps extends KeyComboTagProps {
...
+        (key: string, index: number) => {
+            const keyString = DISPLAY_ALIASES[key.toLowerCase()] ?? key;
+            const icon = getKeyIcon(key);
+            const reactKey = `key-${index}`;
+            return (
+                <kbd className={classNames(Classes.KEY, { [Classes.MODIFIER_KEY]: icon != null })} key={reactKey}>
+                    {icon != null && <Icon icon={icon.icon} title={icon.iconTitle} />}
+                    {keyString}
+                </kbd>
```
cameronjoyner, line 102: @ggdouglas other libraries expose a proper `KBD` component. Have we ever considered this?

packages/core/src/components/hotkeys/keyComboTag.tsx
```diff
@@ -76,50 +76,55 @@ interface KeyComboTagInternalProps extends KeyComboTagProps {
...
-        return (
-            <kbd className={classNames(Classes.KEY, { [Classes.MODIFIER_KEY]: icon != null })} key={reactKey}>
-                {icon != null && <Icon icon={icon.icon} title={icon.iconTitle} />}
-                {keyString}
-            </kbd>
+        },
+        [getKeyIcon],
+    );
+
+    const renderMinimalKey = useCallback(
```
cameronjoyner, line 108: 👀 
```suggestion
    const renderMinimalKey = useMemo(
```

packages/core/src/components/hotkeys/keyComboTag.tsx
```diff
@@ -76,50 +76,55 @@ interface KeyComboTagInternalProps extends KeyComboTagProps {
...
-    };
-
-    private getKeyIcon(key: string) {
-        const { platformOverride } = this.props;
-        const icon = KEY_ICONS[key.toLowerCase()];
-        if (icon?.isMacOnly && !isMac(platformOverride)) {
-            return undefined;
-        }
-        return icon;
-    }
```
cameronjoyner, line 122: This function previously came after its use in code. @mm-wang has encouraged me to follow that pattern. Can we revert back to its original placement in code?

cameronjoyner (commented):
Ok - as I spoke with my close friend, Claude, he actually disagrees with my suggestion to opt for `useMemo`. He takes your side on the use of `useCallback` instead 😓 😰

cameronjoyner (dismissed):

packages/core/src/components/hotkeys/keyComboTag.tsx
ggdouglas: With hooks, this isn't possible. Both `renderKey` and `renderMinimalKey` reference `getKeyIcon` in their `useCallback` dependency arrays, so it has to be declared first. Class methods hoisted so ordering didn't matter, but hook declarations are sequential. I can reorder the others if there's a preferred ordering among `renderKey`/`renderMinimalKey`, but `getKeyIcon` has to come before both.

packages/core/src/components/hotkeys/keyComboTag.tsx
ggdouglas: Yes, FWIW I have thought about this a bit in the past. Opened https://github.com/palantir/blueprint/issues/7973 as a quick issue to consider this for the future. Feel free to add any additional thoughts.

cameronjoyner: TY!

cameronjoyner (approved):

PR #7251, opened 2025-02-11
Revert "Switch from @import to @use (#7157)"

jscheiny:

In https://github.com/palantir/blueprint/pull/7157 we switched our sass from `@import` to `@use` across the board, however this has the side effect of not re-exporting the imported variables to consumers of that sass file. For our `_variables.scss` we need our consumers to have these variables since that's the contract of this file. As a result consumers of core were failing to compile unless they took an explicit dependency on the colors package. Temporarily reverting while we look for a longer term fix.


ggdouglas (approved):

justinbhopper:
@jscheiny @ggdouglas I imagine we might still be theorizing on what the correct long term approach is?

Let me know if I can help brainstorm on ideas.

PR #7252, opened 2025-02-11
Upgrade to ESLint 9

MariaSolOs:

- Upgrade ESLint dependencies to v9.
- Update the ESLint config and plugin packages to the new format.
- Update all configurations file to the new `eslint.config.js` format.

- That the linting experience in this repo hasn't changed.
- That no unexpected regressions in the ESLint Blueprint libraries emerge (but I guess this will be caught by CI?)


.yarn/patches/eslint-plugin-header-npm-3.1.1-402265ec6f.patch
```diff
@@ -0,0 +1,14 @@
+diff --git a/lib/rules/header.js b/lib/rules/header.js
```
MariaSolOs, line 1: Needed to patch the `header` plugin because it doesn't support ESLint 9 yet: https://github.com/Stuk/eslint-plugin-header/issues/59

eslint.config.js
```diff
@@ -0,0 +1,78 @@
+/*
```
MariaSolOs, line 1: The migration of the configuration files was based on https://eslint.org/docs/latest/use/configure/migration-guide#migrate-your-config-file and https://typescript-eslint.io/getting-started/.

packages/eslint-config/eslint-builtin-rules.js
```diff
@@ -1,10 +1,14 @@
-{
+/* !
```
MariaSolOs, line 1: Made these JavaScript files instead of JSON because importing JSON it's still an experimental JS feature.

packages/node-build-scripts/package.json
```diff
@@ -23,8 +23,9 @@
         "cross-spawn": "^7.0.3",
         "css-select": "^5.1.0",
         "dom-serializer": "^2.0.0",
-        "eslint": "^8.57.0",
+        "eslint": "^9.20.0",
         "eslint-formatter-codeframe": "^7.32.1",
+        "eslint-formatter-junit": "^8.3.0",
```
MariaSolOs, line 28: Needed to install this formatter since ESLint removed it from the builtin ones: https://eslint.org/docs/latest/use/migrate-to-9.0.0#-removed-multiple-formatters

packages/select/eslint.config.js
```diff
@@ -0,0 +1,9 @@
+const config = require("../../eslint.config.js");
+const tseslint = require("typescript-eslint");
+
+module.exports = tseslint.config([
+    config,
+    {
+        ignores: ["*.js", "*.d.ts"],
```
MariaSolOs, line 7: Added this configuration to replace the `.eslintignore` file since ESLint no longer uses it: https://eslint.org/docs/latest/use/configure/migration-guide#ignoring-files

packages/datetime/src/components/date-input/dateInput.tsx
```diff
@@ -20,7 +20,7 @@
  * package instead.
  */
 
-/* eslint-disable @typescript-eslint/no-deprecated, @blueprintjs/no-deprecated-components, react-hooks/exhaustive-deps */
+/* eslint-disable @typescript-eslint/no-deprecated, react-hooks/rules-of-hooks, react-hooks/exhaustive-deps */
```
MariaSolOs: Had to disable `rules-of-hooks` in some components because some have names starting with underscore, and the linting rule enforces hook names to begin with a capital letter.

ggdouglas: Doesn't need to be this PR, but we should follow up on these components to rename them to something more compliant with this rule. Having it be disabled is a bit sub-optimal given how important it is.

ggdouglas: Opening #7259 to address this separately

packages/icons/src/loaderUtils.ts
```diff
@@ -29,7 +29,6 @@ export async function wrapWithTimer(taskDescription: string, task: () => Promise
     const shouldMeasure = isNodeEnv("development") && typeof performance !== "undefined";
     let start: number;
 
-    /* eslint-disable no-console */
```
ggdouglas, line 32: Why are we able to remove this ignored block?

packages/eslint-config/index.js
```diff
@@ -13,70 +13,84 @@
...
+const tsEslintRules = require("./typescript-eslint-rules.js");
 
-const eslintBuiltinRules = require("./eslint-builtin-rules.json");
-const eslintPluginRules = require("./eslint-plugin-rules.json");
-const tsEslintRules = require("./typescript-eslint-rules.json");
-
-/**
- * Enables @blueprintjs/eslint-plugin.
- *
- * For TS files, configure typescript-eslint, including type-aware lint rules which use the TS program.
```
ggdouglas, line 25: **nit:** Is this comment still relevant? Can we keep it?

packages/icons/src/loaderUtils.ts
MariaSolOs: I think the disable pragma is already unnecessary as you configure it to allow `info` [here](https://github.com/palantir/blueprint/blob/3f805caf40fd676dc2369daab759aa8d82221c3a/packages/eslint-config/eslint-builtin-rules.json#L20). In this PR I fixed all linting errors and warnings that I saw on CI assuming that they were new errors. I can move this to a separate PR if you want :)

packages/eslint-config/index.js
MariaSolOs: I had removed it since this configuration does much more than enabling the Blueprint plugin and configuring `@typescript-eslint` so I found it a bit misleading (and the configuration feels self-explanatory), but I can keep it if you prefer.

packages/datetime/src/components/date-input/dateInput.tsx
MariaSolOs: Sounds good to me! I will cherry-pick your commits and see if that's enough to satisfy the rule. If so, I'll let you know so that you can merge #7259 and I'll rebase later.

MariaSolOs: @ggdouglas I confirm that I can drop the last commit with your PR :)

packages/eslint-config/index.js
ggdouglas: Ah, no, we can leave it removed in that case. Just wanted to make sure that the drop was intentional.

packages/icons/src/loaderUtils.ts
ggdouglas: Ah, I hadn't noticed that configuration. Thanks for the explanation! I think we can leave this in the current PR.

ggdouglas (approved):

Issue #7193, opened 2025-01-22
Proposal: Replace Physical Direction Values with Logical Equivalents


ggdouglas:
## Summary

To improve support for right-to-left (RTL) languages like Hebrew and Arabic, we propose replacing physical direction values (`"left"` and `"right"`) with logical equivalents (`"start"` and `"end"`) across our components. This aligns with modern CSS practices, specifically the use of [CSS Logical Properties and Values](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_logical_properties_and_values), and enhances consistency in our component APIs.

## Problem Statement

Our components currently use physical direction values (`"left"` and `"right"`) to specify the alignment of text or indicators (e.g., checkboxes, radio buttons) relative to labels. These values become confusing in RTL languages, where layout directions are reversed.

We also have an [`Alignment`](https://github.com/palantir/blueprint/blob/9c16d3a053f347d748ca22cebcec6f9a06a73590/packages/core/src/common/alignment.ts#L18-L22) type shared across components, but its valid values differ. For example:

- In `<Button>`, the `alignText` prop accepts `"left"`, `"right"`, or `"center"`.
- In `<Checkbox>`, the `alignIndicator` prop uses the same `Alignment` type but only `"left"` or `"right"` are valid. Passing `"center"` has no effect since center alignment isn't applicable for the indicator on checkbox.

This inconsistency can lead to confusion and unexpected behavior.

## Proposal

We propose the following changes:

1. **Expand `Alignment` to include logical direction values**
```ts
// Current
type Alignment = "center" | "left" | "right";

// Updated
type Alignment = "center" | "left" | "right" | "start" | "end";
```

2. **Map `"start"` and `"end"` to existing implementations**
- `"start"` maps to `Classes.ALIGN_LEFT`.
- `"end"` maps to `Classes.ALIGN_RIGHT`.

_This mapping currently occurs via [`Classes.alignmentClass(...)`](https://github.com/palantir/blueprint/blob/9c16d3a053f347d748ca22cebcec6f9a06a73590/packages/core/src/common/classes.ts#L387-L397)_

3. **Deprecate physical direction values**
- Mark `"left"` and `"right"` as deprecated in all components using `Alignment`.
- Provide deprecation warnings in development mode.
- Update documentation to promote `"start"` and `"end"`.

4. **Introduce a new `TextAlignment` type**
```ts
type TextAlignment = "center" | "start" | "end";
```
[...]

PR #7255, opened 2025-02-12
Rename rightIcon to endIcon

jscheiny:
#### Checklist

-   [x] Includes tests
-   [x] Update documentation

In this PR we rename the `rightIcon` property of all our components to `endIcon`. We leave the existing `rightIcon` prop in place but deprecated. If a user passes both an `endIcon` and a `rightIcon` then the former takes priority over the latter. This  is in keeping with our effort to replace physical directions (left/right) with logical ones (start/end). See proposal here: https://github.com/palantir/blueprint/issues/7193

**Affects the following components:**
- `<Button>`
- `<CompoundTag>`
- `<Tag>`


packages/core/test/tag/compoundTagTests.tsx
```diff
@@ -33,13 +34,24 @@ describe("<CompoundTag>", () => {
...
-            <CompoundTag icon="tick" rightIcon="airplane" leftContent="Hello">
+            <CompoundTag icon="tick" endIcon="airplane" leftContent="Hello">
                 World
             </CompoundTag>,
         );
         assert.lengthOf(wrapper.find(Icon), 2);
     });
 
+    it("prefers endIcon to rightIcon", () => {
+        const result = render(
```
ggdouglas: I typically opt to use [`container`](https://testing-library.com/docs/react-testing-library/api/#container) off of render to run query methods since it's executed closer to where the component is rendered. Could we use that instead?
```suggestion
        const { container } = render(
```
```ts
expect(container.querySelector(`.${Classes.ICON}-airplane`)).to.exist; 
expect(container.querySelector(`.${Classes.ICON}-add`)).not.to.exist;
```

packages/core/test/tag/compoundTagTests.tsx
```diff
@@ -33,13 +34,24 @@ describe("<CompoundTag>", () => {
...
+            // eslint-disable-next-line @typescript-eslint/no-deprecated
+            <CompoundTag endIcon="airplane" rightIcon="tick" leftContent="Hello">
+                World
+            </CompoundTag>,
+        );
+        expect(result.baseElement.querySelector(`.${Classes.ICON}-airplane`)).to.exist;
+        expect(result.baseElement.querySelector(`.${Classes.ICON}-add`)).not.to.exist;
```
ggdouglas: The test setup above renders `rightIcon="tick"`, but we're testing for the absence of the "add" icon. Should this be:
```suggestion
        expect(result.baseElement.querySelector(`.${Classes.ICON}-tick`)).not.to.exist;
```

packages/core/test/tag/compoundTagTests.tsx
```diff
@@ -33,13 +34,24 @@ describe("<CompoundTag>", () => {
...
+    });
```
ggdouglas, line 55: Also, it might be slightly more robust/clean to opt for testids here vs. querying the dom structure. You'll be able to use some of RTL's screen methods, and it will test fewer implementation details (such as the structure/name of the class names)

```tsx
it("prefers endIcon to rightIcon", () => {
    const endIcon = <Icon icon="airplane" data-testid="endIcon" />;
    const rightIcon = <Icon icon="tick" data-testid="rightIcon" />;
    render(
        // eslint-disable-next-line @typescript-eslint/no-deprecated
        <CompoundTag endIcon={endIcon} rightIcon={rightIcon} leftContent="Hello">
            World
        </CompoundTag>,
    );

    expect(screen.getByTestId("endIcon")).to.exist;
    expect(screen.queryByTestId("rightIcon")).not.to.exist;
});
```

ggdouglas (dismissed):

ggdouglas (approved):

PR #7256, opened 2025-02-12
fix(ToastCreateAsyncExample): must add props to `Example`

bvandrc:
- Fixes issue when running the docs app locally.

- Also add info to the `README` for Windows development.

- Also perform some slight refactoring in the webpack config files to make them all in sync style-wise


ggdouglas (approved):
Thanks for the fixes!

PR #7258, opened 2025-02-12
bugfix(Section): section header collapse button

bvandrc:
Fixes an issue with this PR: https://github.com/palantir/blueprint/pull/7144 . See comment below.


packages/core/src/components/section/_section.scss
```diff
@@ -59,6 +59,12 @@ $section-card-padding-compact: $pt-grid-size * 1.5 !default;
       margin-left: auto;
     }
 
+    &-collapse-caret {
```
bvandrc, line 62: fixes small position change to the caret that was introduced in https://github.com/palantir/blueprint/pull/7144

packages/core/src/components/section/section.tsx
```diff
@@ -208,9 +208,9 @@ export const Section: React.FC<SectionProps> = React.forwardRef((props, ref) =>
                                     aria-expanded={isCollapsed}
                                     aria-controls={sectionId}
                                     aria-label={isCollapsed ? "expand section" : "collapse section"}
-                                    onClick={toggleIsCollapsed}
+                                    // no OnClick, click event triggers header below
```
bvandrc, line 211: This PR https://github.com/palantir/blueprint/pull/7144 actually had a bug. Clicking (or hitting `enter` on) this `span` caused the section to not collapse, because I didn't include `event.preventDefault()` here, so the event was being passed down to the header below, meaning `toggleIsCollapsed` was called twice in a row immediately.

Could either add `e.preventDefault()`, or, just thought we'd leave the `onClick` out (clicking clicks the header below)

jscheiny (approved):

PR #7281, opened 2025-02-27
[Docs] Refactor Select Examples

ggdouglas:
Updates the following documentation examples:
- **Select:** [Before](https://blueprintjs.com/docs/#select/select-component) / After
- **Suggest:** [Before](https://blueprintjs.com/docs/#select/suggest) / After
- **MultiSelect:** [Before](https://blueprintjs.com/docs/#select/multi-select) / After
- **Omnibar:** [Before](https://blueprintjs.com/docs/#select/omnibar) / After

### DIFF
--- a/packages/docs-app/src/examples/select-examples/omnibarExample.tsx
+++ b/packages/docs-app/src/examples/select-examples/omnibarExample.tsx
@@ -1,5 +1,5 @@
 /*
- * Copyright 2017 Palantir Technologies, Inc. All rights reserved.
+ * Copyright 2025 Palantir Technologies, Inc. All rights reserved.
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
  *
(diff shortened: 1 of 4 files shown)


packages/docs-app/src/examples/select-examples/omnibarExample.tsx
```diff
@@ -41,125 +30,92 @@ import {
...
-    private handleAllowCreateChange = handleBooleanChange(allowCreate => this.setState({ allowCreate }));
+    const [allowCreate, setAllowCreate] = React.useState(false);
+    const [isOpen, setIsOpen] = React.useState(false);
+    const [overlayHasBackdrop, setOverlayHasBackdrop] = React.useState(true);
+    const [resetOnSelect, setResetOnSelect] = React.useState(true);
 
-    private handleOverlayHasBackdropChange = handleBooleanChange(overlayHasBackdrop =>
-        this.setState({ overlayHasBackdrop }),
+    const toaster = React.useMemo(
+        () => OverlayToaster.createAsync({ className: classNames({ [Classes.DARK]: useDarkTheme }) }),
```
ggdouglas, line 42: Wasn't 100% sure if this is the best way to reimplement this Toaster. We generally [recommend](https://blueprintjs.com/docs/#core/components/toast.overlaytoaster) creating a shared instance with `OverlayToaster.createAsync`, but this particular example switches the styles of the toast depending on dark mode.

We could alternatively use the `OverlayToaster` with a ref (more similar to before), but I figured I'd try the more current, recommended approach first.

jscheiny (approved):
The overlay toaster implementation looks good to me. Went through all the examples, everything seems to be working as expected.

PR #7282, opened 2025-02-27
fix: size prop should be optional

bvandrc:
In the recent changes (https://github.com/palantir/blueprint/pull/7228), the `size` prop should be optional, but it's not. The fact that there is a `@default` in the JSDoc, yet it is required, is contradictory.


ggdouglas (approved):
Thanks for the catch, @bvandercar-vt!

PR #7307, opened 2025-03-03
fix: enum export

bvandrc:
Since these were recently changed to be enums and types, we need to do `export` instead of `export type` so that we can use it as a constant.

Currently, when we try to do.

```tsx
import { Size } from '@blueprintjs/core'

...

<Button>
   size={Size.LARGE}
```

We get the error `"'Size' cannot be used as a value because it was exported using 'export type'." `


ggdouglas (approved):
LGTM. Thanks for the catch!

jscheiny (approved):

PR #7308, opened 2025-03-04
dev: convert icons dist scripts from bash to mjs

bvandrc:
Have been having trouble with the `.sh` scripts on windows, even though I have done `npm config set script-shell "C:\\Program Files\\git\\bin\\bash.exe"`. By converting these scripts to `.mjs`, we have a more universal script that is also easier to read and also matches the fact that most of the other scripts in this folder are alraeady `.mjs`.


bvandrc:
@ggdouglas  rebased to latest.

PR #7315, opened 2025-03-05
break(Icon): Delete iconSize prop

jscheiny:

Removed the deprecated `iconSize` prop from the `icon` component. The `size` prop should be used instead.


ggdouglas:
~~Are we able to remove the blocklisted `iconSize` prop from here as well?~~ https://github.com/palantir/blueprint/blob/04acffbeda3afd28347d913109eec55ff878bcb8/packages/core/src/common/props.ts#L141

**Edit:** Nvm, I suppose we still use `iconSize` on components like `<NonIdealState>`

ggdouglas (approved):

PR #7327, opened 2025-03-07
chore: Rename TableBody2 -> TableBody

jscheiny:
This component is only used internally and thus this is not a breaking change.


packages/table/src/tableBody2.tsx
```diff
@@ -19,20 +19,54 @@ import * as React from "react";
...
+import type { Locator } from "./locator";
 import { type Region, Regions } from "./regions";
-import type { TableBodyProps } from "./tableBody";
-import { TableBodyCells } from "./tableBodyCells";
+import { TableBodyCells, type TableBodyCellsProps } from "./tableBodyCells";
 
 const DEEP_COMPARE_KEYS: Array<keyof TableBodyProps> = ["selectedRegions"];
 
-export class TableBody2 extends AbstractComponent<TableBodyProps> {
+export interface TableBodyProps extends SelectableProps, TableBodyCellsProps {
```
jscheiny, line 35: Moved this type here from `tableBody.tsx` since that results in the smallest diff. In a follow up PR I will rename `tableBody2.tsx`. Didn't do it in this PR as that would result in a bad diff/blame.

packages/table/src/table2.tsx
```diff
@@ -62,7 +62,7 @@ import {
     resizeRowsByTallestCell,
 } from "./resizeRows";
 import { compareChildren, getHotkeysFromProps, isSelectionModeEnabled } from "./table2Utils";
-import { TableBody2 } from "./tableBody2";
+import { TableBody } from "./tableBody2";
```
ggdouglas, line 65: Should the filename also be renamed?
```suggestion
import { TableBody } from "./tableBody";
```

ggdouglas (approved):

PR #7328, opened 2025-03-07
break(Table2): Rename to Table

jscheiny:
This PR renames `Table2` to `Table` and `Table2Props` to `TableProps`. It then exposes the old v2 variants of the names as aliases for the new values.


packages/table/src/table2Utils.ts
```diff
@@ -20,10 +20,33 @@ import type { HotkeyConfig } from "@blueprintjs/core";
...
+function maybeClampValue(value: number | undefined, max: number) {
+    return value === undefined ? 0 : Utils.clamp(value, 0, max);
+}
+
+export function hasLoadingOption(loadingOptions: string[] | undefined, loadingOption: string) {
+    if (loadingOptions === undefined) {
+        return false;
+    }
+    return loadingOptions.indexOf(loadingOption) >= 0;
+}
```
jscheiny, line 48: These are moved here from `tableUtils.tsx` because this results in the smallest diff to combine these files. I will rename this file in a follow up PR.

packages/table-dev-app/src/features.tsx
```diff
@@ -468,8 +467,8 @@ const bodyContextMenuRenderer = (context: MenuContext) => {
     );
 };
 
-const table2Root = ReactDOM.createRoot(document.getElementById("table-2"));
-table2Root.render(
+const tableRoot = ReactDOM.createRoot(document.getElementById("table-2"));
+tableRoot.render(
```
ggdouglas: **nit:** The `table2Root` here refers to the named order of these React roots `table1Root`, `table2Root`, ... `table9Root`, so we should keep this variable named as is. I suspect it got caught up in the find/replace.

jscheiny: Oh nice catch, thanks

ggdouglas (approved):

PR #7334, opened 2025-03-10
break(PanelStack2): Rename to PanelStack

jscheiny:
This PR renames `PanelStack2` to `PanelStack` and `PanelStack2Props` to `PanelStackProps`. This also contains backward compatible aliases for the v2 names to the current names for migration purposes. In addition all class name constants now have back compat aliases, but the class names themselves retain the `2` in the string. For example:

- `Classes.PANEL_STACK` is `"${ns}-panel-stack2`
- `Classes.PANEL_STACK2` is the same but is deprecated

This way, current consumers of `PanelStack2` don't have to change their css to continue to use `PanelStack2` but can update the JS constants that they use.


packages/core/src/common/classes.ts
```diff
@@ -277,10 +277,19 @@ export const OVERLAY_SCROLL_CONTAINER = `${OVERLAY}-scroll-container`;
 export const OVERLAY_START_FOCUS_TRAP = `${OVERLAY}-start-focus-trap`;
 export const OVERLAY_END_FOCUS_TRAP = `${OVERLAY}-end-focus-trap`;
 
-export const PANEL_STACK2 = `${NS}-panel-stack2`;
-export const PANEL_STACK2_HEADER = `${PANEL_STACK2}-header`;
-export const PANEL_STACK2_HEADER_BACK = `${PANEL_STACK2}-header-back`;
-export const PANEL_STACK2_VIEW = `${PANEL_STACK2}-view`;
+export const PANEL_STACK = `${NS}-panel-stack2`;
```
ggdouglas, line 280: Would it make sense to add forwards compatibility to the class names here? For example, we have these classes tied to `.#{$ns}-panel-stack2`. Would it make sense to additionally style to the non-v2 class name as well?
https://github.com/palantir/blueprint/blob/1bf736ac5830d5d01e2f9d7f0140f057f2fe05f7/packages/core/src/components/panel-stack/_panel-stack.scss#L8-L11
```diff
+.#{$ns}-panel-stack,
 .#{$ns}-panel-stack2 {
   overflow: hidden;
   position: relative;
}

ggdouglas (approved):

packages/core/src/common/classes.ts
jscheiny: I like it, let me examine that idea in a follow up.

PR #7336, opened 2025-03-11
test: use testing-library waitFor instead of arbitrary wait times

bvandrc:


ggdouglas (approved):
These changes look great. Thanks for updating these tests!

Issue #7293, opened 2025-03-02
Consolidate `@blueprintjs/datetime` and `@blueprintjs/datetime2` packages


ggdouglas:
Migrate v3 `@blueprintjs/datetime2` components into `@blueprintjs/datetime` and replace legacy components.

The result is that `@blueprintjs/datetime` will consist of the following components:
- [`DatePicker3`](https://blueprintjs.com/docs/#datetime2/date-picker3) → `DatePicker`
- [`DateInput3`](https://blueprintjs.com/docs/#datetime2/date-input3) → `DateInput`
- [`DateRangePicker3`](https://blueprintjs.com/docs/#datetime2/date-range-picker3) → `DateRangePicker`
- [`DateRangeInput3`](https://blueprintjs.com/docs/#datetime2/date-range-input3) → `DateRangeInput`
- [`TimePicker`](https://blueprintjs.com/docs/#datetime/timepicker) (no change)
- [`TimezoneSelect`](https://blueprintjs.com/docs/#datetime/timezone-select) (no change)

We will also drop the following from  `@blueprintjs/datetime2` since they are reexported from `@blueprintjs/datetime`
- `DateInput2`
- `DateInput2Props`
- `DateRangeInput2`
- `DateRangeInput2Props`

## Tasks
- Sever dependency of `@blueprintjs/datetime2` on `@blueprintjs/datetime`
  - Move over components and utilities that are imported from `@blueprintjs/datetime` into `@blueprintjs/datetime2`.
  - Note: this will involve having code duplicated briefly during the transition
  - Severing the dependency will make it easier to migrate `@blueprintjs/datetime` to use the latest components without having to deal with reexported code
- `@blueprintjs/datetime`
  - Update `react-day-picker` dependency from `7.4.9` to `8.10.0`
  - Add React 18 to `peerDependencies`

## References

https://github.com/palantir/blueprint/blob/f900a36b35c612d7e4f945ec51cb9f1618698b51/packages/datetime2/src/components/date-input3/date-input3.md?plain=1#L14C1-L17

- [**DateInput3** is a replacement for DateInput and will replace it in Blueprint v6.](https://github.com/palantir/blueprint/blob/f900a36b35c612d7e4f945ec51cb9f1618698b51/packages/datetime2/src/components/date-input3/date-input3.md?plain=1#L14C1-L14C83)
  
[https://blueprintjs.com/docs/#datetime2](https://blueprintjs.com/docs/#datetime2)  

[...]

PR #7337, opened 2025-03-11
Fix type errors hidden by disabled `strictNullChecks` in `@blueprintjs/datetime` package

ggdouglas:
Part of #7293

### Proposed Changes

The datetime package has [`strictNullChecks`](https://www.typescriptlang.org/tsconfig/#strictNullChecks) set to `false`, which hides some type errors in the package.

https://github.com/palantir/blueprint/blob/37c3c66a445c91ca214f9e89b15140fd5c1751b3/packages/datetime/src/tsconfig.json#L3-L6

This seems to have been introduced to datetime's tsconfig in #4351

This config value is slightly problematic because `@blueprintjs/datetime2` does _**not**_ have `strictNullChecks` disabled:
https://github.com/palantir/blueprint/blob/37c3c66a445c91ca214f9e89b15140fd5c1751b3/packages/datetime2/src/tsconfig.json#L3-L6

This leads to type compilation errors when attempting to pull components/utilities from datetime into datetime2. It seems that we can either:

1. Keep `strictNullChecks` set to false when porting datetime2 components into datetime
2. Temporarily set `strictNullChecks` to true and fix type compilation errors for files relevant to datetime2 components (this PR)

This PR was created by temporarily setting `strictNullChecks` to `true` in `@blueprintjs/datetime`, showing the compilation errors and fixing ones relevant to components in `@blueprintjs/datetime2`.

### Reviewers should focus on:

Existing Datetime components should pass tests and should not have any regressions in functionality.

### DIFF
--- a/packages/datetime/src/common/dateRangeSelectionStrategy.ts
+++ b/packages/datetime/src/common/dateRangeSelectionStrategy.ts
@@ -74,7 +74,7 @@ export class DateRangeSelectionStrategy {
             nextDateRange = this.createRangeForBoundary(boundary, nextBoundaryDate, null);
         } else if (boundaryDate == null && otherBoundaryDate != null) {
             if (isSameDay(day, otherBoundaryDate)) {
-                let nextDate: Date;
+                let nextDate: Date | null;
                 if (allowSingleDayRange) {
                     nextBoundary = boundary;
                     nextDate = otherBoundaryDate;
@@ -92,18 +92,19 @@ export class DateRangeSelectionStrategy {
             }
         } else {
             // both boundaryDate and otherBoundaryDate are already defined
-            if (isSameDay(boundaryDate, day)) {
-                const isSingleDayRangeSelected = isSameDay(boundaryDate, otherBoundaryDate);
+            if (boundaryDate != null && isSameDay(boundaryDate, day)) {
+                const isSingleDayRangeSelected =
+                    otherBoundaryDate != null && isSameDay(boundaryDate, otherBoundaryDate);
                 const nextOtherBoundaryDate = isSingleDayRangeSelected ? null : otherBoundaryDate;
                 nextBoundary = boundary;
                 nextDateRange = this.createRangeForBoundary(boundary, null, nextOtherBoundaryDate);
-            } else if (isSameDay(day, otherBoundaryDate)) {
+            } else if (otherBoundaryDate != null && isSameDay(day, otherBoundaryDate)) {
                 const [nextBoundaryDate, nextOtherBoundaryDate] = allowSingleDayRange
                     ? [otherBoundaryDate, otherBoundaryDate]
                     : [boundaryDate, null];
                 nextBoundary = allowSingleDayRange ? boundary : otherBoundary;
                 nextDateRange = this.createRangeForBoundary(boundary, nextBoundaryDate, nextOtherBoundaryDate);
-            } else if (this.isOverlappingOtherBoundary(boundary, day, otherBoundaryDate)) {
+            } else if (otherBoundaryDate != null && this.isOverlappingOtherBoundary(boundary, day, otherBoundaryDate)) {
                 nextBoundary = boundary;
                 nextDateRange = this.createRangeForBoundary(boundary, day, null);
             } else {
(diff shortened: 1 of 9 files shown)


packages/datetime/src/common/dateRangeSelectionStrategy.ts
```diff
@@ -74,7 +74,7 @@ export class DateRangeSelectionStrategy {
             nextDateRange = this.createRangeForBoundary(boundary, nextBoundaryDate, null);
         } else if (boundaryDate == null && otherBoundaryDate != null) {
             if (isSameDay(day, otherBoundaryDate)) {
-                let nextDate: Date;
+                let nextDate: Date | null;
```
jscheiny, line 77: Curious why the choice of `null` here as opposed to `undefined`?

packages/datetime/src/common/dateUtils.ts
```diff
@@ -160,9 +164,9 @@ export function isTimeSameOrAfter(date: Date, dateToCompare: Date): boolean {
 /**
  * @returns a Date at the exact time-wise midpoint between startDate and endDate
  */
-export function getDateBetween(dateRange: DateRange) {
-    const start = dateRange[0].getTime();
-    const end = dateRange[1].getTime();
+export function getDateBetween(dateRange: DateRange): Date {
+    const start = dateRange[0]!.getTime();
+    const end = dateRange[1]!.getTime();
```
ggdouglas, line 172: A bit stumped on how to resolve this one (hence the non-null assertion operators)

DateRange has the following type:
https://github.com/palantir/blueprint/blob/37c3c66a445c91ca214f9e89b15140fd5c1751b3/packages/datetime/src/common/dateRange.ts#L17

So both `dateRange[0]` and `dateRange[1]` are both possibly `null` here. I'm not sure what default values that we could use since we're looking for a date "between" two possibly null dates.

For additional context, this function is only ever used in the `getInitialMonth` function of datetime components, ex:

https://github.com/palantir/blueprint/blob/37c3c66a445c91ca214f9e89b15140fd5c1751b3/packages/datetime2/src/components/date-picker3/datePicker3.tsx#L411

https://github.com/palantir/blueprint/blob/37c3c66a445c91ca214f9e89b15140fd5c1751b3/packages/datetime2/src/components/date-range-picker3/dateRangePicker3.tsx#L483

jscheiny (commented):
Overall LGTM, one question

packages/datetime/src/common/dateUtils.ts
jscheiny: Could we change the signature here to accept `NonNullDateRange` instead?

packages/datetime/src/common/dateRangeSelectionStrategy.ts
ggdouglas: Ended up with `null` here since there is an assignment a few lines down in this block:
https://github.com/palantir/blueprint/blob/a6ab776166550911e9a6b74d9eb96b042224054b/packages/datetime/src/common/dateRangeSelectionStrategy.ts#L83
It's also relevant to the parameter passed into `createRangeForBoundary`:

https://github.com/palantir/blueprint/blob/a6ab776166550911e9a6b74d9eb96b042224054b/packages/datetime/src/common/dateRangeSelectionStrategy.ts#L85

https://github.com/palantir/blueprint/blob/a6ab776166550911e9a6b74d9eb96b042224054b/packages/datetime/src/common/dateRangeSelectionStrategy.ts#L167-L184

packages/datetime/src/common/dateUtils.ts
ggdouglas: Hm, maybe! This works for `DateRangePicker3` since we already have the non-null assertions passed in the parameter. The `DatePicker3` component does not provide a `NonNullDateRange` however:
https://github.com/palantir/blueprint/blob/37c3c66a445c91ca214f9e89b15140fd5c1751b3/packages/datetime2/src/components/date-picker3/datePicker3.tsx#L400-L413

We'd need to either modify the behavior or add non-null assertions there. I like where this is going, but I'm more inclined to do it as cleanup after the datetime/datetime2 packages have been consolidated/merged.

jscheiny: Okay, I'm fine with this in the short term, but passing in a `NonNullDateRange` and adding in external checks feels the right away to do this long term. Maybe add a todo here?

ggdouglas: Added in d41b3abc23e1477539ce1a0ea2cf935b2e6d3c11

jscheiny: This one seems fine, but generally prefer to not put todos in docs comments

jscheiny (approved):

PR #7341, opened 2025-03-13
Add test for DateFormatProps

ggdouglas:
#### Proposed changes

Preflight, part of #7293

Adds explicit test for `DateFormatProps` to avoid test coverage failures when consolidating datetime/datetime2 code.

### DIFF
--- a/packages/datetime/test/common/dateFormatPropsTests.ts
+++ b/packages/datetime/test/common/dateFormatPropsTests.ts
@@ -14,26 +14,25 @@ describe("DateFormatProps", () => {
     describe("getFormattedDateString", () => {
         it("should return an empty string for null date", () => {
             const testDate: Date | null = null;
-            const props = { formatDate };
 
-            expect(getFormattedDateString(testDate, props)).to.equal("");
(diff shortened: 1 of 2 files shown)


packages/datetime/test/common/dateFormatPropsTests.ts
```diff
@@ -0,0 +1,60 @@
...
+import { getFormattedDateString } from "../../src/common/dateFormatProps";
+import { Months } from "../../src/common/months";
+
+const formatDate = (date: Date) => format(date, "yyyy-MM-dd");
+
+describe("DateFormatProps", () => {
+    describe("getFormattedDateString", () => {
+        it("should return an empty string for null date", () => {
+            const testDate: Date | null = null;
+            const props = { formatDate };
```
jscheiny: Nit: Here and below, add a type annotation for these prop object literals or inline them into the function call. Context: https://pages.github.palantir.build/palantir/documentation/#/foundry/develop/frontend/forge-style-guide?id=annotate-types-of-variables-assigned-from-object-literals

jscheiny (approved):

PR #7359, opened 2025-03-13
[Datetime] Delete legacy datetime components

ggdouglas:
#### Proposed changes:

Removes legacy v1 components, docs, and associated tests.


packages/datetime/test/common/dayPickerTestUtils.ts
```diff
@@ -14,18 +14,17 @@
  * limitations under the License.
  */
 
-/** @fileoverview test utils for react-day-picker v7 */
+/** @fileoverview test utils for react-day-picker v8 */
```
ggdouglas, line 17: Diff here is a result of deleting existing `dayPickerTestUtils.ts` and renaming `dayPicker8TestUtils.ts`. See https://github.com/palantir/blueprint/pull/7357#discussion_r1994196270

packages/datetime/src/components/date-picker3/datePickerUtils.ts
```diff
@@ -14,10 +14,22 @@
  * limitations under the License.
  */
 
+import { Months } from "../../common";
 import { getFormattedDateString } from "../../common/dateFormatProps";
 import { measureTextWidth } from "../../common/utils";
 
-import { getDefaultMaxDate, getDefaultMinDate } from "./datePickerCore";
```
ggdouglas, line 20: These were inlined since the other utilities from `datePickerCore.tsx` are no longer needed.

jscheiny (approved):

PR #7360, opened 2025-03-13
[Datetime] Upgrade `react-day-picker` at `v7.4.9` to `^8.10.0` and remove temporary v8 dependency

ggdouglas:
#### Proposed changes:

FLUP to #7359. Now that we have removed all legacy datetime components using `react-day-picker@7.4.9`, we can now remove that version and patch and upgrade to `react-day-picker@^8.10.0` to be consistent with datetime2.


jscheiny (approved):

packages/datetime/src/common/datePickerBaseProps.ts
```diff
@@ -70,11 +70,6 @@ export interface DatePickerBaseProps {
      */
     locale?: string;
 
-    /**
-     * Collection of functions that provide internationalization support.
-     */
-    localeUtils?: typeof LocaleUtils;
```
ggdouglas, line 76: `LocaleUtils` is no longer available from `react-day-picker`, and thus `localeUtils` is no longer accepted as a prop.

See: https://github.com/palantir/blueprint/wiki/react-day-picker-8-migration#locale

PR #7369, opened 2025-03-14
deprecation: Deprecate Tabs.Tab alias

jscheiny:
This PR deprecates the `Tabs.Tab` component in favor of just `Tab`. I think this is rarely used as is.

### DIFF
--- a/packages/core/src/components/tabs/tabs.tsx
+++ b/packages/core/src/components/tabs/tabs.tsx
@@ -126,6 +126,7 @@ export interface TabsState {
  * @see https://blueprintjs.com/docs/#core/components/tabs
  */
 export class Tabs extends AbstractPureComponent<TabsProps, TabsState> {
+    /** @deprecated Use `Tab` instead */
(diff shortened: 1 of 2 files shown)


packages/core/src/components/tabs/tabs.tsx
```diff
@@ -126,6 +126,7 @@ export interface TabsState {
  * @see https://blueprintjs.com/docs/#core/components/tabs
  */
 export class Tabs extends AbstractPureComponent<TabsProps, TabsState> {
+    /** @deprecated Use `Tab` instead */
```
ggdouglas: This seems a bit vague/confusing by itself since the thing we are deprecating is also called "Tab". Maybe we should reference the Tab **_component_** here (or link to it?)

jscheiny: Good point, updated the comment, let me know what you think.

ggdouglas (approved):

PR #7372, opened 2025-03-17
[docs] Merge and clean up datetime docs

ggdouglas:
#### Proposed changes

Updates `datetime` docs to reflect new components ported from `datetime2`. Also deletes docs in `datetime2`.

#### Checklist

-   [ ] Includes tests
-   [x] Update documentation

Datetime docs should be logically consistent and current. There should be no broken links or references to old/outdated patterns from the legacy datetime components.


packages/datetime/src/components/date-picker/date-picker.md
```diff
@@ -1,43 +1,28 @@
----
-tag: new
----
+@# DatePicker
 
-@# DatePicker3
+**DatePicker** renders a UI to choose a single date and (optionally) a time of day.
+It is built on top of the [react-day-picker v8](https://daypicker.dev/v8) library.
+Time selection is enabled by the [TimePicker](#datetime/timepicker) component.
```
jscheiny, line 5: Do we still care about sharing this information? This feels like is only relevant when comparing datetime and datetime2.

jscheiny (dismissed):
One minor comment

packages/datetime/src/components/date-picker/date-picker.md
ggdouglas: Yeah, fair question. I kept the references around since we explicitly mentioned it on the original DatePicker docs: https://github.com/palantir/blueprint/blob/d356c8eea907c1ffc84ef626188c90a3698153fc/packages/datetime/src/components/date-picker/datepicker.md?plain=1#L21-L22

Perhaps we can keep that reference in the updated docs and remove it for the other components.

**Edit:** 487fa1dc2ade6f40dadeefea3d88a18c2de0a664

jscheiny (approved):

PR #7374, opened 2025-03-17
break(Portal): Remove legacy context options

jscheiny:
Removes the legacy portal context options in favor of `PortalProvider`


packages/core/src/common/utils/reactUtils.ts
```diff
@@ -123,8 +123,3 @@ export function isElementOfType<P = {}>(
         element.type.displayName === ComponentType.displayName
     );
 }
-
-export function isReact18OrHigher(): boolean {
```
ggdouglas, line 127: Glad we can get rid of this check now 🙂

ggdouglas (approved):

PR #7375, opened 2025-03-17
[Datetime] Update `DatePickerShortcutMenu` to a functional component

ggdouglas:
#### Proposed changes:

Updates the `DatePickerShortcutMenu` component to a functional component.

### DIFF
--- a/packages/datetime/src/components/date-picker/datePicker.tsx
+++ b/packages/datetime/src/components/date-picker/datePicker.tsx
@@ -229,7 +229,11 @@ export class DatePicker extends DateFnsLocalizedComponent<DatePickerProps, DateP
         }
 
         const { selectedShortcutIndex } = this.state;
-        const { maxDate, minDate, timePrecision } = this.props;
+        const {
+            maxDate = DatePickerUtils.getDefaultMaxDate(),
+            minDate = DatePickerUtils.getDefaultMinDate(),
+            timePrecision,
+        } = this.props;
         // Reuse the existing date range shortcuts and only care about start date
         const dateRangeShortcuts: DateRangeShortcut[] | true =
             shortcuts === true
@@ -242,14 +246,12 @@ export class DatePicker extends DateFnsLocalizedComponent<DatePickerProps, DateP
         return [
             <DatePickerShortcutMenu
                 key="shortcuts"
-                {...{
-                    allowSingleDayRange: true,
-                    maxDate,
-                    minDate,
-                    selectedShortcutIndex,
-                    shortcuts: dateRangeShortcuts,
-                    timePrecision,
-                }}
+                allowSingleDayRange={true}
+                maxDate={maxDate}
+                minDate={minDate}
+                selectedShortcutIndex={selectedShortcutIndex}
+                shortcuts={dateRangeShortcuts}
+                timePrecision={timePrecision}
                 onShortcutClick={this.handleShortcutClick}
                 useSingleDateShortcuts={true}
             />,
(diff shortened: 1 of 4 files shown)


packages/datetime/src/components/date-picker/datePicker.tsx
```diff
@@ -242,14 +246,12 @@ export class DatePicker extends DateFnsLocalizedComponent<DatePickerProps, DateP
         return [
             <DatePickerShortcutMenu
                 key="shortcuts"
-                {...{
```
jscheiny, line 245: LOL, why was it like this? I guess to avoid `x={x}` style props. But still...

packages/datetime/src/components/shortcuts/shortcuts.tsx
```diff
@@ -75,51 +75,47 @@ export interface DatePickerShortcutMenuProps {
...
+        minDate,
+        onShortcutClick,
+        selectedShortcutIndex = -1,
+        timePrecision,
+        useSingleDateShortcuts = false,
+    } = props;
+
+    const shortcuts =
+        props.shortcuts === true
+            ? createDefaultShortcuts(allowSingleDayRange, timePrecision !== undefined, useSingleDateShortcuts)
```
jscheiny: While we're at it?

```suggestion
            ? createDefaultShortcuts(allowSingleDayRange, timePrecision != null, useSingleDateShortcuts)
```

packages/datetime/src/components/shortcuts/shortcuts.tsx
```diff
@@ -75,51 +75,47 @@ export interface DatePickerShortcutMenuProps {
...
+
+    const isShortcutInRange = React.useCallback(
+        (shortcutDateRange: DateRange) => isDayRangeInRange(shortcutDateRange, [minDate, maxDate]),
+        [minDate, maxDate],
+    );
+
+    const handleShortcutClick = React.useCallback(
+        (shortcut: DateRangeShortcut, index: number) => () => onShortcutClick(shortcut, index),
+        [onShortcutClick],
+    );
```
jscheiny: Curious for your thoughts here but generally don't like the get click handler callback format. I'd prefer we have a `<ShortcutMenuItem>` component that takes in `onShortcutClick` and binds a specific `useCallback` for itself.

packages/datetime/src/components/shortcuts/shortcuts.tsx
ggdouglas: d8ed09cdfb06e4f27946ef7cbaaba05b325a61e9

packages/datetime/src/components/shortcuts/shortcuts.tsx
ggdouglas: Yeah, I can get on board with that. Change added in cc02cf3bf2115f0d553152a926a38027b00ee053

ggdouglas:
Will wait for #7383 to merge since the failing tests from `datetime2` will be deleted in that PR.

packages/datetime/src/components/shortcuts/shortcuts.tsx
```diff
@@ -75,52 +75,69 @@ export interface DatePickerShortcutMenuProps {
...
 
+const ShortcutMenuItem: React.FC<ShortcutMenuItemProps> = props => {
+    const { active, index, maxDate, minDate, onShortcutClick, shortcut } = props;
+
+    const handleClick = React.useCallback(() => onShortcutClick(shortcut, index), [index, onShortcutClick, shortcut]);
+
+    const isShortcutInRange = React.useCallback(
+        (shortcutDateRange: DateRange) => isDayRangeInRange(shortcutDateRange, [minDate, maxDate]),
+        [minDate, maxDate],
+    );
```
jscheiny: Now that this is structured this way, I don't think we need a callback here:

```suggestion
    const isShortcutInRange = isDayRangeInRange(shortcut.dateRange, [minDate, maxDate]);
```

ggdouglas: Ah, good catch! Changed in 9de574d4e882cc11da2e4bbb6ef5afc2410a5aed

jscheiny (approved):

PR #7376, opened 2025-03-17
docs: Remove deprecated CSS API docs

jscheiny:
Removes the deprecated CSS API docs from all of our components and the associated css examples. The only one I left in is the icon CSS API, unsure if I want to remove this yet.


packages/core/src/components/navbar/navbar.md
```diff
@@ -59,7 +59,25 @@ If your application is inside a fixed-width container (instead of spanning the e
...
+            <Button icon="home" text="Home" variant="minimal" />
+            <Button icon="document" text="Files" variant="minimal" />
+            <NavbarDivider />
+            <Button icon="user" variant="minimal" />
+            <Button icon="notifications" variant="minimal" />
+            <Button icon="cog" variant="minimal" />
+        </NavbarGroup>
+    </div>
+</Navbar>
+```
```
ggdouglas: Can we remove this second example? Seems to be a duplicate of the code example above it:

Screenshot 2025-03-17 at 21 38 15@2x

ggdouglas (dismissed):

packages/core/src/components/navbar/navbar.md
jscheiny: Oops 🤦

ggdouglas (approved):

PR #7382, opened 2025-03-19
test(Popover): waitFor instead of setTimeout

bvandrc:
Continuation of https://github.com/palantir/blueprint/commit/4c4654d2131b2abe576689b644da85885f2478fb


packages/core/test/popover/popoverTests.tsx
```diff
@@ -255,28 +256,24 @@ describe("<Popover>", () => {
         };
 
         // HACKHACK: skipped test resulting from React 18 upgrade. See: https://github.com/palantir/blueprint/issues/7168
```
ggdouglas: **nit:** Can we also remove the comment referencing this test as skipped?

bvandrc: This comment was removed in the other PR, it's no longer here after rebasing.

ggdouglas (approved):
Thanks for updating these!

PR #7383, opened 2025-03-19
[Datetime] Reexport `datetime2` components from `datetime`

ggdouglas:
#### Proposed changes:

- Removes components, utils, and tests in `datetime2` that were copied to `datetime` when merging the two packages
- Changes exports of `datetime2` to reexport v3 aliased components, types, and class names from `datetime`, eg:
```diff
// datetime2/src/index.ts

-export { DateInput3, type DateInput3Props } from "./components/date-input3/dateInput3";
+export {
+    /** @deprecated import from `@blueprintjs/datetime` instead */
+    DateInput as DateInput3,
+    /** @deprecated import from `@blueprintjs/datetime` instead */
+    type DateInputProps as DateInput3Props,
+} from "@blueprintjs/datetime";
```
- Also adds a new `DayPickerProps` export from `datetime` to be reexported by `datetime2` for consistency


packages/datetime2/src/blueprint-datetime2.scss
```diff
@@ -0,0 +1,7 @@
+// Copyright 2022 Palantir Technologies, Inc. All rights reserved.
+// Licensed under the Apache License, Version 2.0.
+
+@import "common/react-day-picker-overrides";
+@import "components/date-picker3/date-picker3";
+@import "components/date-picker3/date-picker3-caption";
+@import "components/date-range-picker3/date-range-picker3";
```
ggdouglas: Keeping styles intact for now since there doesn't seem to be a straightforward way of reexporting these individually.

After the v6 release is cut, consumers will no longer need to import both the `blueprint-datetime.css` and `blueprint-datetime2.css` stylesheets, so we can eventually delete these when we remove this package in a later major version change.

packages/datetime2/src/index.ts
```diff
@@ -14,11 +14,6 @@
  * limitations under the License.
  */
 
-export type { DayPickerProps } from "./common/reactDayPickerProps";
-export { DatePicker3, type DatePicker3Props } from "./components/date-picker3/datePicker3";
-export { DateInput3, type DateInput3Props } from "./components/date-input3/dateInput3";
-export { DateRangeInput3, type DateRangeInput3Props } from "./components/date-range-input3/dateRangeInput3";
-export { DateRangePicker3, type DateRangePicker3Props } from "./components/date-range-picker3/dateRangePicker3";
 import * as DateInput2MigrationUtils from "./dateInput2MigrationUtils";
```
ggdouglas, line 17: Currently considering migrating `DateInput2MigrationUtils` over to `@blueprintjs/datetime` as well. This would allow consumers to completely drop the `@blueprintjs/datetime2` package in all cases. Will probably do in a FLUP.

jscheiny (approved):

packages/datetime2/src/index.ts
jscheiny: Yeah I like that, then we can deprecate in 6.x and remove in 7.

ggdouglas: https://github.com/palantir/blueprint/pull/7388

PR #7390, opened 2025-03-21
break(OverlayToaster): Use React 18 createRoot for rendering via createAsync

jscheiny:
In this PR we make two breaking changes:
1. Remove the `domUnmounter` field for `OverlayToasterCreateOptions`. This fields was never used anyways, so this is just clean up.
2. Make the default `domRenderer` value for `OverlayToaster.createAsync` use the new React `createRoot` method for rendering. 

In a follow up PR, we will consolidate `create` and `createAsync` to make them have the same behavior and then deprecate `createAsync`.


packages/core/test/toast/overlayToasterTests.tsx
```diff
@@ -101,127 +108,154 @@ describe("OverlayToaster", () => {
...
+            //     const key = toaster.show({ message: "two" });
+            //     // await waitFor(() => { /* noop */ });
+            //     toaster.show({ message: "six" });
+            //     // await waitFor(() => { /* noop*/ }, { timeout: 3 * OVERLAY_TOASTER_DELAY_MS })
+            //     toaster.dismiss(key);
+            //     assert.deepEqual(
+            //         toaster.getToasts().map(t => t.message),
+            //         ["six", "one"],
+            //     );
+            // });
```
ggdouglas: **nit:** remove commented out code

jscheiny: Thanks, fixed now

ggdouglas (approved):

PR #7391, opened 2025-03-24
feat(ContextMenu): Introduce new API for imperative rendering

jscheiny:
This PR introduces a new rendering API for the `showContextMenu` and `hideContextMenu` options and deprecates the old API. The old API will be dropped in Blueprint 6.

Previously these functions would allow users to pass in an object with `domRenderer` and `domUnmounter` options which were implemented via `ReactDOM.render` and `ReactDOM.unmountComponentAtNode` respectively. This API does not work for React 18 which requires tracking a root node between the render and unmount stages. To allow for this, the new API instead allows passing a single `render` function which should return a function that unmounts the rendered node.

The default implementation still uses React 16 compatible defaults as before. In Blueprint 6 we will be making a few breaks here:
1. Remove the `DOMMountOptions<T>` type from the `options` prop of `showContextMenu`
2. Remove the options parameter of `hideContextMenu`
3. Change the default implementation of the `render` method to use React 18 compatible methods. It will look something like:

```tsx
const defaultRenderer: ShowContextMenuDOMRenderer = (element, container) => {
    const root = ReactDOMClient.createRoot(container);
    root.render(element);
    return () => root.unmount();
}

### DIFF
--- a/packages/core/src/components/context-menu/contextMenuSingleton.tsx
+++ b/packages/core/src/components/context-menu/contextMenuSingleton.tsx
@@ -23,8 +23,37 @@ import { OverlaysProvider } from "../../context/overlays/overlaysProvider";
 
 import { ContextMenuPopover, type ContextMenuPopoverProps } from "./contextMenuPopover";
 
-/** DOM element which contains the context menu singleton instance for the imperative ContextMenu APIs. */
-let contextMenuElement: HTMLElement | undefined;
+/**
+ * Options for specifying how an imperatively created context menu should be rendered to the DOM.
+ */
+export interface ShowContextMenuOptions {
+    /**
+     * A new DOM element will be created and appended to this container.
+     *
+     * @default document.body
+     */
+    container?: HTMLElement;
+    /**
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: packages/core/src/components/context-menu/contextMenuSingleton.tsx [hunk 1]
```diff
@@ -23,8 +23,37 @@ import { OverlaysProvider } from "../../context/overlays/overlaysProvider";
...
+     * A new DOM element will be created and appended to this container.
+     *
+     * @default document.body
+     */
+    container?: HTMLElement;
+    /**
+     * A function render the React component onto a newly created DOM element. This should return a function which
+     * unmounts the rendered element from the DOM.
+     */
+    render: ShowContextMenuDOMRenderer;
```
jscheiny, line 40: This parameter is required at the moment to make this type distinguishable from `DOMMountOptions<T>`. In BP6 this will be optional.

ggdouglas (approved):

PR #7396, opened 2025-03-26
[datetime] Rename `dateInput2MigrationUtils` to `dateInputMigrationUtils`

ggdouglas:
#### Proposed changes:
- Rename `dateInput2MigrationUtils` to  `dateInputMigrationUtils` when exported from `@blueprintjs/datetime`
- Alias the re-export from `@blueprintjs/datetime2`


packages/datetime/src/index.ts
```diff
@@ -43,5 +43,5 @@ export { DatePicker, type DatePickerProps } from "./components/date-picker/dateP
 export { DateRangeInput, type DateRangeInputProps } from "./components/date-range-input/dateRangeInput";
 export { DateRangePicker, type DateRangePickerProps } from "./components/date-range-picker/dateRangePicker";
 
-import * as DateInput2MigrationUtils from "./dateInput2MigrationUtils";
-export { DateInput2MigrationUtils };
+import * as DateInputMigrationUtils from "./dateInputMigrationUtils";
+export { DateInputMigrationUtils };
```
jscheiny, line 48: Should we export `DateInput2MigrationUtils` as a deprecated alias?

ggdouglas: Ah, yeah, good call. I'm honestly not sure if there was a plan here for these other than having people eventually port over their components. We should still probably mark as deprecated though.

1e203bff2169689b631d4d479907c0d9ee0cf93e

jscheiny (approved):

PR #7404, opened 2025-03-28
break(ContextMenuPopover): Use React18 to render imperative context menus

jscheiny:
This PR changes the default rendering method for the imperative context menu API to use the React 18 compatible root nodes instead of React 18 `render` and `unmountComponentAtNode`. The breaks in this PR are:

- Switch the default rendering options for `showContextMenu` to use `ReactDOMClient.createRoot`
- Removed `domMounter` and `domUnmounter` fields from the options that can be passed into `showContextMenu`
- Removed options parameter for `hideContextMenu`


packages/core/src/components/context-menu/context-menu-popover.md
```diff
@@ -32,10 +32,15 @@ Two functions are provided as an imperative API for showing and hiding a singlet
...
-export function showContextMenu(props: ContextMenuPopoverProps, options?: ShowContextMenuOptions): void;
-export function hideContextMenu(options?: DOMMountOptions<ContextMenuPopoverProps>): void;
+export function showContextMenu(
+    props: ContextMenuPopoverProps,
+    options?: ShowContextMenuOptions,
+): void;
+export function hideContextMenu(): void;
 ```
 
+@interface ShowContextMenuOptions
```
ggdouglas, line 42: Should we also update the notes below?
```diff
- they create a new React DOM tree via `ReactDOM.render()` (or `ReactDOM.createRoot()` ...
+ they create a new React DOM tree via `ReactDOM.createRoot()` ...
```

jscheiny: Ah good catch, thanks!

ggdouglas (approved):

PR #7419, opened 2025-03-29
[chore] Use new JSX Transform

ggdouglas:
#### Proposed changes

React 17 introduced a [new JSX transform](https://legacy.reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html) that has the following benefits:
> - With the new transform, you can **use JSX without importing React**.
> - Depending on your setup, its compiled output may **slightly improve the bundle size**.
> - It will enable future improvements that **reduce the number of concepts** you need to learn React.

This change switches Blueprint to use the new transform since it is now on React 18.

#### How was this PR created?

This change was made by first running [`update-react-imports`](https://github.com/reactjs/react-codemod?tab=readme-ov-file#update-react-imports) from [`react-codemod`](https://github.com/reactjs/react-codemod). This transformed an initial set of files. The remainder of components that were not processed were edited manually through find and replace.

References:
- [Removing Unused React Imports](https://legacy.reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html#removing-unused-react-imports)


config/tsconfig.web.json
```diff
@@ -1,7 +1,7 @@
 {
     "extends": "./tsconfig.base",
     "compilerOptions": {
-        "jsx": "react",
+        "jsx": "react-jsx",
```
ggdouglas, line 4: https://legacy.reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html#typescript
https://devblogs.microsoft.com/typescript/announcing-typescript-4-1/#jsx-factories
https://www.typescriptlang.org/tsconfig/#jsx

packages/webpack-build-scripts/webpack.config.base.mjs
```diff
@@ -179,7 +177,7 @@ export default {
                             legacyDecorator: true,
                             react: {
                                 refresh: !IS_PRODUCTION,
-                                runtime: "classic",
+                                runtime: "automatic",
```
ggdouglas, line 180: Setting this to "automatic" uses the new JSX runtime module

https://swc.rs/docs/configuration/compilation#jsctransformreactruntime

> Use `runtime: automatic` to use a JSX runtime module (e.g. `react/jsx-runtime` introduced in React 17).

PR #7425, opened 2025-04-03
[core]: Remove deprecated `Classes.SELECT` class from CSS API

### LINKED ISSUE #7427: Remove deprecated `Classes.SELECT` from CSS API
Remove implementation of deprecated `Classes.SELECT` from v6

https://github.com/palantir/blueprint/blob/3ec4b96f68c54a797163469a21e06fc153756669/packages/core/src/common/classes.ts#L181-L182

https://github.com/palantir/blueprint/blob/3ec4b96f68c54a797163469a21e06fc153756669/packages/core/src/components/html-select/_html-select.scss#L100-L107

ggdouglas:

Removes deprecated `Classes.SELECT` from the CSS API along with associated styles.

### DIFF
--- a/packages/core/src/components/html-select/_html-select.scss
+++ b/packages/core/src/components/html-select/_html-select.scss
@@ -67,18 +67,3 @@
     }
   }
 }
-
-.#{$ns}-html-select {
-  .#{$ns}-icon {
-    @extend %pt-select-icon;
-  }
-}
-
-// N.B. this icon implementation is deprecated and will be removed in Blueprint 6.0
-.#{$ns}-select {
-  &::after {
-    @extend %pt-select-icon;
-    @include pt-icon();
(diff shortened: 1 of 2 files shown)


packages/core/src/components/html-select/_html-select.scss
```diff
@@ -67,18 +67,3 @@
     }
   }
 }
-
-.#{$ns}-html-select {
-  .#{$ns}-icon {
-    @extend %pt-select-icon;
-  }
-}
```
ggdouglas, line 75: Removed this also since it duplicates the computed selector above: https://github.com/palantir/blueprint/blob/c09fa6a9956eaed242056c9ff755b5159794c8f0/packages/core/src/components/html-select/_html-select.scss#L23-L26

jscheiny (approved):

PR #7437, opened 2025-04-14
[chore] Refactor Collapse to a functional component

ggdouglas:
Refactors `<Collapse>` from a class-based to a functional component.

[Before](https://blueprintjs.com/docs/#core/components/collapse) | [After](https://output.circle-artifacts.com/output/job/15fcd308-f8c2-4ab9-b3e6-df980c67201f/artifacts/0/packages/docs-app/dist/index.html#core/components/collapse)

**Part of:**
- #6289


packages/core/src/components/collapse/collapse.tsx
```diff
@@ -121,158 +108,144 @@ export enum AnimationStates {
...
+            setHeight("0px");
         }
-    }
-
-    public componentDidUpdate() {
-        if (this.contents == null) {
-            return;
+    }, [isOpen]);
+
+    const contentsRefHandler = (el: HTMLElement | null) => {
```
cameronjoyner: This is fairly nit-picky of me... but I prefer this:
```suggestion
    const contentsRefHandler = (element: HTMLElement | null) => {
```

Feel free to ignore or push back

packages/core/src/components/collapse/collapse.tsx
```diff
@@ -121,158 +108,144 @@ export enum AnimationStates {
...
+            const immediateTimer = setTimeout(() => {
+                setAnimationState(AnimationStates.CLOSING);
+                setHeight("0px");
+                setHeightWhenOpen(clientHeight);
+            });
 
-        const containerStyle = {
-            height: isContentVisible ? this.state.height : undefined,
-            overflowY: isAutoHeight ? "visible" : undefined,
-            // transitions don't work with height: auto
```
cameronjoyner, line 187: How did you overcome this @ggdouglas?

cameronjoyner (commented):
In general I think this is looking good

packages/core/src/components/collapse/collapse.tsx
ggdouglas: Yep, no problem: 4803cf596472c7f7e8bc353b04ee871adb949e09

packages/core/src/components/collapse/collapse.tsx
ggdouglas: Oh, so the styling logic didn't change here, but while looking at this more closely I did notice a major bug. The `contentsRefHandler` was defined as an inline arrow function, so React re-invoked it on every render, resetting `animationState` directly to `CLOSED`, which meant the content disappeared instantly instead of animating through the `CLOSING_START → CLOSING → CLOSED` transition.

I've fixed this in aed257f9941eedfdb42ef7c8b3697be4b049f4e7, thanks for the catch!

packages/core/src/components/collapse/collapse.test.tsx
```diff
@@ -14,60 +14,111 @@
...
-                <div className="removed-child" />
+                <div>Content</div>
             </Collapse>,
         );
-        assert.lengthOf(collapse.find(".removed-child"), 0);
+
+        // Child should be unmounted after the animation completes
+        await waitFor(() => {
+            expect(screen.queryByText("Content")).not.toBeInTheDocument();
+        });
```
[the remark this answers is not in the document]
cameronjoyner: I'm not terribly worried about the time cost

packages/core/src/components/collapse/collapse.test.tsx
```diff
@@ -14,60 +14,111 @@
...
             </Collapse>,
         );
-        assert.lengthOf(collapse.find(".hidden-child"), 1);
+
+        // Child should still be mounted but hidden after the animation completes
+        await waitFor(() => {
+            const child = screen.getByText("Content");
+            expect(child).toBeInTheDocument();
+            expect(child.parentElement).toHaveAttribute("aria-hidden", "true");
+        });
```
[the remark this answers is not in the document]
cameronjoyner: ditto

packages/core/src/components/collapse/collapse.tsx
```diff
@@ -58,19 +58,6 @@ export interface CollapseProps extends Props {
...
-
-    /** The height that should be used for the content animations. This is a CSS value, not just a number. */
-    height: string | undefined;
-
-    /**
-     * The most recent non-zero height (once a height has been measured upon first open; it is undefined until then)
-     */
-    heightWhenOpen: number | undefined;
-}
-
```
cameronjoyner, line 73: What was the motivation to stop exposing this to consumers?

packages/core/src/components/collapse/collapse.tsx
```diff
@@ -121,158 +108,159 @@ export enum AnimationStates {
...
+        }
+    }, []);
 
+    // Handle isOpen prop changes
+    useEffect(() => {
         if (isOpen) {
             switch (animationState) {
                 case AnimationStates.OPEN:
-                    // no-op
-                    break;
```
cameronjoyner, line 142: what specifically allowed us to remove this `break`?

packages/core/src/components/collapse/collapse.tsx
```diff
@@ -121,158 +108,159 @@ export enum AnimationStates {
...
+    // Handle animation state transitions
+    useEffect(() => {
+        if (!contents.current) return undefined;
 
-    public render() {
-        const isContentVisible = this.state.animationState !== AnimationStates.CLOSED;
-        const shouldRenderChildren = isContentVisible || this.props.keepChildrenMounted;
-        const displayWithTransform = isContentVisible && this.state.animationState !== AnimationStates.CLOSING;
-        const isAutoHeight = this.state.height === "auto";
+        if (animationState === AnimationStates.OPEN_START) {
```
cameronjoyner: Feel free to ignore this... but if we have a fixed enum of `AnimationState`(s) should we do a `switch` instead?

packages/core/src/components/collapse/collapse.tsx
cameronjoyner: 👏

packages/core/src/components/collapse/collapse.tsx
```diff
@@ -121,158 +108,159 @@ export enum AnimationStates {
...
-        component: "div",
-        isOpen: false,
-        keepChildrenMounted: false,
-        transitionDuration: 200,
-    };
-
-    public static getDerivedStateFromProps(props: CollapseProps, state: CollapseState) {
-        const { isOpen } = props;
-        const { animationState } = state;
+export const Collapse: React.FC<CollapseProps> = ({
```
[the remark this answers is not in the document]
ggdouglas: Not valid.

packages/core/src/components/collapse/collapse.tsx
ggdouglas: `CollapseState` was never in `components/index.ts`, so removing it isn't a breaking change. With hooks the state is just split across individual `useState` calls instead.

packages/core/src/components/collapse/collapse.tsx
ggdouglas: The two no-op cases are combined as fall-throughs now: `case OPEN: case OPENING: break;`. The old comments were referencing class internals (`Collapse#onDelayedStateChange())` that don't exist anymore, so they didn't carry over.

packages/core/src/components/collapse/collapse.tsx
ggdouglas: Good call! Switched it to a `switch`. 9aeabc47661fbaefa126b5c7f29b5aca15f6f3ff

cameronjoyner (approved):

PR #7441, opened 2025-04-16
Fix button group stacking context

### LINKED ISSUE #7440: ButtonGroup hovers over absolutely positioned elements
#### Environment

-   **Package version(s)**: bp5
-   **Operating System**: mac
-   **Browser name and version**: any

#### Code Sandbox
Code sandbox is down but pretty easy to repro on BP docs

#### Steps to reproduce

Image
- Set some element to be absolutely position
- Observe how button group floats over the absolutely positioned element, without specifying a z-index to do so

Image
resetting stacking context causes button group to not float over absolutely positioned context

#### Actual behavior
ButtonGroups all have a z-index on the button elements within, causing them to appear over other absolutely position context

#### Expected behavior
Button groups should only appear to float above absolutely positioned content if I as a BP consumer specify a z-index higher than my other content or include them later in the DOM tree as a sibling

#### Possible solution
PR incoming - the stacking context needs to be reset at the button group level. It looks like this was done for control groups [here](https://github.com/palantir/blueprint/issues/700) but was never raised for button groups.

#### Other info
[This PR](https://github.com/palantir/blueprint/pull/592/files#diff-4ee502502f5c7a82b4d82de58a473f9a1a0f24d9299fe09388bd02b212bea0edL66) introduced this issue, by switching from `z-index: 0` which was previously resetting the stacking context, to using the `$control-group-stack`, but with button groups not receiving any z-index: 0 due to inputs being defined earlier in that list.

evansjohnson:
#### Checklist

-   [n/a] Includes tests
-   [n/a] Update documentation

Reset stacking context of button group component

Could this unintentionally override any consumer land styling?

#### Screenshot
Screenshot 2025-04-15 at 11 37 47 PM
Screenshot 2025-04-15 at 11 37 41 PM

This PR ended up using the `translateZ(0)` method to reset the stacking context like we do for control groups but anything that resets the stacking context should be valid.

### DIFF
--- a/packages/core/src/components/button/_button-group.scss
+++ b/packages/core/src/components/button/_button-group.scss
@@ -43,7 +43,7 @@ Styleguide button-group
 
 .#{$ns}-button-group {
   display: inline-flex;
-  z-index: 0; // needed for a fresh stacking context for z-indexes applied in below styles
+  @include new-render-layer(); // fresh stacking context for z-indexes applied below
 
   .#{$ns}-button {
     // ensure button can never be smaller than its default size
--- a/packages/core/src/components/button/_button-group.scss
+++ b/packages/core/src/components/button/_button-group.scss
@@ -43,7 +43,7 @@ Styleguide button-group
 
 .#{$ns}-button-group {
   display: inline-flex;
-  @include new-render-layer(); // fresh stacking context for z-indexes applied below
+  z-index: 0; // fresh stacking context for z-indexes applied below
 
   .#{$ns}-button {
     // ensure button can never be smaller than its default size
--- a/packages/core/src/components/button/_button-group.scss
+++ b/packages/core/src/components/button/_button-group.scss
@@ -43,6 +43,7 @@ Styleguide button-group
 
 .#{$ns}-button-group {
   display: inline-flex;
+  z-index: 0; // needed for a fresh stacking context for z-indexes applied in below styles
 
   .#{$ns}-button {
     // ensure button can never be smaller than its default size
--- a/packages/core/src/components/button/_button-group.scss
+++ b/packages/core/src/components/button/_button-group.scss
@@ -43,7 +43,9 @@ Styleguide button-group
 
 .#{$ns}-button-group {
   display: inline-flex;
-  z-index: 0; // fresh stacking context for z-indexes applied below
+
+  // fresh stacking context for z-indexes applied below
+  z-index: 0;
 
   .#{$ns}-button {
     // ensure button can never be smaller than its default size
--- a/packages/core/changelog/@unreleased/pr-7441.v2.yml
+++ b/packages/core/changelog/@unreleased/pr-7441.v2.yml
@@ -1,0 +1,5 @@
+type: fix
+fix:
+  description: Fix button group stacking context
+  links:
+  - https://github.com/palantir/blueprint/pull/7441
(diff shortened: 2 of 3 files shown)


evansjohnson:
Screenshot 2025-04-15 at 11 44 31 PM

an LLM response when asking `I need to reset a z-index stacking context in CSS, are there any performance considerations between z-index: 0 and transform: translateZ(0)? which would be more performant?`

seeing this I'm tempted to say we should at least avoid introducing more `translateZ(0)`, specifically for button groups which there may be a lot of rendered at once

we could separately consider switching the implementation to reset the control group stacking context but that has been in place for awhile

packages/core/src/components/button/_button-group.scss
```diff
@@ -44,6 +44,9 @@ Styleguide button-group
 .#{$ns}-button-group {
   display: inline-flex;
 
+  // fresh stacking context for z-indexes applied below
+  z-index: 0;
```
ggdouglas, line 48: The button group is positioned **`static`** by default, so should we also provide a **`position: relative`** here? Otherwise, in a vacuum, this `z-index` won't have an effect:

Screenshot 2025-04-16 at 11 10 17@2x

---

If the button group is a child element of a flexbox parent, however, it does apply. I'm a little fuzzy on the spec here, but I believe this has something to do with [flex item z-ordering](https://www.w3.org/TR/css-flexbox-1/#painting).

Here's a quick comparison: https://codesandbox.io/p/sandbox/q4v2ch

Screenshot 2025-04-16 at 11 30 50@2x

evansjohnson: oh great catch here!

is the assumption that this is safe to do because consumers that rely on the current `static` position should not have set something like `left: 100px` since it would not make any impact with the current `static` position unless they have already specified an explicit position?

also wondering if somehow BP classes are weaker selectors than user styles - ex could this wipe out a user defined `position: absolute`

---

I hadn't heard of it but from looking at the [MDN stacking context docs](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_positioned_layout/Stacking_context#features_creating_stacking_contexts) [`isolation: isolate`](https://developer.mozilla.org/en-US/docs/Web/CSS/isolation) seems like exactly what we want - any thoughts on that?

evansjohnson: Screenshot 2025-04-16 at 4 45 48 PM

it appears to work from the docs test

ggdouglas: > also wondering if somehow BP classes are weaker selectors than user styles - ex could this wipe out a user defined `position: absolute`

Hm, I'm not as worried about this since we're fixing an existing buggy behavior in the process, but it is a fair point.

> I hadn't heard of it but from looking at the [MDN stacking context docs](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_positioned_layout/Stacking_context#features_creating_stacking_contexts) [isolation: isolate](https://developer.mozilla.org/en-US/docs/Web/CSS/isolation) seems like exactly what we want - any thoughts on that?

I've only ever used `isolation` for purposes related to isolating elements from `mix-blend-mode`, but I don't see why we couldn't use it for `z-index` also. I suppose, as you've said, the benefit here is that there is much less chance of inadvertently overriding user styles. I'm on board with that! Suggest also adding an explainer comment to clarify intent.

evansjohnson: I suppose we already do it for control groups but any concerns about suddenly creating a new stacking context (which I believe creates a new render layer) suddenly introducing perf issues if many button groups are used?

I think isolating the stacking context is needed in some way unless BP takes a global approach to z-indexes, but it might be safer to do in a major version with an opt-out available to avoid creating a stacking context. I don't have a gauge on if it could become an issue at 100 or 10k or if this isn't a concern at all really.

PR #7477, opened 2025-06-10
[datetime] Export `TIMEZONES` and `MINIMAL_TIMEZONES`

ggdouglas:
Export `TIMEZONEs` and `MINIMAL_TIMEZONEs` lists to allow consumers to construct customized timezone selection components.

Also include example in demo app on how to leverage these exports to handle more advanced/custom use cases with timezones in DateInput:

Screenshot 2026-01-07 at 14 04 57@2x

### DIFF
--- a/packages/datetime/src/common/timezoneItems.ts
+++ b/packages/datetime/src/common/timezoneItems.ts
@@ -313,8 +313,14 @@ const minimalTimezonesWithoutOffset: TimezoneWithoutOffset[] = [
     { label: "Kiritimati", ianaCode: "Pacific/Kiritimati" },
 ];
 
-/** List of standard timezones */
-export const TIMEZONE_ITEMS = timezonesWithoutOffset.map(tz => lookupTimezoneOffset(tz));
+/**
+ * Comprehensive list of timezones with their IANA codes, labels, and UTC offsets.
+ * Useful for building custom timezone selection components.
+ */
+export const TIMEZONES = timezonesWithoutOffset.map(tz => lookupTimezoneOffset(tz));
 
-/** Minimal list of timezones */
-export const MINIMAL_TIMEZONE_ITEMS = minimalTimezonesWithoutOffset.map(tz => lookupTimezoneOffset(tz));
+/**
+ * Curated minimal list of timezones covering major world regions.
+ * Useful for simplified timezone selection UI where a full list would be overwhelming.
+ */
+export const MINIMAL_TIMEZONES = minimalTimezonesWithoutOffset.map(tz => lookupTimezoneOffset(tz));
--- a/packages/datetime/src/index.ts
+++ b/packages/datetime/src/index.ts
@@ -27,7 +27,7 @@ export { MonthAndYear } from "./common/monthAndYear";
 export type { DayPickerProps } from "./common/reactDayPickerProps";
 export type { TimePickerProps } from "./common/timePickerProps";
 export { TimePrecision } from "./common/timePrecision";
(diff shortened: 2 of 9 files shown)


packages/datetime/src/index.ts
```diff
@@ -27,6 +27,7 @@ export { MonthAndYear } from "./common/monthAndYear";
 export type { DayPickerProps } from "./common/reactDayPickerProps";
 export type { TimePickerProps } from "./common/timePickerProps";
 export { TimePrecision } from "./common/timePrecision";
+export { TIMEZONE_ITEMS, MINIMAL_TIMEZONE_ITEMS } from "./common/timezoneItems";
```
benrucker: Do you think the names are crisp enough for an addition to the API? We could maybe consider condensing them to something like `TIMEZONES` and `MINIMAL_TIMEZONES`, maybe also TSDocs on the fields. What do you think?

benrucker:
Think we could also export `formatTimezone` to go along with this? Tho would also have to export `getTimezoneMetadata` to get the input to `formatTimezone`.

Edit: thinking about this more, it could be nice to export a function that takes a IANA timezone string and a timezone format and outputs the nicely formatted string without needing to do the lookup to TimezoneWithNames beforehand. 

Although, my team's needs on this have changed, so honestly we can just close this PR if wanted as it's not pressing for me anymore.

packages/datetime/src/common/timezoneItems.ts
```diff
@@ -313,8 +313,14 @@ const minimalTimezonesWithoutOffset: TimezoneWithoutOffset[] = [
...
+ */
+export const TIMEZONES = timezonesWithoutOffset.map(tz => lookupTimezoneOffset(tz));
 
-/** Minimal list of timezones */
-export const MINIMAL_TIMEZONE_ITEMS = minimalTimezonesWithoutOffset.map(tz => lookupTimezoneOffset(tz));
+/**
+ * Curated minimal list of timezones covering major world regions.
+ * Useful for simplified timezone selection UI where a full list would be overwhelming.
+ */
+export const MINIMAL_TIMEZONES = minimalTimezonesWithoutOffset.map(tz => lookupTimezoneOffset(tz));
```
mm-wang, line 326: **suggestion, non-blocking**: I'd perhaps aim for a single array of timezones that we generate minimal timezones from, but I don't feel super strongly about it and it doesn't affect this specific work.

packages/datetime/src/common/timezoneItems.ts
```diff
@@ -313,8 +313,14 @@ const minimalTimezonesWithoutOffset: TimezoneWithoutOffset[] = [
     { label: "Kiritimati", ianaCode: "Pacific/Kiritimati" },
 ];
 
-/** List of standard timezones */
-export const TIMEZONE_ITEMS = timezonesWithoutOffset.map(tz => lookupTimezoneOffset(tz));
+/**
+ * Comprehensive list of timezones with their IANA codes, labels, and UTC offsets.
+ * Useful for building custom timezone selection components.
+ */
+export const TIMEZONES = timezonesWithoutOffset.map(tz => lookupTimezoneOffset(tz));
```
mm-wang, line 320: **general note**: `lookupTimezoneOffset` is a bit of a misnomer based on the object that method returns...

mm-wang (commented):
Is the expectation that, in the demo app, if the timezone is changed that the time is then *not* changed? See the video below.

If so, I'm wondering if that is expected. My experience as a user of these types of inputs before is that the time changes with the time zone.

https://github.com/user-attachments/assets/a7cd6b34-12a6-4ef9-887e-2da5b09935e3

PR #7490, opened 2025-06-23
[core] feat(Tag): a11y names for tags via button roles for tags with onClick behavior

dstulski:

Tag with onClick behaviors have their roles set to "button" to ensure a11y readers are compatible with tag-style buttons

#### Screenshot

Ensures tag-style buttons comply with: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/button_role


palantirtech:
Thanks for your interest in [palantir/blueprint](https://github.com/palantir/blueprint), @dstulski! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

packages/core/src/components/tag/tag.tsx
```diff
@@ -135,7 +135,12 @@ export const Tag: React.FC<TagProps> = React.forwardRef((props, ref) => {
     );
 
     return (
-        <span {...removeNonHTMLProps(htmlProps)} {...interactiveProps} className={tagClasses}>
+        <span
+            {...removeNonHTMLProps(htmlProps)}
+            {...interactiveProps}
+            className={tagClasses}
+            role={interactiveProps.onClick !== undefined ? "button" : undefined}
```
ggdouglas: Can we reuse the `isInteractive` variable from above? Seems like it would read a bit more directly here:
```suggestion
            role={isInteractive ? "button" : undefined}
```

ggdouglas (approved):

PR #7499, opened 2025-07-02
[chore] Update Config to Use New JSX Transform

ggdouglas:
## Changes

- Updates base, web/docs, and test configurations to use the [new JSX transform](https://legacy.reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html).
- Also updates any components where the React import is no longer used (enforced by linter).

Configuration changes are committed in 3ac7db9fc27faf649cd63658d66ada4c498ee218 and 4b7efad512613d6203a7eb3af9ef918e01fdb735.

This PR is a scoped down version of https://github.com/palantir/blueprint/pull/7419 that modifies only files essential to updating to use the new transform.

The following FLUP to this PR (https://github.com/palantir/blueprint/pull/7500) modifies the remaining files in the repo.

## Context

This new transform has a few advantages:

- It simplifies and modernizes syntax around React imports
- Compiled output from the transform may slightly improve the bundle size
- React will eventually require the new transform (see [note](https://react.dev/blog/2024/04/25/react-19-upgrade-guide#new-jsx-transform-is-now-required) in React 19 upgrade guide).

### DIFF
--- a/config/tsconfig.web.json
+++ b/config/tsconfig.web.json
@@ -1,7 +1,7 @@
 {
     "extends": "./tsconfig.base",
     "compilerOptions": {
-        "jsx": "react",
+        "jsx": "react-jsx",
         "lib": ["dom", "dom.iterable", "es2019"],
         "module": "es2020",
         "moduleResolution": "node",
--- a/packages/core/src/components/dialog/dialogStepButton.tsx
+++ b/packages/core/src/components/dialog/dialogStepButton.tsx
@@ -14,8 +14,6 @@
  * limitations under the License.
  */
 
-import * as React from "react";
-
 import type { ButtonSharedPropsAndAttributes } from "../button/buttonProps";
 import { AnchorButton } from "../button/buttons";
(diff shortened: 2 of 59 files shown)


config/tsconfig.web.json
```diff
@@ -1,7 +1,7 @@
 {
     "extends": "./tsconfig.base",
     "compilerOptions": {
-        "jsx": "react",
+        "jsx": "react-jsx",
```
ggdouglas, line 4: https://www.typescriptlang.org/tsconfig/#jsx

> - react-jsx: Emit .js files with the JSX changed to _jsx calls optimized for production
> - react: Emit .js files with JSX changed to the equivalent React.createElement calls

https://www.typescriptlang.org/docs/handbook/jsx.html

> "react (classic runtime), react-jsx (automatic runtime)"

The automatic runtime is the new transform (see also ["Manual Babel Setup"](https://legacy.reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html#how-to-upgrade-to-the-new-jsx-transform))

invliD (approved):

PR #7501, opened 2025-07-02
Feature/divider compact prop

liubrandon:
## Summary
  - Add a `compact` prop to the `Divider` component that sets margin to 0
  - Update documentation to reflect the new prop
  - Add SCSS styling for the compact variant
  - Update the interactive playground example to demonstrate the compact prop

  ## Test plan
  - Verify the Divider renders correctly with `compact={true}`
  - Verify the Divider has margin: 0 when compact prop is set
  - Verify the interactive playground allows toggling the compact prop
  - Verify existing Divider functionality works as expected

Smoketest (replace anything that says minimal with compact, just didn't update video):

https://github.com/user-attachments/assets/a30de8d7-4089-4562-bf84-7a9429cb8e3a

### DIFF
--- a/packages/core/src/components/divider/_divider.scss
+++ b/packages/core/src/components/divider/_divider.scss
@@ -16,7 +16,7 @@ $divider-margin: $pt-grid-size * 0.5 !default;
     border-color: $pt-dark-divider-white;
   }
 
-  &-compact {
+  .#{$ns}-compact {
     margin: 0;
   }
 }
--- a/packages/core/src/components/divider/_divider.scss
+++ b/packages/core/src/components/divider/_divider.scss
@@ -15,4 +15,8 @@ $divider-margin: $pt-grid-size * 0.5 !default;
   .#{$ns}-dark & {
     border-color: $pt-dark-divider-white;
   }
+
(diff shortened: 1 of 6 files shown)


packages/core/src/components/divider/_divider.scss
```diff
@@ -15,4 +15,8 @@ $divider-margin: $pt-grid-size * 0.5 !default;
   .#{$ns}-dark & {
     border-color: $pt-dark-divider-white;
   }
+
+  &-compact {
```
ggdouglas: I'd suggest we name this class off of the top level `bp*` namespace rather than nesting it under `divider`.
```suggestion
  .#{$ns}-compact {
```
This would be more in line with other components that have this "compact" prop/class like Callout and Card:
https://github.com/palantir/blueprint/blob/1782045c7837b73b67cf743241214f59f46c6015/packages/core/src/components/callout/_callout.scss#L62-L73
https://github.com/palantir/blueprint/blob/1782045c7837b73b67cf743241214f59f46c6015/packages/core/src/components/card/_card.scss#L43-L45

PR #7516, opened 2025-07-25
[DateInput]: Remove `defaultProps` and consolidate datetime constants

ggdouglas:
#### Part of #7518

- Replaces `defaultProps` on DateInput in favor of ES6 default parameters.
- Also consolidates datetime constant values to be reused as default values across datetime components.
- Removes `DateInputPropsWithDefaults` type.


packages/datetime/src/components/dateConstants.ts
```diff
@@ -0,0 +1,16 @@
+/* !
```
mm-wang, line 1: **suggestion**: `dateContants.ts` --> `dateConstants.ts` (sp).

packages/datetime/src/components/date-input/dateInput.tsx
```diff
@@ -35,11 +35,11 @@ import { getDefaultDateFnsFormat } from "../../common/dateFnsFormatUtils";
 import { useDateFnsLocale } from "../../common/dateFnsLocaleUtils";
 import type { ReactDayPickerSingleProps } from "../../common/reactDayPickerProps";
 import { DatePicker } from "../date-picker/datePicker";
-import { DatePickerUtils } from "../date-picker/datePickerUtils";
+import { INVALID_DATE_MESSAGE, LOCALE, MAX_DATE, MIN_DATE, OUT_OF_RANGE_MESSAGE } from "../dateContants";
```
mm-wang: **suggestion**: `dateContants.ts` --> `dateConstants.ts` (sp).

packages/datetime/src/components/dateConstants.ts
```diff
@@ -0,0 +1,16 @@
...
+ */
+
+import { DatePickerUtils } from "./date-picker/datePickerUtils";
+
+export const LOCALE = "en-US";
+
+export const MAX_DATE = DatePickerUtils.getDefaultMaxDate();
+export const MIN_DATE = DatePickerUtils.getDefaultMinDate();
+
+export const INVALID_DATE_MESSAGE = "Invalid date";
```
mm-wang, line 12: **question**: Do these need i18n?

packages/datetime/src/components/date-range-picker/dateRangePicker.tsx
```diff
@@ -27,7 +27,7 @@ import { DateRangeSelectionStrategy } from "../../common/dateRangeSelectionStrat
 import { combineModifiers, HOVERED_RANGE_MODIFIER } from "../../common/dayPickerModifiers";
 import { MonthAndYear } from "../../common/monthAndYear";
 import { DatePickerProvider } from "../date-picker/datePickerContext";
-import { DatePickerUtils } from "../date-picker/datePickerUtils";
+import { LOCALE, MAX_DATE, MIN_DATE } from "../dateContants";
```
mm-wang: **suggestion**: `dateContants.ts` --> `dateConstants.ts` (sp).

packages/datetime/test/common/dateFormatPropsTests.ts
```diff
@@ -7,6 +7,7 @@ import { format } from "date-fns";
 
 import { getFormattedDateString } from "../../src/common/dateFormatProps";
 import { Months } from "../../src/common/months";
+import { OUT_OF_RANGE_MESSAGE } from "../../src/components/dateContants";
```
mm-wang: **suggestion**: `dateContants.ts` --> `dateConstants.ts` (sp).

mm-wang (commented):
LGTM, just a file spelling update - I tried to flag each place I found the filename to make it easier. Would be curious to know if we use i18n for any error messages or just assume English since we're mostly considering devs.

packages/datetime/src/components/dateConstants.ts
ggdouglas: Ah, good catch! Fixed here: d1a047f33b6fd63ac1dd501094bae772fd8f1c75

packages/datetime/src/components/dateConstants.ts
ggdouglas: Generally, we leave default strings for placeholders and descriptions in English and put the onus on the consumer to provide translated strings directly to the component if they need to. There currently is no built-in i18n in Blueprint.

In this particular case, this string can be overwritten by providing a value to the `invalidDateMessage` prop.

PR #7541, opened 2025-08-12
Migrate `$pt-grid-size` values to `$pt-spacing`

ggdouglas:
## Changes

This PR migrates Blueprint's spacing system from the 10px-based `$pt-grid-size` to a new 4px-based `$pt-spacing` variable. All existing spacing values have been converted to equivalent `$pt-spacing` calculations, ensuring no visual changes. The legacy `$pt-grid-size` and `$half-grid-size` both remain for backward compatibility. Raw pixel values were also migrated for consistency, making it easier to identify and update inconsistent spacing in future work.

Component spacing updates will follow in #7542.

https://github.com/palantir/blueprint/blob/c036e394b174273c3f5346bd093dbdc7b7205209/packages/core/src/common/_variables.scss#L23-L33

### Conversion Table

The following table roughly describes the new format of the changes. The conversion from `$pt-grid-size` to `$pt-spacing` is achieved by multiplying by `2.5` (10px/4px).

| px Value | `$pt-grid-size` (10px base) | `$pt-spacing` (4px base) |
| -------- | ------------------------- | ---------------------- |
| `2px`      | `$pt-grid-size * 0.2`       | `$pt-spacing * 0.5`     |
| `3px`      | `$pt-grid-size * 0.3`       | `$pt-spacing * 0.75`     |
| **`4px`**      | **`$pt-grid-size * 0.4`**       | **`$pt-spacing`**     |
| `5px`      | `$pt-grid-size * 0.5`       | `$pt-spacing * 1.25`     |
| `6px`      | `$pt-grid-size * 0.6`       | `$pt-spacing * 1.5`     |
| `7px`      | `$pt-grid-size * 0.7`       | `$pt-spacing * 1.75`     |
| `8px`      | `$pt-grid-size * 0.8`       | `$pt-spacing * 2`     |
| **`10px`**     | **`$pt-grid-size`**             | **`$pt-spacing * 2.5`**     |
| `11px`     | `$pt-grid-size * 1.1`       | `$pt-spacing * 2.75`     |
| `12px`     | `$pt-grid-size * 1.2`       | `$pt-spacing * 3`     |
| `13px`     | `$pt-grid-size * 1.3`       | `$pt-spacing * 3.25`     |
| `14px`     | `$pt-grid-size * 1.4`       | `$pt-spacing * 3.5`     |
| `15px`     | `$pt-grid-size * 1.5`       | `$pt-spacing * 3.75`     |
| `16px`     | `$pt-grid-size * 1.6`       | `$pt-spacing * 4`     |
| `18px`     | `$pt-grid-size * 1.8`       | `$pt-spacing * 4.5`     |
| `20px`     | `$pt-grid-size * 2`       | `$pt-spacing * 5`     |
| `22px`     | `$pt-grid-size * 2.2`       | `$pt-spacing * 5.5`     |
| `24px`     | `$pt-grid-size * 2.4`       | `$pt-spacing * 6`     |
| `25px`     | `$pt-grid-size * 2.5`       | `$pt-spacing * 6.25`     |
| `30px`     | `$pt-grid-size * 3`       | `$pt-spacing * 7.5`     |
| `36px`     | `$pt-grid-size * 3.6`       | `$pt-spacing * 9`     |
| `40px`     | `$pt-grid-size * 4`       | `$pt-spacing * 10`     |
| `50px`     | `$pt-grid-size * 5`       | `$pt-spacing * 12.5`     |

etc.

- ~~Does the naming of this new variable (`$bp-spacing`) make sense? Are there alternatives we want to consider?~~
  - Renamed to `$pt-spacing`
- Regression testing to ensure no errors were introduced while converting values

### DIFF
--- a/packages/core/src/common/_variables.scss
+++ b/packages/core/src/common/_variables.scss
@@ -55,7 +55,7 @@ $pt-font-family-monospace: monospace !default;
 
 $pt-font-size: $pt-grid-size * 1.4 !default;
 $pt-font-size-large: $pt-grid-size * 1.6 !default;
-$pt-font-size-small: $pt-grid-size * 1.2 !default;
+$pt-font-size-small: $bp-spacing * 3 !default;
 
 // a little bit extra to ensure the height comes out to just over 18px (and browser rounds to 18px)
 // equivalent to: math.div($pt-grid-size * 1.8, $pt-font-size) + 0.0001
--- a/packages/core/src/common/_variables.scss
+++ b/packages/core/src/common/_variables.scss
@@ -24,6 +24,8 @@ $bp-ns: $ns;
 // (so other variables can use it to define themselves)
 $pt-grid-size: 10px !default;
 
+$bp-spacing: 4px !default;
+
 // Icon variables
 
 $icons16-family: "blueprint-icons-16" !default;
--- a/packages/core/src/common/_variables.scss
+++ b/packages/core/src/common/_variables.scss
@@ -66,13 +66,13 @@ $pt-line-height: 1.28581 !default;
 $pt-border-radius: $bp-spacing * 0.5 !default;
 
 // Buttons
-$pt-button-height: $pt-grid-size * 3 !default;
+$pt-button-height: $bp-spacing * 7.5 !default;
 $pt-button-height-small: $bp-spacing * 6 !default;
 $pt-button-height-smaller: $bp-spacing * 5 !default;
 $pt-button-height-large: $pt-grid-size * 4 !default;
 
 // Inputs
-$pt-input-height: $pt-grid-size * 3 !default;
+$pt-input-height: $bp-spacing * 7.5 !default;
 $pt-input-height-large: $pt-grid-size * 4 !default;
 $pt-input-height-small: $bp-spacing * 6 !default;
 
--- a/packages/core/src/common/_variables.scss
+++ b/packages/core/src/common/_variables.scss
@@ -32,7 +32,7 @@ $icons16-family: "blueprint-icons-16" !default;
 $icons20-family: "blueprint-icons-20" !default;
 
 $pt-icon-size-standard: $bp-spacing * 4 !default;
-$pt-icon-size-large: 20px !default;
+$pt-icon-size-large: $bp-spacing * 5 !default;
 
 // Typography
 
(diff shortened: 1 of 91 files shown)


packages/core/src/common/_variables.scss
```diff
@@ -24,13 +24,15 @@ $bp-ns: $ns;
 // (so other variables can use it to define themselves)
 $pt-grid-size: 10px !default;
 
+$bp-spacing: 4px !default;
```
invliD: Why bp-spacing instead of pt-spacing? That creates inconsistency with all the other variables, and I'm not sure the prefix change is worth it.

packages/core/src/common/_variables.scss
```diff
@@ -51,32 +53,31 @@ $pt-font-family:
...
 // a little bit extra to ensure the height comes out to just over 18px (and browser rounds to 18px)
-// equivalent to: math.div($pt-grid-size * 1.8, $pt-font-size) + 0.0001
+// equivalent to: math.div($bp-spacing * 4.5, $pt-font-size) + 0.0001
 $pt-line-height: 1.28581 !default;
 
 // Grids & dimensions
 
-// equivalent to: floor(math.div($pt-grid-size, 5))
-$pt-border-radius: 2px !default;
+$pt-border-radius: $bp-spacing * 0.5 !default;
```
invliD: Should the radius depend on spacing automatically? I feel like these are separate design decisions that aren't necessarily coupled (e.g. I could imaging increasing bp-spacing to 6px, but that doesn't mean border-radius has to go up to 3px).

ggdouglas: Fair point. I'll revert this particular change for now. I think it would make sense to define multiple border-radius sizes at some point (e.g. sm, md, lg) to gain more consistency, but we don't need to bundle that change with this one. 793b6fbadaf2458ac566cdb1468c7d651e086bfa

packages/core/src/common/_variables.scss
ggdouglas: I'm not particularly fussed on the naming here, but I agree with wanting consistency. Changed in 00154e5203e5c8c414a538767bbc647f5fea136f.

I do think it's a bit odd to have `pt-*` prefixed variables in Blueprint considering that external users don't necessarily have any association with Palantir. When writing the variable name originally, I thought it might be a good way to differentiate between older and newer variables, but I agree consistency is more important for now.

invliD: Sounds good. If we ever want to change the prefix, we should do it in a big bang (probably with one major version of having both) so that consumers can use them consistently.

invliD (dismissed):

PR #7543, opened 2025-08-13
MenuItem and QueryList ScreenReader Accessibility Improvements

caitlinmcdougall:
#### Checklist

-   [ ] Includes tests
-   [ ] Update documentation

ScreenReader accessibility improvements for the ItemRenderer, MenuItem, QueryList and Select components

**Added proper ARIA relationships and controls**:
- Added id prop to ItemRenderer interface to enable aria-activedescendant pattern
- Enhanced MenuItem component to accept and use the id prop
- Added aria-controls pointing to the listbox in select component
- Proper aria-activedescendant that references the currently focused option

**Improved focus management in QueryList**:

- Added listboxHasVisualFocus state to track when arrow key navigation is active
- Only set aria-activedescendant when users navigate with arrow keys, not during typing
- Added handleInputFocus and handleInputBlur methods to manage this behavior
- Clear visual focus state when input receives focus or user starts typing

**Updated films example**:
- Modified getFilmItemProps return type to use Omit<MenuItemProps, "key"> to prevent React key conflicts
- Added conditional id prop spreading in the props object

Ensuring this allows for backwards compatibility with other users of the component.


packages/select/src/__examples__/films.tsx
```diff
@@ -138,12 +138,13 @@ export const TOP_100_FILMS: Film[] = [
  */
 export function getFilmItemProps(
     film: Film,
-    { handleClick, handleFocus, modifiers, query, ref }: ItemRendererProps,
-): MenuItemProps & React.Attributes {
+    itemProps: ItemRendererProps,
+): Omit<MenuItemProps, "key"> & React.Attributes {
```
ggdouglas: If we're omitting `key` from `MenuItemProps` for the return type, we should also remove `React.Attributes` since the intersection of that type just ends up adding it back:
```ts
interface Attributes {
    key?: Key | null | undefined;
}
```
```suggestion
): Omit<MenuItemProps, "key"> {
```

I'm supportive of this fix for the docs examples, but could we split that part out into its own PR?

packages/select/src/components/query-list/queryList.tsx
```diff
@@ -14,6 +14,8 @@
  * limitations under the License.
  */
 
+import * as React from "react";
```
ggdouglas: This React import can be removed:
```suggestion

```

packages/select/src/components/query-list/queryList.tsx
```diff
@@ -199,6 +234,9 @@ export class QueryList<T> extends AbstractComponent<QueryListProps<T>, QueryList
     public constructor(props: QueryListProps<T>) {
         super(props);
 
+        // Generate unique ID for accessibility
+        this.listId = props.listId ?? `bp5-query-list-${Utils.uniqueId("ql")}`;
```
ggdouglas: Suggest removing the hardcoded version prefix and replacing with the following:
```suggestion
        this.listId = props.listId ?? Utils.uniqueId("bp-query-list");
```

packages/select/src/components/query-list/queryList.tsx
```diff
@@ -555,8 +613,23 @@ export class QueryList<T> extends AbstractComponent<QueryListProps<T>, QueryList
         onKeyUp?.(event);
     };
 
+    private handleInputFocus = (_event: React.FocusEvent<HTMLInputElement>) => {
+        // When input gains focus, we don't immediately set visual focus state
+        // Visual focus state is only set when the user navigates with arrow keys
+        // This ensures screen readers only announce activedescendant when appropriate
+    };
```
ggdouglas: Why are we defining this empty focus event handler?

packages/select/src/components/query-list/queryList.tsx
```diff
@@ -229,10 +275,15 @@ export class QueryList<T> extends AbstractComponent<QueryListProps<T>, QueryList
                 ...spreadableState,
                 items,
                 itemsParentRef: this.refHandlers.itemsParent,
-                menuProps,
+                menuProps: {
+                    ...menuProps,
+                    id: this.listId,
+                    role: "listbox",
```
ggdouglas: The default `itemListRenderer` implementation already includes this role. Do we need to include it again here?

https://github.com/palantir/blueprint/blob/3029aa28fce586576f0ea110d9d21975c2695a09/packages/select/src/components/query-list/queryList.tsx#L367

packages/select/src/components/query-list/queryList.tsx
```diff
@@ -229,10 +275,15 @@ export class QueryList<T> extends AbstractComponent<QueryListProps<T>, QueryList
                 ...spreadableState,
                 items,
                 itemsParentRef: this.refHandlers.itemsParent,
-                menuProps,
+                menuProps: {
+                    ...menuProps,
+                    id: this.listId,
```
ggdouglas, line 273: Won't setting an `id` here replace the id referenced by the outermost combobox that already defines `aria-controls`? I'm a bit confused as to what we're trying to accomplish. Here's a before/after of these changes on a simple example to illustrate:

| Before | After |
|--------|--------|
| before | after |

https://codesandbox.io/p/sandbox/v4x4r8

packages/select/src/components/select/select.tsx
```diff
@@ -175,13 +175,16 @@ export class Select<T> extends AbstractPureComponent<SelectProps<T>, SelectState
 
         const input = (
             <InputGroup
+                aria-activedescendant={listProps.activeItemId}
```
ggdouglas, line 179: Why are we setting [`aria-activedescendant`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-activedescendant) on the filter input group? Shouldn't we be setting this on the outermost combobox element?

Here's the WAI [Select-Only Combobox Example](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/examples/combobox-select-only) for reference:

Screenshot 2025-08-13 at 12 14 27@2x

ggdouglas (commented):
I'm a bit confused as to what the specific accessibility improvements here are. Can you provide some examples of what you're trying to achieve?

Generally supportive of allowing `id` to be passed on to MenuItem and for setting `aria-activedescendant` (though I have questions regarding how this is currently implemented).

packages/select/src/components/select/select.tsx
caitlinmcdougall: The aria-activedescendant is on the InputGroup because that's the element that has keyboard focus when users navigate the options with arrow keys. The outermost combobox element loses focus once the popover opens and the InputGroup is focused. Happy to revise if this doesn't make sense.

packages/select/src/components/query-list/queryList.tsx
caitlinmcdougall: Missed this, will remove thank you!

packages/select/src/components/query-list/queryList.tsx
caitlinmcdougall: Will remove thank you

packages/select/src/components/query-list/queryList.tsx
caitlinmcdougall: Will remove thank you

packages/select/src/__examples__/films.tsx
caitlinmcdougall: That makes sense, thank you. I mostly am keeping in the docs examples as it is useful for my testing. Unless I'm not supposed to test on the docs examples? What is the better way to test?

packages/select/src/components/select/select.tsx
caitlinmcdougall: I have moved the other aria details to the InputGroup now to be consistent

packages/select/src/components/query-list/queryList.tsx
```diff
@@ -229,10 +266,17 @@ export class QueryList<T> extends AbstractComponent<QueryListProps<T>, QueryList
                 ...spreadableState,
                 items,
                 itemsParentRef: this.refHandlers.itemsParent,
-                menuProps,
+                menuProps: {
+                    ...menuProps,
+                    // Hide listbox from screen readers when typing (only show when navigating with arrows)
+                    // This prevents VoiceOver from announcing fragmented updates during typing/backspacing
+                    "aria-hidden": !this.state.listboxHasVisualFocus,
```
ggdouglas: Can you explain the intent behind the `listboxHasVisualFocus` state?

Also, interestingly, Chrome warns about using `aria-hidden` for this and recommends using [inert](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/inert) instead.

Screenshot 2025-11-19 at 13 27 17@2x

> Blocked aria-hidden on an element because its descendant retained focus. The focus must not be hidden from assistive technology users. Avoid using aria-hidden on a focused element or its ancestor. Consider using the inert attribute instead, which will also prevent focus. For more details, see the aria-hidden section of the WAI-ARIA specification at https://w3c.github.io/aria/#aria-hidden.

anonymous-username602: The aria-hidden stops the screen reader from trying to read both the filter box and the film names at the same time, which would otherwise result in this:
image

anonymous-username602:
Before:
https://github.com/user-attachments/assets/cc116939-9be8-4ba6-966b-1296d58f1799


After:
https://github.com/user-attachments/assets/fd95dc88-ec25-4952-9cc4-fbf4e6a89fce

packages/select/src/__examples__/filmSelect.tsx
```diff
@@ -72,9 +72,12 @@ export function FilmSelect({ allowCreate = false, fill, ...restProps }: FilmSele
             }
             return (
                 <MenuItem
+                    aria-label={`${film.rank}. ${film.title} (${film.year})`}
```
ggdouglas: **nit:** What is the purpose of adding an additional `aria-label` example here in addition to the MenuItem text?

packages/select/src/__examples__/filmSelect.tsx
```diff
@@ -72,9 +72,12 @@ export function FilmSelect({ allowCreate = false, fill, ...restProps }: FilmSele
...
+                    key={film.rank}
                     {...getFilmItemProps(film, props)}
                     roleStructure="listoption"
                     selected={film === selectedFilm}
+                    tagName="div"
```
ggdouglas, line 79: **nit:** Why use a div instead of the default anchor tag?

packages/select/src/components/query-list/queryList.tsx
```diff
@@ -229,10 +266,17 @@ export class QueryList<T> extends AbstractComponent<QueryListProps<T>, QueryList
                 ...spreadableState,
                 items,
                 itemsParentRef: this.refHandlers.itemsParent,
-                menuProps,
+                menuProps: {
+                    ...menuProps,
+                    // Hide listbox from screen readers when typing (only show when navigating with arrows)
+                    // This prevents VoiceOver from announcing fragmented updates during typing/backspacing
```
ggdouglas: **nit:** Can we replace references to "VoiceOver" with the more generic term "screen reader"?
```suggestion
                    // This prevents screen readers from announcing fragmented updates during typing/backspacing
```

ggdouglas (dismissed):
A few questions, but this generally seems like a good set of improvements on the current component. 👍

packages/select/src/__examples__/filmSelect.tsx
anonymous-username602: When on the default tag it will read out 'Link, 1, Shawshank Redemption', with the tag Div it will read out '1, Shawshank Redemption'. Since they aren't functionally links the latter seems better practice.

ggdouglas (approved):

packages/select/src/components/select/select.tsx
```diff
@@ -249,7 +257,7 @@ export class Select<T> extends AbstractPureComponent<SelectProps<T>, SelectState
                     ),
                     onKeyUp: this.withPopoverTargetPropsHandler("keyup", isOpen ? handleKeyUp : undefined),
                     ref,
-                    role: "combobox",
+                    role: filterable ? undefined : "combobox",
```
bvandrc, line 260: @caitlinmcdougall @ggdouglas  this change has created aXe a11y failures because now this element has several properties that are not allowed on an item that has no role. The failure problem is "aria-expanded", but other properties no longer apply that should. 

IMO this change wasn't the right move; an a11y user would probably expect, in a form, for the item that opens the list to be the `combobox`, regardless of the filter input.

Also, this can't be `button` because there is a child that is a `button`.

PR #7554, opened 2025-08-27
fix: tag role on interactive

bvandrc:
#### Checklist

-   [N/A] Includes tests
-   [N/A] Update documentation

`Tag` should not get `role="button"` when is only stylistically interactive (has `interactive` prop but not `onClick`). Also, should allow overwriting of `role` prop if want to use custom.

Proper a11y role.

#### Screenshot

N/A


bvandrc:
@ggdouglas this has been rebased to latest.

PR #7556, opened 2025-08-29
Fix toast jitter

sbsto:
#### Checklist

-   [ ] Includes tests (N/A)
-   [ ] Update documentation (N/A)

Make the toast animation a little more smooth by modifying the css to use relative height rather than absolute and have the toasts start off screen.

The CSS file.

#### Screenshot

BEFORE
https://github.com/user-attachments/assets/b7f8a883-8893-4075-abe0-f4daed1c0e45

AFTER
https://github.com/user-attachments/assets/201e584a-5d4b-486c-a293-3014b6e82e7c

### DIFF
--- a/packages/core/src/components/toast/_toast.scss
+++ b/packages/core/src/components/toast/_toast.scss
@@ -203,8 +203,6 @@ $toast-intent-colors: (
   /* stylelint-disable-next-line declaration-no-important */
   display: flex !important;
   flex-direction: column;
-  // Remove gap to prevent immediate spacing changes
-  // gap: $toast-margin;
   left: 0;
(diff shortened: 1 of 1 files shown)


packages/core/src/components/toast/_toast.scss
```diff
@@ -138,6 +139,10 @@ $toast-intent-colors: (
   /* stylelint-disable-next-line declaration-no-important */
   position: relative !important;
 
+  &:last-child {
+    margin-bottom: 0;
+  }
```
sbsto, line 144: No need for margin on the bottom I think?

sbsto:
while this fixed the jitter, it introduced a different issue

PR #7572, opened 2025-09-17
[chore] Move Popover props to separate module and eliminate circular imports

ggdouglas:
This PR refactors the Popover component structure to resolve circular import dependencies and improve code organization. This is a pure refactoring change with no functional modifications to the component behavior.

### Changes

- Move `PopoverProps` interface and `PopoverInteractionKind` enum from the main Popover component into a separate `popoverProps.ts` module
- Replace direct Tooltip import in Popover to eliminate an import cycle
- Additionally clean up other unused ESLint directives

### DIFF
--- a/packages/core/src/common/alignment.ts
+++ b/packages/core/src/common/alignment.ts
@@ -35,5 +35,5 @@ export const TextAlignment = {
     END: "end" as const,
     START: "start" as const,
 };
-// eslint-disable-next-line @typescript-eslint/no-redeclare
+
 export type TextAlignment = (typeof TextAlignment)[keyof typeof TextAlignment];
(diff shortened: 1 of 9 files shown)


packages/core/src/common/alignment.ts
```diff
@@ -35,5 +35,5 @@ export const TextAlignment = {
     END: "end" as const,
     START: "start" as const,
 };
-// eslint-disable-next-line @typescript-eslint/no-redeclare
```
ggdouglas, line 38: Not strictly related to Popover/Tooltip, noticed this while running lint locally

invliD (approved):

Issue #7139, opened 2025-01-02
Deprecation of React-Popper


bppdddqqqq:
#### Environment

Not essential

#### Feature request

As of 4.12.2024, the `React-Popper` is deprecated, with no direct recommendation for replacement, (but it hints at using floating-ui). Due to the deprecation this blocks `@blueprintjs/core` from fully supporting React 19, since library is primarily used in `Popover` component.

Source: https://github.com/floating-ui/react-popper

Together with breaking changes in React 19 regarding Refs, I am confident to state that the behavior of React-Popper is unstable on React 19. This can be further validated by testing the `Select` component, which fails to detect proper bounding boxes and renders the dropdown menu outside the component and rather on the corner of the Window.

yanndinendal:
Maybe the README could link to this migration guide: https://floating-ui.com/docs/migration

Issue #7115, opened 2024-12-08
Add support for React 19


cacabo:
#### Environment

- __Package version(s)__: `"@blueprintjs/core": "^5.11.0"`
- __Browser and OS versions__: n/a

#### Feature request

I want to leverage React 19 in my app which uses Blueprint in order to benefit from the React compiler, other performance improvements, and the new hooks.

Blueprint currently requires:

```text
npm error peer react@"^16.8 || 17 || 18" from @blueprintjs/core@5.13.1
```

I can use the beta compiler runtime until Blueprint can upgrade:

```jsonc
// package.json
{
    // ...
    "dependencies": {
        "react": "^18.3.1",
        "react-compiler-runtime": "^19.0.0-beta-63b359f-20241101",
        // ...
    }
}
```

bofa:
I would love react 19 support to be able to leverage the improved custom elements integration.

smeng9:
Currently there are still a lot of use cases of `ReactDOM.render` in the codebase which react 19 dropped support of. All of these need to be migrated to use `createRoot`

bradchristensen:
> Currently there are still a lot of use cases of `ReactDOM.render` in the codebase which react 19 dropped support of. All of these need to be migrated to use `createRoot`

I just had a quick scan through the repository and it looks like all the references to `ReactDOM.render` are either in tests/documentation, or they can be replaced on your app side if you provide a `domRenderer` prop to the relevant component. So I don't think this is a blocker, but it would be nice if Blueprint could either update to use the updated `createRoot` method by default, or provide an automatic backwards-compatible implementation.

bppdddqqqq:
@bradchristensen I don't think so, see #7139

bradchristensen:
> @bradchristensen I don't think so, see #7139

Oh of course - I hadn't considered dependencies, sorry. Good catch!

bppdddqqqq:
> > @bradchristensen I don't think so, see #7139
> 
> Oh of course - I hadn't considered dependencies, sorry. Good catch!

[...]

PR #7573, opened 2025-09-18
feat: Add PopoverNext component built on Floating UI

ggdouglas:
### Summary

Introduce `PopoverNext` as a modern, Floating UI–based replacement for `Popover`. The goal is near drop-in API compatibility to ease adoption, while migrating from Popper ["modifiers"](https://popper.js.org/docs/v2/modifiers) to Floating UI ["middleware"](https://floating-ui.com/docs/middleware). This change also helps to unblock React 19 support as part of #7115.

- **Why**: Popper v2 via `react-popper` isn’t actively maintained, and the ecosystem has moved to Floating UI. React 19 support is blocked by the legacy stack.
- **What**: New `PopoverNext` component, docs, tests, and examples. Existing `Popover` component marked deprecated.
- **How**:
  - Ported the existing Popover from a class-based to a functional component, preserving existing behavior where possible.
  - Maintained feature parity while switching the positioning engine to Floating UI.
  - Exposed `middleware` as options instead of `modifiers` in new `PopoverNext` API.

Screenshot 2025-09-18 at 13 42 19@2x

### Migration
- **Prop compatibility**
  - Preserved where feasible: `children`, `renderTarget`, `content`, `isOpen`/`defaultIsOpen`, `onInteraction`, `onClose`, `usePortal`, `hasBackdrop`, `interactionKind`, `minimal`, `matchTargetWidth`, etc.
- **Auto behavior**
  - The `"auto"` placement options are not supported in `PopoverNext`.
  - Instead, for automatic placement behavior, omit `placement` entirely. When `placement` is undefined, `PopoverNext` uses Floating UI's automatic placement by default (via `autoPlacement`). See the docs: [Floating UI autoPlacement](https://floating-ui.com/docs/autoPlacement).
- **Defaults and behavior notes**
  - Boundary defaults differ:
    - `Popover`: `boundary="clippingParents"`
    - `PopoverNext`: `boundary="clippingAncestors"`
    - To match legacy behavior precisely, set `boundary="clippingParents"` on `PopoverNext`.
  - Dismiss and accessibility are preserved:
    - `Classes.POPOVER_DISMISS`, `Classes.POPOVER_DISMISS_OVERRIDE`, `captureDismiss`
    - `aria-haspopup`/`aria-expanded`, backdrop + focus behavior via `Overlay2`
- **Modifiers → Middleware mapping**
  - `flip` → `flip`
  - `preventOverflow` → `shift` (use `boundary`/`rootBoundary`)
  - `offset` → `offset` (`mainAxis`, `crossAxis`)
  - `arrow` → `arrow`
- **Practical migration steps**
  - Switch imports to `PopoverNext`.
  - Use the same props; modify `placement`, `boundary`, etc. where needed
  - Translate any Popper modifiers to Floating UI middleware (migration utilities forthcoming)

### Follow-ups
- Migrate internal usages (`Tooltip`, menus, selects) to `PopoverNext`.
- Provide codemods and a migration wiki page as adoption ramps.
- Remove legacy `Popover` in a future major version

### References
- Origin issue: #7115
- #7139
- Popper v2: [docs](https://popper.js.org/docs/v2), [react-popper](https://popper.js.org/react-popper)
- Floating UI: [site](https://floating-ui.com), [migration guide](https://floating-ui.com/docs/migration#migration)


packages/core/src/components/popover-next/popoverTarget.tsx
```diff
@@ -0,0 +1,152 @@
...
+        const floatingProps = floatingData.getReferenceProps();
+
+        target = renderTarget({
+            ...ownTargetProps,
+            ...childTargetProps,
+            // Apply Floating UI's interaction props for renderTarget case
+            ...floatingProps,
+            className: classNames(ownTargetProps.className, targetModifierClasses),
+            // if the consumer renders a tooltip target, it's their responsibility to disable that tooltip
+            // when *this* popover is open
```
invliD: This seems like an odd place for this comment.

packages/core/src/components/popover-next/popoverTarget.tsx
```diff
@@ -0,0 +1,152 @@
...
+        const childTarget = Utils.ensureElement(Children.toArray(children)[0]);
+
+        if (childTarget === undefined) {
+            return null;
+        }
+
+        const clonedTarget = cloneElement(childTarget, {
+            ...childTargetProps,
+            className: classNames(childTarget.props.className, targetModifierClasses),
+            disabled: isOpen && isTooltipElement(childTarget) ? true : childTarget.props.disabled,
```
invliD: Let's just use a regular boolean expression instead of a ternary.
```suggestion
            disabled: (isOpen && isTooltipElement(childTarget)) || childTarget.props.disabled,
```

packages/core/src/components/popover-next/popoverTarget.tsx
```diff
@@ -0,0 +1,152 @@
...
+import { Children, cloneElement, createElement, forwardRef } from "react";
+
+import { Classes, DISPLAYNAME_PREFIX, mergeRefs, Utils } from "../../common";
+import { PopoverInteractionKind } from "../popover/popoverProps";
+import type { PopoverClickTargetHandlers, PopoverHoverTargetHandlers } from "../popover/popoverSharedProps";
+
+import type { PopoverNextProps } from "./popoverNextProps";
+import { type usePopover } from "./usePopover";
+
+interface PopoverTargetProps extends PopoverNextProps {
```
invliD: Might be cleaner to define the exact props this component takes instead of taking a superset of the entire popover.

packages/core/src/components/popover-next/popoverTarget.tsx
```diff
@@ -0,0 +1,152 @@
...
+        className,
+        floatingData,
+        disabled = false,
+        fill = false,
+        handleMouseEnter,
+        handleMouseLeave,
+        handleTargetBlur,
+        handleTargetContextMenu,
+        handleTargetFocus,
+        interactionKind = PopoverInteractionKind.CLICK,
```
invliD: Can we do this default on the popover level instead of the target?

packages/core/src/components/popover-next/popover-next.md
```diff
@@ -0,0 +1,369 @@
...
+in different DOM layout depending on your application's needs:
+
+-   The simplest way to specify a target is via `children`. Provide a single React child to
+    `<PopoverNext>` and the component will render that child wrapped in a `@ns-popover-target` HTML element.
+    This wrapper is configured with event handling logic necessary for the PopoverNext to function. Its tag name
+    (e.g. `div`, `span`) and props can be customized with the `targetTagName` and `targetProps` props, respectively.
+
+-   A more advanced API is available through the `renderTarget` prop. Here, PopoverNext provides you with all the
+    information necessary to render a functional popover with a [render prop](https://reactjs.org/docs/render-props.html).
+    You are responsible for then propogating that information with an
```
invliD:
```suggestion
    You are responsible for then propagating that information with an
```

packages/core/src/components/popover-next/popover-next.md
```diff
@@ -0,0 +1,369 @@
...
+    [object spread](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax#spread_in_object_literals)
+    to the `React.JSX.Element` returned from `renderTarget`.
```
invliD: This sentence is quite confusing. You're responsible for spreading information into a JSX.Element? You should spread it into the props, not the resulting React object. This also sounds like you _call_ renderTarget, and then you spread the data to the returned object, but you actually return the object after spreading the props.

packages/core/src/components/popover-next/popover-next.md
```diff
@@ -0,0 +1,369 @@
...
+
+    -   If the rendered element is _not_ a native HTML element, you must ensure that it supports the
+        `className`, `ref`, and `tabIndex` props (i.e. renders them out to the DOM).
```
invliD: "into" maybe? Never heard "render out" before…
```suggestion
        `className`, `ref`, and `tabIndex` props (i.e. renders them into the DOM).
```

packages/core/src/components/popover-next/popover-next.md
```diff
@@ -0,0 +1,369 @@
...
+                <Button {...targetProps} intent="primary" text="PopoverNext target" />
+            )}
+        />
+    );
+};
+```
+
+@### Placement
+
+The `placement` prop controls the popover's position relative to the target. PopoverNext passes this prop directly to Floating UI; it uses the same semantics and supported values [as shown here in the docs](https://floating-ui.com/docs/computePosition#placement).
```
invliD: I think we should abstract this away. We can still refer to the Floating UI docs, but we should repeat the relevant parts (assuming the below is not complete) and remove "PopoverNext passes this prop directly to Floating UI"—that's implementation detail and may change in the future if they break API and we don't want to.

packages/core/src/components/popover-next/popover-next.md
```diff
@@ -0,0 +1,369 @@
...
+@### Middleware
+
+Middleware allow us to customize Floating UI's positioning behavior. **PopoverNext** configures several of Floating UI's built-in middleware to handle things such as flipping, preventing overflow from a boundary element, and positioning the arrow.
+
+You may override the default middleware with the `middleware` prop, which is an object with key-value pairs representing the middleware name and its options object, respectively. See the [Floating UI middleware docs page](https://floating-ui.com/docs/middleware) for more info.
+
+<div class="@ns-callout @ns-intent-warning @ns-icon-warning-sign @ns-callout-has-body-content">
+    <h5 class="@ns-heading">Auto placement requires autoPlacement or flip middleware</h5>
+
+Be careful when disabling the "autoPlacement" or "flip" middleware, since automatic placement relies on them. If you _do_ decide to disable these middleware, be sure to also specify a placement which is specific (not automatic).
```
invliD, line 175: Do we need to support this API? It might be best to keep the simplest possible API for now until we know we need more, and even then maybe make it more semantic than just exposing all of Floating UI's API. That way, if we ever have to switch underlying libraries again, we might be able to get away without an API break next time.

packages/core/src/components/popover-next/popover-next.md
```diff
@@ -0,0 +1,369 @@
...
+    overflow-y: auto;
+}
+```
+
+@reactExample PopoverNextSizingExample
+
+@### Minimal style
+
+You can create a minimal popover by setting `minimal={true}`.
+This removes the arrow from the popover and makes the transitions more subtle.
```
invliD: I believe we've talked about this one before, but do we want to split these two into separate props now that we're redoing the API anyway?

packages/core/src/components/popover-next/popoverNextProps.ts
```diff
@@ -0,0 +1,72 @@
...
+    InlineOptions,
+    OffsetOptions,
+    ShiftOptions,
+    SizeOptions,
+} from "@floating-ui/react";
+
+import type { PopoverProps } from "../popover/popoverProps";
+import type { DefaultPopoverTargetHTMLProps } from "../popover/popoverSharedProps";
+
+export type { FloatingBoundary, FloatingPlacement };
```
invliD: We should probably export our own types that correspond to these. That would avoid accidentally propagating a breaking change if floating UI ever made one, and allows us to abstract away the use of the library at least in our API.

packages/core/src/components/popover-next/popoverNextProps.ts
```diff
@@ -0,0 +1,72 @@
...
+    size: Partial<SizeOptions>;
+}>;
+
+/**
+ * Props interface for PopoverNext component using Floating UI types.
+ * This extends the original PopoverProps but replaces Popper-specific types with Floating UI equivalents.
+ */
+export interface PopoverNextProps<T extends DefaultPopoverTargetHTMLProps = DefaultPopoverTargetHTMLProps>
+    extends Omit<PopoverProps<T>, "boundary" | "modifiers" | "placement" | "rootBoundary"> {
+    boundary?: FloatingBoundary;
```
invliD: Since this is public API we should probably redefine the actual types here. It'll make it easier to eventually delete Popover in favor of PopoverNext, and also makes this more readable.

packages/core/src/components/popover-next/popoverPopup.tsx
```diff
@@ -0,0 +1,169 @@
...
+                    )}
+                    <div className={Classes.POPOVER_CONTENT}>{content}</div>
+                </div>
+            </div>
+        </Overlay2>
+    );
+}
+
+function cssPropertyToString(value: string | number | undefined): string {
+    return value != null && !isNaN(Number(value)) ? `${value}px` : "";
```
invliD: What happens if you set `left: ` (the empty string case)? Should we set the property at all in that case?

packages/core/src/components/popover-next/floatingUtils.ts
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    // Positioning middleware (autoPlacement and flip are mutually exclusive)
+    // autoPlacement uses "most space" strategy, flip uses "fallback when no space" strategy
+    if (overrides?.autoPlacement && overrides?.flip) {
+        console.warn("autoPlacement and flip are mutually exclusive. Using autoPlacement and ignoring flip.");
+        middlewares.push(autoPlacement(overrides.autoPlacement));
+    } else if (overrides?.autoPlacement) {
+        middlewares.push(autoPlacement(overrides.autoPlacement));
+    } else if (overrides?.flip) {
```
mm-wang: **nit/question**: I can tell that this is meant to be very explicit, but I was wondering if there's a reason why this isn't defaulting to pushing `autoPlacement(overrides.autoPlacement))` in an `else` block:
```
} else if (overrides?.flip) {
        middlewares.push(flip(overrides.flip));
} else {
        middlewares.push(autoPlacement(overrides.autoPlacement);
}
```
Is it because you might not have the overrides for placement at all? Mostly asking for my own understanding

packages/core/src/components/popover-next/floatingUtils.ts
```diff
@@ -0,0 +1,67 @@
...
+ * Converts a middleware configuration object to a properly ordered array of Floating UI middleware.
+ *
+ * @param overrides - Configuration object containing middleware options
+ * @returns Array of Floating UI middleware instances in the correct order
+ *
+ * @see https://floating-ui.com/docs/middleware#ordering
+ */
+export function convertMiddlewareConfigToArray(overrides: MiddlewareConfig | undefined): Middleware[] {
+    const middlewares: Middleware[] = [];
+
```
mm-wang, line 19: **nit/question**: Are there any other overrides than the ones below? If not, could you return early here? That might save checking if overrides exist below, but I'm asking more for my understanding.

packages/core/src/components/popover-next/popoverNext.tsx
```diff
@@ -0,0 +1,423 @@
...
+import { PopoverTarget } from "./popoverTarget";
+import { usePopover } from "./usePopover";
+
+export interface PopoverNextRef {
+    reposition(): void;
+}
+
+export const PopoverNext = forwardRef<PopoverNextRef, PopoverNextProps>((props, ref) => {
+    const {
+        boundary = "clippingAncestors",
```
mm-wang, line 27: **note**: Thought it was odd that Floating UI didn't have an enum for this, but it looks like it's literally the string: https://floating-ui.com/docs/detectoverflow#boundary.

packages/core/src/components/popover-next/popoverNext.tsx
```diff
@@ -0,0 +1,423 @@
...
+    ]);
+
+    const [hasDarkParent, setHasDarkParent] = useState(false);
+    const [isClosingViaEscapeKeypress, setIsClosingViaEscapeKeypress] = useState(false);
+
+    const arrowRef = useRef(null);
+    const cancelOpenTimeout = useRef<(() => void) | undefined>(undefined);
+    const isMouseInTargetOrPopover = useRef(false);
+    const lostFocusOnSamePage = useRef(true);
+    const targetRef = createRef<HTMLElement>();
```
mm-wang: **question**: Why is this using `createRef` vs `useRef`?

packages/core/src/components/popover-next/popoverNext.tsx
```diff
@@ -0,0 +1,423 @@
...
+    const timeoutIds = useRef<number[]>([]);
+
+    const isArrowEnabled = !minimal || middlewareOverrides?.arrow !== undefined;
+    const isControlled = isOpen !== undefined;
+    const isContentEmpty = content == null || Utils.isEmptyString(content);
+
+    const computedIsOpen = disabled ? false : isOpen ?? defaultIsOpen;
```
mm-wang: **nit**: `const computedIsOpen = disabled ? false : (isOpen ?? defaultIsOpen)` for clarity, although I believe the latter evaluates first anyway, so really not a big deal

packages/core/src/components/popover-next/popoverNextProps.ts
mm-wang: If Floating UI made a breaking change, and we were to rely on the updated version, wouldn't we want to have that change included?

packages/core/src/components/popover-next/popoverTarget.tsx
mm-wang: +1

packages/core/src/components/popover-next/popoverTarget.tsx
ggdouglas: Hm, this is an artifact from being ported over from the original component without being recontextualized. I can try to move it somewhere more appropriate. fce1b5d261e5019223d8526d6db40ba9a1d9b03b
https://github.com/palantir/blueprint/blob/d8783f00b3ea2a637b62d6afcd9890d41a8614e3/packages/core/src/components/popover/popover.tsx#L333-L334

packages/core/src/components/popover-next/popover-next.md
ggdouglas: Hm, yeah, another one that I copied over from the original Popover docs:
https://github.com/palantir/blueprint/blob/d8783f00b3ea2a637b62d6afcd9890d41a8614e3/packages/core/src/components/popover/popover.md?plain=1#L64-L66

I'll try to come up with some better wording here. fce1b5d261e5019223d8526d6db40ba9a1d9b03b

packages/core/src/components/popover-next/popover-next.md
ggdouglas: fce1b5d261e5019223d8526d6db40ba9a1d9b03b

packages/core/src/components/popover-next/popover-next.md
ggdouglas: fce1b5d261e5019223d8526d6db40ba9a1d9b03b

packages/core/src/components/popover-next/popover-next.md
ggdouglas: I can get on board with that; first crack at breaking some of this out into our own docs explicitly: fce1b5d261e5019223d8526d6db40ba9a1d9b03b

packages/core/src/components/popover-next/popoverTarget.tsx
ggdouglas: bef744acd3ada851303151d9a9e5aebb9f76ab6c

packages/core/src/components/popover-next/popoverTarget.tsx
ggdouglas: Yep! 62ee536d382a7bd26cbd2a945df7bb1e7ee0b2c8

packages/core/src/components/popover-next/popoverTarget.tsx
ggdouglas: Added a couple changes related to limiting the scope of these props: dbf5012f7a9c831d372feb467e03ab9cd3bb6d26 00801418ed415055b4595074d12bc9066d15e96f

packages/core/src/components/popover-next/popoverNextProps.ts
ggdouglas: Turns out that `FloatingBoundary` no longer needed to be re-exported. Created our own type for Placement: 7dd3241d3238a319bde36653f2e9b091d15503d3

packages/core/src/components/popover-next/popoverNextProps.ts
ggdouglas: I can get on board with this 👍  00801418ed415055b4595074d12bc9066d15e96f

packages/core/src/components/popover-next/popoverPopup.tsx
```diff
@@ -0,0 +1,169 @@
...
+            [Classes.POPOVER_CAPTURING_DISMISS]: captureDismiss,
+            [Classes.POPOVER_MATCH_TARGET_WIDTH]: matchTargetWidth,
+            [Classes.POPOVER_REFERENCE_HIDDEN]: isReferenceHidden,
+            [Classes.POPOVER_POPPER_ESCAPED]: hasPopperEscaped,
+        },
+        `${Classes.POPOVER_CONTENT_PLACEMENT}-${basePlacement}`,
+        popoverClassName,
+    );
+
+    const defaultAutoFocus = isHoverInteractionKind ? false : undefined;
```
mm-wang, line 101: **question**: What is the difference between `false` and `undefined` here?

mm-wang (commented):
Reviewed for syntax and understanding, mostly tried to get context and an example of how we structure these components. I'd love to hear more about how I can use and test this.

packages/core/src/components/popover-next/popoverNextProps.ts
invliD: > If Floating UI made a breaking change, and we were to rely on the updated version, wouldn't we want to have that change included?

Probably not, because that would require us to major bump Blueprint, and we'd likely not want to do that at the time prescribed by Floating UI. By making our own type, we can abstract that connection away, and have a conversion function that keeps our API stable, and we can then break ours at our convenience at the next Blueprint major bump.

packages/core/src/components/popover-next/popoverNext.tsx
invliD: Same as my comment above, this should probably just be `!disabled && (isOpen ?? defaultIsOpen)`. Ternaries with booleans in both arms just make the expression more difficult to follow.

packages/core/src/components/popover-next/floatingUtils.ts
ggdouglas: Hmm, it's entirely possible that this got overcomplicated at some point in porting over the logic. Looking at where this is used in `PopoverNext`, it looks like it's always going to be provided with either `autoPlacement `or `flip` in its default config. Your suggestion makes sense to me. 06463f9ba453c4575024143003fdc4b528f87b92

packages/core/src/components/popover-next/popoverNext.tsx
ggdouglas: Hm, good point! I didn't notice this when porting over logic from `Popover` but `useRef` is more appropriate here to maintain a stable ref. Changed in 290909173015874702c4b52666fbcffdd640de1b

packages/core/src/components/popover-next/popoverNext.tsx
ggdouglas: Hm, I can agree on the point of clarity, but it seems that Prettier ends up stripping these parens out regardless.

packages/core/src/components/popover-next/floatingUtils.ts
ggdouglas: Are you asking in regards to the need for optional chaining here? Looks like we can actually remove checks that `overrides` exists since we currently only use this utility function in one place where it's always provided. 9b3be1305dfcbd0a4e8d66097eeac863a6f9b853

Not sure if that helps clarify though. Let me know if I misunderstood!

packages/core/src/components/popover-next/popoverNext.tsx
invliD: That's fine, I'm not too worried about the parens, just about a ternary with more boolean expressions in the two arms.

packages/core/src/components/popover-next/popoverPopup.tsx
ggdouglas: This logic was originally defined in the original `Popover` component here:
https://github.com/palantir/blueprint/blob/d8783f00b3ea2a637b62d6afcd9890d41a8614e3/packages/core/src/components/popover/popover.tsx#L415

My understanding of this is that for hover interactions, we want to explicitly disable auto-focus (`false`) because we don't want hover to steal focus from the user's current input. For click interactions, we want to use the default behavior which likely enables auto-focus when appropriate.

packages/core/src/components/popover-next/popover-next.md
ggdouglas: Yeah, there are a few things to split out here:
- Presence of the arrow
- Offset of the popup element from the target
- Animation style

I've split `minimal` into two separate props, `arrow` and `animation`. The `arrow` prop controls the display of the arrow and applies the correct amount of offset to accommodate the arrow. The `animation` prop solely controls the style of the animation ("scale" vs. "minimal"). 86cd4c0b15ec6e6950d3d98a5b63e01e7672f7ac

packages/core/src/components/popover-next/popover-next.md
ggdouglas: I get the concern about keeping the API simple and future-proof, but I'm thinking we should probably keep the middleware prop for now.

The main thing is that we're going to need this flexibility pretty much immediately. Looking at the current codebase, submenus and tooltips already rely on custom modifiers for things like `offset`, `flip`, and `preventOverflow`.

https://github.com/palantir/blueprint/blob/031a998e066f234067e46ca37d4e5ad192c51b42/packages/core/src/components/menu/menuItem.tsx#L308-L314

https://github.com/palantir/blueprint/blob/031a998e066f234067e46ca37d4e5ad192c51b42/packages/core/src/components/tooltip/tooltip.tsx#L126-L135

Without exposing middleware, we'd either have to recreate all these specific APIs (which feels like more work for less flexibility) or continually add on to functionality during the migration.

I also think the middleware concept is actually pretty well-designed already. Floating UI's API is well-documented, and it's tough to anticipate all the positioning edge cases we might need to handle. A more semantic API would probably just end up being a thin wrapper around middleware anyway, and then we'd be maintaining our own abstraction layer.

On the future-proofing front, if we do need to switch libraries again someday, the middleware concept seems pretty universal. Popper had modifiers, Floating UI has middleware, so there's probably a pattern here. We could potentially create a compatibility layer that maps between different systems if needed.

What's your take? Are there specific middleware use cases that concern you, or do you have a suggestion on a clean wrapper we could use here?

packages/core/src/components/popover-next/popover-next.md
```diff
@@ -0,0 +1,400 @@
...
+
+<div class="@ns-callout @ns-intent-warning @ns-icon-warning-sign @ns-callout-has-body-content">
+    <h5 class="@ns-heading">
+
+[OverlaysProvider](#core/context/overlays-provider) recommended
+
+</h5>
+
+This component renders an **Overlay2** which works best inside a React tree which includes an
+**OverlaysProvider**. Blueprint v5.x includes a backwards-compatibile shim which allows this context
```
mm-wang, line 30: **question**: Why do we need Overlay2? Is it for interactions with other overlays?

packages/core/src/components/popover-next/usePopover.ts
```diff
@@ -0,0 +1,99 @@
...
+
+            // Always call the external callback if provided
+            if (onOpenChange) {
+                onOpenChange(nextOpen, event);
+            }
+        },
+        [onOpenChange, isControlled],
+    );
+
+    const data = useFloating({
```
mm-wang, line 73: **question**: Did we want to expose the floating strategy? https://floating-ui.com/docs/useFloating#strategy

packages/core/src/components/popover-next/popoverNext.tsx
```diff
@@ -0,0 +1,415 @@
...
+        matchTargetWidth = false,
+        middleware: middlewareOverrides,
+        onClose,
+        onInteraction,
+        openOnTargetFocus = true,
+        placement,
+        position,
+        renderTarget,
+        rootBoundary,
+        shouldReturnFocusOnClose = false,
```
mm-wang: **suggestion**: I thought this was expected to be `true` to satisfy a11y requirements?
```suggestion
        shouldReturnFocusOnClose = true,
```

packages/core/src/components/popover-next/popoverNext.tsx
```diff
@@ -0,0 +1,415 @@
...
+        isOpen,
+        matchTargetWidth = false,
+        middleware: middlewareOverrides,
+        onClose,
+        onInteraction,
+        openOnTargetFocus = true,
+        placement,
+        position,
+        renderTarget,
+        rootBoundary,
```
mm-wang, line 46: **question**: Is this necessary, given that we have `boundary`? Could we instead extend boundary's capabilities? I would expect to try to minimize this API as opposed to expose everything.

packages/core/src/components/popover-next/popoverNext.tsx
```diff
@@ -0,0 +1,415 @@
...
+        isOpen,
+        onInteraction,
+        placement,
+        position,
+        renderTarget,
+        targetProps,
+        usePortal,
+    ]);
+
+    const [hasDarkParent, setHasDarkParent] = useState(false);
```
mm-wang, line 92: **question**: Can we get rid of this? I'm not seeing usages unless I'm misreading this...

packages/core/src/components/popover-next/popoverNext.tsx
```diff
@@ -0,0 +1,415 @@
...
+        hoverOpenDelay = 150,
+        interactionKind = PopoverInteractionKind.CLICK,
+        isOpen,
+        matchTargetWidth = false,
+        middleware: middlewareOverrides,
+        onClose,
+        onInteraction,
+        openOnTargetFocus = true,
+        placement,
+        position,
```
mm-wang: **question**: What is the difference between `position` and `placement`? I thought position was determined by placement.

packages/core/src/components/popover-next/popoverNext.tsx
```diff
@@ -0,0 +1,415 @@
...
+    });
+
+    useImperativeHandle(
+        ref,
+        () => ({
+            reposition: () => {
+                floatingData.update();
+            },
+        }),
+        [floatingData],
```
mm-wang, line 156: **suggestion**: Limit dependency to `floatingData.update`?
```suggestion
        [floatingData.update],
```

packages/core/src/components/popover-next/popoverNext.tsx
```diff
@@ -0,0 +1,415 @@
...
+                      },
+                  }
+                : undefined,
+        };
+        const mergedMiddleware: MiddlewareConfig = { ...defaultMiddleware, ...middlewareOverrides };
+        return convertMiddlewareConfigToArray(mergedMiddleware);
+    }, [placement, boundary, rootBoundary, arrow, arrowRef, matchTargetWidth, middlewareOverrides]);
+
+    const floatingData = usePopover({
+        canEscapeKeyClose,
```
mm-wang, line 134: **question**: Should `positioningStrategy` be set here?

packages/core/src/components/popover-next/popoverNext.tsx
```diff
@@ -0,0 +1,415 @@
...
+            ) {
+                setOpenState(false, event);
+            }
+        },
+        [setOpenState, targetRef],
+    );
+
+    const updateDarkParent = useCallback(() => {
+        if (usePortal && floatingData.isOpen) {
+            setHasDarkParent(targetRef.current?.closest(`.${Classes.DARK}`) != null);
```
mm-wang, line 348: **question**: How are we planning to handle when we have the data attribute? This is assuming we still need `hasDarkParent`

packages/core/src/components/popover-next/popoverNextProps.ts
```diff
@@ -0,0 +1,434 @@
...
+    /**
+     * When enabled, clicks inside a `Classes.POPOVER_DISMISS` element
+     * will only close the current popover and not outer popovers.
+     * When disabled, the current popover and any ancestor popovers will be closed.
+     *
+     * @default false
+     */
+    captureDismiss?: boolean;
+
+    /** Whether the popover can be closed by pressing the Escape key. */
```
mm-wang, line 74: **question**: Should we even allow this prop? Shouldn't we always expect that a user can use `Esc` to close a popover?

packages/core/src/components/popover-next/popoverNextProps.ts
```diff
@@ -0,0 +1,434 @@
...
+     */
+    hoverOpenDelay?: number;
+
+    /**
+     * Whether a popover that uses a `Portal` should automatically inherit the
+     * dark theme from its parent.
+     *
+     * @default true
+     */
+    inheritDarkTheme?: boolean;
```
mm-wang, line 141: **suggestion**: This should not be a component prop, IMO. I think this should be done at the CSS level.

packages/core/src/components/popover-next/popoverNextProps.ts
```diff
@@ -0,0 +1,434 @@
...
+        // Popover<TProps, "click" | "hover">. Instead of discriminating, we union the different possible event handlers
+        // that may be passed (they are all optional properties anyway).
+        props: PopoverRenderTargetProps & PopoverHoverTargetHandlers<T> & PopoverClickTargetHandlers<T>,
+    ) => React.JSX.Element;
+
+    /**
+     * A root boundary element supplied to the positioning middleware.
+     * This is a shorthand for overriding Floating UI middleware options with the `middleware` prop.
+     */
+    rootBoundary?: PopoverNextRootBoundary;
```
mm-wang, line 287: **suggestion**: I don't know that I'd expose this.

packages/core/src/components/popover-next/popoverNextProps.ts
```diff
@@ -0,0 +1,434 @@
...
+
+    /**
+     * Whether the application should return focus to the last active element in the
+     * document after this popover closes.
+     *
+     * This is automatically set (overridden) to:
+     *  - `false` for hover interaction popovers
```
mm-wang: **question**: Why would this be false for tooltips? I understand that it should be true always: https://www.w3.org/WAI/ARIA/apg/patterns/tooltip/#:~:text=A%20tooltip%20is%20a%20popup%20that%20displays,dismissed%20when%20it%20no%20longer%20has%20focus

packages/core/src/components/popover-next/popoverPopup.tsx
```diff
@@ -0,0 +1,162 @@
...
+
+    const shouldReturnFocusOnClose = isHoverInteractionKind
```
mm-wang, line 107: Noted my thoughts above.

packages/core/src/components/popover-next/popoverTarget.tsx
```diff
@@ -0,0 +1,140 @@
+/* !
+ * (c) Copyright 2025 Palantir Technologies Inc. All rights reserved.
```
mm-wang: **nit**: All 2026 now (sorry!)

mm-wang (commented):
This is my first pass at review. I haven't taken a look at tests, examples, or styling. I also have only done a cursory review of popup + target, and probably still need to re-review. But here are some comments thus far.

I'd like to try to minimize the API boundary if possible - the current component has a pretty large surface area and this is increasing it with some added props. I'd love to know whether we can make choices for the user about what is actually useful to expose.

I also wanted to check - is there a way that we're detecting simulated clicks, similar to the old version?
```
isSimulatedButtonClick = (e: React.MouseEvent<HTMLElement> | React.KeyboardEvent<HTMLElement>) => {
        return !e.isTrusted && (e.target as HTMLElement).matches(`.${Classes.BUTTON}`);
    };
```

mm-wang:
I also want to check on a couple of things:
* Can we add the delays to open + close to the examples? That'd be good to note for a "kitchen sink" model
* Can we add a context menu opening another context menu to check that behavior with focus?

packages/core/src/components/popover-next/popoverNextProps.ts
mm-wang: Looks like we already do so 🤷

packages/core/src/components/popover-next/popoverNextProps.ts
mm-wang: In the spirit of trying for pure parity with `Popover`, fine to ignore this. Longer term would prefer to separate styling from the component API.

packages/core/src/components/popover-next/popoverNext.tsx
mm-wang: Same thoughts as https://github.com/palantir/blueprint/pull/7573#discussion_r2747568860

packages/core/src/components/popover-next/popoverNext.tsx
mm-wang: Sounds like we're already exposing this right now, so this doesn't help optimize.

packages/core/src/components/popover-next/popover-next.md
mm-wang: Interested to see porting this over in another implementation of this.

packages/core/src/components/popover-next/popoverTarget.tsx
ggdouglas: Updated in cabca9ce3cebbe1cea99827d49b3d671f7bc3d3c

packages/core/src/components/popover-next/popover-next.md
ggdouglas: The primary reasons we use [Overlay2](https://blueprintjs.com/docs/#core/components/overlay2) in the original Popover component are:
- Portal functionality
- Focus capturing
- Rendering transitions
- (optionally) Rendering backdrop components

The current set of functionality with Overlay2 was preserved in PopoverNext while porting the component, but it is worth noting that Floating UI can (theoretically) supplant Overlay2 with the features it exposes within the library itself:
- https://floating-ui.com/docs/FloatingPortal
- https://floating-ui.com/docs/FloatingFocusManager
- https://floating-ui.com/docs/useTransition
- https://floating-ui.com/docs/FloatingOverlay

This was out-of-scope for the original intent behind this PR to craft PopoverNext as a drop-in replacement, but noting here for future reference.

packages/core/src/components/popover-next/usePopover.ts
ggdouglas: Ah, good callout. I somehow missed the existing `positioningStrategy` prop on Popover when porting it to PopoverNext. Added in eda60938b0d421762937313a3e03c928f0d7984e for parity.

packages/core/src/components/popover-next/popoverNext.tsx
ggdouglas: We talked a bit offline re: this, but the basic BLUF here is:

- The current Popover component exposes both `boundary` and `rootBoundary`, so we aren't exposing any _additional_ props with PopoverNext
- To clarify the different between the two:
  - [`boundary`](https://floating-ui.com/docs/detectoverflow#boundary) defines local clipping containers (like a scrollable div)
  - [`rootBoundary`](https://floating-ui.com/docs/detectoverflow#rootboundary) defines the _ultimate_ boundary (viewport or entire document)
  - FloatingUI checks overflow against both and uses the most restrictive result

packages/core/src/components/popover-next/popoverNext.tsx
ggdouglas: Seems like `react-hooks/exhaustive-deps` is not happy with that change:

Screenshot 2026-02-02 at 16 20 06@2x

packages/core/src/components/popover-next/popoverNext.tsx
ggdouglas: Added in eda60938b0d421762937313a3e03c928f0d7984e

packages/core/src/components/popover-next/popoverNext.tsx
ggdouglas: I suppose we could update this logic to include the data attribute if/when we make a switch there. I imagine the implementation would look something like:
```ts
// Current: only checks for the class
setHasDarkParent(targetRef.current?.closest(`.${Classes.DARK}`) != null);

// Future: check for class OR data attribute
setHasDarkParent(
    targetRef.current?.closest(`.${Classes.DARK}, [data-theme="dark"]`) != null
);
```

packages/core/src/components/popover-next/popoverNextProps.ts
ggdouglas: We talked a bit offline about the difficulties/nuance around `inheritDarkTheme`. Here's a summary:

The concern is that using JavaScript to control dark mode styling feels like it should be handled purely in CSS. However, there's a fundamental CSS limitation at play here: **CSS selectors can only traverse downward through the DOM tree.** When a popover is rendered into a portal (typically a sibling of the app root, not a descendant), it cannot inherit styles from `.bp6-dark` applied to an ancestor of the _trigger_ element through normal CSS cascade.

A potential CSS-only solution would be using `:root:has(.bp6-dark)` to apply dark styles globally, but this requires the dark mode class to be applied at or near the root level. Currently, Blueprint applications typically apply `.bp6-dark` to a container element wrapping their app, not at the document root.

Future solutions that could eliminate this prop:
- A `ThemeProvider` context that portaled components could consume
- Moving dark mode to a data-theme attribute on `:root` or `<body>`
- Global CSS custom properties for theming

These are larger architectural changes. For now, `inheritDarkTheme` maintains parity with the legacy Popover behavior and allows for easier migrations. We can revisit this as part of a broader theming initiative.

packages/core/src/components/popover-next/popoverNextProps.ts
ggdouglas: See https://github.com/palantir/blueprint/pull/7573#discussion_r2756152951

packages/core/src/components/popover-next/popoverNextProps.ts
ggdouglas: I'm not sure I have enough insight/context into why this was originally made an option on the original Popover component other than to say that it is originally exposed through `OverlayableProps` and is thus a feature of components that use Overlay in Blueprint: https://github.com/palantir/blueprint/blob/90ce7125b971fef9926149c9f86ab16366db00ce/packages/core/src/components/overlay/overlayProps.ts#L27-L32

Considering that this component is the base of many other, more specific components (e.g. Dialog, Tooltip, Select, etc.), I can see the value in having this be optional since there are contexts where you may/may not want to have the widget be closeable on Esc depending on the context. For that reason, I'm in favor of leaving this intact for now.

packages/core/src/components/popover-next/popoverNext.tsx
ggdouglas: Hmm, good question. I originally kept both props here for component API parity, but I now ultimately think we can be a bit more opinionated and port just one for simplicity in the new component.

Here are the docs definitions:

https://github.com/palantir/blueprint/blob/90ce7125b971fef9926149c9f86ab16366db00ce/packages/core/src/components/popover/popoverSharedProps.ts#L256-L267

https://github.com/palantir/blueprint/blob/90ce7125b971fef9926149c9f86ab16366db00ce/packages/core/src/components/popover/popoverSharedProps.ts#L274-L284

Namely calling out this note on `placement`:
> Mutually exclusive with `position` prop. Prefer using this over `position`, as it more closely aligns with Popper.js semantics.

The rendered types look like the following:

For **`placement`**:
- `"top"`, `"top-start"`, `"top-end"`
- `"bottom"`, `"bottom-start"`, `"bottom-end"`
- `"left"`, `"left-start"`, `"left-end"`
- `"right"`, `"right-start"`, `"right-end"`
- `"auto"`, `"auto-start"`, `"auto-end"`

For **`position`**:
- `"top"`, `"top-left"`, `"top-right"`
- `"bottom"`, `"bottom-left"`, `"bottom-right"`
- `"left"`, `"left-top"`, `"left-bottom"`
- `"right"`, `"right-top"`, `"right-bottom"`
- `"auto"`, `"auto-start"`, `"auto-end"`

So `placement` is more of a [Popper.js](https://popper.js.org/docs/v2/constructors/#options) and [Floating UI](https://floating-ui.com/docs/computePosition#placement) concept that uses [logical operators](https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Logical_properties_and_values) for each suffix of the side, whereas `position` is more of a Blueprint-derived concept.

Personally, I think we should opt to roll forward with just `placement` considering that it both:

1. Has better parity with Floating UI types
2. Aligns more generally with our movement towards properties with logical equivalents (see https://github.com/palantir/blueprint/issues/7193)

ggdouglas:
> I also want to check on a couple of things:
> 
> * Can we add the delays to open + close to the examples? That'd be good to note for a "kitchen sink" model
> * Can we add a context menu opening another context menu to check that behavior with focus?

@mm-wang RE: these:

> Can we add the delays to open + close to the examples? That'd be good to note for a "kitchen sink" model

I attempted this locally, but I'm a bit wary of adding more examples to the "kitchen sink" here since it's getting to the point where there are just too many options / it's a bit overwhelming for the docs. It's also only really relevant when the interaction mode is a hover-type, so that makes these controls irrelevant otherwise.

Screenshot 2026-02-02 at 17 57 57

I think I want to defer on adding these, they seem like useful controls, but would maybe be better suited for inclusion in something like Storybook.

> Can we add a context menu opening another context menu to check that behavior with focus?

I'm a bit confused. Can you elaborate on where this should be added?

ContextMenu currently uses the legacy Popover, not PopoverNext. Adding that example would be useful for testing ContextMenu behavior in general, but it's out of scope for this PR. Happy to add it as a follow-up if you'd like, or we can revisit when/if ContextMenu is migrated to use PopoverNext.

packages/core/src/components/popover-next/popoverNext.tsx
ggdouglas: Removed `position` in 229f3df124068c631b17f5849bbe90dd9c9d80dc

packages/core/src/components/popover-next/popoverNextProps.ts
mm-wang: Hmm... all right. I think if we default to true, that is fine, but I do find it odd to not allow it.

packages/core/src/components/popover-next/popoverTarget.tsx
```diff
@@ -0,0 +1,140 @@
...
+        // If, instead, renderTarget is undefined and the target is provided as a child, props.className is
+        // applied to the generated target wrapper element.
+        className: classNames(className, Classes.POPOVER_TARGET, {
+            [Classes.POPOVER_OPEN]: isOpen,
+            // this class is mainly useful for button targets
+            [Classes.ACTIVE]: isOpen && !isControlled && !isHoverInteractionKind,
+        }),
+        ref,
+        ...targetEventHandlers,
+    } satisfies React.HTMLProps<HTMLElement>;
```
mm-wang, line 66: **note**: Interesting, I haven't seen `satisfies` used. TIL

packages/core/src/components/popover-next/popoverPopup.tsx
```diff
@@ -0,0 +1,164 @@
...
+    // Override shouldReturnFocusOnClose based on interaction type:
+    // - Hover interactions: always false (focus should stay on trigger per WCAG)
+    // - ESC keypress: always true (keyboard users need focus returned)
+    // - Otherwise: use prop value (defaults to true)
+    const shouldReturnFocusOnClose = isHoverInteractionKind
+        ? false
+        : isClosingViaEscapeKeypress || props.shouldReturnFocusOnClose;
+
+    return (
+        <Overlay2
```
mm-wang, line 112: **non-blocking comment**: Would very much love to see the future where we don't use Overlay at all here to minimize this API!

packages/core/src/components/popover-next/popoverTarget.tsx
```diff
@@ -0,0 +1,140 @@
...
+    return target;
+});
+
+PopoverTarget.displayName = `${DISPLAYNAME_PREFIX}.PopoverTarget`;
+
+/**
+ * Check if a React element is a Tooltip component by comparing its displayName.
+ * This avoids importing the Tooltip component directly, which would create a dependency cycle.
+ */
+function isTooltipElement(element: React.ReactElement): boolean {
```
mm-wang, line 132: **non-blocking comment**: This is interesting, something I'd be keen to remove if we choose to have Tooltip built differently.

packages/core/test/popover-next/popoverNextTests.tsx
```diff
@@ -0,0 +1,1466 @@
...
+            expect(screen.getByRole("button", { name: "dismiss" })).to.exist;
+
+            userEvent.click(screen.getByRole("button", { name: "dismiss" }));
+
+            await waitFor(() => expect(screen.queryByRole("button", { name: "dismiss" })).not.to.exist);
+            await waitFor(() => expect(screen.queryByRole("button", { name: "inner target" })).not.to.exist);
+        });
+    });
+
+    describe("key interactions on Button target", () => {
```
mm-wang, line 1475: **non-blocking comment**: It would be cool to see if we can set up some combination tests, such as Popover with MenuItems, although I think that they are likely better suited to an integration test.

packages/core/test/popover-next/popoverNextTests.tsx
```diff
@@ -0,0 +1,1466 @@
...
+
+            userEvent.keyboard("{Escape}");
+
+            await waitFor(() => {
+                expect(screen.queryByText("content")).not.to.exist;
+            });
+        });
+    });
+
+    describe("focus management when shouldReturnFocusOnClose={true}", () => {
```
mm-wang: **non-blocking comment**: I suspect we'll need to opposite test now that `shouldReturnFocusOnClose=
{true}` is the default

packages/core/test/popover-next/popoverNextTests.tsx
```diff
@@ -0,0 +1,1466 @@
...
+                    <Button text="target" />
+                </PopoverNext>,
+            );
+
+            expect(container.querySelector("[aria-haspopup]")).to.be.null;
+        });
+    });
+
+    describe("basic functionality", () => {
+        it.skip("inherits dark theme from trigger ancestor", () => {
```
mm-wang: **question**: Why is this skipped already?

packages/core/test/popover-next/popoverNextTests.tsx
```diff
@@ -0,0 +1,1466 @@
...
+                render(
+                    <PopoverNext content="content" interactionKind="hover" disabled={true}>
+                        <Button text="target" />
+                    </PopoverNext>,
+                );
+
+                expect(screen.getByRole("button", { name: "target" }).getAttribute("tabindex")).to.be.null;
+            });
+
+            it.skip("opens popover on target focus when interactionKind is HOVER", async () => {
```
mm-wang: **question**: Why is this skipped already (similar to others)?

packages/core/test/popover-next/popoverNextTests.tsx
```diff
@@ -0,0 +1,1466 @@
...
+                );
+                const targetButton = screen.getByRole("button", { name: "target" });
+
+                targetButton.focus();
+
+                await waitFor(() => expect(screen.getByText("content")).to.exist);
+                expect(targetButton).to.equal(document.activeElement);
+            });
+
+            it.skip("opens popover on target focus when interactionKind is HOVER_TARGET_ONLY", async () => {
```
mm-wang: **question**: Why is this skipped already (similar to others)?

packages/core/test/popover-next/popoverNextTests.tsx
```diff
@@ -0,0 +1,1466 @@
...
+            );
+
+            await waitFor(() => expect(screen.getByRole("button", { name: "close" })).to.exist);
+
+            userEvent.click(screen.getByRole("button", { name: "close" }));
+
+            await waitFor(() => expect(screen.queryByRole("button", { name: "close" })).not.to.exist);
+        });
+
+        it.skip("pressing Escape closes popover when canEscapeKeyClose=true and usePortal=false", async () => {
```
mm-wang: **question**: Why is this skipped already (similar to others)?

packages/core/test/popover-next/popoverNextTests.tsx
```diff
@@ -0,0 +1,1466 @@
...
+import sinon from "sinon";
+
+import type { PopoverInteractionKind } from "../../lib/esm/components/popover/popoverProps";
+import { Classes } from "../../src";
+import * as Errors from "../../src/common/errors";
+import { Button, PopupKind, Tooltip } from "../../src/components";
+import { PopoverNext } from "../../src/components/popover-next/popoverNext";
+import { hasClass } from "../utils";
+
+describe("<PopoverNext>", () => {
```
mm-wang, line 16: **suggestion, non-blocking**: It might be nice to test target + popup behaviors in a different file, but not necessary for this to move forward.

packages/core/test/popover-next/popoverNextTests.tsx
mm-wang: **question**: Why is this skipped already? (Similar to others)

packages/core/test/popover-next/popoverNextTests.tsx
mm-wang: **question**: Why is this skipped already? (Similar to others)

packages/core/test/popover-next/popoverNextTests.tsx
mm-wang: **question**: Why is this skipped already? (Similar to others)

packages/core/test/popover-next/popoverNextTests.tsx
mm-wang, line 16: **suggestion, non-blocking**: It might be nice to have some tests that specifically have to do with `popoverPopup` and `popoverTarget` behavior in separate files.

mm-wang (dismissed):
A saga! Looks good, I have some nonblocking comments and notes for the future. In general I think this is fine to put in and to start decomping the next components to see what changes happen.

Chatted about the use of ResizeObserver not necessary here as Floating UI should support those interactions now.

packages/core/src/components/popover-next/popoverTarget.tsx
ggdouglas: Yes, this was originally split out/added in #7572 to eliminate some of the import cycle issues between Popover and Tooltip. I agree that the coupling would be nice to eventually get rid of.

packages/core/test/popover-next/popoverNextTests.tsx
ggdouglas: Good question. Many of these were skipped for various reasons in the original Popover tests, and I'd carried them over for posterity. However, not that we're using vitest in core, it's much easier/more possible to unskip these. Done in 06774d8eb4b50e70b1b1e9a73c241c496cc4ab63

packages/core/test/popover-next/popoverNextTests.tsx
ggdouglas: Sure, should be fairly straightforward to add fabfed46388b8d21a77b1e89f06435bab374fd12

packages/core/test/popover-next/popoverNextTests.tsx
ggdouglas: These hover interaction tests in particular needed to be modified to reduce flake. There were issues with the assertions and the async behavior of the focus trapping behavior.

PR #7581, opened 2025-10-02
Fix react warning from Toaster (#7574)

### LINKED ISSUE #7574: React JSX 'key' warning from OverlayToaster
#### Environment

-   **Package version(s)**: 6.2.0+
-   **Operating System**: n/a
-   **Browser name and version**: n/a

#### Code Sandbox

Link to a minimal repro: https://codesandbox.io/p/sandbox/red-night-wyht2g

#### Steps to reproduce

1. Show any toast notification while in dev mode

#### Actual behavior

React JSX gives the following warning:

Image

`Warning: A props object containing a "key" prop is being spread into JSX`

#### Possible solution

This issue is a subtle one.  With #7500, Blueprint shifted to using react-jsx transform.  This causes some new JSX validations to run in dev-mode.  One of those validations catches when a `key` prop leaks into the component, which is happening from `OverlayToaster`.  

I believe the solution is to add an explicit `key={toastProps.key}` within the `renderToast` method.  You'll need to subtract key out from the props being spread in.  This causes the key prop to get passed in differently in the compiled JSX code.

```ts
private renderToast = (toast: ToastOptions) => {
  const { key, ...toastProps } = toast;
  return <Toast key={key} {...toastProps} onDismiss={this.getDismissHandler(toast)} />;
};
```

cc: @ggdouglas

justinbhopper:
#### Fixes #7574

Fixes a react warning that is occurring when the Toaster renders a toast.  This warning started happening from #7500, when Blueprint shifted to using react-jsx transform.  This activated new JSX validations.

#### Checklist

-   [ ] Includes tests
-   [ ] Update documentation

This separates the `key` out from the spread operator, ensuring the compiled code doesn't result in the react key passed in as part of the props object.

### DIFF
--- a/packages/core/src/components/toast/overlayToaster.tsx
+++ b/packages/core/src/components/toast/overlayToaster.tsx
@@ -295,7 +295,8 @@ export class OverlayToaster extends AbstractPureComponent<OverlayToasterProps, O
     }
 
     private renderToast = (toast: ToastOptions) => {
-        return <Toast {...toast} onDismiss={this.getDismissHandler(toast)} />;
+        const { key, ...toastProps } = toast;
+        return <Toast key={key} {...toastProps} onDismiss={this.getDismissHandler(toast)} />;
     };
 
     private createToastOptions(props: ToastProps, key = `toast-${this.toastId++}`) {


ggdouglas (approved):
Thanks for the investigation and fix. 🙂

PR #7585, opened 2025-10-08
[FR] Add ensureParentOverflowVisible prop to prevent slider label clipping

### LINKED ISSUE #6765: Slider popover in hidden overflow div
#### Environment

- __Package version(s)__: Blueprint 4 (issue exists on 5 as well)

#### Question

We have a case where we need a blueprint slider within a div. This div needs to have the overflow-x: hidden property.
The min and max values of the slider can be arbitrarily large. The width of the slider, however, needs to remain consistent regardless of the min and max values.
With very large numbers, both the axis labels and the popover containing the value of the slider overflows outside of the div it is in. (In the screenshot the min and max are -1,000,000 and 1,000,000 respectively)

Bildschirmfoto 2024-04-12 um 09 23 54

We have a workaround that moves the min and max label to within the bounds of the div, but we are unable to find a solution that works for the popover. 

Is there any built-in functionality or workaround that can be suggested to either:
- Allow the popover of the slider pop out above the overflowing div
- As soon as the popover is about to leave the bounds of the div that it sticks and expands towards the center of the div

Here is a sandbox to hopefully better illustrate the issue:
https://codesandbox.io/p/sandbox/constrained-blueprint-slider-l42kx6

aworld1:

Added a new optional `useHandleLabelPortal` prop to `SliderBaseProps` that fixes an issue where slider handle labels (tooltips showing current values) get clipped when the slider is placed inside containers with `overflow: hidden`. This is particularly problematic with sliders that have large min/max ranges (e.g., -1,000,000 to 1,000,000) where the label text is wide.

### Technical implementation:
  - Added `ensureParentOverflowVisible?: boolean` prop to `SliderBaseProps` interface (defaults to `false` for backward
  compatibility)
  - Modified Handle component to programmatically set parent container's overflow style to visible when the prop is enabled
  - Walks up the DOM tree from the slider element to find all ancestor elements with `overflow: hidden/auto/scroll` and
  temporarily modifies them
  - Stores original overflow values and restores them when the prop is disabled or component unmounts
  - Works with all slider variants (`Slider`, `RangeSlider`, `MultiSlider`) and both orientations (horizontal, vertical)

### Files modified:
- `packages/core/src/components/slider/multiSlider.tsx` - Added prop definition and default value
- `packages/core/src/components/slider/handle.tsx` - Added DOM traversal logic to modify parent overflow styles

#### Screenshot

Screenshot 2025-10-09 at 8 47 32 PM

### DIFF
--- a/packages/core/src/components/slider/handle.tsx
+++ b/packages/core/src/components/slider/handle.tsx
@@ -116,34 +116,34 @@ export class Handle extends AbstractPureComponent<InternalHandleProps, HandleSta
         // Also include the styling that would normally come from .bp5-slider-handle .bp5-slider-label
         const style: React.CSSProperties = vertical
             ? {
-                  position: "fixed",
+                  // Tooltip-like styling from SCSS (lines 141-157 in _slider.scss)
+                  background: "rgb(47, 51, 56)", // $dark-gray5
+                  borderRadius: "3px",
+                  boxShadow:
+                      "0 0 0 1px rgba(17, 20, 24, 0.1), 0 2px 4px rgba(17, 20, 24, 0.2), 0 8px 24px rgba(17, 20, 24, 0.2)",
+                  color: "rgb(216, 219, 223)", // $light-gray5
+                  fontSize: "12px",
                   left: `${handleRect.right + labelOffset}px`,
+                  lineHeight: "1",
+                  padding: "4px 8px",
+                  position: "fixed",
                   top: `${handleRect.top + handleRect.height / 2}px`,
                   transform: "translateY(-50%)",
-                  // Tooltip-like styling from SCSS (lines 141-157 in _slider.scss)
-                  background: "rgb(47, 51, 56)", // $dark-gray5
-                  borderRadius: "3px",
-                  boxShadow:
-                      "0 0 0 1px rgba(17, 20, 24, 0.1), 0 2px 4px rgba(17, 20, 24, 0.2), 0 8px 24px rgba(17, 20, 24, 0.2)",
-                  color: "rgb(216, 219, 223)", // $light-gray5
-                  padding: "4px 8px",
-                  fontSize: "12px",
-                  lineHeight: "1",
               }
             : {
-                  position: "fixed",
-                  left: `${handleRect.left + handleRect.width / 2}px`,
-                  top: `${handleRect.top}px`,
-                  transform: `translate(-50%, ${labelOffset}px)`,
                   // Tooltip-like styling from SCSS (lines 141-157 in _slider.scss)
                   background: "rgb(47, 51, 56)", // $dark-gray5
                   borderRadius: "3px",
                   boxShadow:
                       "0 0 0 1px rgba(17, 20, 24, 0.1), 0 2px 4px rgba(17, 20, 24, 0.2), 0 8px 24px rgba(17, 20, 24, 0.2)",
                   color: "rgb(216, 219, 223)", // $light-gray5
-                  padding: "4px 8px",
                   fontSize: "12px",
+                  left: `${handleRect.left + handleRect.width / 2}px`,
                   lineHeight: "1",
+                  padding: "4px 8px",
+                  position: "fixed",
+                  top: `${handleRect.top}px`,
+                  transform: `translate(-50%, ${labelOffset}px)`,
               };
 
         return (
--- a/packages/core/src/components/slider/handle.tsx
+++ b/packages/core/src/components/slider/handle.tsx
@@ -18,6 +18,7 @@ import classNames from "classnames";
 
 import { AbstractPureComponent, Classes, Utils } from "../../common";
 import { DISPLAYNAME_PREFIX } from "../../common/props";
+import { Portal } from "../portal/portal";
 
 import type { HandleProps } from "./handleProps";
 import { formatPercentage } from "./sliderUtils";
@@ -34,6 +35,7 @@ export interface InternalHandleProps extends HandleProps {
     tickSize: number;
     tickSizeRatio: number;
     vertical: boolean;
+    usePortal?: boolean;
 }
 
 export interface HandleState {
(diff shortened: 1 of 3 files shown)


packages/core/src/components/slider/handle.tsx
```diff
@@ -65,29 +67,91 @@ export class Handle extends AbstractPureComponent<InternalHandleProps, HandleSta
...
+        // Position the label to match inline positioning
+        // Inline labels use: position: absolute; transform: translate(-50%, $label-offset)
+        // where $label-offset = $handle-size + $pt-spacing (16px + 4px = 20px)
+        // The transform is applied from the handle's position, so we need to calculate from handle's top
+        const labelOffset = 20; // matches $label-offset in SCSS
+
+        // Also include the styling that would normally come from .bp5-slider-handle .bp5-slider-label
+        const style: React.CSSProperties = vertical
+            ? {
+                  // Tooltip-like styling from SCSS (lines 141-157 in _slider.scss)
```
ggdouglas: I'm not a fan of duplicating all those inline styles from `.bp6-slider-label`. The two implementations could drift apart pretty easily. How about we add a class to the portal (something like `.bp6-slider-portal`) to handle the majority of these styles via the existing class instead?

__The idea:__

1. Add `className="bp6-slider-portal"` to the Portal component
2. Update the SCSS to target both `.bp6-slider-handle .bp6-slider-label` AND `.bp6-slider-portal .bp6-slider-label` with the shared styles
3. Keep only the positioning stuff (position, top, left, transform) as inline styles since those need to be calculated on the fly

packages/core/src/components/slider/handle.tsx
```diff
@@ -65,29 +67,91 @@ export class Handle extends AbstractPureComponent<InternalHandleProps, HandleSta
...
+                  top: `${handleRect.top}px`,
+                  transform: `translate(-50%, ${labelOffset}px)`,
+              };
+
+        return (
+            <Portal>
+                <span className={Classes.SLIDER_LABEL} style={style}>
+                    {label}
+                </span>
+            </Portal>
```
ggdouglas: Since we're using a portal, we unfortunately also lose the `.bp6-dark` class scope that gives the label styles in dark mode. I think we should probably address this.

Screenshot 2025-10-09 at 14 01 43@2x

Perhaps we could use a method similar to the one we use in Popover with `hasDarkParent`: https://github.com/palantir/blueprint/blob/031a998e066f234067e46ca37d4e5ad192c51b42/packages/core/src/components/popover/popover.tsx#L700-L705

packages/core/src/components/slider/handle.tsx
```diff
@@ -65,29 +67,91 @@ export class Handle extends AbstractPureComponent<InternalHandleProps, HandleSta
...
+        );
+    }
+
+    private renderLabelPortal(label: React.JSX.Element | string) {
+        if (this.handleElement == null) {
+            return null;
+        }
+
+        // Calculate the position of the label based on the handle's current position
+        const handleRect = this.handleElement.getBoundingClientRect();
```
ggdouglas: I think there are a few issues with using `getBoundingClientRect()` here for calculating label position. The positioning is only calculated once during render, but the handle position can change dynamically without triggering a re-render. For example, when scrolling or resizing the page:

Screenshot 2025-10-09 at 14 12 06

Screenshot 2025-10-09 at 14 11 30

packages/core/src/components/slider/multiSlider.tsx
```diff
@@ -119,6 +119,17 @@ export interface SliderBaseProps extends Props, IntentProps {
...
+    /**
+     * Whether to use a `Portal` to render the handle labels.
+     *
+     * This can be useful to prevent overflow issues in containers with `overflow: hidden`.
+     * When enabled, handle labels will render in a separate React root outside the DOM hierarchy,
+     * allowing them to break out of containers with overflow constraints.
+     *
+     * @default false
+     */
+    useHandleLabelPortal?: boolean;
```
ggdouglas: Can we add this prop to the [Slider component docs example](https://github.com/palantir/blueprint/blob/031a998e066f234067e46ca37d4e5ad192c51b42/packages/docs-app/src/examples/core-examples/sliderExample.tsx)? This would make it easier to review/test this change.

packages/core/src/components/slider/handle.tsx
```diff
@@ -65,29 +67,91 @@ export class Handle extends AbstractPureComponent<InternalHandleProps, HandleSta
...
+                  boxShadow:
+                      "0 0 0 1px rgba(17, 20, 24, 0.1), 0 2px 4px rgba(17, 20, 24, 0.2), 0 8px 24px rgba(17, 20, 24, 0.2)",
+                  color: "rgb(216, 219, 223)", // $light-gray5
+                  fontSize: "12px",
+                  left: `${handleRect.right + labelOffset}px`,
+                  lineHeight: "1",
+                  padding: "4px 8px",
+                  position: "fixed",
+                  top: `${handleRect.top + handleRect.height / 2}px`,
+                  transform: "translateY(-50%)",
```
ggdouglas: Calculation seems to be a bit off here with respect to positioning the label next to a handle on a `vertical` Slider.

Screenshot 2025-10-09 at 14 24 56@2x

ggdouglas (changes requested):

aworld1:
@ggdouglas I have made some changes to make this severely less complicated. Instead of using portals, which proved to be exceptionally challenging to mimic the example formatting that we had for the slider handle's existing styling, I went with a CSS trick that does the job. I noticed that `overflow: visible;` on the container of the slider fixes this issue, however there is no way to force the user to make styling changes to their parent divs. So, I created a flag that programmatically sets the overflow properties of the parent, causing the handle to show through.

I have temporarily modified the docs to show a section that contains a demo of this directly, which can be removed pre-merge (see screenshot in PR description).

